PR #13, opened 2017-12-04
travis: update go versions

dvrkps:


danderson:
Wow, I've been specifying the Go versions incorrectly! Thank you for fixing it!

PR #25, opened 2017-12-08
Added *.pyc to gitignore

xnaveira:
The fabfile.pyc in my git status is driving me crazy :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 vendor/
+*.pyc


danderson:
The travis build failure is an unrelated race condition, so that's fine. However, you do need to sign the CLA before I can hit the merge button. Yes, it's a bit silly for a 1-line gitignore change, but those are the rules sadly.

xnaveira:
Yeah, it was kind of silly PR such a small change :)

Thanks!

PR #31, opened 2017-12-09
Add arp-speaker

miekg:
This adds a arp-speaker that will make an LB IP exist on the LAN by
responding to ARP request for that IP. It needs to run on a single node
and will use that node's MAC address as the destination for the service
IP.

No config changes are needed, arp-speaker will consume bgp-speaker
config and only use what it needs.

This PR only *adds* arp-speaker, it has not been hooked up to be
actually used by metallb.

Also: not tested \0/ and there is no periodic routine sending out unsolicated ARP responses.

### DIFF
--- a/arp-speaker/main.go
+++ b/arp-speaker/main.go
@@ -1,0 +1,227 @@
+// Copyright 2017 Google Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"errors"
+	"flag"
+	"fmt"
+	"net"
+	"os"
+
+	"go.universe.tf/metallb/internal/allocator"
+	"go.universe.tf/metallb/internal/arp"
+	"go.universe.tf/metallb/internal/config"
+	"go.universe.tf/metallb/internal/k8s"
+
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+
+	"k8s.io/api/core/v1"
+)
+
+type controller struct {
+	myIP   net.IP
+	myNode string
+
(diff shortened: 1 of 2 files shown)


miekg:
somewhat meta: there is also no way to switch to this LB instead of bgp one.
As discussed: lots of code duplication.

miekg:
I'm eager to test this out, but I need to setup minikube or buy some new/extra Pi hardware

arp-speaker/main.go
```diff
@@ -0,0 +1,227 @@
...
+		glog.Infof("%s: skipped, waiting for config", name)
+		return nil
+	}
+
+	if len(svc.Status.LoadBalancer.Ingress) != 1 {
+		glog.Infof("%s: no IP allocated by controller", name)
+		return c.deleteBalancer(name, "no IP allocated by controller")
+	}
+
+	// Should we advertise? Yes, if externalTrafficPolicy is Cluster,
```
danderson: Note for future changes: the arp logic needs to run leader election, and only 1 node should advertise. Sadly ARP doesn't really support LB, and things won't work right if there's >1 ARP responder for a VIP.

But, this is just a copypasta  import for now, no need to change it for this CL.

internal/arp/arp.go
```diff
@@ -0,0 +1,153 @@
...
+		client:       client,
+		c:            make(chan *arp.Packet),
+		ips:          make(map[string]net.IP),
+	}, nil
+}
+
+// Start starts the announcer, making it listen on the interface for ARP requests.
+func (a *Announce) Start() {
+	// Read packet from the wire.
+	go func() {
```
danderson: Curious: why the goroutines here?

arp-speaker/main.go
miekg: I was hoping to postpone this until every thing is more polished, although leader-election in a k8s might be trivial to implement.

internal/arp/arp.go
miekg: Grand plan that fell apart in the last minute. There is no reason why this should be very different from: https://raw.githubusercontent.com/mdlayher/arp/master/cmd/proxyarpd/main.go
i.e. just a `for` loop reading and responding.

arp-speaker/main.go
danderson: There is https://godoc.org/k8s.io/client-go/tools/leaderelection . It's a library that uses k8s Endpoints objects to implement leader election, similar to how Chubby does it (try to atomically write the leader's IP to the file, if you succeed, you are the leader).

It's fine to do this whenever you want :)

PR #35, opened 2017-12-10
Cleanup: arp-speaker

miekg:
Make arp-speaker simpler, implement leaderelection and provide the
framework for sending unsolicitated ARPs on acquiring leadership
~~(actually sending not implemented yet).~~

I needed access to the k8s client, hence the internal/k8s/kus.go change,
the rest is all local to the arp stuff.

All not tested; but this should give me a base to allow some local
testing.

### DIFF
--- a/internal/arp/arp.go
+++ b/internal/arp/arp.go
@@ -119,7 +119,16 @@ func (a *Announce) Announce(ip net.IP) bool {
 
 // Unsolicited returns a slice of ARP responses that can be send out as unsolicited ARPs.
 func (a *Announce) Unsolicited() []*arp.Packet {
-	return nil
+	a.RLock()
+	defer a.RUnlock()
+
+	arps := []*arp.Packet{}
+	for _, ip := range a.ips {
+		if a, err := arp.NewPacket(arp.OperationReply, a.hardwareAddr, ip, ethernet.Broadcast, ip); err == nil {
+			arps = append(arps, a)
+		}
+	}
+	return arps
 }
 
 // interfaceByIP returns the interface that has ip.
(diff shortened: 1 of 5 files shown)


internal/arp/arp.go
```diff
@@ -94,15 +89,14 @@ func (a *Announce) Close() error {
...
-// Now only uses an net.IP.
+// SetBalancer implementes adds ip to the set of announced address.
 func (a *Announce) SetBalancer(name string, ip net.IP) {
 	a.Lock()
 	defer a.Unlock()
 	a.ips[name] = ip
 }
 
-// DeleteBalancer deletes...
+// DeleteBalancer an address from the set of address we should announce.
```
danderson: Missing word: deletes. Github + timezones makes review RTT annoying, so I'll just fix it up in a followup commit

internal/k8s/k8s.go
```diff
@@ -214,6 +214,9 @@ func (c *Client) Errorf(svc *v1.Service, kind, msg string, args ...interface{})
 	c.events.Eventf(svc, v1.EventTypeWarning, kind, msg, args...)
 }
 
+// ClientCoreV1 returns the embeded k8s client from Client.
+func (c *Client) ClientCoreV1() v1core.CoreV1Interface { return c.client.CoreV1() }
```
danderson: I'm not a fan of this, because it leaks the whole k8s API to all clients. That'll make it significantly harder to test arp-speaker without having to deal with k8s machinery.

Maybe some time later, move the k8s-ish leader election code into internal/k8s, so we can keep the k8s-specific implementation bits hidden?

But as an incremental step, this is fine, no action required in this PR.

miekg:
Moving the master election into internal/k8s sgtm.

PR #38, opened 2017-12-11
Move leadership code into internal/k8s

miekg:
This moves the code into k8s.

*cleanup*

Fixes no issue, but discussued offline.


miekg:
So.. I've been reading the source for the leadership election stuff, but I'm slightly unclear:
a) should `le.Run()` be started in a go-routine or does it do this by itself?
b) does it `Run()` exit and do I need to restart it?

I *think* a) No b) No

Issue #47, opened 2017-12-13
No Makefiles?


miekg:
Is that fabfile.py thing the replacement for make in this project?
i.e. If I send a PR with arp-speaker build + docker foo in a Makefile will that be accepted?

danderson:
Fabric is a make replacement, effectively. It's in python, so it lets me do things like use context managers and fancy string substitution that are harder to do in make. I am misusing it slightly, it's really designed for doing stuff over SSH to remote systems, I'm using the 10% of Fabric that lets you run local commands.

I'm not *deeply* committed to it, it's just a tool I'm comfortable with so I used it. If you think you can reproduce its current functionality in a makefile, I'm open to switching to make. But I am opposed to having makefile *and* fabfile, we should use one or the other, not both.

xnaveira:
I know that you don't like both but what about calling makefiles from the fabfile? I mean, let makefiles do one thing only, which is the build, and then use the fabfile to build the stuff through calling the makefiles and then do the deploy which is what fabric is supposed to be doing anyway? Best of both worlds...

miekg:
I mostly have .PHONY targets in the Makefile, so not really using that `make` feature :)

As long as someone is willing to rewrite my Makefiles to Python, I happy with it :)

danderson:
@xnaveira We're both using Make and Fabric "wrong", as a way to run shell scripts with some forms of error handling. I don't think mixing both will get us anything useful :).

@miekg I'm more than happy to be on the hook for porting your build steps into fabfile.py, if you're okay using it. Now that 0.2.0 is released, my next focus is getting multi-arch builds into master.

miekg:
sgtm.

happy to report the multi arch stuff is working (allthough everything is crashlooping :-) )

PR #48, opened 2017-12-13
WIP: Test Docker multi-arch image

miekg:
Test PR to feel the waters; untested; this add a Makefile and docker file that builds a multi arch docker image - I'll be using this to test metallb at home.

### DIFF
--- a/arp-speaker/Dockerfile
+++ b/arp-speaker/Dockerfile
@@ -1,5 +1,5 @@
 FROM alpine:latest
 MAINTAINER Miek Gieben <miek@miek.nl> @miekg
+
 ADD arp-speaker /arp-speaker
-USER nobody:nobody
 ENTRYPOINT ["/arp-speaker", "--logtostderr"]
--- a/arp-speaker/Dockerfile
+++ b/arp-speaker/Dockerfile
@@ -2,7 +2,6 @@ FROM alpine:latest
 MAINTAINER Miek Gieben <miek@miek.nl> @miekg
 
 ADD arp-speaker /arp-speaker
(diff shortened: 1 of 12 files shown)


miekg:
fyi: just jotting down some docs for me

For my testing to works I also need to give the controller same treatment and then change the yaml to use my docker registry and (I think) add `hostNetwork = true`

arp-speaker/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM alpine:latest
```
danderson: Check out the contents of the `dockerfiles/` directory, it has some templates for dev+prod containers for the metallb binaries. At some point you might want to try using those templates, which might work as well?

arp-speaker/Makefile
```diff
@@ -0,0 +1,61 @@
+SYSTEM:=
```
danderson: I have to admit I'm really not a fan of the makefile format. I know it's super portable and available anywhere, but ugh it's such write-only code :(. That's why I used Fabric in the end, python is at least semi-maintainable.

But as I said in #47, if we can get all the stuff currently in fabfile into a makefile... I guess we could maybe switch to makefile. I'll hate it, but if it makes metallb easier to hack on...

miekg:
Note: I have no expectation that any of this will be merged. Mostly me doing my thing whilst debugging arp on arm and having this mostly as an fyi.

danderson:
This is a great FYI, and thank you for posting it! I'll translate it back into fabfile once I have a free moment, but this is a great guide for what the fabric command needs to do.

miekg:
Code changes have been merged seperately. The Makefile and Dockerfiles are now in #59.

Closing this here.

PR #58, opened 2017-12-14
internal/arp: initial tests for Announce.Run logic

mdlayher:

Initial tests for the ARP announcer read loop.

Fixes # n/a

This works, but I think I should probably rework my ARP package a bit to not depend on an interface handle, since it mostly cares about addresses anyway.

### DIFF
--- a/internal/arp/arp.go
+++ b/internal/arp/arp.go
@@ -41,43 +41,64 @@ func New(ip net.IP) (*Announce, error) {
 	}, nil
 }
 
+// A dropReason is a reason an Announce drops an incoming ARP packet.
+type dropReason int
+
+// Possible dropReason values.
+const (
+	dropReasonNone dropReason = iota
+	dropReasonReadError
+	dropReasonARPReply
+	dropReasonEthernetDestination
+	dropReasonAnnounceIP
+	dropReasonNotLeader
+)
+
 // Run starts the announcer, making it listen on the interface for ARP requests. It only responds to these
 // requests when a.leader is set to true, i.e. we are the current cluster wide leader for sending ARPs.
 func (a *Announce) Run() {
 	for {
-		pkt, eth, err := a.client.Read()
-
-		if err != nil {
-			continue
-		}
-
-		// Ignore ARP replies.
-		if pkt.Operation != arp.OperationRequest {
-			continue
-		}
-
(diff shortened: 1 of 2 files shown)


internal/arp/arp_test.go
```diff
@@ -0,0 +1,169 @@
...
+		t.Fatalf("failed to create ARP client: %v", err)
+	}
+
+	a := &Announce{
+		hardwareAddr: ifi.HardwareAddr,
+		client:       c,
+		ips:          make(map[string]net.IP),
+	}
+
+	uc, err := net.DialUDP("udp", nil, pc.LocalAddr().(*net.UDPAddr))
```
danderson: This fails on travis - looks like the packetconn sets the IP portion of the address to the empty address (`[::]`), so the dial fails?

Seems to fail consistently on 1.8, 1.9, 1.10beta1. Can you reproduce locally?

mdlayher: Interesting, it works just fine locally with 1.9.2 but I'll play with it more tomorrow.

internal/arp/arp.go
```diff
@@ -41,43 +41,64 @@ func New(ip net.IP) (*Announce, error) {
 	}, nil
 }
 
+// A dropReason is a reason an Announce drops an incoming ARP.
```
danderson: missing word: "an incoming ARP packet"

danderson:
I like the way you're doing the testing. Only remaining question is why travis doesn't like that DialUDP...

internal/arp/arp_test.go
miekg: doesn't get `:0` you an IP v6 addr? that fails on travis?

mdlayher: Updated to use `localhost:0` instead and it works in Travis now.

danderson: Turns out this is travis-ci/travis-ci#5359 . IPv6 connectivity in test containers used to work, then they deliberately broke it because they don't *officially* support it :(

I guess I need to think about where to rehome the CI/CD stuff some day...

miekg:
[ Quoting <notifications@github.com> in "Re: [google/metallb] internal/arp: ..." ]
>I guess I need to think about where to rehome the CI/CD stuff some day...

Once you do, please tell, getting daily announed by travis, slow and no ipv6

PR #59, opened 2017-12-15
Dockerfile and Makefiles

miekg:
This PR has no commits other than adding Makefiles and Dockerfiles.


danderson:
I'm working on doing this with Fabric right now. There's some complication because test-bgp-router won't build correctly on !amd64, so I'm finding a workaround for that now. I'm using this PR as a reference, but unless I fail with fabric, I'd prefer to not merge makefiles.

PR #61, opened 2017-12-15
Add disable auto assign flag

### LINKED ISSUE #4: Add a flag to disable automatic allocation in an address pool
In some environments, you'll have some large address pools of "cheap" IPs (e.g. RFC1918), and some smaller pools of "expensive" IPs (e.g. public IPv4 addresses leased on the grey market).

Right now, MetalLB will allocate IPs from any address pool that has free addresses. This might end up burning "expensive" addresses for services that don't need it. Of course, you can override this selection wiith `loadBalancerIP` and `metallb.universe.tf/address-pool`, but this is error-prone if the default is still to do something expensive.

Address pool configs should have an optional parameter, `auto-assign`, defaulting to `true`. When turned off, MetalLB will not allocate from the given pool unless that pool has been explicitly requested via `loadBalancerIP` or the `address-pool` annotation.

msoderberg:
A proposed implementation of the "disable auto assign" feature brought up in #4 

**Which issue(s) this PR fixes**:
Fixes #4 

Defaulting the flag to true required some changes in the test configurations. Maybe it would be cleaner to name it `disable-auto-assign` so we can go with the zero bool value if the flag has not been set in the YAML?

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -83,7 +83,8 @@ type Pool struct {
 	// unusable, for maximum compatibility with ancient parts of the
 	// internet.
 	AvoidBuggyIPs bool
-	// Prevents IP addresses to be automatically assigned from this pool.
+	// If false, prevents IP addresses to be automatically assigned
+	// from this pool.
 	AutoAssign bool
 	// When an IP is allocated from this pool, how should it be
 	// translated into BGP announcements?
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -39,6 +39,7 @@ type configFile struct {
 		Name           string
 		CIDR           []string
 		AvoidBuggyIPs  bool `yaml:"avoid-buggy-ips"`
+		AutoAssign     *bool `yaml:"auto-assign"`
 		Advertisements []struct {
 			AggregationLength *int `yaml:"aggregation-length"`
 			LocalPref         *uint32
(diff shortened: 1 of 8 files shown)


internal/config/config.go
```diff
@@ -82,6 +83,8 @@ type Pool struct {
 	// unusable, for maximum compatibility with ancient parts of the
 	// internet.
 	AvoidBuggyIPs bool
+	// Prevents IP addresses to be automatically assigned from this pool.
```
danderson: Can you start the docstring with "if false, ..." here? Otherwise this docstring sounds like I have to set this = true to prevent auto-allocation.

manifests/example-config.yaml
```diff
@@ -43,6 +43,10 @@ data:
       # smurf protection. Such devices have become fairly rare, but
       # the option is here if you encounter serving issues.
       avoid-buggy-ips: true
+      # (optional) If false, MetalLB will not automatically allocate
```
danderson: Can you say `(optional, default true)`, to make it clear that you have to disable it?

msoderberg:
Thanks for the feedback! I've updated accordingly. I added a section in the usage docs but I basically copied your description from #4. Feel free to edit as you wish before you merge :)

danderson:
Thank you for contributing! Merged.

PR #63, opened 2017-12-15
make DeleteBalancer part of the controller interface

### LINKED ISSUE #30: extend interface with DeleteBalancer?
Currently we have SetBalancer and SetConfig. I *think* (nothing coded) yet, that DeleteBalancer makes sense as well.

miekg:
This makes DeleteBalancer a thing in the interface and calls it
from internal/k8s. This drops the reason string - the caller should log
the reason and then call DeleteBalancer.

Abstracting things even more mean bgp/arp speaker can easily become 1
binary.

Updated {bgp,arp}-speaker.

Fixes #30

### DIFF
--- a/arp-speaker/main.go
+++ b/arp-speaker/main.go
@@ -44,10 +44,6 @@ type controller struct {
 }
 
 func (c *controller) SetBalancer(name string, svc *v1.Service, eps *v1.Endpoints) error {
-	if svc == nil {
-		return c.deleteBalancer(name, "service deleted")
-	}
-
 	if svc.Spec.Type != "LoadBalancer" {
 		return nil
 	}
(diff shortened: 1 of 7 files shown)


miekg:
hmmm, the testing may needs some tweaks - I've been deleting test cases don't make sense anymore - because DeleteBalancer is there now - but that is probably not the best way forward...

miekg:
speaker_test.go:681: "Delete peer": unexpected advertisement state (-want +got)

still fails

bgp-speaker/speaker_test.go
```diff
@@ -590,24 +590,26 @@ func TestBGPSppeaker(t *testing.T) {
...
 
-		{
-			desc:     "Delete svc",
-			balancer: "test1",
-			wantAds: map[string][]*bgp.Advertisement{
-				"1.2.3.4:0": []*bgp.Advertisement{
-					{
-						Prefix:  ipnet("10.20.30.5/32"),
-						NextHop: net.ParseIP("1.2.3.4"),
+		/*
```
danderson: I don't understand why you're deleting these tests. This change should be a no-op, if it's not there's a logic bug because we're no longer deleting services correctly.

arp-speaker/main.go
```diff
@@ -62,32 +58,32 @@ func (c *controller) SetBalancer(name string, svc *v1.Service, eps *v1.Endpoints
 
 	if len(svc.Status.LoadBalancer.Ingress) != 1 {
 		glog.Infof("%s: no IP allocated by controller", name)
-		return c.deleteBalancer(name, "no IP allocated by controller")
+		return c.DeleteBalancer(name)
```
danderson: I'm not super happy about losing the reason field here, it was pretty vital when debugging why services are not working correctly - otherwise all you know is that your service got deleted, but you don't know which of the 6-7 codepaths decided to delete it.

I think we should try to keep the reason around. And based on that... I'm wondering if this change actually makes sense, because all we're doing is making a method public and removing one "if svc == nil { ... }"... WDYT?

miekg: I'm fine with leaving things as-is.

bgp-speaker/speaker_test.go
miekg: The test is still calling SetBalancer, but  now with `nil`s which is doesn't check for anymore.

But this PR seems more trouble then it's worth; happy to drop.

PR #66, opened 2017-12-15
arp-speaker: more logging and other fixes

miekg:
* more logging
* send unsolicitated arps every 10 second
* make leader election times longer

### DIFF
--- a/internal/arp/arp.go
+++ b/internal/arp/arp.go
@@ -18,6 +18,7 @@ type Announce struct {
 	client       *arp.Client
 	ips          map[string]net.IP // map containing IPs we should announce
 	sync.RWMutex                   // protects ips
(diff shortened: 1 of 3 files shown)


internal/arp/arp.go
```diff
@@ -58,7 +60,10 @@ const (
 // requests when a.leader is set to true, i.e. we are the current cluster wide leader for sending ARPs.
 func (a *Announce) Run() {
 	for {
-		a.readPacket()
+		dropped := a.readPacket()
+		if dropped != dropReasonNone {
+			glog.Infof("ARP packet dropped %d", dropped)
```
danderson: Some day, this might be too spammy. But for now it's worth it.

PR #75, opened 2017-12-16
config: allow protocol selection

miekg:
This introduces protocol into the config with 3 allowed values: "arp",
"bgp" and "rip", this specifices what protocol the "speaker" binary
speaks.

### DIFF
--- a/arp-speaker/main.go
+++ b/arp-speaker/main.go
@@ -44,6 +44,10 @@ type controller struct {
 }
 
 func (c *controller) SetBalancer(name string, svc *v1.Service, eps *v1.Endpoints) error {
+	if a.config.Protocol != config.ProtoARP {
+		return nil
+	}
+
 	if svc == nil {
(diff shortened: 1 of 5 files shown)


miekg:
@danderson this is a first step to merging the speakers into one binary - if you like it that is.

miekg:
#76 has all the fixed that this PR doesn't have

danderson:
I think this branch got merged in the other PR, I see all the same code.

miekg:
Yes. In branched this of the another branch. End result should be ok
though, I hope?

danderson:
Yes, but this got merged already. When I merged the other PR, it pulled in this code as well. If you look at the code in master, it already has these changes (that's also why the PR now complains about merge conflicts).

So I think we can just close this PR, this code is already merged.

danderson:
To be clear: the result lgtm, I was just making sure I can close this PR without merging, because the code is already merged.

PR #76, opened 2017-12-16
Merge speakers

miekg:
Merge both speakers into 1 binary.


miekg:
if c.config is `nil` this now crashes (test exposes this), because we only act when we have config. Not sure how we deal with this "service deleted" when we would tick everything off from the config?
~~~
	if c.config.Protocol != config.ProtoBGP {
		return nil
	}

	if svc == nil {
		return c.deleteBalancer(name, "service deleted")
	}

	if svc.Spec.Type != "LoadBalancer" {
		return nil
	}

	glog.Infof("%s: start update", name)
	defer glog.Infof("%s: end update", name)

	if c.config == nil {
		glog.Infof("%s: skipped, waiting for config", name)
		return nil
	}
~~~

miekg:
hmm, it could be harmless. Anyway, I'll wait until you wake up :)

miekg:
Note we *shoudn't* need two k8s clients.

manifests/example-config.yaml
```diff
@@ -5,6 +5,7 @@ metadata:
   name: config
 data:
   config: |
+    protocol: "bgp"
```
danderson: Can you make this field per-address-pool? In my cluster, I want some address pools advertised over BGP, and others advertised over ARP :)

Another open question: for a single address pool, does it make sense to advertise several protocols? As in, `protocol: ["bgp", "arp"] ? I'm not sure if it makes sense, but I don't know. WDYT?

speaker/main.go
```diff
@@ -0,0 +1,102 @@
...
+
+	// Setup both ARP and BGP clients and speakers, config decides what is being done runtime.
+
+	// BGP
+	cb, err := newBGPController(myIP, *myNode)
+	if err != nil {
+		glog.Fatalf("Error getting BGP controller: %s", err)
+	}
+
+	clientb, err := k8s.NewClient(name(config.ProtoBGP), *master, *kubeconfig, cb, true)
```
danderson: TODO future cleanup, we can probably have a single k8s client and a single leader elector for all protocols, and do the fanout in this code.

Each k8s client maintains its own cache of all the objects, and a separate connection to the k8s apiserver, so currently this code is consuming 2x the ram it needs, and making 2x the number of necessary connection (scaling concern).

Again, no need to change this right now, I'm just braindumping things we should do later.

PR #86, opened 2017-12-17
Make proto a property of the pool

### LINKED ISSUE #80: Protocol per address-pool

feature request

**What happened**:

The protocol selection is now global, it makes more sense to push this down to the address-pools so that you can mix and match.

@danderson:
> Can you make this field per-address-pool? In my cluster, I want some address pools advertised over BGP, and others advertised over ARP :)

> Another open question: for a single address pool, does it make sense to advertise several protocols? As in, `protocol: ["bgp", "arp"] ? I'm not sure if it makes sense, but I don't know. WDYT?

Not sure about the mutiple protocols, smells a bit like premeture optimization; we can leave that until some asks for it.

miekg:
Move proto down to the address pools, also drop the Proto prefix on ProtoBGP, so it is just
BGP, looks nicer.

Fixes #80

### DIFF
--- a/speaker/bgp_controller.go
+++ b/speaker/bgp_controller.go
@@ -68,10 +68,6 @@ func (c *bgpController) SetBalancer(name string, svc *v1.Service, eps *v1.Endpoi
 		return nil
 	}
 
-	if c.config.Protocol != config.ProtoBGP {
-		return nil
-	}
-
 	if len(svc.Status.LoadBalancer.Ingress) != 1 {
 		glog.Infof("%s: no IP allocated by controller", name)
 		return c.deleteBalancer(name, "no IP allocated by controller")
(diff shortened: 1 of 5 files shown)


speaker/bgp_controller.go
```diff
@@ -101,6 +97,11 @@ func (c *bgpController) SetBalancer(name string, svc *v1.Service, eps *v1.Endpoi
 		glog.Errorf("%s: could not find pool %q that definitely should exist!", name, poolName)
 	}
 
+	if pool.Protocol != config.BGP {
+		glog.Errorf("%s: protocol %s in pool is not set to ARP", name, pool.Protocol)
```
danderson: s/ARP/BGP/, please fix in followup?

PR #99, opened 2017-12-17
Split makefile push and set-image

miekg:
If I push, say amd64, I don't want set-image because these images
will be of the wrong arch. Split the targets

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,6 +54,9 @@ push-manifests:
 .PHONY: push
 push: gen-image-targets
 	+make -f $(MK_IMAGE_TARGETS) $(foreach bin,$(BINARIES),$(bin)/$(ARCH))
+
+.PHONY: set-image
+set-image:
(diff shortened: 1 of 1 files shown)


miekg:
This won't work because TAG is a timestamp....

danderson:
What's your use case for this? `make push` is supposed to be specifically "build just the arch I need for my test cluster, and deploy to that cluster". In that scenario, it doesn't make sense to separate push and set-image... And as you saw, separating them breaks the timestamp-based image tags (necessary so that k8s actually redeploys the pods).

So... Why do need this? Maybe there's a different way to do it...

miekg:
Ack I was building amd64 to just push to the registry. What target should I use for that?

danderson:
Aah, I see. Long term, will you ever need to push amd64-only? If no, you can `make all-arch-images` to push images for all architectures, and also push the multi-arch manifest.

Do you think we need to keep a make command for pushing single-arch images (other than the the "dev iteration" use case) ? If so, we should make a command separate from `make push`, and leave `make push` as-is so that it correctly does timestamp-based pushes.

PR #111, opened 2017-12-19
Arp network

### LINKED ISSUE #110: Avoid enclosing network/broadcast addresses in ARP mode

Feature request

**What happened**:

Let's say my LAN is 192.168.16.0/24. The lower end of that is my DHCP range, so I set up MetalLB to advertise with ARP, on 192.168.16.192/26. This covers the addresses from 192.168.16.192 to 192.168.16.255.

Note that the .255 IP is actually not usable, because that's the broadcast address of the enclosing /24.

We need to somehow teach MetalLB how to avoid those addresses. In a perfect world, we'd just infer the usable range from the NIC's CIDR range on the hosts, but (a) only the speaker knows that information, (b) it might vary between hosts on non-trivial configurations, (c) it's maybe a bit too magic.

What other options do we have? We could add an "avoid-cidr-edges" knob, on by default, that makes ARP mode automatically avoid the network and broadcast addresses of its CIDR ranges. That probably burns 1 IP too many in some cases, but it should be safe enough to be on by default.

We could add an arp-specific option to say "This is the size of the containing L2 subnet", so that MetalLB can automatically derive usable/unusable IPs from that. That ensures no IP is wasted, but might be a bit more complex to explain to users.

danderson:

Adds the `arp-network` pool attribute, and teaches MetalLB to not allocate network+broadcast IPs in ARP mode.

Fixes #110

### DIFF
--- a/internal/allocator/allocator.go
+++ b/internal/allocator/allocator.go
@@ -72,23 +72,35 @@ func poolCount(p *config.Pool) int64 {
 	for _, cidr := range p.CIDR {
 		o, b := cidr.Mask.Size()
 		sz := int64(math.Pow(2, float64(b-o)))
+		avoidedFirst, avoidedLast := false, false
 		if p.AvoidBuggyIPs {
 			if o <= 24 {
 				// A pair of buggy IPs occur for each /24 present in the range.
 				buggies := int64(math.Pow(2, float64(24-o))) * 2
 				sz -= buggies
+				avoidedFirst, avoidedLast = true, true
 			} else {
 				// Ranges smaller than /24 contain 1 buggy IP if they
 				// start/end on a /24 boundary, otherwise they contain
 				// none.
-				ip := cidr.IP.Mask(cidr.Mask).To4()
-				if ipConfusesBuggyFirmwares(ip) {
+				if ipConfusesBuggyFirmwares(firstIP(cidr)) {
+					avoidedFirst = true
 					sz--
 				}
-				ip[3] += byte(math.Pow(2, float64(b-o)))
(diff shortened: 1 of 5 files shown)


internal/allocator/allocator.go
```diff
@@ -72,25 +72,37 @@ func poolCount(p *config.Pool) int64 {
 	for _, cidr := range p.CIDR {
 		o, b := cidr.Mask.Size()
 		sz := int64(math.Pow(2, float64(b-o)))
+		avoidedFirst, avoidedLast := false, false
```
miekg: Present tense for these: avoidFirst, avoidLast ?
(super nit)

internal/allocator/allocator_test.go
```diff
@@ -482,29 +738,10 @@ func assignedPool(a *Allocator, svc string) string {
...
-			panic(fmt.Sprintf("malformed CIDR %q", cidr))
-		}
-		ret.CIDR = append(ret.CIDR, n)
-	}
-	return map[string]*config.Pool{
-		name: ret,
+func ipnet(s string) *net.IPNet {
+	_, n, err := net.ParseCIDR(s)
+	if err != nil {
+		panic(err)
```
miekg: `panic`? Either we will never hit this, or otherwise just return the error and deal with it?

danderson: If we hit this, it means the test source code has an invalid IP network in it. It should never trigger when testing the code, it's just a shortcut to construct a *net.IPNet.

internal/allocator/allocator.go
danderson: Done.

Issue #83, opened 2017-12-16
NDP speaker implementation


jpds:
**Is this a bug report or a feature request?**: Feature

**What happened**: I noticed that metallb has gained an ARP speaker, which is great for IPv4, however we'll need a NDP speaker for IPv6 once #7 is implemented.

danderson:
Thanks for the report!

Agreed, IPv6 support implies v6 support for all the protocol stacks. My gut feeling is that it'll happen after v0.3. In 0.3 we're adding all the various protocols with compat for k8s 1.8 (i.e. no v6 support). Once we've done that, adding v6 support before k8s 1.10 graduates that support to beta is the logical next step.

miekg:
Is there a good Go library that implements NPD?
@mdlayher's ARP lib, might be whole implementation in MetalLB a breeze.

Implementation wise it looks we're settling on:
* `<proto>_controller.go` in the speaker/ dir
* `internal/<proto>` for protocol details

I'm happy to review PRs. As an intermediate step we could just announce a fake IPv6 address to see if that parts works. Actually hooking it up to k8s and making that all work, will then be step 2.

mdlayher:
I'd actually love to build out an NDP package and may have a need for it for work, but I can't commit to doing so immediately. Best bet for now is probably another small internal package.

mdlayher:
I'm still pretty new to IPv6 but I also implemented DHCPV6: https://github.com/mdlayher/dhcp6

I wonder if prefix delegation would fit with MetalLB as well. Maybe it'll be necessary!

mdlayher:
Looks like I might be getting into NDP in the next couple of days actually, so feel free to assign to me.

Initial game plan is to start putting together a package either on my account or on DO's with a couple of message types set up.

Once that's done, I can at least try to do the non Kubernetes portions since I don't understand the Kubernetes API very well yet.

danderson:
I'm showing that I suck at github... Can I assign issues to non-committers? The assign autocomplete doesn't show you.

mdlayher:
Unsure! Accepted your invitation now.

danderson:
Okay, that did it, looks like I have to invite people to collaborate before I can assign them issues. A bit cumbersome, but we'll make it work :)

Your plan to start with an NDP package SGTM. For reference you may want to consult `internal/arp` to see what kind of API metallb will eventually want, to gracefully plug into the k8s side of things.

[...]

PR #118, opened 2017-12-20
Initial NDP support

mdlayher:

Initial NDP speaker, although it isn't wired up anywhere as of yet.

Updates #83.

I can't seem to figure out why glide tries to update every single thing when I try to vendor this, so I vendored it manually as of now.   I am guessing @danderson and I are on different glide revisions.

/cc @miekg @danderson


internal/ndp/ndp.go
```diff
@@ -0,0 +1,217 @@
...
+	defer a.RUnlock()
+
+	solicited := false
+	for _, ip := range a.ips {
+		_ = a.advertise(net.IPv6linklocalallnodes, ip, solicited)
+	}
+}
+
+// interfaceByIP returns the interface that has ip.
+func interfaceByIP(ip net.IP) (*net.Interface, error) {
```
miekg: As a followup we should start with moving this to internal/interface so both arp and ndp can use it.

mdlayher: Agreed.  I made no effort to dedup as I assume @danderson has put more thought into the abstractions.

danderson:
Thanks! I'll come in behind this merge and do a full vendor update, so that everything is happily synced up.

PR #119, opened 2017-12-20
Start new internal/iface package

miekg:
Allows to mimimize the similarties between ARP and NDP. Also put
dropReasons and there and made them public.

### DIFF
--- a/internal/iface/iface.go
+++ b/internal/iface/iface.go
@@ -1,0 +1,53 @@
+package iface
+
+import (
+	"fmt"
+	"net"
+
+	"github.com/golang/glog"
+)
+
+// DropReason is the reason why a packet was dropped.
+type DropReason int
+
+// Various reasons why a packet was dropped.
+const (
+	DropReasonNone DropReason = iota
+	DropReasonError
+	DropReasonARPReply
+	DropReasonMessageType
+	DropReasonNoSourceLL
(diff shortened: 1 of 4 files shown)


danderson:
CircleCI build fails, looks like you forgot to commit the internal/iface package?

miekg:
of course I forgot that ...

miekg:
... and also forgot to gut ndp for this function.

internal/iface/iface.go
```diff
@@ -0,0 +1,52 @@
...
+	"fmt"
+	"net"
+
+	"github.com/golang/glog"
+)
+
+// DropReason is the reason why a packet was dropped.
+type DropReason int
+
+const (
```
danderson: Constant block needs a docstring (or each constant separately, but the entire block is probably enough)

danderson:
Almost there... Build and test works, now only lint checks fail.

miekg:
linted and squashed

danderson:
golint is still sad about that constant block :(

BTW netlify deploy failure is my fault, I'm refactoring the website and there's dead links that make hugo sad. I'll fix that separately.

Approving the change because I have to run out, please make CircleCI happy before merging :)

danderson:
Ah, and it's happy now! I was looking at the previous build.

PR #127, opened 2018-01-05
Fix usage doc service example: "annotations" field should be part of …

nyaxt:
…ObjectMeta, not Service.

The "Service" yaml example in the documentation cannot be deployed as is.


danderson:
Thank you for the PR!

PR #129, opened 2018-01-10
vendor: bump github.com/mdlayher/ndp

mdlayher:

Updates `ndp` to the latest master, with common message types and fixed options handling.

Fixes # n/a

Even if this isn't in use yet, it's good to ensure that the newest version of `ndp` is used.


glide.yaml
```diff
@@ -4,6 +4,7 @@ import:
 - package: github.com/mdlayher/arp
 - package: github.com/mdlayher/ethernet
 - package: github.com/mdlayher/ndp
+  version: ff6d60d8d99386184158616d7302471581335768
```
danderson: Do you need to explicitly lock the version? I tend to prefer leaving it unlocked so that `glide up` automatically pulls the latest, because I will inevitably forget to maintain version pins :(

mdlayher: I'm no good at glide. I just did this because I couldn't figure out how to get it to update just my package otherwise, heh.

danderson: If you just run `glide up`, it should update everything to the latest version, respecting other constraints.

Then again, version management is NP-hard, so it's possible that my attempts to make k8s and gobgp cooperate has borked something else...

mdlayher:
Can't seem to get it to work locally without specifying a SHA:

```
% glide up
// ...
[INFO]  Replacing existing vendor dependencies
[INFO]  Versions did not change. Skipping glide.lock update.
[INFO]  Project relies on 69 dependencies.
```

Will leave up to you if you want to merge or not.

danderson:
I hate glide/dep/NP-hard version management, and I fervently wish that Go had never opened that can of worms.

Issue #137, opened 2018-01-15
ARP speaker: multiple masters


miekg:
**Is this a bug report or a feature request?**:

bug

**What happened**:

All my 3 arp-speakers are master

**What you expected to happen**:

Only 1 should be master

**How to reproduce it (as minimally and precisely as possible)**:

Deploy metallb with ARP enabled.

**Anything else we need to know?**:

Probably.

**Environment**:
- MetalLB version: master
- Kubernetes version: 1.9.1

miekg:
https://github.com/kubernetes/kubernetes/issues/23731

danderson:
We could try using https://github.com/kubernetes/contrib/tree/master/election , instead of client-go's leader election logic. The contrib version seems simpler (single callback for "current leader is X"), so maybe it's also less terrible?...

danderson:
Oh god, never mind, it uses client-go's library internally... Everything is terrible.

But maybe it provides some clues about what we're doing wrong?...

miekg:
I'm going to look at a few issues on this and maybe ping mike, because I'm running out of idea.
(I thought we might be blocking the OnStartedLeading cb, but that doesn't make much sense  - and this crap *was* working. I've upgrades my k8s version, maybe that jinxed it)

danderson:
I'll try digging into this as well tomorrow. I can reproduce this on my cluster, I have 3 ARP leaders and a bunch of error message sadness in my logs. I'll see if I can find anything.

One thing that comes to mind: I think a recent glide update pulled in v6.0.0 of the k8s client. We were running 5.0.x before, so maybe something broke in there, or the API guarantees/requirements changed, or... something.

miekg:
client-go leadership's code hasn't seen any major changes:
~~~
% git lg *
fe985a55 - Autogenerate BUILD files (3 weeks ago) Jeff Grafton
c92755ea - update BUILD files (3 months ago) Jeff Grafton
5da217e5 - Use buildozer to delete licenses() rules except under third_party/ (5 months ago) Jeff Grafton
fa2ceb74 - Use buildozer to remove deprecated automanaged tags (5 months ago) Jeff Grafton
f921a739 - Run hack/update-bazel.sh to generate BUILD files (5 months ago) Jeff Grafton
684e17ef - Use a client interface instead of a concrete type (6 months ago) Clayton Coleman
0e9b63a3 - Log error when failed to renew lease. (6 months ago) liaoyw
[...]

PR #138, opened 2018-01-15
arp leadership: cosmetic changes

miekg:
Put all leadership state changes in leadership.go and double check
if we're really the leader when sending unsolicitated responses and in
various other places.


miekg:
What I can't wrap my head around is that `OnStoppedLeading` doesn't seem to get called.

miekg:
Ref #137

PR #139, opened 2018-01-15
Dont return

miekg:


miekg:
I'm not at home, so I can't test this right. I think I'm starting to understand how to use this API.
(probably going send a couple of PRs to fix the documentation of the damn thing)

ref #137

PR #140, opened 2018-01-15
Give each instance of speaker a unique identity for leader election.

danderson:
Each replica of speaker must have a distinct identity for leader election, because that's the key they use to decide who is master. We were accidentally setting the identity to the same thing in all instances of speaker, so they all "became" leader simultaneously.

### DIFF
--- a/internal/k8s/leader.go
+++ b/internal/k8s/leader.go
@@ -24,10 +24,17 @@ func (c *Client) NewLeaderElector(a *arp.Announce, lockName string, identity str
 	leader.Set(-1)
 
 	lec := le.LeaderElectionConfig{
-		Lock:          lock,
-		LeaseDuration: 30 * time.Minute,
-		RenewDeadline: 10 * time.Second,
-		RetryPeriod:   5 * time.Second,
+		Lock: lock,
+		// Time before the lock expires and other replicas can try to
+		// become leader.
+		LeaseDuration: 10 * time.Second,
+		// Elected leader refreshes the lock this often. It should be
+		// less than LeaseDuration, so that even if there is clock
+		// skew in the cluster the leader can keep their lock.
+		RenewDeadline: 5 * time.Second,
+		// Time to wait between operation retries, if we get an error
+		// back.
+		RetryPeriod: 1 * time.Second,
 		Callbacks: le.LeaderCallbacks{
 			OnStartedLeading: func(stop <-chan struct{}) {
 				glog.Infof("Host %s acquiring leadership", hostname)
--- a/internal/k8s/leader.go
+++ b/internal/k8s/leader.go
@@ -14,9 +14,9 @@ import (
 )
 
 // NewLeaderElector returns a new LeaderElector used for endpoint leader election using c.
-func (c *Client) NewLeaderElector(a *arp.Announce, identity string) (*le.LeaderElector, error) {
+func (c *Client) NewLeaderElector(a *arp.Announce, lockName string, identity string) (*le.LeaderElector, error) {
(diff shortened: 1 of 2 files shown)


miekg:
Note this used twice:
~~~
	conf := rl.ResourceLockConfig{Identity: identity, EventRecorder: &noopEvent{}}
	lock, err := rl.New(rl.EndpointsResourceLock, "metallb-system", identity, c.client.CoreV1(), conf)
	if err != nil {
		return nil, err
	}
~~~
I've tried using how  in `rl.New` and that tries to create locks on node0, node1, node2, which are different. Pretty sure *one of these* needs to be the IP though.

mdlayher:
Brutal. :) LGTM though.

miekg:
I think this:
~~~
diff --git a/internal/k8s/leader.go b/internal/k8s/leader.go
index 5f7d58f..953b23e 100644
--- a/internal/k8s/leader.go
+++ b/internal/k8s/leader.go
@@ -15,7 +15,7 @@ import (
 
 // NewLeaderElector returns a new LeaderElector used for endpoint leader election using c.
 func (c *Client) NewLeaderElector(a *arp.Announce, identity string) (*le.LeaderElector, error) {
-       conf := rl.ResourceLockConfig{Identity: identity, EventRecorder: &noopEvent{}}
+       conf := rl.ResourceLockConfig{Identity: hostname, EventRecorder: &noopEvent{}}
        lock, err := rl.New(rl.EndpointsResourceLock, "metallb-system", identity, c.client.CoreV1(), conf)
        if err != nil {
                return nil, err
~~~
But I thought I tried it, trying again.

danderson:
Crap, I see what you mean, this creates 3 distinct locks. Yeah, we need to split this parameter into 2, one the lock name, the other the replica ID.

miekg:
let me fix my DNS first, and then I'll try the above patch (from my comment, not this PR)

danderson:
You're right, your diff is all that's needed. Updating the PR.

danderson:
The other problem is that, right now, we're setting the lease duration to 30min. I think that means ARP will take 30min to failover?...

miekg:
played with those timers as well (documentation is basically non-existent for this API), but that can be much lower Minute -> Second

danderson:
@miekg PTAL, added a second change that adjusts the timers and documents what they do (based on some source code comments in various places, argh terrible documentation).

miekg:
This needs rbac changes.

miekg:
on your master branch is old(er).

We now also log:
`glog.Infof("Host %s lost leadership", hostname)`

Should we use hostname or IP address for the identity?

danderson:
Why do we need RBAC changes? We're still editing the same lock as before, and speaker is allowed to edit any endpoints in the metallb-system namespace.

Good call on hostname, rebasing and changing to hostname...

miekg:
re: rbac: not sure anymore; lock name should be the same indeed.

Ok, ship it

danderson:
OK, did some more cargo culting of the timers, and switched lock identity to hostname.

Leader election is now working correctly: leader renews their lease every 5s, and if there's no renewal for 10s another pod takes over. PTAL?

miekg:
*uploading* and *downloading* containers.

miekg:
Yep, fixed:
~~~
Annotations:  control-plane.alpha.kubernetes.io/leader={"holderIdentity":"node2","leaseDurationSeconds":10,"acquireTime":"2018-01-15T19:53:37Z","renewTime":"2018-01-15T19:55:16Z","leaderTransitions":0}
~~~

PR #149, opened 2018-01-25
vendor: add github.com/mikioh/ipaddr

mdlayher:

Vendors in an IP address prefix and address walking package to replace the existing IPv4 only implementations.

Fixes # n/a

I figure this will be easier than making some of the current allocator functions also deal with IPv6.


danderson:
I'm not generally a fan of pulling in more dependencies, because of my fear of Go becoming another left-pad quagmire... But in this case, mikioh's package is high quality, and probably better tested. LGTM.

mdlayher:
Yeah, my thought process is that mikioh would probably know best about how to deal with IP addresses in Go, so might as well leverage his work. :)

PR #151, opened 2018-01-25
speaker: allow speaker to have an IPv6 address

mdlayher:

First pass at threading a bit of IPv6 goodness through the speaker.

Updates #7.

I tried to make things a bit more extensible and general, and got a head start on plumbing a bit of IPv6 addressing through.  It probably needs more work though.

### DIFF
--- a/speaker/main.go
+++ b/speaker/main.go
@@ -25,6 +25,7 @@ import (
 	"go.universe.tf/metallb/internal/arp"
 	"go.universe.tf/metallb/internal/bgp"
 	"go.universe.tf/metallb/internal/config"
+	"go.universe.tf/metallb/internal/iface"
 	"go.universe.tf/metallb/internal/k8s"
 	"go.universe.tf/metallb/internal/version"
 	"k8s.io/api/core/v1"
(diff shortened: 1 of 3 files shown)


mdlayher:
Note that IPv6 link local addresses are a little tricky because they can be the same on multiple links, so you  need to specify an interface as well.  I'll get back to this once I have a more clear picture of what this should look like.  Just looking for first thoughts now.

mdlayher:
A thought: it might make sense just to keep the "node IP" part so I can use https://github.com/google/metallb/blob/master/internal/iface/iface.go#L26 to get the interface used for NDP and the IPv6 link local address as well.  But this would all be implicit on the fact that `METALLB_NODE_IP` determines the IPv4, IPv6, and interface information for the speaker.  Thoughts?

danderson:
CircleCI is failing with a lint complaint:

speaker/main.go:92::warning: declaration of "myIP6" shadows declaration at speaker/main.go:90 (vetshadow)

mdlayher:
Ah, easy enough to fix. What do you think otherwise? Particularly the IPv6 configuration bit. Does kubernetes expose that so we can use it in manifests like IPv4?

speaker/main.go
```diff
@@ -103,31 +132,41 @@ type controller struct {
...
-func newController(myIP net.IP, myNode string, noARP bool) (*controller, error) {
-	var arpCtrl *arpController
-	if !noARP {
-		arpAnn, err := arp.New(myIP)
+type controllerConfig struct {
+	IP4    net.IP
+	IP6    net.IP
+	MyNode string
+
+	DisableARP bool
```
danderson: Please add a comment that this is for testing only. I do still want to remove it at some point.

mdlayher:
I've revised this to accommodate the fact that node IP might be an IPv4 or IPv6 address, and we should use that to infer the interface and then create the speaker configurations.

PR #194, opened 2018-03-14
WIP: Remove metallb-system namespace

### LINKED ISSUE #193: Remove hardcoded namespace - make metallb namespace-agnostic
I'd consider the hardcoded namespace a bug, making metallb independent from a configured namespace a feature.

**What happened**:
I deployed metallb in the `default` namespace.

The speakers not failing to elect a leader with the recurring error message:
```
E0313 10:35:36.630874       1 leaderelection.go:228] error initially creating leader election record: namespaces "metallb-system" not found
```

Deploy metallb in the "default" namespace, with `--config-ns default` set on both the controller and the speakers.

I'm using the helm chart of the currenlty pending PR kubernetes/charts/pull/3974 and I'm the the process of patching it. 

**Environment**:
- MetalLB version: 0.4.3
- Kubernetes version: 1.9.2
- BGP router type/version: none, using arp
- OS (e.g. from /etc/os-release): CentOS 7
- Kernel (e.g. `uname -a`): 3.10.0-693.2.2.el7.x86_64

simt2:
This is a first draft of how to remove the hardcoded  `metallb-system` namespace and make metallb namespace-agnostic.

* Adds a function `InClusterNamespace` to `internal/k8s/` to detect the namespace the pod is currently running in
* Uses above function to determine the namespace for config and leader election. 
* Removes the `--config-ns` parameter. 
* Updates the manifests in `manifests/` to use the default namespace. This is just as POC we might just keep the `metallb-namespace` in there - it does not matter with the changes anymore. I did not update the helm chart because it will probably be maintained in kubernetes/charts in the near future.

Fixes #193

~~Please elaborate on why the `--config-ns` parameter is necessary. Should the metallb controller be able to run outside of cluster? I can not think of any other scenarios. If this is not the case, it should probably be removed and default to namespace metallb is running in.~~

### DIFF
--- a/internal/k8s/k8s.go
+++ b/internal/k8s/k8s.go
@@ -3,7 +3,10 @@ package k8s // import "go.universe.tf/metallb/internal/k8s"
 import (
 	"errors"
 	"fmt"
+	"io/ioutil"
 	"net/http"
+	"os"
+	"strings"
 	"time"
 
 	"go.universe.tf/metallb/internal/config"
@@ -25,8 +28,9 @@ import (
 // Client watches a Kubernetes cluster and translates events into
 // Controller method calls.
 type Client struct {
-	client *kubernetes.Clientset
-	events record.EventRecorder
+	client    *kubernetes.Clientset
+	events    record.EventRecorder
+	namespace string
 
 	queue workqueue.RateLimitingInterface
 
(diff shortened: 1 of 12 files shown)


internal/k8s/k8s.go
```diff
@@ -435,3 +438,20 @@ func (c *Client) sync(key interface{}) error {
 		panic(fmt.Errorf("unknown key type for %#v (%T)", key, key))
 	}
 }
+
+// InClusterNamespace retrieves the namespace metallb is running in, when
+// running in-cluster.
+func InClusterNamespace() (string, error) {
```
simt2: This is based on [kubernetes/client-go](https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/client_config.go#L497). I have briefly investigated if we should build the k8s client in a way that gives access to this function but did not want to alter too much code in `internal/k8s/`. I believe that this is currently the better solution.

danderson:
Submitted something similar, with some tweaks to code layout and (not) updating the static manifests. Thank you for the PR though! You showed me the correct way of getting at the "current" namespace.

simt2:
I don't see what keeping the custom namespace in the development setup achieves except that you need to specify it when debugging via `kubectl`, but I guess it's your project.

PR #228, opened 2018-04-14
Listen on all structurally valid interfaces, regardless of addresses.

### LINKED ISSUE #226: Listen on all interfaces, even ones with no IPs
Slack discussion made me realize...

The multi-interface code currently only listens on interfaces with >=1 routable IP (rfc1918 or global scope). However, there is a use case where you would have an interface on a "public" LAN that is marked up, but has no dedicated node IP... And layer2 mode would still work fine, all it needs is an interface that is up.

So, let's just change the filter to listen on all up interfaces unconditionally.

danderson:
There is a valid use case for having interfaces that are "services-only",
i.e. are marked up, but don't have global unicast addresses of their own.
Such addresses can still participate in layer2 mode advertisements, and
it can significantly save on the number of IPs in use on the network.

This change therefore listens on all interfaces that are up and support
the features required by ARP/NDP (e.g. broadcast, multicast, existence
of a link-local address...).

Fixes #226

### DIFF
--- a/internal/layer2/announcer.go
+++ b/internal/layer2/announcer.go
@@ -58,19 +58,23 @@ func (a *Announce) updateInterfaces() error {
 			return fmt.Errorf("couldn't get addresses for %q: %s", ifi.Name, err)
 		}
 
+		if ifi.Flags&net.FlagUp == 0 {
+			continue
+		}
+		if ifi.Flags&net.FlagBroadcast != 0 {
+			keepARP[ifi.Index] = true
+		}
+
 		for _, a := range addrs {
 			ipaddr, ok := a.(*net.IPNet)
 			if !ok {
 				continue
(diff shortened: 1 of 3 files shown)


danderson:
Are you running the master image of metallb/controller as well as the speaker? There was a bug that got fixed yesterday, where the controller never allocated IPs. This seems to match your symptoms for pending IPs, so we should start by fixing that. If you are running the master build on controller, can you delete the controller pod to force it to pull a new image?

Once we have IP allocations working again, we can look at why the listeners seem to be skipping your interface. Off the top of my head, the main reason why it'd skip the interface is if it's in the "down" state (check `ip link` for link status). Even if there is no IP configured on the interface, it needs to be marked up so that it can receive packets from the network.

Issue #215, opened 2018-04-04
Custom field in BGP protocol  config


ghorofamike:
**Is this a bug report or a feature request?**:
Feature Request

**What happened**:
We would with to be able to set a password field  in the BGP configuration, needed when working with some peers:  (https://github.com/google/metallb/blob/7489989faf9a3cc47da8b55c8b60ec7a4db6f40b/internal/config/config.go#L39)[Like Vultr's BGP]  which accept a password field to authenticate a user, see the  `protocol bgp vultr`  section in documentation link above, there is a `password`  line. 

**What you expected to happen**:
We cant set the password, so we will obviously not be able to use their BGP 

**How to reproduce it (as minimally and precisely as possible)**:
Always

**Anything else we need to know?**:
Not sure if there is a downside, but we only see the upsides. 

**Environment**:
- MetalLB version: 0.5.0
- Kubernetes version: 1.9X
- BGP router type/version: OSS Bird
- OS (e.g. from /etc/os-release): Fedora
- Kernel (e.g. `uname -a`):  4.15.12

danderson:
This looks like a great feature to have.

General overview of implementation:

 - [ ] Add the field to the `Peer` struct in internal/config, and add some config parsing tests (are there any constraints on the format of the BGP password? I don't know)
 - [ ] Add the password to the `Session` struct in internal/bgp, and set the TCP_MD5 socket option in `Session.connect()` if the password is non-empty. For this, you probably need to be familiar with low-level networking and syscall interfaces in Go, because Go doesn't have an easy API to set the TCP-MD5 key.
 - [ ] Connect things up in speaker/bgp_controller.go: take the password out of the config objects and pass them into the BGP session creations.
 - [ ] Add documentation: update `manifests/example-config.yaml`, check if the website needs an update (probably minor edits, but no need for a full doc section about BGP passwords).
 - [ ] Add to release notes for 0.6.

Are you interested in contributing a PR for this? I'm happy to implement it myself, but if you want to do it, I would be happy to do code reviews and give you advice!

ghorofamike:
So i have made some attempt, not yet ready for a PR since its not functional, but see here
[code](https://github.com/ghorofamike/metallb/blob/custom_param/internal/bgp/bgp.go#L454)
i need some pointers for solving this error: 
```
E0410 12:19:51.481824       1 bgp.go:52] Connecting to "169.254.169.254:179": setting TCP_MD5 sockopt to "169.254.169.254:179": invalid argument
```
[...]

PR #236, opened 2018-04-15
This PR adds support for TCP MD5 auth in metallb.

ghorofamike:
The default net.Conn object is replaced by  low level
syscalls that create the FD's set the sockopts directly,
as TCP MD5  sockopts have to be set before the connection
is made.

This PR attempts to provide a new feature TCP MD5  BGP auth, 

Fixes # #215 

This is of course a my first interaction with GOLANG,  not perfect, but done is better than missing. 
There is code and inspiration from gopgp, expect some similarities.

### DIFF
--- a/internal/bgp/bgp.go
+++ b/internal/bgp/bgp.go
@@ -11,6 +11,10 @@ import (
 	"reflect"
 	"sync"
 	"time"
+	"syscall"
+	"unsafe"
+        "os"
+	"strconv"
 
 	"github.com/go-kit/kit/log"
 )
@@ -25,6 +29,7 @@ type Session struct {
 	peerASN  uint32
 	holdTime time.Duration
 	logger   log.Logger
+	password     string
 
 	newHoldTime chan bool
 	backoff     backoff
@@ -144,6 +149,7 @@ func (s *Session) sendUpdates() bool {
 }
 
 // connect establishes the BGP session with the peer.
+// sets TCP_MD5 sockopt if password is !="", 
 func (s *Session) connect() error {
 	s.mu.Lock()
 	defer s.mu.Unlock()
@@ -155,8 +161,21 @@ func (s *Session) connect() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 
-	var d net.Dialer
-	conn, err := d.DialContext(ctx, "tcp", s.addr)
+	laddr, err := net.ResolveTCPAddr("tcp", net.JoinHostPort("0.0.0.0", "0")) 
+	var conn net.Conn
+	d := TCPDialer{
+		Dialer: net.Dialer{
+			LocalAddr: laddr,
+			Timeout:  10*time.Second,
+		},
+		AuthPassword: s.password ,
+        }
+
+        d.Ttl = 10
+	tcphost, portstr , err:= net.SplitHostPort( s.addr )
+	port, err := strconv.Atoi(portstr)
+        conn, err = d.DialTCP(tcphost, port )
+
 	if err != nil {
 		return fmt.Errorf("dial %q: %s", s.addr, err)
 	}
@@ -284,7 +303,7 @@ func (s *Session) sendKeepalive() error {
 //
 // The session will immediately try to connect and synchronize its
 // local state with the peer.
-func New(l log.Logger, addr string, asn uint32, routerID net.IP, peerASN uint32, holdTime time.Duration) (*Session, error) {
+func New(l log.Logger, addr string, asn uint32, routerID net.IP, peerASN uint32, holdTime time.Duration, password string) (*Session, error) {
 	ret := &Session{
 		addr:        addr,
 		asn:         asn,
@@ -294,6 +313,7 @@ func New(l log.Logger, addr string, asn uint32, routerID net.IP, peerASN uint32,
 		logger:      log.With(l, "peer", addr, "localASN", asn, "peerASN", peerASN),
 		newHoldTime: make(chan bool, 1),
 		advertised:  map[string]*Advertisement{},
+		password:    password,
 	}
 	ret.cond = sync.NewCond(&ret.mu)
 	go ret.sendKeepalives()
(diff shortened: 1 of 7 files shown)


ghorofamike:
I signed it!

internal/bgp/bgp.go
```diff
@@ -427,3 +447,211 @@ func (a *Advertisement) Equal(b *Advertisement) bool {
...
+
+	sockType := syscall.SOCK_STREAM | syscall.SOCK_CLOEXEC | syscall.SOCK_NONBLOCK
+	proto := 0
+	fd, err := syscall.Socket(family, sockType, proto)
+	if err != nil {
+		return nil, err
+	}
+	fi := os.NewFile(uintptr(fd), "")
+	defer fi.Close()
+	// A new socket was created so we must close it before this
```
danderson: Put this comment block before the fi := os.NewFile. You're justifying why you wrapped the file descriptor in a file, that comment should be introducing the code that makes the change.

internal/bgp/bgp.go
```diff
@@ -427,3 +447,211 @@ func (a *Advertisement) Equal(b *Advertisement) bool {
...
+	// The TTL value to set outgoing connection.
+	TTL uint8
+}
+
+// DialTCP does the part of creating a connection manually,  including setting the
+// proper TCP MD5 options when the password is not empty. Works by manupulating
+// the low level FD's, skipping the net.Conn API as it has not hooks to set
+// the neccessary sockopts for TCP MD5.
+func (d *TCPDialer) DialTCP(tcphost string, port int) (*net.TCPConn, error) {
+	var family int
```
danderson: The convention in Go is to declare variables as close as possible to where they are used. In this case, move the declarations of `family` and `ra` to just before `if raddr.IP.To4() != nil`.

internal/bgp/bgp.go
```diff
@@ -427,3 +447,211 @@ func (a *Advertisement) Equal(b *Advertisement) bool {
...
+	var ra, la syscall.Sockaddr
+	laddr, err := net.ResolveTCPAddr("tcp", d.LocalAddr.String())
```
danderson: You can simplify this piece away: in the caller, you create an address that is "0.0.0.0:0". This just create an address that is all zero bytes. So, you can just remove this entirely, and also remove the copy() of laddr down below.

One downside is that if we ever create a dialer with initialized local addresses, we won't dial correctly... However, we don't need that today, and I'd rather aim for less/simpler code until we definitely need it (and hopefully before then, Go 1.11 will be out and we can just use the builtin pre-dial socket settings code).

internal/bgp/bgp.go
```diff
@@ -427,3 +447,211 @@ func (a *Advertisement) Equal(b *Advertisement) bool {
...
+
+//TCPDialer  represents the connection
+type TCPDialer struct {
+	net.Dialer
+
+	// MD5 authentication password.
+	AuthPassword string
+
+	// The TTL value to set outgoing connection.
+	TTL uint8
```
danderson: This option doesn't seem related to the TCP-MD5 change, and setting the TTL to a static value different from the system default can lead to surprising behavior. I would suggest just removing this, unless you have a specific need for it.

In general, I try to keep "generic" options and tweaks out of MetalLB unless someone demonstrates a specific need for it, to limit the additional complexity in the code, and the number of different combinations that have to be tested.

internal/bgp/bgp.go
```diff
@@ -155,8 +161,21 @@ func (s *Session) connect() error {
...
+	d := TCPDialer{
+		Dialer: net.Dialer{
+			LocalAddr: laddr,
+			Timeout:   10 * time.Second,
+		},
+		AuthPassword: s.password,
+	}
+
+	d.TTL = 10
+	tcphost, portstr, err := net.SplitHostPort(s.addr)
```
danderson: Move this into the DialTCP function. Convention in Go is that dial takes a string address. The dialer is the one that cares about breaking up the definition, so it should be the one doing stuff with the address.

This is especially true because, inside DialTCP, you immediately do `net.JoinHostPort` to reconstruct an address string, and pass it to net.ResolveTCPAddr... So you're doing a bunch of work for nothing here :)

internal/bgp/bgp.go
```diff
@@ -155,8 +161,21 @@ func (s *Session) connect() error {
...
+	port, err := strconv.Atoi(portstr)
+	conn, err = d.DialTCP(tcphost, port)
```
danderson: The code you replaced used `ctx` to pass in a timeout to the overall connection logic. The replacement code no longer does that, so we've lost functionality here. Can you plumb `ctx` into DialTCP, and use the deadline it provides for various timeouts?

internal/bgp/bgp.go
```diff
@@ -427,3 +447,211 @@ func (a *Advertisement) Equal(b *Advertisement) bool {
...
+	if err != nil {
+		return nil, fmt.Errorf("invalid remote address: %s ", err)
+	}
+
+	if raddr.IP.To4() != nil {
+		family = syscall.AF_INET
+		rsockaddr := &syscall.SockaddrInet4{Port: port}
+		copy(rsockaddr.Addr[:], raddr.IP.To4())
+		ra = rsockaddr
+		lsockaddr := &syscall.SockaddrInet4{}
```
danderson: Per above, you can just get rid of all the local address stuff here. The BGP dialing code doesn't need to provide local addresses, so we should just simplify the logic.

This is an internal package for MetalLB only, so we don't need to support everything that a "real" dialer would support, we just need exactly what MetalLB needs.

internal/bgp/bgp.go
```diff
@@ -427,3 +447,211 @@ func (a *Advertisement) Equal(b *Advertisement) bool {
...
+
+	// The TTL value to set outgoing connection.
+	TTL uint8
+}
+
+// DialTCP does the part of creating a connection manually,  including setting the
+// proper TCP MD5 options when the password is not empty. Works by manupulating
+// the low level FD's, skipping the net.Conn API as it has not hooks to set
+// the neccessary sockopts for TCP MD5.
+func (d *TCPDialer) DialTCP(tcphost string, port int) (*net.TCPConn, error) {
```
danderson: Return a `net.Conn` here, not a `*net.TCPConn`. The calling code just needs a generic connection interface, it doesn't need the TCP-specific implementation.

internal/bgp/bgp.go
```diff
@@ -427,3 +447,211 @@ func (a *Advertisement) Equal(b *Advertisement) bool {
...
+		if err = setsockoptIPTTL(fd, family, int(d.TTL)); err != nil {
+			return nil, err
+		}
+	}
+
+	if err = syscall.Bind(fd, la); err != nil {
+		return nil, os.NewSyscallError("bind", err)
+	}
+
+	newTCPConn := func(fi *os.File) (*net.TCPConn, error) {
```
danderson: Make this return a `net.Conn`, as explained above - and then this function is so simple that you can probably get rid of it completely. You call this function in 2 places, and in both you call `return newTCPConn(fi)`. Instead, you can just `return net.FileConn(fi)`, and you're done.

internal/bgp/bgp.go
```diff
@@ -427,3 +447,211 @@ func (a *Advertisement) Equal(b *Advertisement) bool {
...
+			return nil, errs
+		}
+		return conn.(*net.TCPConn), errs
+	}
+
+	err = syscall.Connect(fd, ra)
+	switch err {
+	case syscall.EINPROGRESS, syscall.EALREADY, syscall.EINTR:
+		// do timeout handling
+	case nil, syscall.EISCONN:
```
danderson: I don't understand this. AFAICT, that error happens only if we call Connect on an already-connected socket. This is impossible in this code. Am I missing something, does this error mean something else?

internal/bgp/bgp.go
```diff
@@ -427,3 +447,211 @@ func (a *Advertisement) Equal(b *Advertisement) bool {
...
+		return newTCPConn(fi)
+	default:
+		return nil, os.NewSyscallError("connect", err)
+	}
+
+	epfd, err := syscall.EpollCreate1(syscall.EPOLL_CLOEXEC)
```
danderson: Is this code taken from somewhere else? I'm having a hard time following the logic behind creating a poller, and what it's trying to do in general. I _think_ it's trying to handle a non-blocking connect operation, but it's hard for me to follow. Do you have references for what this code is doing?

manifests/example-config.yaml
```diff
@@ -25,6 +25,9 @@ data:
       # to the node IP address. Generally only useful when you need to peer with
       # another BGP router running on the same machine as MetalLB.
       router-id: 1.2.3.4
+      # (optional) Password for  TCPMD5  authenticated BGP sessions
```
danderson: Remove extra spaces around "TCPMD5"

ghorofamike:
I have clean out as much cruft as i could safely remove without loosing functionality, 
but there of course remains some ulyginess int he FD handling. Now about the linter
test, is there some flag in golang to ommit the warning? we cant remove any of the struct fields
i would believe.

PR #239, opened 2018-04-18
Add brackets on IPv6  (fix too many colons in address) 

valentin2105:

It add brackets if the IP is an IPv6 one. 
It's a first step to be able to go forward in IPv6. What do you think ? 

Fixes #

parts of https://github.com/google/metallb/issues/7

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -251,8 +251,8 @@ func parsePeer(p peer) (*Peer, error) {
 	if ip == nil {
 		return nil, fmt.Errorf("invalid peer IP %q", p.Addr)
 	}
-	if strings.ContainsAny(ip, ":") == true {
-		ip := fmt.Printf("[%#s]", p.Addr)
+	if strings.ContainsAny(p.Addr, ":") == true {
+		ip, _ := fmt.Printf("[%#s]", p.Addr)
 	}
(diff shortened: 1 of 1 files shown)


ghorofamike:
@valentin2105  would you mind fixing  the error thats failing in `CI`?
```
internal/config/config.go:254:24: cannot use ip (type net.IP) as type string in argument to strings.ContainsAny
internal/config/config.go:255:19: multiple-value fmt.Printf() in single-value context
```

internal/config/config.go
```diff
@@ -251,6 +251,9 @@ func parsePeer(p peer) (*Peer, error) {
 	if ip == nil {
 		return nil, fmt.Errorf("invalid peer IP %q", p.Addr)
 	}
+	if strings.ContainsAny(p.Addr, ":") == true {
```
danderson: I wouldn't put this change here. Instead, look for the call to `bgpNew` in speaker/bgp_controller.go . There is a `Sprintf` call there that builds the address, and it should be using `net.JoinHostPort` instead. That will make the stdlib add bracket as needed, depending on the address family etc.

valentin2105: Ok thanks for the tips, I will close this PR and take a time for check theses functions. 

:+1:

PR #265, opened 2018-06-21
Added a note in layer 2 mode concepts page on comparison with other a…

srampal:
…pproaches like VRRP

This PR adds a note comparing metallb's layer2 mode with other similar solutions such as VRRP and KeepAliveD. This was the first open question I had when looking at metallb for the first time and I expect others will have this question as well so would be good to add this note to make this point clear and also shed a bit more insight on the architecture of metallb. In future, we can add further architecture documentation for clarity of both layer 2 and bgp modes as appropriate so that users can quickly compare with other similar solutions.

Fixes #

This was briefly discussed on the metallb slack channel with Dave Anderson


srampal:
@danderson or others PTAL

PR #270, opened 2018-06-27
Less post processing of helm chart

plockc:
The implicit namespace required post processing when making the
manifest, made it explicit in the chart and remove some post processing

helm specific labels are conditionally hidden when
value of manifest is true

updated README to suggest the --namespace argument

Changes to the manifest Makefile target:
* builds in a single pipeline
* much more readable
* some string replacements are no longer needed

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,15 +53,18 @@ all:
 MANIFESTFILE=manifests/metallb.yaml
 .PHONY: manifest
 manifest:
-	cat manifests/namespace.yaml >$(MANIFESTFILE)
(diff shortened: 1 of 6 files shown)


danderson:
Thanks for this PR! Unfortunately, I don't think I can take it.

I originally implemented the chart->manifest generation in the way you did it here, with --set manifest=true, but that was rejected by the Helm chart maintainers when I tried to upstream it. Their reasoning was that this is never something that Helm users want to do, and furthermore it allows users to violate the helm chart style guide (which requires app, heritage, etc. labels).

The chart in this repo gets submitted verbatim to the stable charts repo on every release, so I don't want to have it deviate into something that requires tweaking when I upstream... Which means I'm stuck with that awful piece of Makefile that strips stuff out by hand :(

I really appreciate you trying to clean that up, because I really dislike that hack. Unfortunately, we can't clean it up by adding things to the Helm chart itself. If you have other ideas on how to improve this, I'd love to hear it!

The other changes you made (highlight --namespace, better linewrap in makefile) are great, and I'd be happy to accept those. Would you prefer to edit this PR, or submit a new one?

chrisplo:
I can submit a new one soon as I get a chance

PR #276, opened 2018-07-13
Remove leader election for layer2 mode.

### LINKED ISSUE #195: Get rid of leader election for layer-2 mode
Layer 2 modes (ARP and NDP) requires a single owner machine for each service IP. Currently we do this by running master election and having the winner own all IPs.

This is a little sub-optimal in several ways:

- Leader election in k8s is kinda expensive in terms of control plane qps
- Electing a machine regardless of what it's running means we are forced to use externalTrafficPolicy=Cluster, so we lose source IP information
- We cannot shard traffic load by IP (arguably this is a feature, but it's not a particularly compelling one)

So, here's a proposal: let's get rid of the leader election, and replace it with a deterministic node selection algorithm. The controller logic remains unchanged (allocates an IP). On the speaker, we would do the following:

- Based on the Endpoints object for the service, construct a list of nodes that have a pod for that service. In python pseudocode, that list would be `[x.node for x in endpoints]`
- Hash the node names and service name together, to produce a service-dependent (but deterministic) set of hashes for the nodes.
- Do a weighted alphabetical sort of the hashes, such that the first element of the list is the alphabetically first hash with the largest number of local pods
- Pick that first element as the "owner" of this service, and make it announce that IP.

This algorithm results in a couple of properties:

- Each IP can be owned by different nodes. In fact, due to the service-dependent hashing, it's likely that services will uniformly distribute throughout the cluster.
- Services will prefer to attach to nodes that have multiple serving pods for the service, to better distribute load.
- There is no explicit locking, similar to consistent hashing in Maglev each speaker simply arrives independently at the same conclusions.
- We can once again allow `externalTrafficPolicy: Local` for layer2 mode services, which removes one of the major downsides of using ARP/NDP today (no client IP visibility).
- One downside is that split-brain is more likely, because if a node gets cut off from the control plane it may not realize that conditions around it have changed, and so we might end up with multiple machines thinking that they own an IP. We can either accept this as a tradeoff (ARP and NDP behave _somewhat_ gracefully in the presence of a split brain), or keep a concept similar to leader election that just pings a lease in the cluster, and speakers who don't see that lease increasing stop all advertisement on the assumption that they've lose communications with the control plane. This however has the significant downside that the cluster will stop all announcements if the control plane goes down, rather than gracefully keep announcing the last known state. I think I would prefer just accepting the split brain in that case.

@miekg @mdlayher Thoughts on this proposal?

### LINKED ISSUE #257: Enable externalTrafficPolicy=Local for layer2 mode
**Is this a bug report or a feature request?**: Feature request

**What happened**:
Having no clue about BGP, I opted for the Layer 2 setup.
All works fine except one little detail. Layer 2 does not honor the externalTrafficPolicy (as documented). Unfortunately this is a show stopper in case of web services since you have to know the source IP for any production web service for many reasons (from simple stats to identifying abuse/hack attempts etc).  Could this be changed to use NodePort instead of ClusterIP?

Well it works as expected :)

It would be great to get the source IP on REMOTE_ADDR on the ingress controller (ingress-nginx in my case). This is possible if metallb would honor externalTrafficPolicy and the ingress-controller is deployed as a DaemonSet on all nodes running a metallb speaker. 

Metallb is configured as a loadbalancer
Ingress-nginx is configured to handle ingress
nginx pod is configured to display a phpinfo(); page.

The phpinfo page gives cluster specific IPs as $_SERVER['REMOTE_ADDR'] and $_SERVER['HTTP_X_FORWARDED_FOR']

$_SERVER['HTTP_X_FORWARDED_FOR']	10.244.0.0
$_SERVER['REMOTE_ADDR']	10.244.1.80
(Note: 10.244.1.80 is the IP of nginx-ingress-controller)

**Environment**:
- MetalLB version: 0.62
- Kubernetes version: 1.9.7
- BGP router type/version: n/a - layer 2 mode used
- OS (e.g. from /etc/os-release): Ubuntu 16.04.4 LTS
- Kernel (e.g. `uname -a`):4.4.0-122-generic

PS: Thank you for the great work!! :)

danderson:
Fixes #195, fixes #257


speaker/layer2_controller.go
```diff
@@ -25,12 +28,69 @@ import (
...
+
+		for _, ep := range subset.NotReadyAddresses {
+			if ep.NodeName == nil {
+				continue
+			}
+			usable[*ep.NodeName] = false
+		}
+	}
+
+	ret := make([]string, 0, len(usable))
```
mdlayher: Total nit, but since you may use less capacity based on the `ok` state in the loop below, would it be best to just let append do its thing and grow this as needed, eg `var ret []string`?

speaker/bgp_controller.go
```diff
@@ -80,6 +81,44 @@ newPeers:
...
+	for _, r := range ready {
+		if r {
+			// At least one fully healthy endpoint on this machine.
+			return true
+		}
+	}
+	return false
+}
+
+func (c *bgpController) ShouldAnnounce(l log.Logger, name string, svc *v1.Service, eps *v1.Endpoints) string {
```
mdlayher: Nit: I tend to prefer an integer enum for this kind of thing rather than a "magic" string, but up to you.

website/content/release-notes/_index.md
```diff
@@ -9,8 +9,25 @@ weight: 7
 
 Action required if updating from 0.6.x:
 
+- MetalLB no longer does leader election, so you can delete a number
+  of k8s resources associated with that. This is just a cleanup,
+  nothing bad happens if you leave the resources orphaned in your
+  cluster. Depending on your installation method, some of these may
+  have already been cleaned up for you.
+  - `kubectl delete -nmetallb-system endpoints metallb-speaker`
```
mdlayher: Nit: does kubectl allow flags with no spaces before the argument?

danderson: Yup, it does! :)

speaker/bgp_controller.go
danderson: Hrm. Tried this, but it just ended up making the rest of the code less clear. I agree that I don't like magic strings though, so that suggests this interface needs fixing.

speaker/layer2_controller.go
danderson: Good call, done.

danderson:
Thanks for the review! Merging once the tests become happy.

PR #313, opened 2018-09-14
Fix for issue #301

### LINKED ISSUE #301: spec.loadBalancerIP always fail for ipv6
An unnecessary `To4()` [here](https://github.com/google/metallb/blob/20b10dbb53fe0bea67453be6acdc42eb7c22109f/controller/service.go#L123) causes specifying a `loadBalancerIP` to fail always for ipv6.

metallb version: v0.7.2

uablrek:
Remove a check for ipv4 to allow also ipv6 addresses.

This PR removes a check that loadBalancerIP is ipv4.
It is needed to allow also ipv6 addresses. 

Fixes #301 

This is a very tiny fix. It is necessary for using the controller in an ipv6 k8s cluster but it is also a test of the PR procedure, especially CLA which should work now...


uablrek:
I signed it!

PR #336, opened 2018-11-09
Generate a router-id from a hash on the hostname as fallback

uablrek:

This may be necessary in a true ipv6-only cluster if there are no ipv4 addresses and no router-id is specified by configuration.

Since the router-id is generated there is a probability for collisions. This is however very small and (more important) predictable beforehand. Since the router-is's are dependent on hostname only and totally deterministic it is easy to write an offline program that with 100% accuracy can tell if there will be a collision given the hostnames as input.

See also; https://github.com/projectcalico/calico/issues/2229

A step towards fix for #7 

This PR also corrects a bug in PR [#318](https://github.com/google/metallb/pull/318), only the first address n the interface was checked. That happens to be the ipv4 address so tests pass but just out of luck.

### DIFF
--- a/internal/bgp/bgp.go
+++ b/internal/bgp/bgp.go
@@ -1,10 +1,12 @@
 package bgp // import "go.universe.tf/metallb/internal/bgp"
 
 import (
+	"bytes"
 	"context"
 	"encoding/binary"
 	"errors"
 	"fmt"
+	"hash/crc32"
 	"io"
 	"io/ioutil"
 	"net"
@@ -25,6 +27,7 @@ var errClosed = errors.New("session closed")
 type Session struct {
 	asn      uint32
 	routerID net.IP // May be nil, meaning "derive from context"
+	myNode   string
 	addr     string
 	peerASN  uint32
 	holdTime time.Duration
@@ -184,11 +187,7 @@ func (s *Session) connect() error {
 
 	routerID := s.routerID
 	if routerID == nil {
-		routerID = getRouterID(s.defaultNextHop)
-		if routerID == nil {
-			conn.Close()
-			return fmt.Errorf("cannot automatically derive router ID for IPv6 connection to %q", s.addr)
-		}
+		routerID = getRouterID(s.defaultNextHop, s.myNode)
 	}
 
 	if err = sendOpen(conn, s.asn, routerID, s.holdTime); err != nil {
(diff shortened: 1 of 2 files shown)


uablrek:
A test with hard-coded hash router-id and hostnames "vm-001" - "vm-004" gives;
```
# (on the router)
> gobgp neighbor -j | jq . | grep router
      "remote-router-id": "131.9.185.108"
      "remote-router-id": "182.172.218.133"
      "remote-router-id": "21.57.190.27"
      "remote-router-id": "57.88.176.245"
```

With the `thash` program (below) it is possible to check the router-ids;
```
> thash vm-001
Routerid:  57.88.176.245
> thash vm-002
Routerid:  131.9.185.108
```

The `thash` program;

```go
package main

import (
        "fmt"
        "net"
        "os"
        "hash/crc32"
        "bytes"
        "encoding/binary"
)

func hashRouterId(hostname string) net.IP {
        buf := new(bytes.Buffer)
        binary.Write(buf, binary.LittleEndian, crc32.ChecksumIEEE([]byte(hostname)))
        return net.IP(buf.Bytes())
}

func collisionTest(n int) {
        m := make(map[uint32]string)
        var s string
        var h uint32
        for i := 0; i < n; i++ {
                s = fmt.Sprintf("vm-%06d", i)
                h = crc32.ChecksumIEEE([]byte(s))
                if t, ok := m[h]; ok {
                        fmt.Println("COLLISION: ", s, t)
                } else {
                        m[h] = s
                }
                if i % 100000 == 0 {
                        fmt.Println("Host: ", s)
                }
        }
}

func main() {
        if len(os.Args) > 1 {
                fmt.Println("Routerid: ", hashRouterId(os.Args[1]))
        } else {
                collisionTest(1000000)
        }
}
```

njuptlzf:
Hello, I have a question about the use,

I understand that this change makes IPv6 only environment no longer provide router ID, but startup still reports an error. 

 cat /etc/kubernetes/calico_env/env
IP=none
IP6=1:2:3:4::a68

2020-06-18 12:29:05.631 [ERROR][61] resource.go 285: Error from checkcmd: "bird: /etc/calico/confd/config/.bird.cfg105155733, line 7: syntax error\n"
2020-06-18 12:29:05.631 [INFO][61] resource.go 223: Check failed, but file does not yet exist - create anyway
2020-06-18 12:29:05.632 [INFO][59] daemon.go 217: Created datastore client
2020-06-18 12:29:05.633 [ERROR][61] resource.go 285: Error from checkcmd: "bird: /etc/calico/confd/config/.bird6.cfg633115132, line 8: syntax error\n"

containers:
  - args:
    - cat /etc/calico/env/env && export $(cat /etc/calico/env/env) && start_runit
    command:
    - /bin/sh
    - -c
    env:
    - name: ETCD_ENDPOINTS
      valueFrom:
        configMapKeyRef:
          key: etcd_endpoints
          name: calico-config
    - name: CALICO_NETWORKING_BACKEND
      valueFrom:
        configMapKeyRef:
          key: calico_backend
          name: calico-config
    - name: CLUSTER_TYPE
      value: k8s,bgp
    - name: CALICO_DISABLE_FILE_LOGGING
      value: "true"
    - name: CALICO_K8S_NODE_REF
      valueFrom:
        fieldRef:
          apiVersion: v1
          fieldPath: spec.nodeName


Can't it be defaulted? How to use this function

Issue #351, opened 2018-11-28
Layer 2: Metallb shouldn't reply on interface with NOARP flag


kvaps:
**Is this a bug report or a feature request?**:

Feature request

**What happened**:

Speaker sends replies to any interface, even if arp disabled for it 

**What you expected to happen**:

If interface have NOARP flag set, speaker should ignore it.

**How to reproduce it (as minimally and precisely as possible)**:

```
ip link set dev <interface> arp off
```
Then try to arping

**Anything else we need to know?**:

Related #349

**Environment**:
- MetalLB version:
- Kubernetes version:
- BGP router type/version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):

kvaps:
@danderson , what do you thinking about using https://github.com/mickep76/netlink for identify flags on interfaces?

danderson:
I'm not a huge fan of adding a dependency, especially one that uses syscall and unsafe. Those are hard to get right, even for experienced Go programmers.

The one netlink library I trust is the one by @mdlayher . It's very well tested, and Matt's got a bunch of experience with kernel interfacing. But AFAICT, he hasn't implemented interface reading, so we'd have to add that on top of the base netlink (or genetlink? Not sure) package.

I mean, with that said, if I get a PR with anything, I'll probably merge it. If it's not using mdlayher's netlink packages, I'll just file an extra bug to remind me to change it later ;)

kvaps:
Ok, right now, I'm trying to parse noarp from `/sys/class/net/<interface>/flags` using only standard libs

danderson:
Hmm, actually, it looks like that repository is mostly just copy/pasting the linux implementation of net.Interfaces, and adding support for more flags. If we're going to just do that anyway, I would prefer it if we just copy/paste our own implementation using https://golang.org/src/net/interface_linux.go as a reference. That way we can tailor it to exactly the data we need.

But like I said, if you're sending a PR, I'll merge it if it's not hugely unsafe. Worst case I'll just file a bug to think about improving it later, but that's not a big deal.

kvaps:
HI, I have success with another netlink library: `github.com/vishvananda/netlink`. It's well tested too.
[...]

PR #350, opened 2018-11-28
Layer 2: do not listen slave interfaces

kvaps:
**What this PR does / why we need it**:

This PR disables listening on interfaces which have master

**Which issue(s) this PR fixes**:
Fixes #349

**Special notes for your reviewer**:

PR #352, opened 2018-11-28
Layer 2: do not listen slave and noarp interfaces (via netlink)

### LINKED ISSUE #349: Layer 2: MetalLB shouldn't listen on interfaces which have master

BUG

**What happened**:

My nodes have a lot of interfaces, eg. 4 hardware interfaces, 2 bonding, bridge, vlans, and veth interfaces.

When I use arping, I have 8 replies from the metallb speaker. From each interface which passing the traffic.

Eg if you have the next scheme:
```
   |       |
[eth0]   [eth1]
    \     /
    [bond0]
       |
     [br0]
```

I will get answer from all of them. That's totally wrong because connectivity will not work if ip address will be associated with interface which have master (`eth0`, `eth1` or `bond0`)

I expected answer only from `br0` interface.

**Environment**:
- MetalLB version:
- Kubernetes version:
- BGP router type/version:
- OS (e.g. from /etc/os-release):
- Kernel (e.g. `uname -a`):

kvaps:

This PR disables listening on interfaces which have master or NOARP flag set.
It also overrides previous check from #350 

**Which issue(s) this PR fixes** :

Fixes #349, #351

There is new library included: `github.com/vishvananda/netlink`. It's well tested too.
Kubernetes uses it in [ipvs proxier](https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/ipvs/netlink_linux.go) and [kubenet](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/network/kubenet/kubenet_linux.go) projects.

### DIFF
--- a/internal/layer2/announcer.go
+++ b/internal/layer2/announcer.go
@@ -6,6 +6,8 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/log"
+	"github.com/vishvananda/netlink"
+	"golang.org/x/sys/unix"
 )
 
 // Announce is used to "announce" new IPs mapped to the node's MAC address.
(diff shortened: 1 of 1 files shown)


internal/layer2/announcer.go
```diff
@@ -60,11 +61,19 @@ func (a *Announce) updateInterfaces() {
 			l.Log("op", "getAddresses", "error", err, "msg", "couldn't get addresses for interface")
 			return
 		}
+		link, err := netlink.LinkByName(ifi.Name)
+		if err != nil {
```
kvaps: Do we need this error check here?

internal/layer2/announcer.go
```diff
@@ -60,11 +61,19 @@ func (a *Announce) updateInterfaces() {
...
+			l.Log("op", "getAttrs", "error", err, "msg", "couldn't get attributes for interface")
+			return
+		}
```
kvaps: Do we need this error check here?

kvaps:
Close due #353

PR #358, opened 2018-12-04
fix broken link

igoihman:

Romana's documentation link was broken and was updated with the right address.


igoihman:
I signed it!

danderson:
Thank you!

PR #382, opened 2019-01-16
Support for PodSecurityPolicy

hatt:
Add PodSecurityPolicy to Helm chart and rendered manifest. For clusters with a PSP admissions controller, the default Capabilities will prevent the Speaker daemon set from running.

Capabilities were tested with eBFP tracing on Speaker, `NET_ADMIN` and `SYS_ADMIN` are required (on my `x86_64` CentOS 7 cluster anyway). In actual testing, the caps don't really matter because the process is running as root, they only need to be set so the policy can be selected properly. `RunAsUser` has been left as any until the image itself is able to use a non-root user. This should be tightened up when possible.

Also fixed a bug in the Makefile so `sed` on Mac works. Spacing between template blocks is still kinda weird and it's possible to export the PSP manifest independently, but I'm not sure how to keep the RBAC binding for it independent since it's an extension of the default. PSP has been enabled in the Makefile but I can disable this if changing the default isn't wanted. As far as I know, having a PSP without the controller will just ignore it though.

**Example of rendering just the PodSecurityPolicy template:**

````
$ cd helm-charts
$ helm template \
       --namespace metallb-system \
       --set psp.create=true \
       --execute templates/psp.yaml . > ../psp.yaml
````

### DIFF
--- a/helm-chart/values.yaml
+++ b/helm-chart/values.yaml
@@ -23,6 +23,11 @@ rbac:
   # create specifies whether to install and use RBAC rules.
   create: true
 
+psp:
+  # create specifies whether to install and use Pod Security Policies.
+  # This only apples to L2 mode.
+  create: true
+
 prometheus:
   # scrape annotations specifies whether to add Prometheus metric
   # auto-collection annotations to pods. See
--- a/helm-chart/values.yaml
+++ b/helm-chart/values.yaml
@@ -25,7 +25,6 @@ rbac:
 
 psp:
   # create specifies whether to install and use Pod Security Policies.
-  # This only apples to L2 mode.
   create: true
 
 prometheus:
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by hatt** [comment]
I signed it!

File: helm-chart/values.yaml [hunk 1]
```diff
@@ -23,6 +23,11 @@ rbac:
   # create specifies whether to install and use RBAC rules.
   create: true
 
+psp:
+  # create specifies whether to install and use Pod Security Policies.
+  # This only apples to L2 mode.
```
danderson: Can you remove this line? It'll just confuse the issue, most people don't know enough about the internals of MetalLB to make sense of this, it'll just create more support burden.

danderson:
Oh and I guess drop the changes to the Makefile, I replaced all that crud with a Go binary that makes the right substitutions.

helm-chart/values.yaml
hatt: No worries, removed this and also got rid of my changes to the Makefile.

PR #387, opened 2019-01-22
WIP Add end 2 end tests using kubeadm-dind-cluster

### LINKED ISSUE #192: Add e2e tests running on Virtuakube
Should set up full e2e regression tests that spin up a full cluster on GCP, using various network providers and cluster settings, and runs a series of basic functionality tests.

I finally have a working Terraform config that can stand up an IPv4 cluster using {flannel, calico, romana, weave} for networking, with a working layer-2 network that can be used to test both BGP and ARP mode.

Currently, IPv6 is still a bit borked because of https://github.com/kubernetes/kubernetes/pull/58769 , so we can't test NDP just yet (plus, calico's config would need a bunch of tweaking). But, once the framework is in place, we can go there.

aojea:

This PR adds e2e tests using containers to deploy a multinode k8s cluster and 2 external BGP routers.
The current code executes the following steps:

- Deploys a k8s cluster with 1 master and 2 workers
- Deploys 2 external BGP routers in the same network that the k8s cluster
- Installs metalLB
- Configures BGP sessions between metalLB and the external BGP routers
- Deploys a nginx service in the k8s cluster and exposes it using a LoadBalancer
- Checks that the IP assigned to the LoadBalancer has been learned by the external BGP routers

Fixes #192 

I guess we should define first:

- [ ] Test matrix against k8s, i.e. what k8s versions do we want to include
- [ ] testing framework: gingko, bash, bats, ...
- [ ] test plan, scenarios, periodic jobs, ...

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -45,8 +45,86 @@ jobs:
       - run: make ci-prepare
       - run: make gen-image-targets TAG=${CIRCLE_BRANCH:-${CIRCLE_TAG}} REGISTRY=metallb
       - run: make -f Makefile.image-targets test-bgp-router
+  e2e-dind-test-1.13:
+    working_directory: /go/src/go.universe.tf/metallb
+    docker:
+      - image: ubuntu:16.04
+    environment:
+      DOCKER_VERSION: "17.03.0-ce"
+      DIND_K8S_VERSION: v1.13
+      CNI_PLUGIN: bridge
+      DIND_CRI: docker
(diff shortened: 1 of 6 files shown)


danderson:
This is cool, but I'm afraid I'm not going to merge it. I've just added VM-based tests that allow for very rapid iteration, which I care about more than supporting environments where VMs cannot run :(

aojea:
Fair enough :+1:

PR #401, opened 2019-02-15
Remove namespace definition in test manifest.

briantopping:
This stops accidental deletion of entire namespace when yaml is deleted.

When the pre-patch yaml is deleted, it will delete the namespace of an operational cluster. Per convo w/ @danderson on Slack.

### DIFF
--- a/manifests/test-bgp-router.yaml
+++ b/manifests/test-bgp-router.yaml
@@ -1,9 +1,3 @@
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: metallb-system
-
----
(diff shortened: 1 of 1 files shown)


danderson:
Hmm, as written, this will actually break the tutorial, because we deploy this file before installing MetalLB, so it'll fail because the namespace doesn't exist. Could you instead break this out into a namespace.yaml, and update the tutorial to apply both namespace and test-bgp-router.yaml at construction, but only delete test-bgp-router.yaml ? That way we still create the NS early, but don't destroy it.

PR #415, opened 2019-03-06
Adding reconfigure on delete

### LINKED ISSUE #413: Configuration isn't reapplied after service deletion
In my cluster, we are adding and removing layer-2 address-pools fairly frequently.  Sometimes we accidentally update the configuration to remove an address pool before the last service using the pool has released it.  Since MetalLB doesn't re-read the ConfigMap after the service is deleted, we need to push the unchanged ConfigMap again to trigger a reload.

After reading the code, I *think* the right thing to do is trigger a reload of the ConfigMap with SetConfig() within the main controller after the deletion has happened.  I'm thinking around line 98

https://github.com/google/metallb/blob/e5dd238516dbc6ed39c08965ea7594a20da374c1/controller/main.go#L98

Before I tackle this and offer a PR, can someone more familiar with this codebase comment on the approach?  Is there a better place for this behavior?  Is there a better function to call?

alexlovelltroy:
My first PR attempting a one-line change to address automatic reconfiguration after deleting a Service LoadBalancer

Fixes #413 

maybe...

### DIFF
--- a/controller/main.go
+++ b/controller/main.go
@@ -98,6 +98,7 @@ func (c *controller) SetBalancer(l log.Logger, name string, svcRo *v1.Service, _
 func (c *controller) deleteBalancer(l log.Logger, name string) {
 	if c.ips.Unassign(name) {
 		l.Log("event", "serviceDeleted", "msg", "service deleted")
+		c.SetConfig(l, c.config)
 	}
 }
 


controller/main.go
```diff
@@ -98,6 +98,7 @@ func (c *controller) SetBalancer(l log.Logger, name string, svcRo *v1.Service, _
 func (c *controller) deleteBalancer(l log.Logger, name string) {
 	if c.ips.Unassign(name) {
 		l.Log("event", "serviceDeleted", "msg", "service deleted")
+		c.SetConfig(l, c.config)
```
dave-tock: Sadly, this is not going to do it. The "reprocess all other services for changes" logic happens in the caller of these functions, in `internal/k8s/`. That package contains all the k8s goop to talk to the cluster and fire these callbacks. In particular, when SetConfig returns, it returns true if the config load was successful, and the framework in internal/k8s interprets that to mean "okay I need to fire SetBalancer for all known services in case they changed".

We need to effectively do something similar for the SetBalancer calls from internal/k8s, i.e. give it a way to say "this was successful (which is the current true/false return), and also please trigger a reprocess of all other services"

Several ways to do that. One would be to make DeleteBalancer its own call in the internal/k8s interface, so that it just Knows to reprocess all services on successful delete. Another would be to make SetBalancer return two bools, (success, processAll). Success is true if the update was processed correctly (false triggers a backoff+retry), processAll is this new signal that does a similar behavior to SetConfig.

I'm okay with either of these approaches, with a _slight_ preference for making DeleteBalancer a first-class method in the internal/k8s interface. WDYT?

alexlovelltroy: Ahhh.  I wondered about the interaction between this and internal/k8s/k8s.go.

When I looked at it though, it wasn't clear to me where to add a DeleteBalancer call in there.  If I understand the flow, when a Service is deleted from kubernetes k8s/k8s.go will receive a notification as cfg.ServiceChanged.  Is that right?

danderson: Yeah, the way internal/k8s works is that it subscribes to information about a bunch of objects (all Service and Endpoints objects in the cluster). When you delete a service, the sync logic receives an event for that service name, looks it up, and the client says "no such object". At which point we call ServiceChanged with a nil object, and the code in controller+speaker interpret that as "oh the service doesn't exist any more."

look for the sync() function in internal/k8s/k8s.go, that's the meat of it. In the svcKey case, there's an `if !exists` branch. That's where you'd want to invoke this new cfg.ServiceDeleted callback, rather than cfg.ServiceChanged as it does today.

danderson:
Implemented this myself. Deleting an LB now reprocesses all known services, to try and reassign the freed IP.

Issue #422, opened 2019-04-06
Match BGP connection source address to router ID


briantopping:
### What happened: 
Peering BIRD and MLB on the same host node was possible by fiddling with the addresses that BIRD and MLB expect to see each other on. The processes were peering, but the origin for the propagated routes from MLB was somehow showing as the upstream peer to the BIRD instance, not the BIRD host itself. This caused routing loops as soon as the unaggregated MLB routes were propagated.

### What you expected to happen:
The goal was that the upstream would know what host to direct traffic at per address. But with the routing loop, there was no way to get that properly working.

### Resolution
Forthcoming small patch will set the source address of the TCP connection to the router ID specified in configuration, rather than relying on the Go resolver to provide one. With the patch, MLB connects to BIRD from the matching router ID. This apparently avoids some issue that might more properly be rejected by it for a session opening.

russellb:
The associated PR was closed over a year ago, so I assume this issue is no longer needed.

briantopping:
@russellb Problem still exists, unfortunately. I just keep manually patching on my end when I need to upgrade.

russellb:
> @russellb Problem still exists, unfortunately. I just keep manually patching on my end when I need to upgrade.

Sorry about that.  I'm closing stuff a bit quickly in a cleanup round.  Should the PR be reopened then?  Is that the patch you're carrying to make your environment work at each upgrade?

briantopping:
> Should the PR be reopened then? Is that the patch you're carrying to make your environment work at each upgrade?

Probably. I don't remember what the objections were, but I've been using it cleanly for a long time.

russellb:
> > Should the PR be reopened then? Is that the patch you're carrying to make your environment work at each upgrade?
> 
> Probably. I don't remember what the objections were, but I've been using it cleanly for a long time.

It looks like more of a misunderstanding about whether it was still needed?

briantopping:
Not really sure. I could update the PR if that's desirable.

briantopping:
Closed in deference to #670

PR #423, opened 2019-04-06
Fix for #422: originate sessions from the interface matching the router ID

briantopping:
See #422

### DIFF
--- a/internal/bgp/bgp.go
+++ b/internal/bgp/bgp.go
@@ -168,7 +168,15 @@ func (s *Session) connect() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 	deadline, _ := ctx.Deadline()
-	conn, err := dialMD5(ctx, s.addr, s.password)
+
+	laddr := "[::]:0"
+	if (s.routerID.To4() != nil) {
+		laddr = fmt.Sprintf("%s:0", s.routerID.String())
+	} else if (s.routerID.To16() != nil) {
+		laddr = fmt.Sprintf("[%s]:0", s.routerID.String())
+	}
+
+	conn, err := dialMD5(ctx, laddr, s.addr, s.password)
 	if err != nil {
 		return fmt.Errorf("dial %q: %s", s.addr, err)
 	}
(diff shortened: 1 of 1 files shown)


internal/bgp/bgp.go
```diff
@@ -168,7 +168,15 @@ func (s *Session) connect() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 	deadline, _ := ctx.Deadline()
-	conn, err := dialMD5(ctx, s.addr, s.password)
+
+	laddr := "[::]:0"
+	if (s.routerID.To4() != nil) {
+		laddr = fmt.Sprintf("%s:0", s.routerID.String())
+	} else if (s.routerID.To16() != nil) {
```
danderson: Router IDs cannot be IPv6 addresses in BGP, so I think you can remove this branch. `s.routerID != nil` should imply `To4() != nil`

internal/bgp/bgp.go
```diff
@@ -168,7 +168,15 @@ func (s *Session) connect() error {
...
+		laddr = fmt.Sprintf("[%s]:0", s.routerID.String())
```
danderson: The BGP spec doesn't require the router ID to match any local interfaces. What happens if I specify a router ID of 1.2.3.4, and that address doesn't exist locally? I think there needs to be another check here, to see if the router ID is actually configured on a local interface (`net.Interfaces()` lets you find that out).

briantopping:
Thanks @danderson, the problem arose in peering with BIRD 2.0.2 ( I didn’t test other versions) where BIRD would get extremely confused about upstream routes if the source address of the socket conflicted with another router that was using that address as a router ID. 

I filed the issue with BIRD as well (seems this is done on their users list). The response there was a mild irritation about addresses that are not on interfaces. I think there is some merit to this perspective. I spent some time with the BIRD source, but it’s Autotools-based and CLion was having none of it. It uses a bison/flex stack to execute their filter language and I simply wasn’t able to crack the parse tree executor to figure out where a fix was on that end. 

You may be correct that these additional lines are unnecessary to interoperate with vendor gear that does not run containers on the same node, but it’s a slightly more challenging situation when routes are being sourced from the same hardware that the routing protocol instance is running on. Cisco has a commodity hardware router stack, but I don’t have a license to test it. 

It may be that this patch is less concerning behind a feature flag. I’m happy to add that if you wish.

briantopping:
Was there something about the PR that was out-of-spec or was the fix applied without this source? Just curious why it was closed with no comment.

danderson:
Sorry, I was going through inactive PRs and closing them. I assumed you'd dropped this, clearly mistakenly!

Reading back, I didn't think you were blocked on anything, but it sounds like I'm reading wrong. Sorry! What do you need from me to get this submitted?

briantopping:
No worries! 

I don’t remember what I solved with it, I was just trying to figure out if there was some problem with the PR or it was already applied. Was mostly trying to help others avoid the same trap.

briantopping:
Not exactly sure, I will have to look through the branch.

PR #428, opened 2019-04-23
Remove vendor

miekg:
This removes the vendor directory and only depends on mod.go to pull
down the dependencies.

I'm not 100% what a combined go modules and vendor dir does (maybe do
you? If so just close this PR).


danderson:
go.mod + vendor just makes the build compatible with non-module-aware versions of Go, `go mod vendor` builds vendor/ based on the contents of modules.

Now that modules have been around for a while, I think it's fine to delete the vendor/ dir. Maybe someday I'll go back and do a hard history edit to drop the hundreds of MiB of crap :)

PR #429, opened 2019-04-23
Remove Trailing YAML Section Delimiter

byronmccollum:
The trailing YAML section delimiter causes some YAML parsers to fail when parsing the file. This can be seen with the Ansible K8S module specifically:

https://github.com/ansible/ansible/issues/47081#issuecomment-451982493


danderson:
This file is autogenerated from the Helm chart, so this change will get reverted next time I generate the manifest. Fixed with f1de6b2c instead.

PR #436, opened 2019-05-24
Update the ASNs to match the tutorial-1.yaml manifest

AshleyDumaine:
Was running through the tutorial for BGP mode on minikube and saw that the ASNs in the snippet were incorrect compared to the tutorial-1.yaml manifest. This should fix that.


danderson:
Thank you!

PR #446, opened 2019-07-04
Add ability to configure metrics host address.

trevex:
MetalLB exposes its metrics on every public endpoint, because the host port is used.
However it is desirable to reduce this to e.g. the internal node IP.

This PR adds the capability to specify the listener host address of the speaker metrics.

Both a flat `--host` and the environment variable `METALLB_HOST` can be used. The environment variable can then be automatically populated via:
```
        - name: METALLB_HOST
          valueFrom:
            fieldRef:
              fieldPath: status.hostIP
```

A public image we used for testing can be found here: https://hub.docker.com/r/trevex/speaker

### DIFF
--- a/internal/k8s/k8s.go
+++ b/internal/k8s/k8s.go
@@ -53,6 +53,7 @@ type Config struct {
 	ProcessName   string
 	ConfigMapName string
 	NodeName      string
+	MetricsHost   string
 	MetricsPort   int
 	ReadEndpoints bool
 	Logger        log.Logger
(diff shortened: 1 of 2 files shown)


danderson:
Thanks! Please also send a PR to edit the helm chart in this repository, so we can set this by default. I think it's fine to restrict metrics to the k8s node IP by default. Really what I would like is a way to have a pod IP just for these metrics, the speaker's privileged position makes it a bit hard...

PR #447, opened 2019-07-08
Added steps to make it work in OpenShift 4

e-minguez:
I've been able to make MetalLB work in OpenShift 4 and wrote about it here https://www.underkube.com/posts/metallb-on-ocp4-baremetal/

This PR just adds some notes on how to patch the current kubernetes manifests to make it work.

I was wondering if submit this PR or a PR to include a specific manifest file with the fixes, but I believe this is more educational.

Thanks!


danderson:
Thanks!

PR #457, opened 2019-07-24
allow ingress hostname reserved

joesonw:
Preserve hostname in Service.Status.LoadBalancer.Ingress, thus metallb can be used with custom controllers to provide a domain to it.

### DIFF
--- a/controller/service.go
+++ b/controller/service.go
@@ -26,6 +26,7 @@ import (
 
 func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service) bool {
 	var lbIP net.IP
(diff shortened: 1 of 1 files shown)


danderson:
I don't think this is correct.

The hostname field for ServiceStatus is there for cloud LB implementations that provision load-balancers by DNS name instead of by IP. If that field exists, that means there is already an LB controller in the cluster that is provisioning load-balancers. MetalLB can't run alongside another controller, they will just fight for control over the object.

If you're trying to provision domain names associated with the IP, I think you need to store that information somewhere else, the hostname field of the Service object is not semantically the right place to put that.

PR #461, opened 2019-07-28
Fixed helm chart configInline ref

rickard-von-essen:
See https://github.com/danderson/metallb/blob/master/helm-chart/values.yaml#L20


danderson:
Thank you!

PR #466, opened 2019-08-18
Add dual-stack support in the controller

### LINKED ISSUE #464: Feature request; Support for dual-stack in the controller
## Background

Support for ipv4/ipv6 dual-stack as described in the dual-stack [KEP](https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20180612-ipv4-ipv6-dual-stack.md) is well on it's way. [Phase 1](https://github.com/kubernetes/kubernetes/pull/73977) is already merged and will be in k8s 1.16. Phase 2 is partially implemented in PR https://github.com/kubernetes/kubernetes/pull/79386 which is currently under review. With this PR in place the services will have a new attribute `serviceIPFamily` which defines the ip family. Example;
```
apiVersion: v1
kind: Service
metadata:
  name: mconnect-ipv6
spec:
  ipFamily: IPv6
  selector:
    app: mconnect
  ports:
  - port: 5001
  type: LoadBalancer
```
The `serviceIPFamily` can have values (names may change after the review);
* Default
* IPv6
* IPv4

The `ClusterIP` will be assigned from the selected family.

## Feature request

The metallb `controller` must be extended to maintain pools for both ipv4 and ipv6 and assign `loadBalancerIP` depending on family of the `ClusterIP` (the `serviceIPFamily` should not be examined).

I know this issue is "ahead of time" but work can start now to be prepared when dual-stack is released in alpha. I think there are many `metallb` users that want to try dual-stack early (I am one).

I think the `controller` and `speaker` should be handled independently for dual-stack support. Also I think L2 and BGP mode in the `speaker` shall be handled independently. Dual-stack for L2 is probably not too difficult, but BGP in `metallb` does not currently support ipv6 so it will be a much harder task.

uablrek:
Fixes #464 

- [x] Add a `isIPv6` parameter to the pool Allocate functions
- [x] Check `isIPv6` and allocate from a pool with the correct family
- [x] Add test for the new function in `allocator_test.go`
- [x] Add controller tests
- [x] Investigate statistics and update if necessary https://github.com/danderson/metallb/pull/466#issuecomment-533507879

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -126,6 +126,16 @@ func TestControllerMutation(t *testing.T) {
 				AutoAssign: false,
 				CIDR:       []*net.IPNet{ipnet("3.4.5.6/32")},
 			},
+			"pool3": {
+				Protocol:   config.BGP,
+				AutoAssign: true,
+				CIDR:       []*net.IPNet{ipnet("1000::/127")},
+			},
+			"pool4": {
+				Protocol:   config.Layer2,
+				AutoAssign: false,
+				CIDR:       []*net.IPNet{ipnet("2000::1/128")},
+			},
 		},
 	}
 
@@ -402,6 +412,76 @@ func TestControllerMutation(t *testing.T) {
 					ClusterIP:             "1.2.3.4",
 				},
 				Status: statusAssigned("3.4.5.6"),
+			},
+		},
+
+		{
+			desc: "No ClusterIP",
+			in: &v1.Service{
+				Spec: v1.ServiceSpec{
+					Type: "LoadBalancer",
+				},
+			},
+			wantErr: false,
+		},
+
+		{
+			desc: "request IP from wrong ip-family (ipv4)",
+			in: &v1.Service{
+				Spec: v1.ServiceSpec{
+					Type:           "LoadBalancer",
+					LoadBalancerIP: "1.2.3.1",
+					ClusterIP:      "3000::1",
+				},
+			},
+			wantErr: true,
+		},
+
+		{
+			desc: "request IP from wrong ip-family (ipv6)",
+			in: &v1.Service{
+				Spec: v1.ServiceSpec{
+					Type:           "LoadBalancer",
+					LoadBalancerIP: "1000::",
+					ClusterIP:      "1.2.3.4",
+				},
+			},
+			wantErr: true,
+		},
+
+		{
+			desc: "IP from wrong ip-family (ipv6) assigned",
+			in: &v1.Service{
+				Spec: v1.ServiceSpec{
+					Type:      "LoadBalancer",
+					ClusterIP: "1.2.3.4",
+				},
+				Status: statusAssigned("1000::"),
+			},
+			want: &v1.Service{
+				Spec: v1.ServiceSpec{
+					Type:      "LoadBalancer",
+					ClusterIP: "1.2.3.4",
+				},
+				Status: statusAssigned("1.2.3.0"),
+			},
+		},
+
+		{
+			desc: "IP from wrong ip-family (ipv4) assigned",
+			in: &v1.Service{
+				Spec: v1.ServiceSpec{
+					Type:      "LoadBalancer",
+					ClusterIP: "3000::1",
+				},
+				Status: statusAssigned("1.2.3.0"),
+			},
+			want: &v1.Service{
+				Spec: v1.ServiceSpec{
+					Type:      "LoadBalancer",
+					ClusterIP: "3000::1",
+				},
+				Status: statusAssigned("1000::"),
 			},
 		},
 	}
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -166,12 +166,14 @@ func TestControllerMutation(t *testing.T) {
 			desc: "simple LoadBalancer",
 			in: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type: "LoadBalancer",
+					Type:      "LoadBalancer",
+					ClusterIP: "1.2.3.4",
 				},
 			},
 			want: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type: "LoadBalancer",
+					ClusterIP: "1.2.3.4",
+					Type:      "LoadBalancer",
 				},
 				Status: statusAssigned("1.2.3.0"),
 			},
@@ -181,12 +183,14 @@ func TestControllerMutation(t *testing.T) {
 			desc: "request specific IP",
 			in: &v1.Service{
 				Spec: v1.ServiceSpec{
+					ClusterIP:      "1.2.3.4",
 					Type:           "LoadBalancer",
 					LoadBalancerIP: "1.2.3.1",
 				},
 			},
 			want: &v1.Service{
 				Spec: v1.ServiceSpec{
+					ClusterIP:      "1.2.3.4",
 					Type:           "LoadBalancer",
 					LoadBalancerIP: "1.2.3.1",
 				},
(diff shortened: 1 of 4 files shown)


uablrek:
The functions can now be tested on a k8s cluster with PR https://github.com/kubernetes/kubernetes/pull/79386 applied.

#### Metallb config;
```
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: default
  name: config
data:
  config: |
    peers:
    address-pools:
    - name: default
      protocol: bgp
      addresses:
      - 10.0.0.0/28
      - 1000::/124
```
CIRDs with different families defined for the pool.

Services with different families are created;
```
apiVersion: v1
kind: Service
metadata:
  name: mconnect-ipv4
spec:
  serviceIPFamily: IPv4Service
  selector:
    app: mconnect
  ports:
  - port: 5001
  type: LoadBalancer
---
apiVersion: v1
kind: Service
metadata:
  name: mconnect-ipv6
spec:
  serviceIPFamily: IPv6Service
  selector:
    app: mconnect
  ports:
  - port: 5001
  type: LoadBalancer
```

The `controller` will assign loadBalancerIP from the right family;
```
# kubectl get svc
NAME            TYPE           CLUSTER-IP        EXTERNAL-IP   PORT(S)          AGE
coredns         ClusterIP      12.0.0.2          <none>        53/UDP,53/TCP    39s
kubernetes      ClusterIP      12.0.0.1          <none>        443/TCP          40s
mconnect-ipv4   LoadBalancer   12.0.98.75        10.0.0.0      5001:30063/TCP   9s
mconnect-ipv6   LoadBalancer   fd00:4000::d742   1000::        5001:31771/TCP   9s
```

uablrek:
Dual-stack with `metallb` in L2 mode works out-of-tne-box!

Use the (updated) manifests above (I use [mconnect](https://github.com/Nordix/mconnect) as test program), then from a cluster-external machine (but connected to the cluster sub-net) do;
```
ip ro add 10.0.0.0/28 dev eth1
ip -6 ro add 1000::/124 dev eth1
```
to tell Linux to use L2 to find the addresses on the `eth1` network, then;
```
vm-201 ~ # mconnect -address 10.0.0.0:5001 -nconn 100
Failed connects; 0
Failed reads; 0
mconnect-deployment-64554c9d9f-cjxzz 25
mconnect-deployment-64554c9d9f-qxkcg 25
mconnect-deployment-64554c9d9f-4cptb 25
mconnect-deployment-64554c9d9f-lbgsx 25
vm-201 ~ # mconnect -address [1000::]:5001 -nconn 100
Failed connects; 0
Failed reads; 0
mconnect-deployment-64554c9d9f-4cptb 25
mconnect-deployment-64554c9d9f-cjxzz 25
mconnect-deployment-64554c9d9f-qxkcg 25
mconnect-deployment-64554c9d9f-lbgsx 25
```
This is really cool :smiley:

uablrek:
About the design item;
> Investigate statistics and update if necessary

The `poolCount()` function (and others?) needs to be divided into families. This will have an impact on metrics API's and UX I guess, so I will not begin with this on my own.

@danderson Can you please take a look at this.

For ipv6 the "normal" range you get is /64, but it is possible to assign even larger ranges, then the `int64` will be insufficient. But over maxint64 the count is uninterresting and should be interpreted as "unlimited". But some logic (and test) is necessary.

controller/service.go
```diff
@@ -128,6 +128,13 @@ func (c *controller) clearServiceState(key string, svc *v1.Service) {
 }
 
 func (c *controller) allocateIP(key string, svc *v1.Service) (net.IP, error) {
+	clusterIP := net.ParseIP(svc.Spec.ClusterIP)
+	if clusterIP == nil {
+		// Note: "clusterIP: None" is not allowed for "type: LoadBalancer", so this is a fault
+		return nil, fmt.Errorf("invalid ClusterIP [%s], can't determine family", svc.Spec.ClusterIP)
```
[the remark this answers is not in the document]
uablrek: The k8s main dual-stack PR [#79386](https://github.com/kubernetes/kubernetes/pull/79386) has now been merged to "master", so dual-stack/alpha will be in 1.16 :smile: 

After reviews is was decided that the `ipFamily` is mandatory and will always be "IPv4" or "IPv6", and nothing else ("Cluster" or "Default" was removed). The field is also immutable.

So as opposed to before the `ipFamily`now shall be used to determine the service family.

But is I change this this PR **must** be built against k8s branch "master" or "release-1.16", and require k8s >=1.16.

I propose that we continue to use the family of `ClusterIP`

uablrek:
@danderson Please check https://github.com/danderson/metallb/pull/466/commits/4841c66ab866a658991584346786857bcdcdcd22

I don't know the implications of the return value. I changed true->false to make the controller unit test PASS.

controller/service.go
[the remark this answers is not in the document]
danderson: Using ClusterIP LGTM, I like universal compatibility :)

controller/service.go
```diff
@@ -128,6 +144,13 @@ func (c *controller) clearServiceState(key string, svc *v1.Service) {
 }
 
 func (c *controller) allocateIP(key string, svc *v1.Service) (net.IP, error) {
+	clusterIP := net.ParseIP(svc.Spec.ClusterIP)
+	if clusterIP == nil {
+		// (we should never get here)
```
danderson: Can you add "because the caller ensured that Spec.ClusterIP != nil" ? So that in 2 years we remember why it should never get here :)

danderson:
Code change looks good. I'm on a plane and my battery is about to die, when I land I'll patch in the change and figure out what the controller tests are mad about.

uablrek:
@danderson Updated after review

I dropped the faulty commit fo return value in convergeBalancer rather than revert it.
I squashed some test commits and corrected the comment.

uablrek:
Added a few controller tests. The ipv4 tests are not repeated for ipv6.

uablrek:
If a >62-bit mask is specified in an ipv6  pool math.MaxInt64 is returned from `poolCount()`. I think to specify a /64 ipv6 range in an pool will be common since it is the "standard" range.

To get meaningful returns from `poolCount()` users should really separate ipv6 and ipv4 pools even though it is not necessary.

This should probably be documented somewhere.

uablrek:
@danderson Please take a look at this;

The `convergeBalancer()` now returns "true" when the ClusterIP is unset or malformed;

https://github.com/Nordix/metallb/blob/c48b5a77e1ad73cfd944f3c1df465b7d298f6e84/controller/service.go#L42-L47

But in the test I had to specify `wantErr: false,`;

https://github.com/Nordix/metallb/blob/c48b5a77e1ad73cfd944f3c1df465b7d298f6e84/controller/controller_test.go#L418-L426

controller/controller_test.go
```diff
@@ -373,17 +401,89 @@ func TestControllerMutation(t *testing.T) {
...
+		{
+			desc: "request IP from wrong ip-family (ipv4)",
+			in: &v1.Service{
+				Spec: v1.ServiceSpec{
+					Type:           "LoadBalancer",
+					LoadBalancerIP: "1.2.3.1",
+					ClusterIP:      "3000::1",
+				},
+			},
+			wantErr: true,
```
danderson: This didn't make sense to me, but then I re-read the code, and wantErr means something completely different than what I thought! I thought it meant "there was a transient error, the k8s control logic should retry". But in fact it means "we logged a warning event to complain that our goal is impossible to satisfy", which is exactly correct for this state.

So, disregard my past comments about "this should not error", I was wrong :)

internal/allocator/allocator.go
```diff
@@ -303,6 +319,11 @@ func poolCount(p *config.Pool) int64 {
 	var total int64
 	for _, cidr := range p.CIDR {
 		o, b := cidr.Mask.Size()
+		if b-o >= 62 {
+			// An enormous ipv6 range is allocated which will never run out.
+			// Just return max to avoid any math errors.
+			return math.MaxInt64
```
danderson: It's funny the kind of problems we get with IPv6! :)

danderson:
Sold! Thank you so much for taking the time to make this change!

PR #467, opened 2019-08-21
Documentation: update provider list

SpencerSharkey:
I've been using MetalLB successfuly on [Vultr](https://www.vultr.com/) for the past few weeks. They [support BGP](https://www.vultr.com/features/bgp/) on all their virtual products in all locations. Setup was a breeze, it really "just worked".

I made amendments to the OVH listing as well, as I have been using MetalLB in Layer2 mode with a [vRack](https://www.ovh.com/world/solutions/vrack/) for quite some time. It requires that you purchase an IP block add it to the vRack, but after that you can use the IPs in an address pool with no problems so far.

Thanks for the hard work on MetalLB!


danderson:
Thanks!

PR #470, opened 2019-08-26
Implement MP-BGP for IPv6 announcements

Marlinc:
A basic implementation of MP-BGP for just IPv6 announcements. This allows IPv6 service addresses to be announced using BGP.

I've tried to keep the implementation as simple as possible and have only done MP-BGP for IPv6. I tried to avoid touching the IPv4 codepath.

Todo:
- [ ] IPv6 withdraw

### DIFF
--- a/internal/bgp/messages.go
+++ b/internal/bgp/messages.go
@@ -10,6 +10,15 @@ import (
 	"time"
 )
 
+const (
+	addressFamilyIpv4 uint16 = 1
+	addressFamilyIpv6 uint16 = 2
+	subsequentAddressFamilyUnicast uint8 = 1
+
+	mpReachNLRI = 14
+	mpUnreachNLRI = 15
+)
+
 func sendOpen(w io.Writer, asn uint32, routerID net.IP, holdTime time.Duration) error {
 	if routerID.To4() == nil {
 		panic("non-ipv4 address used as RouterID")
@@ -297,11 +306,18 @@ func sendUpdate(w io.Writer, asn uint32, ibgp bool, defaultNextHop net.IP, adv *
 		return err
 	}
 	l := b.Len()
+	legacyAnnouncements := []*net.IPNet{adv.Prefix}
+	if adv.Prefix.IP == nil {
+	    return nil
+	}
+	if len(adv.Prefix.IP) == net.IPv6len {
+		legacyAnnouncements = []*net.IPNet{}
+	}
 	if err := encodePathAttrs(&b, asn, ibgp, defaultNextHop, adv); err != nil {
 		return err
 	}
 	binary.BigEndian.PutUint16(b.Bytes()[21:23], uint16(b.Len()-l))
-	encodePrefixes(&b, []*net.IPNet{adv.Prefix})
+	encodePrefixes(&b, legacyAnnouncements)
 	binary.BigEndian.PutUint16(b.Bytes()[16:18], uint16(b.Len()))
 
 	if _, err := io.Copy(w, &b); err != nil {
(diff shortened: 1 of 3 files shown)


internal/bgp/messages.go
```diff
@@ -297,11 +304,18 @@ func sendUpdate(w io.Writer, asn uint32, ibgp bool, defaultNextHop net.IP, adv *
 		return err
 	}
 	l := b.Len()
+	legacyAnnouncements := []*net.IPNet{adv.Prefix}
+	if adv.Prefix.IP == nil {
+	    return nil
+	}
+	if len(adv.Prefix.IP) == net.IPv6len {
```
danderson: You need to do a bit more work here unfortunately, because IPv4 addresses can be encoded into 16 bytes in Go. So, you can't know if the IP is IPv4 or IPv6 just based on length.

Instead, you should do:

```go
isIPv6 := adv.Prefix.IP.To4() == nil
```

This tries to convert the IP to the canonical 4-byte IPv4 encoding, and returns nil if that's not possible (i.e. the address is IPv6).

Then you can use `if (isIPv6) ...` in checks.

speaker/bgp_controller.go
```diff
@@ -210,7 +210,10 @@ func (c *bgpController) syncPeers(l log.Logger) error {
 func (c *bgpController) SetBalancer(l log.Logger, name string, lbIP net.IP, pool *config.Pool) error {
 	c.svcAds[name] = nil
 	for _, adCfg := range pool.BGPAdvertisements {
-		m := net.CIDRMask(adCfg.AggregationLength, 32)
+		m := net.CIDRMask(128, 128)
+		if len(lbIP) == net.IPv4len {
```
danderson: same here, you need to check if ip.To4() != nil if you want to check for IPv4-ness.

speaker/bgp_controller.go
```diff
@@ -210,7 +210,10 @@ func (c *bgpController) syncPeers(l log.Logger) error {
...
+			m = net.CIDRMask(adCfg.AggregationLength, 32)
```
danderson: We should probably support aggregation length in IPv6 as well, my guess is that people will want to advertise a covering /64 prefix as well.

internal/bgp/messages.go
Marlinc: This should now be fixed, thank you!

Marlinc:
Hey @danderson,

I have now tested and implemented withdrawals IPv6 using MP-BGP. I have also moved over IPv4 to using MP-BPG although I don't really have the ability to test this, I don't have any IPv4 clusters. I have also modified the IPv6/IPv4 check as you suggested.

One thing I'm worried about is the unit tests, I noticed some files that direct binary data in them, probably the protocol data. It was already pretty difficult to get this code going but I'm gonna try and see how far I can get with getting the unit tests up and running.

danderson:
The changes for IPv4 announcements don't seem to be pushed? I still see legacy announcement fields passed around.

For the testdata, it's just pcaps of a BGP Open message to test decoding the peer's message, it's fine to ignore. If the code can talk to a BGP router and send routes, that's good enough.

ppmathis:
@Marlinc Do you intend to continue working on this PR? I'm very interested in using MetalLB in BGP mode with IPv4 + IPv6, especially now that dual-stack support has been released with k8s 1.17.

Marlinc:
@ppmathis unfortunately not any time soon. Might be better if someone else picks this up.

ppmathis:
@Marlinc Thank you for the work you've done so far. Could you push your most recent changes introducing IPv4 announcements using MP-BGP? You've talked about this a while ago, but it seems like these changes didn't get pushed to this PR. I'd like to take a look at it afterwards to see if I can move this PR forward.

jbrisko:
@ppmathis Have you made any plans to work on this? I also am looking to implement this into a cluster. Did you ever get the latest commits from @Marlinc? Thanks everyone for your work on this!

ppmathis:
@jbrisko I am more than happy to give this a shot and will take a closer look next week. Unfortunately I have not heard back from @Marlinc so far, so I will take the currently available code in this PR as a base if there are no updates.

Marlinc:
Hey @jbrisko and @ppmathis, sorry for the late response. I haven't had the time to look into this anymore and I ran into quite a few issues as was/is my first time implementing a binary protocol. Unfortunately priorities have shifted at work so I won't have the time to work on this in the near future. So @ppmathis go ahead, I think there will be quite a few people who will be very happy with this feature.

PR #476, opened 2019-09-06
change config name to actual default name

oskapt:
The docs point to a configmap called `config` but the helm installation looks for a configmap called `metallb-config`. Copying and pasting from the docs results in a metallb install that doesn't work.

   accepted.
   the PR is merged.

### DIFF
--- a/website/content/configuration/_index.md
+++ b/website/content/configuration/_index.md
@@ -38,7 +38,7 @@ apiVersion: v1
 kind: ConfigMap
 metadata:
   namespace: metallb-system
-  name: config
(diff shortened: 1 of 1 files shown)


danderson:
Unfortunately this is a mismatch between Helm style and common sense for non-helm users: the configmap _is_ called `config` if you don't install MetalLB via Helm, and non-Helm is the primary install method. So, this is correct, albeit I agree annoying for Helm users starting out.

danderson:
Did 5fff8b37 instead, which at least tells Helm users about the mismatch.

PR #502, opened 2019-12-05
Address Pool with node selector filtering

elliotmoso:
In our infrastructure it will be two networks (internal and public), only some nodes will have access to the public network and this scenario was not covered by metallb.

I have implemented such using node-selectors in address pools.
In some complex scenario I think this would be useful for the community.

I have added an example in manifests/example-layer2-filter.yaml
```apiVersion: v1
kind: ConfigMap
metadata:
  namespace: metallb-system
  name: config
data:
  config: |
    address-pools:
    - name: my-ip-space
      protocol: layer2
      addresses:
      - 192.168.5.240/28
      node-selectors:
      - match-labels:
          metallb.universe.tf/rack: test1
    - name: my-ip-space2
      protocol: layer2
      addresses:
      - 192.168.5.224/28
      node-selectors:
      - match-labels:
          metallb.universe.tf/rack: test2
```
This implementation is practically identical to the BGP node-selectors.

### DIFF
--- a/e2etest/manifests/mirror-server.yaml
+++ b/e2etest/manifests/mirror-server.yaml
@@ -1,5 +1,5 @@
 ---
-apiVersion: apps/v1beta2
+apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   namespace: default
--- a/speaker/main.go
+++ b/speaker/main.go
@@ -17,6 +17,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"k8s.io/apimachinery/pkg/labels"
 	"net"
 	"os"
 
@@ -62,10 +63,10 @@ func main() {
 	}
 
 	var (
-		myNode = flag.String("node-name", "", "name of this Kubernetes node")
-		host   = flag.String("host", "", "HTTP host address")
-		port   = flag.Int("port", 80, "HTTP listening port")
-		config = flag.String("config", "config", "Kubernetes ConfigMap containing MetalLB's configuration")
+		myNode  = flag.String("node-name", "", "name of this Kubernetes node")
+		host    = flag.String("host", "", "HTTP host address")
+		port    = flag.Int("port", 80, "HTTP listening port")
+		mlbconf = flag.String("config", "config", "Kubernetes ConfigMap containing MetalLB's configuration")
 	)
 	flag.Parse()
 
@@ -96,7 +97,7 @@ func main() {
 
 	client, err := k8s.New(&k8s.Config{
 		ProcessName:   "metallb-speaker",
-		ConfigMapName: *config,
+		ConfigMapName: *mlbconf,
 		NodeName:      *myNode,
 		Logger:        logger,
 
@@ -127,6 +128,7 @@ type controller struct {
 	protocols map[config.Proto]Protocol
 	announced map[string]config.Proto // service name -> protocol advertising it
 	svcIP     map[string]net.IP       // service name -> assigned IP
+	nodeLabels labels.Set
 }
 
 type controllerConfig struct {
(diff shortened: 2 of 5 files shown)


rata:
Hi! Sorry for the super late response. And thanks again for the PR.

We don't have any clear roadmap yet, but I'd like to consider instead of doing such a one-off knob for this, a more general solution like OPA (https://www.openpolicyagent.org/). That could be used, IIUC haven't actually used it, to write more flexible and fine-grained control policies, that don't need to live inside MetalLB (therefore, we don't need to add a knob for every single policy).

Would using something like OPA be possible for your use case?

russellb:
@rata 

I just came across this PR and wanted to put in my opinion.  :-)

Deploying metallb, especially in layer 2 mode, is currently very simple.  Being able to scope address pools to a subset of nodes would make metallb in this mode work better in a broader set of environments.  I also feel like a node selector is a pretty well established pattern and not too complex of an implementation.  The idea gets a +1 from me.  Exploring integration with something like OPA sounds interesting too, but I would hate to depend on it for anything that seems like a very common use case.

I haven't reviewed the actual implementation yet, but will do that next.

e2etest/manifests/mirror-server.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-apiVersion: apps/v1beta2
+apiVersion: apps/v1
```
russellb, line 2: I would at least split this into its own commit since it's not related to the feature being added.  You could also put it in its own PR so it can merge without blocking on the feature going in.

speaker/main.go
```diff
@@ -62,10 +63,10 @@ func main() {
...
 
 	var (
-		myNode = flag.String("node-name", "", "name of this Kubernetes node")
-		host   = flag.String("host", "", "HTTP host address")
-		port   = flag.Int("port", 80, "HTTP listening port")
-		config = flag.String("config", "config", "Kubernetes ConfigMap containing MetalLB's configuration")
+		myNode  = flag.String("node-name", "", "name of this Kubernetes node")
+		host    = flag.String("host", "", "HTTP host address")
+		port    = flag.Int("port", 80, "HTTP listening port")
+		mlbconf = flag.String("config", "config", "Kubernetes ConfigMap containing MetalLB's configuration")
```
russellb, line 69: why the variable name change?  I'd limit the commit to what's necessary to add the feature, and optionally put any refactoring in another commit.

speaker/main.go
```diff
@@ -314,6 +326,12 @@ func (c *controller) SetConfig(l log.Logger, cfg *config.Config) k8s.SyncState {
 }
 
 func (c *controller) SetNode(l log.Logger, node *v1.Node) k8s.SyncState {
+	nodeLabels := node.Labels
+	if nodeLabels == nil {
+		nodeLabels = map[string]string{}
+	}
+	ns := labels.Set(nodeLabels)
+	c.nodeLabels = ns
```
russellb, line 334: It looks like this is going to conflict with code in `speaker/bgp_controller.go`.  Take a look at `SetNode()` in there.

speaker/main.go
```diff
@@ -209,6 +211,16 @@ func (c *controller) SetBalancer(l log.Logger, name string, svc *v1.Service, eps
...
+	validForNode := false
+	for _,ns := range pool.NodeSelectors {
+		if ns.Matches(c.nodeLabels) {
+			validForNode = true
+		}
+	}
+	if !validForNode {
+		l.Log("op", "setBalancer", "error", "assigned IP not allowed in this node by config")
+		return c.deleteBalancer(l, name, "ipNotAllowed")
+	}
```
russellb, line 223: I see how this will do the desired Node filtering, but I think it will leave some Services orphaned with no Node at all, at least in layer 2 mode.

The algorithm used in the layer 2 controller will need to be updated to account for this.  I think it could decide that the Node that should be the leader is one not allowed by the node selector policy.  It will also need to account for the possibility that none of the endpoints are active on Nodes that are allowed for a given address pool.

russellb:
@elliotmoso I know you submitted this a long time ago.  Let me know if you're not interested in pursuing this any further and maybe I can pick it up.

speaker/main.go
elliotmoso: This was because config variable conflicts with the imported "go.universe.tf/metallb/internal/config" library.
I'll commit this in annother PR

elliotmoso:
Hi @russellb, The code i wrote doesn't work quite well. I'll have to redo it again. You can close it. I'll  commit a new PR

russellb:
> Hi @russellb, The code i wrote doesn't work quite well. I'll have to redo it again. You can close it. I'll commit a new PR

Thanks for the follow-up!

Issue #458, opened 2019-07-25
Use policy/v1beta1 API space for the PodSecurityPolicy


danderson:
https://kubernetes.io/blog/2019/07/18/api-deprecations-in-1-16/

PR #503, opened 2019-12-10
Fix the apiGroup of PodSecurityPolicy in Roles

masa213f:
MetalLB seems already ready for k8s 1.16.
But I think, it is better to update the API group of PodSecurityPolicy in roles.

ref #458


champtar:
@daxmc99 can you also look at this one

PR #516, opened 2020-01-06
modify kustomization.yaml

binoue:
Metallb is now using `kustomize` to generate Kubernetes resource files since [v0.8.3](https://metallb.universe.tf/release-notes/#version-0-8-3)

I noticed that in the manifest directory both `namespace.yaml` and `metallb.yaml` contain the same namespace resource, but `metallb.yaml` does not need the namespace resource if `kustomization.yaml` merge both files.

It does not only reduce codes but also beneficial for the users who want to use `metallb` resources in a different namespace. 
(They do not need to edit the upsteam yaml file and only need to edit `kustomization.yaml`)

Therefore, this pull request made the following changes.
- delete namespace resource from `metallb.yaml`
- modify `kustomization.yaml` to  merge `metallb.yaml` and `namespace.yaml`

### DIFF
--- a/manifests/kustomization.yaml
+++ b/manifests/kustomization.yaml
@@ -1,2 +1,3 @@
 resources:
   - metallb.yaml
+  - namespace.yaml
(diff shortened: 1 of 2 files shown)


danderson:
Thanks!

Issue #552, opened 2020-03-22
Make dev environment work


johananl:
Running `inv dev-env` doesn't seem to work on current `main` (e82a77d4). A couple of recent changes seem to have broken the dev environment:

- In 514091a2 the k8s namespace resource got moved to a kustomization and is no longer created in `manifests/metallb.yaml`.
- In #527, HashiCorp memberlist support was added, which requires a secret to be present in the `metallb-system` namespace in order for speakers to converge successfully.

In addition, looks like the environment assumes the developer uses kind `v0.5.1` or earlier, which is a pre-release. The latest kind version at time of writing is `v0.7.0`. Among other improvements, this version simplifies kubeconfig handling, which allows us to remove some code from `tasks.py`. We may want to update the environment to use the latest kind.

I'll open a PR to handle the above.

rata:
Thanks! :)

johananl:
https://github.com/metallb/metallb/pull/553 open.

I guess we aren't running `inv dev-env` in the CI, right? Maybe we should to avoid this kind of breaks in the future.

danderson:
`inv dev-env` requires an environment that can run Docker, which is rare in CIs. I think currently on CircleCI we run in containers, so docker-in-docker doesn't work. I think CircleCI supports VMs for free, but it'll make all CI much slower.

Sounds like a good idea to test this, I'm just not sure how to do it well.

PR #553, opened 2020-03-22
Update dev environment

johananl:
Fixes https://github.com/metallb/metallb/issues/552.

This PR updates the dev environment following recent changes to the MetalLB code and k8s manifests. It also updates `tasks.yaml` for kind `v0.7.0`.

To test the PR, ensure you have kind `v0.7.0` installed, then run `inv dev-env`.

P.S. Should we mention the kind dependency in the docs somewhere? I had to guess the required version in order to get `inv dev-env` to work...

### DIFF
--- a/e2etest/manifests/mirror-server.yaml
+++ b/e2etest/manifests/mirror-server.yaml
@@ -1,5 +1,5 @@
 ---
-apiVersion: apps/v1beta2
+apiVersion: apps/v1
 kind: DaemonSet
 metadata:
   namespace: default
--- a/e2etest/manifests/mirror-server.yaml
+++ b/e2etest/manifests/mirror-server.yaml
@@ -17,7 +17,7 @@ spec:
     spec:
       containers:
       - name: mirror
-        image: metallb/mirror-server:master
+        image: metallb/mirror-server:main
         env:
         - name: NODE_NAME
           valueFrom:
--- a/tasks.py
+++ b/tasks.py
@@ -1,7 +1,8 @@
 import glob
-import os.path
+import os
 import re
 import semver
+import shutil
 import sys
 import yaml
 import tempfile
@@ -196,14 +197,22 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None):
     run("kind load docker-image --name={} metallb/mirror-server:dev-{}".format(name, architecture), echo=True)
 
     run("kubectl delete po -nmetallb-system --all", echo=True)
-    with open("manifests/metallb.yaml") as f:
-        manifest = f.read()
-    manifest = manifest.replace(":main", ":dev-{}".format(architecture))
-    manifest = manifest.replace("imagePullPolicy: Always", "imagePullPolicy: Never")
-    with tempfile.NamedTemporaryFile() as tmp:
-        tmp.write(manifest.encode("utf-8"))
-        tmp.flush()
-        run("kubectl apply -f {}".format(tmp.name), echo=True)
+
+    manifests_dir = os.getcwd() + "/manifests"
+    with tempfile.TemporaryDirectory() as tmpdir:
+        # Copy namespace manifest.
+        shutil.copy(manifests_dir + "/namespace.yaml", tmpdir)
+
+        with open(manifests_dir + "/metallb.yaml") as f:
+            manifest = f.read()
+        manifest = manifest.replace(":main", ":dev-{}".format(architecture))
+        manifest = manifest.replace("imagePullPolicy: Always", "imagePullPolicy: Never")
+        with open(tmpdir + "/metallb.yaml", "w") as f:
+            f.write(manifest)
+            f.flush()
+
+        run("kubectl apply -f {}/namespace.yaml".format(tmpdir), echo=True)
+        run("kubectl apply -f {}/metallb.yaml".format(tmpdir), echo=True)
 
     with open("e2etest/manifests/mirror-server.yaml") as f:
         manifest = f.read()
--- a/tasks.py
+++ b/tasks.py
@@ -211,7 +211,22 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None):
             f.write(manifest)
             f.flush()
 
+        # Create memberlist secret.
+        secret = """---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: memberlist
+  namespace: metallb-system
+stringData:
+  secretkey: verysecurelol"""
+
+        with open(tmpdir + "/secret.yaml", "w") as f:
+            f.write(secret)
+            f.flush()
+
         run("kubectl apply -f {}/namespace.yaml".format(tmpdir), echo=True)
+        run("kubectl apply -f {}/secret.yaml".format(tmpdir), echo=True)
         run("kubectl apply -f {}/metallb.yaml".format(tmpdir), echo=True)
 
     with open("e2etest/manifests/mirror-server.yaml") as f:
--- a/tasks.py
+++ b/tasks.py
@@ -187,10 +187,8 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None):
             tmp.flush()
             run("kind create cluster --name={} --config={}".format(name, tmp.name), pty=True, echo=True)
 
-    config = run("kind get kubeconfig-path --name={}".format(name), hide=True).stdout.strip()
-    env = {"KUBECONFIG": config}
     if mk_cluster and cni:
-        run("kubectl apply -f e2etest/manifests/{}.yaml".format(cni), echo=True, env=env)
+        run("kubectl apply -f e2etest/manifests/{}.yaml".format(cni), echo=True)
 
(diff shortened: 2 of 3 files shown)


tasks.py
```diff
@@ -196,14 +197,21 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None):
...
+
+        with open(tmpdir + "/metallb.yaml") as f:
+            manifest = f.read()
+        manifest = manifest.replace(":main", ":dev-{}".format(architecture))
+        manifest = manifest.replace("imagePullPolicy: Always", "imagePullPolicy: Never")
+        with open(tmpdir + "/metallb.yaml", "w") as f:
+            f.write(manifest)
+            f.flush()
+
+        run("kubectl apply -k {}".format(tmpdir), echo=True)
```
rata: Just curious (no strong opinion if using kustomize or not): why use kustomize instead of `kubectl apply -f manifests/namespace.yaml` first and then the metallb.yaml?

I mean, both should do the same now and I'm not sure we want to kustomize something in the dev env via kustomize. What was your reasoning for using kustomize instead of the applying the two yamls (`namespace.yaml` and `metallb.yaml`) present in the manifest folder?

tasks.py
```diff
@@ -211,7 +211,21 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None):
...
             f.flush()
 
+        # Create memberlist secret.
+        secret = """apiVersion: v1
+kind: Secret
+metadata:
+  name: memberlist
+  namespace: metallb-system
+data:
+  secretkey: dmVyeXNlY3VyZWxvbAo="""
```
rata: he, lol. Does it make sense to use plaintext for the secret? Seems easier in case you want to debug something

e2etest/manifests/mirror-server.yaml
```diff
@@ -17,7 +17,7 @@ spec:
     spec:
       containers:
       - name: mirror
-        image: metallb/mirror-server:master
+        image: metallb/mirror-server:main
```
rata: Not sure I follow the commit message here (just in case, as GH really doesn't help with this is the commit `s/master/main/g in mirror-server.yaml`).

Sorry if it's a trivial question (haven't tried the inv script yet), but does it generate the `metallb/mirror-server:main` image locally and, with this patch, it doesn't reach for dockerhub to get it? Because the `metallb/mirror-server:main` surely exist on dockerhub

tasks.py
johananl: I don't see a big difference honestly. I'm fine with switching to `stringData` if you find this better.

e2etest/manifests/mirror-server.yaml
johananl: This is fallout from some changes done a while ago following switching from `master` to `main`. The commits which introduced the changes are mentioned in my commit message. What I've tried to do here is to get the dev environment working again by adding a missing rename of `master` to `main` in `mirror-server.yaml`.

`inv dev-env` builds the Go code, generates Docker images for the controller and the speaker, uploads the images to the [kind](https://github.com/kubernetes-sigs/kind/) cluster and applies the manifests while setting `imagePullPolicy` to `Never`. This ensures you always run your dev version on the dev environment.

tasks.py
johananl: I've switched to Kustomize following 514091a2. I'm not sure about the reasoning behind this commit since its message doesn't explain the change, however - yes, we can definitely do `kubectl apply -f` on the relevant manifests only.

My *personal* rationale for this refactor was: assuming the `manifests/` dir keeps changing, it's more generic to operate on the entire dir rather than on individual manifests. Using Kustomize it's safe to do so since only the manifests explicitly mentioned in `kustomizations.yaml` are applied.

Ideally we probably want to deploy MetalLB to the dev environment as similarly as possible to the way users deploy on real clusters. I am not sure what the canonical way for deploying MetalLB is since the [docs](https://metallb.universe.tf/installation/) mention both a Kustomize and a non-Kustomize options. In any case I don't have a strong opinion here besides saying we should have a working dev environment :-)

johananl:
>Just some simple questions, mostly LGTM (I'd like to test this locally before merging, though, or if someone else tests this too).

If you have kind `v0.7.0` installed as well as [Invoke](https://www.pyinvoke.org/), it should be as simple as running `inv dev-env`.

>Also, maybe we can improve docs to mention kind and some brief words on how to set this up? There is a contributing section, maybe some more words there can help others in the future?

Agree 100%, I couldn't find a complete documentation for the dev workflow. I can add a section to the docs in this PR if that makes sense.

tasks.py
johananl: Actually the rationale for 514091a is explained in the PRs description: https://github.com/metallb/metallb/pull/516. Sorry :-|

e2etest/manifests/mirror-server.yaml
rata: Ohh, thanks for the explanation! :)

tasks.py
rata: I slightly would prefer that, yes. Seems easier to debug and to read with the `"""` ending

tasks.py
rata: Good point, I agree. And using kustomize may be simpler to maintain (although I doubt the manifest will change much... but who knows), then :ok_hand:

tasks.py
johananl: Done. I've also tried to make the YAML snippet more readable by starting the document with `---` so that the first line doesn't look out of place.

rata: Nice :)

johananl:
I've improved the dev docs a bit to mention kind and explain how to deploy to it.

tasks.py
```diff
@@ -187,40 +188,53 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None):
...
-    manifest = manifest.replace("imagePullPolicy: Always", "imagePullPolicy: Never")
-    with tempfile.NamedTemporaryFile() as tmp:
-        tmp.write(manifest.encode("utf-8"))
-        tmp.flush()
-        run("kubectl apply -f {}".format(tmp.name), echo=True, env=env)
+
+    manifests_dir = os.getcwd() + "/manifests"
+    with tempfile.TemporaryDirectory() as tmpdir:
+        # Copy manifests dir to temp dir.
+        shutil.copytree(manifests_dir, tmpdir, dirs_exist_ok=True)
```
rata: `dirs_exist_ok` was introduced in python 3.8 (for example, python 3.8 was uploaded to debain unstable at the end of Nov 2019). The reference [here](https://docs.python.org/3/library/shutil.html) says it was introduced in 3.8

My local env fails because I'm using python 3.7, for example. I'd prefer to not force such a new python version for this to work. Also, taking into account commit 19cb437d12399e67d9b870b0675fd640f3ed0975 that added a workaround for python2/python3, it seems that this is an effort that we are already doing.

Can you please remove that argument and find a workaround that works with older python versions? :-) Not sure how old, I trust your judgement, but not "the latests python release only" would be great :)

website/content/community/_index.md
```diff
@@ -105,9 +105,14 @@ To develop MetalLB, you'll need a couple of pieces of software:
   tool)
 - [Docker](https://www.docker.com/docker-community), the container
   running system
+- [kind](https://github.com/kubernetes-sigs/kind), a lightweight Kubernetes cluster running in Docker
 - [Kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/), the Kubernetes commandline interface
 - [Invoke](https://www.pyinvoke.org) to drive the build system
 
+>NOTE: The development environment was tested with **kind `v0.7.0`**. Older
+>versions may not work since there have been breaking changes between minor
+>versions.
```
rata: mental note: hopefully now as the `kind` command we run are rather simple, they will work with newer releases too. So, for now, this note seems fine and no extra complications (like checks in the code, etc.) needed :-)

e2etest/manifests/mirror-server.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-apiVersion: apps/v1beta2
+apiVersion: apps/v1
```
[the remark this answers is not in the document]
johananl: Thanks. I actually wasn't sure about this one, and got lost in the k8s changelog trying to figure out exactly what happened :grimacing: I'll rephrase.

tasks.py
johananl: Thanks. Indeed I'm using Python 3.8 and wasn't aware this is a new flag. I'll fix this.

johananl: Eventually I've opted for applying plain manifests (namespace, secret and MetalLB) rather than an entire directory. The reason is - this seems to be the main way for deploying MetalLB according to the [docs](https://metallb.universe.tf/installation/).

johananl: I've verified the current code works with the following Python versions:

- `3.8.2`
- `3.7.4`
- `3.5.6`

Python 2 is EOL so I don't think we should worry about it.

Issue #561, opened 2020-03-28
[Question] How to avoid service disruptions in L2 mode


liuyuan10:
Currently in L2 mode, master is selected by sorting the hashing of node+service name so that each speaker can do this by themselves. One downside I can see is that when a node is added, it may change  master of some services (even all is possible) causing service disruptions. What's worse is when a node is restarted, it causes two service disruptions in a row for those with masters on the node.

So I'm wondering if we can have sticky master selection in some way. One straightforward way is to do master selection at a centralized place (like the controller) which can stick a service to a node. The reason not doing so must have been clear but some quick answers are appreciated.

rata:
Just curious, how much time does this disruption take and using which OS? From https://metallb.universe.tf/concepts/layer2/:

> All modern versions of major OSes (Windows, Mac, Linux) implement layer 2 failover correctly, so the only situation where issues may happen is with older or less common OSes.

Just to understand the impact of the current behavior. Of course, improving is always good and thanks for noticing this use case that can be improved :-)

liuyuan10:
Here I'm talking about existing TCP connections got disrupted because connection state is missing on a new master. Gateway doesn't play a role here. When a new node shows up, some of services may be migrated to this node but existing TCP connection states are not (kept in conntrack in most cases). If we can have sticky master election, a new node will not causing such migration

champtar:
Grafana team implemented a small KV store on top of Memberlist: https://grafana.com/blog/2020/03/25/how-were-using-gossip-to-improve-cortex-and-loki-availability/
That might be a way to exchange info and have "sticky" nodes

guoh1988:
@liuyuan10 
你好 
在k8s service ExternalTrafficPolicy =Local 下LB IP会将流量固定一个节点上的pod，当这个pod重启或者退出，会将流量切到其他节点上的pod，但是当这个节点上pod起来还是会将流量切回来，多了一次不必要的中断
如192.168.1.105/192.168.1.106两个udp测试pod程序，发现在k8s service ExternalTrafficPolicy =Local模式下，流量一般都是从105上走，当105上的程序退出或者重启，流量会切到106上去，但当105上的pod起来还会切到105，这个中间多了一次中断
在local模式下的lb的主节点好像默认选择105，不太清楚当中的算法

russellb:
It looks like this discussion has been quiet for a good while now, so I'm going to close this out.  #metallb on kubernetes slack is another good place to discuss further.  Thanks!

Issue #5, opened 2017-11-28
Do you use MetalLB? Tell us!


danderson:
This is not an issue so much as a lightweight way of gathering information on who is using MetalLB. This is mostly to satisfy our curiosity, but might also help us decide how to evolve the project.

So, if you use MetalLB for something, please chime in here and tell us more!

rsanders:
We're not using it yet, but our K8S clusters are 100% metallic, running in our data centers and on customer premises. We're going to be trying this out in the new year.

Our deployment would in most cases have this peer directly with our routers. We're a mostly Cisco shop. Which BGP implementations have you tested with?

danderson:
So far, I've only tested with OSS BGP implementations, mainly [BIRD](http://bird.network.cz). I used to have some Juniper and Cisco hardware that could speak BGP, but I got rid of them :(.

With that said, currently MetalLB speaks a very conventional, "has existed for ever" dialect of BGP, so it should work fine. I foresee two possible issues interacting with Cisco's implementation:
- Some old Cisco devices use an unusual encoding for capability advertisements in the OPEN message. AFAICT, they stopped doing this in the late 1990s, so it shouldn't be a problem. If it is a problem, it's a trivial patch to fix it.
- Cisco _may_ refuse to speak to a BGPv4 speaker that uses the original route encoding for IPv4 (which the RFCs say is still 100% correct), and instead requires all route advertisements to use the MP-BGP encoding. I'm in the middle of making MetalLB speak MP-BGP for dual-stack support anyway, so hopefully by the time you test MetalLB this should not be a problem.

If you hit any interop problems, please file bugs! Ideally with a pcap that captures the initial exchange of BGP OPEN messages and a few UPDATE messages, so that I can add regression tests.

xnaveira:
A colleague brought my attention to this project recently. We're running bare metal k8s on bgp enabled hosts and pods so this looks very interesting. Unfortunately we are using 32-bit AS numbers in our setup so we can't test it right away but it looks promising!

danderson:
@xnaveira MetalLB supports 32-bit AS numbers! It's the one BGP extension that's implemented in v0.1.0. v0.2.0 will also support MP-BGP, so if you use Quagga or Cisco devices, you'll want to wait for 0.2.0.

xnaveira:
That is awesome @danderson , i'll give it a try asap then.

tangjoe:
[...]

PR #562, opened 2020-03-29
[RFC] Fix conflicting ARP when IP is shared

### LINKED ISSUE #558: Multiple services with same IP announced multiple times
When using two services (TCP+UDP) with the `Local` traffic policy, the same shared IP and the same set of pods (as [described in the documentation](https://metallb.universe.tf/usage/#ip-address-sharing)), MetalLB in Layer 2 mode will sometimes announce the LB IP from two different nodes.

A workaround is setting`externalIPs` on one of the services to the shared virtual IP, and using MetalLB on just one of the services.

This was previously reported in https://github.com/metallb/metallb/issues/530#issuecomment-599134715.

liuyuan10:
This is just a RFC to show the idea. Not tested yet.

When IP is shared by two services, it may be annonced from different
nodes causing conflicting arp responses.

This commit fixes by using service IP in the hash instead of service
name so that service sharing the same IP will have the same master.

For traffic cluster services, all nodes should be usable instead of
those running pods, so that services sharing IPs have the same set of
usable nodes.

Fixed   #558

### DIFF
--- a/speaker/layer2_controller.go
+++ b/speaker/layer2_controller.go
@@ -24,7 +24,7 @@ import (
 	"github.com/hashicorp/memberlist"
 	"go.universe.tf/metallb/internal/config"
 	"go.universe.tf/metallb/internal/layer2"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 )
 
 type layer2Controller struct {
@@ -39,7 +39,7 @@ func (c *layer2Controller) SetConfig(log.Logger, *config.Config) error {
 
 // usableNodes returns all nodes that have at least one fully ready
 // endpoint on them.
-func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
+func usableNodes(isTrafficLocal bool, eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
 	var activeNodes map[string]bool
 	if mList != nil {
 		activeNodes = map[string]bool{}
@@ -49,20 +49,25 @@ func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
 	}
 
 	usable := map[string]bool{}
-	for _, subset := range eps.Subsets {
-		for _, ep := range subset.Addresses {
-			if ep.NodeName == nil {
-				continue
-			}
-			if activeNodes != nil {
-				if _, ok := activeNodes[*ep.NodeName]; !ok {
+	if isTrafficLocal {
+		for _, subset := range eps.Subsets {
+			for _, ep := range subset.Addresses {
+				if ep.NodeName == nil {
 					continue
 				}
-			}
-			if _, ok := usable[*ep.NodeName]; !ok {
-				usable[*ep.NodeName] = true
+				if activeNodes != nil {
+					if _, ok := activeNodes[*ep.NodeName]; !ok {
+						continue
+					}
+				}
+				if _, ok := usable[*ep.NodeName]; !ok {
+					usable[*ep.NodeName] = true
+				}
 			}
 		}
+	} else {
+		// For non traffic local services, all nodes are usable.
+		usable = activeNodes
 	}
 
 	var ret []string
@@ -75,14 +80,14 @@ func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
 	return ret
 }
 
-func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, svc *v1.Service, eps *v1.Endpoints) string {
-	nodes := usableNodes(eps, c.mList)
-	// Sort the slice by the hash of node + service name. This
-	// produces an ordering of ready nodes that is unique to this
-	// service.
+func (c *layer2Controller) ShouldAnnounce(l log.Logger, svc *v1.Service, eps *v1.Endpoints, lbIP net.IP) string {
+	nodes := usableNodes(svc.Spec.ExternalTrafficPolicy == v1.ServiceExternalTrafficPolicyTypeLocal, eps, c.mList)
+	// Sort the slice by the hash of node + service IP. This
+	// produces an ordering of ready nodes that is unique to the IP
+	// so that all services sharing the same IP are annonced from the same node.
 	sort.Slice(nodes, func(i, j int) bool {
-		hi := sha256.Sum256([]byte(nodes[i] + "#" + name))
-		hj := sha256.Sum256([]byte(nodes[j] + "#" + name))
+		hi := sha256.Sum256([]byte(nodes[i] + "#" + lbIP.String()))
+		hj := sha256.Sum256([]byte(nodes[j] + "#" + lbIP.String()))
 
 		return bytes.Compare(hi[:], hj[:]) < 0
 	})
--- a/speaker/main.go
+++ b/speaker/main.go
@@ -320,7 +320,7 @@ func (c *controller) SetBalancer(l gokitlog.Logger, name string, svc *v1.Service
 		return c.deleteBalancer(l, name, "internalError")
 	}
 
-	if deleteReason := handler.ShouldAnnounce(l, name, svc, eps); deleteReason != "" {
+	if deleteReason := handler.ShouldAnnounce(l, svc, eps, lbIP); deleteReason != "" {
 		return c.deleteBalancer(l, name, deleteReason)
 	}
 
@@ -423,7 +423,7 @@ func (c *controller) SetNode(l gokitlog.Logger, node *v1.Node) k8s.SyncState {
 // A Protocol can advertise an IP address.
 type Protocol interface {
 	SetConfig(gokitlog.Logger, *config.Config) error
-	ShouldAnnounce(gokitlog.Logger, string, *v1.Service, *v1.Endpoints) string
+	ShouldAnnounce(gokitlog.Logger, *v1.Service, *v1.Endpoints, net.IP) string
 	SetBalancer(gokitlog.Logger, string, net.IP, *config.Pool) error
 	DeleteBalancer(gokitlog.Logger, string, string) error
 	SetNode(gokitlog.Logger, *v1.Node) error
--- a/speaker/bgp_controller.go
+++ b/speaker/bgp_controller.go
@@ -25,7 +25,7 @@ import (
 
 	"go.universe.tf/metallb/internal/bgp"
 	"go.universe.tf/metallb/internal/config"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/go-kit/kit/log"
@@ -136,7 +136,7 @@ func healthyEndpointExists(eps *v1.Endpoints) bool {
 	return false
 }
 
-func (c *bgpController) ShouldAnnounce(l log.Logger, name string, svc *v1.Service, eps *v1.Endpoints) string {
+func (c *bgpController) ShouldAnnounce(l log.Logger, svc *v1.Service, eps *v1.Endpoints, _ net.IP) string {
 	// Should we advertise?
 	// Yes, if externalTrafficPolicy is
 	//  Cluster && any healthy endpoint exists


champtar:
Now that I think about it a bit more, we have 2 cases:
1) 1 set of pods with both tcp and udp, that want to have both behind the same IP
2) N set of pods that want to share the IP
Case 1 is pretty easy, just need to use `metallb.universe.tf/allow-shared-ip` or the IP instead of the name.
Case 2 we must have:
- ExternalTrafficPolicy == Cluster (or an affinity so that the pods run together)
- all the nodes where pods are running from the N sets of pods (all nodes also works)

I prefer to always send the traffic to a node that has a pod, but this is not mandatory

liuyuan10:
If we want to keep sending traffic to a node with a pod, then we have to calculate nodes with all N set of pods running. Also we need to make sure when assigning IP, the services must have at least one common node to be able to share the IP. All of those makes the implementation complex. On the other side, saving one hop for 1/(num_of_pods) traffic is not critical at least to me. Just two my cents.

speaker/layer2_controller.go
```diff
@@ -49,20 +49,25 @@ func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
...
+					}
+				}
+				if _, ok := usable[*ep.NodeName]; !ok {
+					usable[*ep.NodeName] = true
+				}
 			}
 		}
+	} else {
+		// For non traffic local services, all nodes are usable.
+		usable = activeNodes
```
rata: Hmm, if the speaker daemonset is deployed with a `nodeSelector:` then this can also fail? Let me elaborate:

Let's say 4 nodes cluster (A, B, C and D nodes), speaker ds is on nodes A and B (due to a `nodeSelector:` in the speaker ds yaml added by the user) but with this new order, nodes C and D are also a candidate. If they are elected, but the speaker doesn't run on it, it will not announce. If the reasoning is correct, though, this problem existed before the PR too (for example, not overlaping speaker pods and svc pods due to node selectors). But just trying to see ways this might fail :-)

I'm not familiar with the code and haven't even used L2 before, so this is all very theoretical. I might be missing something very obvious, sorry if that is the case :)

liuyuan10: The activeNodes list is built from memberlist active nodes so it will only include nodes with metalLB running. I don't see it's an issue.

speaker/main.go
```diff
@@ -320,7 +320,7 @@ func (c *controller) SetBalancer(l gokitlog.Logger, name string, svc *v1.Service
 		return c.deleteBalancer(l, name, "internalError")
 	}
 
-	if deleteReason := handler.ShouldAnnounce(l, name, svc, eps); deleteReason != "" {
+	if deleteReason := handler.ShouldAnnounce(l, svc, eps, lbIP); deleteReason != "" {
```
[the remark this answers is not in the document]
liuyuan10: I don't care about where to get the IP. from the svc object is also fine.

liuyuan10:
> What do you think?

I'm using this RFC to present my idea and what's more important is to see if it fixes the issue without breaking others. I've tested it briefly and it's working for me. I can't come up with a case to break things so please do let me know.

Overall I think the issue is a must fix otherwise, we can't claim IP sharing is supported in L2 mode (for both traffic local and cluster services).

speaker/layer2_controller.go
rata: Ohh, that seems right. Thanks for noticing! :)

rata:
Thanks again for the PR and the time, really :-)

I fully agree, we must fix this.

I'm with little time this week, though, to look in detail into this PR. If @champtar can have a look, it will be great. I will allocate time during next weekend or next week to review this.

Sorry for the lack of time now and thanks again :)

champtar:
@liuyuan10 @rata I'll try to have a careful look this week

rata:
Ok, so I'll try to have a look this week. I'd like to avoid changing the semantics heavily (as small as the patch can be, the semantics are heavily impacted by such a change) several ways in a row, though. Because we might as well need to change them for https://github.com/metallb/metallb/issues/561, I'd like to take that into account when doing this change, if possible.

If we can manage this change to be "obviously right" or something and reduce impact on users (i.e. users not using the shared ip thing), it might be ok to change it now and later for the linked issue. But, I think these two issues (this bug the PR tries to address and the other one) leans towards in a direction to improve more heavily how this is done.

Thanks again for the patch, will have a look in a few days :-)

liuyuan10:
Yes. This patch has a behavior change (master on node with pod vs any node). I anyway will have to apply this patch in our own fork because we may not run metallb on every node (it's too troublesome to move all nodes into the same L2 domain). So only allowing master with pod doesn't make sense in our case. Also I think there is use case where metallb only runs on linux nodes while other nodes are windows. It won't work without this patch. That's why I personally like the new behavior more than the old one.

rata:
@liuyuan10 yeah, exactly in that regard this is a big change, and I'll have to ask you for some patience (sorry!) as I am new in the maintainer role and want to start small, being sure I understand in detail the changes I merge and their impact.

However, real-world issues without known workaround are very important too. Just to be clear, what is the issue you just mentioned? Is it that you have a cluster with windows and linux nodes, will use MetalLB to expose pods, but that can't work today using layer 2 because you are not able to expose services for pods that only run on windows? (Because MetalLB uses nodes where the pod is running to expose the IP, and that of course won't happen as the pod run on windows).

And a curious question: can I ask where are you using MetalLB? Is it a home cluster, work, etc.? Feel free to share your usage here too: https://github.com/metallb/metallb/issues/5 :)

liuyuan10:
I'm just doing evaluations based on my use case. I don't use window so
that's only one of the use case I can see doesn't work. In my use case, I
can't put all nodes into a single L2 so I need only run metallab in part of
the cluster which doesn't work as described.

rata:
Oh, okay. That makes perfect sense too. I guess your use case is only Layer 2, right? And can you confirm also that you don't mind about the source IP the pods sees? as if using kube-proxy in iptables mode is used and pods are not running in the local node, pods won't see the client IP but the node IP as source IP.

liuyuan10:
Yeah. that's the downside I know of but I can live with that.

champtar:
I started to try to fix the issues, and ended up with the same fix as @liuyuan10 :)
Here the changes:
1) using IP instead of service name:
default master node will change but will still be collocated with a pod, so I'm fine with it
we might be affected by #584 while updating from older version !
2) using all the active nodes when `externalTrafficPolicy: Cluster`:
this is a behavior change for L2, but it make sense to me, as with it IP sharing for different selector works out of the box, and also having speakers on only part of the cluster.
looking at the code it's what BGP is doing.

if we really want to have all the options available, we can add an annotation like `metallb.universe.tf/balancer-affinity: preferNodeWithPods` to use with `externalTrafficPolicy: Cluster`
An important point is that it makes memberlist mandatory (not the case right now)

@rata in this order, I would vote to have #584 fixed, make memberlist mandatory, have this PR cleanup and merged, then add `metallb.universe.tf/balancer-affinity: preferNodeWithPods`

russellb:
> I started to try to fix the issues, and ended up with the same fix as @liuyuan10 :)
> Here the changes:
> 
>     1. using IP instead of service name:
>        default master node will change but will still be collocated with a pod, so I'm fine with it
>        we might be affected by #584 while updating from older version !
> 
>     2. using all the active nodes when `externalTrafficPolicy: Cluster`:
>        this is a behavior change for L2, but it make sense to me, as with it IP sharing for different selector works out of the box, and also having speakers on only part of the cluster.
>        looking at the code it's what BGP is doing.
> 
> 
> if we really want to have all the options available, we can add an annotation like `metallb.universe.tf/balancer-affinity: preferNodeWithPods` to use with `externalTrafficPolicy: Cluster`
> An important point is that it makes memberlist mandatory (not the case right now)
> 
> @rata in this order, I would vote to have #584 fixed, make memberlist mandatory, have this PR cleanup and merged, then add `metallb.universe.tf/balancer-affinity: preferNodeWithPods`

This plan makes sense to me in general.  I wonder why an annotation would be needed for `preferNodeWithPods`?  How about just make that the way it always works with `externalTrafficPolicy: Cluster`?  Is there a downside to doing that?

liuyuan10:
> This plan makes sense to me in general. I wonder why an annotation would be needed for preferNodeWithPods? How about just make that the way it always works with externalTrafficPolicy: Cluster? Is there a downside to doing that?

@russellb I think what you mean is basically #613. but that doesn't help solving the issue of ip sharing. If we want to have that preference while still fixing ip sharing, how about we only enable that preference when annotation "metallb.universe.tf/allow-shared-ip" is not present for externalTrafficPolicy: Cluster services?

russellb:
> > This plan makes sense to me in general. I wonder why an annotation would be needed for preferNodeWithPods? How about just make that the way it always works with externalTrafficPolicy: Cluster? Is there a downside to doing that?
> 
> @russellb I think what you mean is basically #613. but that doesn't help solving the issue of ip sharing. If we want to have that preference while still fixing ip sharing, how about we only enable that preference when annotation "metallb.universe.tf/allow-shared-ip" is not present for externalTrafficPolicy: Cluster services?

I was thinking that we would always allow shared IPs and always try to land the IP on a Node where at least one Endpoint for one of the Services resides.

Now that I think about this again, I actually really like simplicity of just using all active Nodes when the traffic policy is `cluster`.

speaker/layer2_controller.go
```diff
@@ -75,14 +80,14 @@ func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
...
+func (c *layer2Controller) ShouldAnnounce(l log.Logger, svc *v1.Service, eps *v1.Endpoints, lbIP net.IP) string {
+	nodes := usableNodes(svc.Spec.ExternalTrafficPolicy == v1.ServiceExternalTrafficPolicyTypeLocal, eps, c.mList)
+	// Sort the slice by the hash of node + service IP. This
+	// produces an ordering of ready nodes that is unique to the IP
+	// so that all services sharing the same IP are annonced from the same node.
 	sort.Slice(nodes, func(i, j int) bool {
-		hi := sha256.Sum256([]byte(nodes[i] + "#" + name))
-		hj := sha256.Sum256([]byte(nodes[j] + "#" + name))
+		hi := sha256.Sum256([]byte(nodes[i] + "#" + lbIP.String()))
+		hj := sha256.Sum256([]byte(nodes[j] + "#" + lbIP.String()))
```
russellb, line 90: We should probably add a release note to let people know that the algorithm changes and will result in many services having their IPs move.  This could cause some brief downtime.

I also imagine this sort of change should only come in a new `Y` release of `X.Y.Z`.

@rata thoughts?

speaker/main.go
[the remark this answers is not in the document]
russellb: @liuyuan10 want to follow up on this suggestion?

speaker/layer2_controller.go
russellb: sounds like this thread can be marked resolved?

liuyuan10:
> Now that I think about this again, I actually really like simplicity of just using all active Nodes when the traffic policy is cluster.

This is great. I like the simplicity too. However, the current change I have here only works for if memberlist is enabled. There are two choices:
1. We change this PR to only fix the issue when using memberlist and make it clear that the bug is only fixed. This will introduce significant behavior change between two modes
2. wait for #662. once that's done, it should be easy to get a list of speaker nodes as active node even memberlist is not used.

I prefer the 2nd choice. what do you think @russellb

russellb:
> > Now that I think about this again, I actually really like simplicity of just using all active Nodes when the traffic policy is cluster.
> 
> This is great. I like the simplicity too. However, the current change I have here only works for if memberlist is enabled. There are two choices:
> 
>     1. We change this PR to only fix the issue when using memberlist and make it clear that the bug is only fixed. This will introduce significant behavior change between two modes
> 
>     2. wait for #662. once that's done, it should be easy to get a list of speaker nodes as active node even memberlist is not used.
> 
> 
> I prefer the 2nd choice. what do you think @russellb

Waiting makes sense.  If we know significant changes are coming in this area, let's let those settle first and then build around it.

rata:
Fully agree, let's revisit this once the other memberlist re-join PR is merged. Thanks!

liuyuan10:
@rata now with all the other issues fixed, I'd like to pick this up. while the change is straightforward, one issue is that the speakerlist doesn't generate the list of nodes in non-ml mode. To achieve the same in non-ml mode, I need to make the speakerlist also work (passing namespace/labels in that mode too). Does this sound good reasonable?

champtar:
@liuyuan10 have a look at #595, it has the needed SpeakerList improvements

rata:
@liuyuan10 Hi! Sorry, I'm away for a few days. @johananl will be more active :)

liuyuan10:
@champtar Yes. your PR is the one I want. guess I will wait for that one.

kvaps:
Hi @liuyuan10 thanks for your PR, I compiled two images with your fix:

```
docker.io/kvaps/metallb-controller:a3047c4d
docker.io/kvaps/metallb-speaker:a3047c4d
```

during the update I had 5 seconds downtime, but now it is working as a charm

liuyuan10:
@kvaps Thanks for verifying. I think the 5 second downtime is not introduced b this change. metallb anyway will cause service disruption during update and 5 second matches what memberlist offers.

kvaps:
Hi, is there any opportunity to get it merged?

shaggykp:
Any idea when this fix will get merged and available in a release?  We are experiencing a similar issue and really like to get a release with this fix.

liuyuan10:
@rata @johananl @daxmc99  looks like more and more people are asking for this. As I said above, it's best to have #595 so that we don't have different implementation based on if memberlist is enabled or not. Any thoughts on how to move it forward?

shaggykp:
Any update on when this being available?  Thanks in advance.

ehlo550:
Hi, 
I am sorry but is there any kind of information if and when this is probalby going to make it into a release?
We are using metallb and are very thankful for this project but currently this is a huge pain for me.

kvaps:
ping @rata?

kvaps:
The problem still persist on v0.10.2

edgsantos:
confirming that @kvaps code fixes the issue for us. It is possible to address the concern of the scope of the change and merge to master?
By scope i mean only being applied when the allow-shared-ip annotation is present

Issue #560, opened 2020-03-28
Speakers do not respond to ARP requests in L2 mode


cmcarey:
I appreciate that troubleshooting specific system configurations is something that belongs in the #metallb slack channel rather than here, however I have found no assistance for a few days there, so as a last resort I must ask here.  I would appreciate even tips on where to look to debug this myself.

Each node in my cluster has an ethernet interface and an address in 10.0.0.0/24, with netmask /16.  Pod CIDR is 10.30.0.0/16, and cluster networking uses Calico.  MetalLB is configured in L2 mode with address pool 10.0.1.1-10.0.1.254.

I have an ingress-nginx service of type LoadBalancer.  MetalLB correctly assigns the IP 10.0.1.1 to it.  When I run `arping` from another device in the network, I see no arp response packets, though I have confirmed every node _does_ receive the arp request packet.  I see nothing indicative in the controller/speaker logs, aside from it confirming that the arp responder was successfully created for the correct interfaces.

I am running MetalLB v0.9.3.  My K8s cluster is on v1.18.0.  The cluster was built using kubeadm, which from what I can tell runs kube-proxy in iptables mode.  Each cluster node is Ubuntu 18.04 LTS.

Again I appreciate this is not the correct avenue for asking for help with this, so I mainly ask the places I should look to figure out what is going wrong - perhaps some checklist could be added to the troubleshooting page on the website.

champtar:
Hi @cmcarey 
Can you "describe" the LoadBalancer object, to confirm on which node it's attributed.
Then please run both tcpdump, to confirm the arp request comes in, and strace the metallb speaker process to see if anything is happening. Maybe you have some firewalling blocking the packets ?

cmcarey:
Thanks for the prompt response @champtar !
Here is the result of `kubectl -n ingress-nginx describe svc ingress-nginx-internal`:

```
Name:                     ingress-nginx-internal
Namespace:                ingress-nginx
Labels:                   app.kubernetes.io/name=ingress-nginx-internal
Annotations:              kubectl.kubernetes.io/last-applied-configuration:
                            {"apiVersion":"v1","kind":"Service","metadata":{"annotations":{},"labels":{"app.kubernetes.io/name":"ingress-nginx-internal"},"name":"ingr...
Selector:                 app.kubernetes.io/name=ingress-nginx-internal
Type:                     LoadBalancer
IP:                       10.100.191.20
LoadBalancer Ingress:     10.0.1.1
[...]

PR #563, opened 2020-03-29
Website: Add extra troubleshooting steps to the website

cmcarey:
I recently ran into an issue where I had a LoadBalancer service that had pending endpoints.  MetalLB was not responding to any ARP requests, but I couldn't see where the issue was (I didn't notice that the endpoints were pending until someone in the Slack channel pointed it out).

This commit indicates that while troubleshooting, you should check that the LoadBalance service has endpoints that are not pending.

Hopefully it will save someone the 10+ hours I spent debugging this issue (and committing the cardinal sin of asking for help in [issues](https://github.com/metallb/metallb/issues/560)) until someone on the Slack channel figured out where I was going wrong!

### DIFF
--- a/website/content/configuration/troubleshooting.md
+++ b/website/content/configuration/troubleshooting.md
@@ -5,14 +5,17 @@ weight: 20
 
 Once you setup a loadbalancer on Kubernetes and it doesn't _seem_ to work right away, here are a few tips to troubleshoot.
 
-SSH into one or more of your nodes and use `arping` and `tcpdump` to verify the ARP requests pass through your network.
+- Check that the LoadBalancer service has endpoints (`kubectl -n <namespace> get endpoints <service>`) that are not pending - if they are, MetalLB will not respond to ARP requests for that service's external IP
+- SSH into one or more of your nodes and use `arping` and `tcpdump` to verify the ARP requests pass through your network
+
+---
 
 Below assumes you setup a loadbalancer with an IP of `192.168.1.240` and then we show usage of the commands and successful
 requests going back and forth.
 
 TL;DR - We found out that the IP `192.168.1.240` is located at the mac `FA:16:3E:5A:39:4C`.
 
-If these requests fail, your network stack does not allow ARP to pass, therefor - we won't know about MetalLB's
+If these requests fail, your network stack does not allow ARP to pass, therefore - we won't know about MetalLB's
 assignment of the IP and can't connect.
 
 ## Tools
@@ -35,7 +38,6 @@ Received 4 response(s)
 ```
 
 `192.168.1.35` is the IP of one of the worker nodes and part of the same subnet.
-
 
 ### tcpdump
 


website/content/configuration/troubleshooting.md
```diff
@@ -5,14 +5,17 @@ weight: 20
 
 Once you setup a loadbalancer on Kubernetes and it doesn't _seem_ to work right away, here are a few tips to troubleshoot.
 
-SSH into one or more of your nodes and use `arping` and `tcpdump` to verify the ARP requests pass through your network.
+- Check that the LoadBalancer service has endpoints (`kubectl get endpoints`) that are not pending - if they are, MetalLB will not respond to ARP requests for that service's external IP
```
rata: That command will probably run fine on most clusters, but just not for the desired service. Can you make it obvious that the service (and maybe ns too) needs to be specified? Like `kubectl -n <namespace> get endpoints <service>`?

rata: I think something common to do, also, is to check that the nodePort works as expected (to rule out other miscommunications, and it would catch also this). What do you think of adding that too?

website/content/configuration/troubleshooting.md
```diff
@@ -5,14 +5,17 @@ weight: 20
...
+- SSH into one or more of your nodes and use `arping` and `tcpdump` to verify the ARP requests pass through your network
+
```
rata: note to self: I see an extra line here. I'd need to check, just in case, if this doesn't break the hugo formatting

website/content/configuration/troubleshooting.md
```diff
@@ -36,7 +39,6 @@ Received 4 response(s)
 
 `192.168.1.35` is the IP of one of the worker nodes and part of the same subnet.
 
-
```
rata: note for me: same here

rata:
Friendly ping? :)

website/content/configuration/troubleshooting.md
cmcarey: Ah, so manually checking that the service works?

website/content/configuration/troubleshooting.md
cmcarey: I added a `---` (hr) after to break up the content a bit, from my experience Markdown likes an extra line between disparate elements, seemed to generate properly when I tested the changes with Hugo.

cmcarey: image

rata: Thanks!

website/content/configuration/troubleshooting.md
rata: Yes, in my little experience on slack users channel, sometimes there is another unrelated issue that would have been catched if the service nodePort was tested (note all services type LoadBalancer are also NodePort) 

Do you feel comfortable to add a step to try that too? If not, I can do it.

cmcarey: Hm, as I've not run into that I'm not sure what the best way to document it would be, sorry.

cmcarey:
:)  Sorry for the slow responses, have been a little caught up in other matters recently.

website/content/configuration/troubleshooting.md
rata: No problem, created an issue to do this later: https://github.com/metallb/metallb/issues/576 then :)

PR #565, opened 2020-03-29
Additional logging when no endpoints defined

### LINKED ISSUE #564: Additional logging could be useful when a service has no endpoints defined
I think there could be some benefit to adding a log line to the sync method inside of k8s.go

In that method there is some processing done on the endpoints retrieved from Kubernetes. In the case of an error retrieving the key a log line is printed however if the key is successfully retrieved but has no available endpoints there are no logs printed that indicate that there was is an issue that would prevent metallb from advertising the load balancer ip that was assigned to the service. 

The actual issue this log would help identify is a configuration error in the service definition where the selector does not match a label in the deployment. When this happens a kubectl get all will show the service with an external ip defined but metallb doesn't respond to arp requests. 
The issue is apparent if you run kubectl get endpoints however as someone who is trying to learn Kubernetes this wasn't something I was familiar with at the time. With the log line I'm suggesting it could provide a hint at where to look for someone like myself. 

Since I resolved my issue I encountered someone else on slack within a few days that ran into the same issue that I did. 

My suggestion is to add a log line inside the if !exists block in the sync method. I did some testing by deploying my custom speaker and as far as I can tell this doesn't print off any logs when the service is defined correctly. 

I can submit a pull request if this sound like a reasonable thing to add. 

James

Matoch:
This will log an error when a service doesn't have any valid endpoints. 

Fixed #564

### DIFF
--- a/internal/k8s/k8s.go
+++ b/internal/k8s/k8s.go
@@ -358,6 +358,7 @@ func (c *Client) sync(key interface{}) SyncState {
 				return SyncStateError
 			}
 			if !exists {
+				l.Log("op", "getEndpoints", "error", "No endpoints for service found", "msg", "run kubectl get endpoints to review service endpoints for service")
 				return c.serviceChanged(l, string(k), nil, nil)
 			}
 			eps = epsIntf.(*v1.Endpoints)


internal/k8s/k8s.go
```diff
@@ -358,6 +358,7 @@ func (c *Client) sync(key interface{}) SyncState {
 				return SyncStateError
 			}
 			if !exists {
+				l.Log("op", "getEndpoints", "error", err, "msg", "no endpoints defined")
```
rata: `err` is nil in this case, right?

Have you run this in your setup? What is the output it shows?

Matoch: Yes err is nil. 

This is the log line that it provides.
`{"caller":"k8s.go:361","error":null,"msg":"no endpoints defined","op":"getEndpoints","service":"default/registry","ts":"2020-03-29T16:11:00.865053157Z"}`

rata: Why print the error when there is no error? That seems buggy to me :)

Matoch: Now I understand why you were asking about err being nil. I can walk you through my thought process. I'm not really sure what else it could/should be. 

Even though metallb doesn't use log levels to determine how verbose logging should be they are typically ALL, TRACE, DEBUG, INFO, WARN, ERROR, FATAL

I can eliminate trace, debug, info, and fatal right off the bat as they just don't make sense for an error message that is relevant to normal operation. 

Warn vs Error could both be reasonable options. When I think of warning it's something that could be wrong but probably doesn't affect normal operation. Error is an indication of a problem that should be dealt with. It's not going to crash but probably isn't going to work the way you expect it to. 

Error just feels like the right answer to me. I'm expecting metallb to respond to the arp requests for the service I created. However under this condition it's not going to do that. 

If not error what do you think would be a more appropriate way to log the information?

Thanks,

James

rata: Sorry I wasn't clear, wasn't talking about that. I mean why print `error: null`. We *know* it is nil, there is no point in printing.

In particular, this part of the code `"error", err,` is what I'm talking about

Matoch: Ah ok. That did not register for me at all. I think having the word error in there is important as it will draw users to it. The logging is essentially key/value pairs so I don't think we can have nothing there so 2 potential options come to mind.
1. I could write some static text there. e.g error, "No endpoints defined" and drop the msg. 
2. I could write some static text there e.g error, "No edmpoints defined" and add a different message with more detail msg, "run kubectl get endpoints to validate service endpoints"

What do you think?

rata: I think a short description and msg seems reasonable :)

Matoch: Ok. I've made another commit for the change. The log now looks like this.
{"caller":"k8s.go:361","error":"No endpoints found","msg":"run kubectl get endpoints to validate that service endpoints exist","op":"getEndpoints","service":"default/registry","ts":"2020-04-01T15:06:47.810393689Z"}

rata: Sorry, thinking about it a little more, what about as msg: `No endpoints for service found`?

champtar: @Matoch maybe also talk about verifying labels and selectors ?

Matoch: @rata Ok. I adjusted the wording again. It will now say error:No endpoints for service found and msg:run kubectl get endpoints to review service endpoints for service.

@champtar I can add that but I wasn't sure of all the possible causes for why endpoints may not exist. My thought was to give them enough info to know that the issue exists outside of metallb and then they can troubleshoot on their own. 

For both of you. I don't do pull requests often. If I read this correctly my additional commits do exist in the same pull request. Can you confirm if the updated wording is showing up for you?

rata: I'd not give troubleshooting or debugging instructions here. I'd just point that no endpoints are available, with the message I suggested: `No endpoints for service found`.

And I'm on the fence if we should do this (as this is really k8s info, not metallb), but I guess we can try to be helpful.

Then, in the troubleshooting guides we can explain more (it can be pods not ready, selector mismatch, etc.) if we want to. But, this is really a kubernetes thing, not very related to metallb, so I think I'd link to this: https://kubernetes.io/docs/tasks/debug-application-cluster/debug-service/

Matoch: @rata I'm certainly fine with keeping it simple. For me I finally figured out my issue as soon as I figured out that there were no endpoints. That gave me enough info to go look up what I needed to verify there was an issue with my service and move forward.

Matoch:
@rata I've never actually tried to squash commits before so I'm not confident I've accomplished what you asked me to do. Commit 2ee84de is my attempt to do it.

rata:
@Matoch no problem, that is not a squash but we can fix that later :) ([Here](https://stackoverflow.com/questions/5189560/squash-my-last-x-commits-together-using-git) seems to be a stackoverflow question for that, in case you want to have a look

internal/k8s/k8s.go
rata: I looked at it today. I found two concerns for now:

 * The patch doesn't work as intended: created a simple service with selector not matching anything and the line is not printed. I used `inv dev-env`, see https://metallb.universe.tf/community/#required-software for software needed to run. I tried using a configmap with just a layer2 IP address
 * It seems that the logic is added in the generic watcher code (shared by controller and speaker) and that the cache key doesn't exist doesn't mean that there are no ready endpoints

Can you please elaborate, so I can reproduce myself, how are you testing this? I might be missing something (starting by bgp/layer2 mode you are using, etc.)

The `ShouldAnnounce()` and it's caller might be a better candidate for this. The function already returns with a reason, and it is aware of the endpoints for the service. Although the reason is not logged (it might be too verbose) now.

Maybe the shared watcher code where you added it can be a better place, I'm not very familiar with it yet, but the code doesn't seem to work at all for a simple test.

Am I missing something?

Matoch: These are the steps I followed to test the code.

cd metallb/speaker
GOOS=linux GOARCH=amd64 go build
cp speaker ~/tmp/metallb-speaker/
cd ~/tmp/metallb-speaker

cat Dockerfile 
FROM metallb/speaker:v0.8.2
COPY speaker /

docker build -t matoch/metallb-speaker2 . && docker push matoch/metallb-speaker2
ssh to where I can run kubectl commands.

wget https://raw.githubusercontent.com/google/metallb/v0.9.3/manifests/metallb.yaml
# replace speaker image with matoch/metallb-speaker2
kubectl apply -f metallb.yaml

# My registry deployment uses the labels run: registry
# My config for the registry loadbalancer has a match label of app: registry just so that the match fails.
# After metallb is running and I load the registries load balancer config I get the following in my logs

{"caller":"k8s.go:361","error":"No endpoints for service found","msg":"run kubectl get endpoints to review service endpoints for service","op":"getEndpoints","service":"default/registry","ts":"2020-04-05T21:01:30.298333756Z"}


Now I'm not sure if the order of operations matters. I generally teardown my registry configs before redeploying metallb with the custom speaker.

 I'll take a look at the inv dev-env. I've only tested it on my cluster.

James

rata: Thanks a lot for the clarification! However, testing it again, such a message will be shown for all services without endpoint, not just services with type load balancer. 

So, if you create a service type clusterip it will show a message. Probably we don't want that, and not at that layer of abstraction. As I mentioned, that is the sync code shared for speaker and controller, that is why you probably see that entry before the `start of service update` one.

I'll paste another two lines diff, but in the general discussion (this is becoming too long and difficult to follow :))

rata:
@Matoch hey! Thanks again for the contribution, as mentioned [here](https://github.com/metallb/metallb/pull/565#discussion_r407774852), what do you think of a diff like the following?

```
$ git diff
index 2209b563..53866f0e 100644
@@ -77,6 +77,10 @@ func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
 
 func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, svc *v1.Service, eps *v1.Endpoints) string {
        nodes := usableNodes(eps, c.mList)
+       if len(nodes) == 0 {
+               return "noEndpoints"
+       }
+
        // Sort the slice by the hash of node + service name. This
        // produces an ordering of ready nodes that is unique to this
        // service.
diff --git a/speaker/main.go b/speaker/main.go
index 63850910..f526e28f 100644
--- a/speaker/main.go
+++ b/speaker/main.go
@@ -321,6 +321,7 @@ func (c *controller) SetBalancer(l gokitlog.Logger, name string, svc *v1.Service
        }
 
        if deleteReason := handler.ShouldAnnounce(l, name, svc, eps); deleteReason != "" {
+               l.Log("event", "serviceNotAnnounced", "msg", deleteReason)
                return c.deleteBalancer(l, name, deleteReason)
        }
 ```

If you look at the bgp code, it already returns a [similar reason](https://github.com/metallb/metallb/blob/9d6de1ee1b6d69a9424c4638c0fea81854b41dbb/speaker/bgp_controller.go#L145-L149) and what this patch does is just add it to layer2 and log it on the higher layer calling those functions.

I'm still not familiar with the code enough to be confident we want that at that layer of abstraction (and if that is the right place to put the log entry, as there are many others, and unsure about the log line itself), but it seems it might be okay. Will need to investigate a little more, sorry :).

In my testing, also, this shows that line between the `start of service update` and `end of service update` (as opposed to your patch that shows it just before it, for the reasons I mentioned about the abstraction layer [here](https://github.com/metallb/metallb/pull/565#discussion_r407774852))

Does such a patch works for you?

Matoch:
@rata I've tested that patch and it seems to work then way I would expect it to. I think it says everything that it needs to as well. 

`{"caller":"main.go:324","event":"serviceNotAnnounced","ip":"10.201.0.150","msg":"noEndpoints","pool":"default","protocol":"layer2","service":"default/registry","ts":"2020-04-13T23:57:09.083451581Z"}`

What do we do from here?

rata:
@Matoch thanks!

I'll ask you to have some patience, if possible, as I am still onboarding in the mainternership role. 

What I'd like to do before merge, is familiarize more with the code to be sure I understand the implications (side effects, like other cases that might spam the logs by adding that line, if that can be done better in another place, if we want to do that at all, if we want an event or a log for that case, if those are the fields we want, etc.).

If you want to have a hard look at the code and explain here in detail the above (why it will not spam the logs for all other possible cases, why there is better than all the other alternatives, why we want an event instead of log line, etc.) it would be great. If not, I'll take a few days to do that myself later :)

Matoch:
I can definitely be patient. I just didn't want things to be waiting on me
if possible.

I'll spend a bit more time looking at the code to see if I can answer some
of those questions but that might take even longer then it will for you to
get to it.

James

rata:
Thanks for the understanding! :)

daxmc99:
This looks like a simple log addition. Looks good to merge

johananl:
I confirm that the line is logged also for `ClusterIP` services so it looks like this might not be the right place to put this log line.

rata:
This PR *as it is* is a no merge. Those lines have nothing to do in the generic watching mechanism. Please see the discussion, I [proposed an alternative][link] in the proper layer of abstraction but lack the time to test it (everything I'm saying is in the PR history). 

[link]: https://github.com/metallb/metallb/pull/565#issuecomment-613146394

johananl:
@rata the diff you suggested changes touches the return value of `ShouldAnnounce()` in layer 2 mode (but not in BGP mode). Isn't this change relevant to both layer 2 and BGP?

Additionally, the scope of this PR is larger than it initially looked to me since apparently we have to change controller logic to distinguish `notOwner` from `noEndpoints`. Does it make sense to make this change? If so, we need to alter the tests as they are currently broken by the proposed patch.

rata:
> @rata the diff you suggested changes touches the return value of `ShouldAnnounce()` in layer 2 mode (but not in BGP mode). Isn't this change relevant to both layer 2 and BGP?

Yes, [bgp does says the reason][link] and I used the same reason for Layer 2 IIRC.

[link]: https://github.com/metallb/metallb/blob/63bbb11239d792178d8e6c01174b53b8d5c505e4/speaker/bgp_controller.go#L145-L149


> Additionally, the scope of this PR is larger than it initially looked to me since apparently we have to change controller logic to distinguish `notOwner` from `noEndpoints`. Does it make sense to make this change? If so, we need to alter the tests as they are currently broken by the proposed patch.

Hmm, are you sure we need that? The idea is to print when no endpoints are present, to when we are not the owner. But yeah, if this patch is not an easy patch, I'd not prioritize it for now.

johananl:
> > Additionally, the scope of this PR is larger than it initially looked to me since apparently we have to change controller logic to distinguish `notOwner` from `noEndpoints`. Does it make sense to make this change? If so, we need to alter the tests as they are currently broken by the proposed patch.
> 
> Hmm, are you sure we need that? The idea is to print when no endpoints are present, to when we are not the owner. But yeah, if this patch is not an easy patch, I'd not prioritize it for now.

My message was in response to https://github.com/metallb/metallb/pull/565#issuecomment-613146394. Your proposed diff changes controller logic, i.e. it returns `noEndpoints` when there are no usable nodes. I'm not saying this isn't an easy patch, just pointing out that it looks like more than a **logging change**. And - `make test` is failing after applying the proposed diff so we have to fix that.

How do we move this forward?

russellb:
This PR stalled out a long time ago and @rata NACKed it as-is.  I'm going to close this out.  Feel free to re-open this one or open a new PR if desired.

PR #570, opened 2020-03-31
Update repository urls

### LINKED ISSUE #567: Replace danderson/metallb with metallb/metallb
We seem to have references to the old location of the repository - `danderson/metallb`. There is a GitHub redirect in place which forwards requests and Git actions to the current repo, however we should probably update these refs to `metallb/metallb` to avoid confusing people.

aditsachde:
This repository used to be located at google/metallb and danderson/metallb. Redirects still work, but this just brings it in line with the current repository location.

---

Fixes #567

### DIFF
--- a/website/content/release-notes/_index.md
+++ b/website/content/release-notes/_index.md
@@ -10,7 +10,7 @@ Bugfixes:
 
 - Fix manifests to use container image version `v0.9.3` instead of `main`. Users
   of `v0.9.2` are encouraged to upgrade, as [manifests included in that
-  release](https://raw.githubusercontent.com/google/metallb/v0.9.2/manifests/metallb.yaml)
+  release](https://raw.githubusercontent.com/metallb/metallb/v0.9.2/manifests/metallb.yaml)
   use an incorrect container image version. Those two images happen to match
   now but, as development continues on `main` branch, they will differ.
 
@@ -84,7 +84,7 @@ Action required:
 
 Bugfixes:
 
-- Fix layer2 node selection when healthy and unhealthy replicas are colocated on a single node. ([#474](https://github.com/danderson/metallb/issues/474))
+- Fix layer2 node selection when healthy and unhealthy replicas are colocated on a single node. ([#474](https://github.com/metallb/metallb/issues/474))
 
 This release includes contributions from David Anderson and Gary Richards.
 
@@ -94,10 +94,10 @@ This release includes contributions from David Anderson and Gary Richards.
 
 Bugfixes:
 
-- Fix the apiGroup for PodSecurityPolicy, for compatibility with Kubernetes 1.16. ([#458](https://github.com/google/metallb/issues/458)).
-- Fix speaker posting events with an empty string as the announcing node name. ([#456](https://github.com/google/metallb/issues/456)).
+- Fix the apiGroup for PodSecurityPolicy, for compatibility with Kubernetes 1.16. ([#458](https://github.com/metallb/metallb/issues/458)).
+- Fix speaker posting events with an empty string as the announcing node name. ([#456](https://github.com/metallb/metallb/issues/456)).
 - Fix RBAC permissions on speaker, to allow it to post events to all
-  namespaces. ([#455](https://github.com/google/metallb/issues/455)).
+  namespaces. ([#455](https://github.com/metallb/metallb/issues/455)).
 
 This release includes contributions from David Anderson.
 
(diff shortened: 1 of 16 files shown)


aditsachde:
Didn't realize there was an issue for this already and a couple more locations to update the urls. Will fix and reopen.

rata:
Hi! Thanks for the PR.

Can I ask what did you run to generate it? I'm mostly interested in that, as if very difficult to review otherwise (I don't know if there might be missing references by just looking at the diff).

Also, this has chances to break links in the website and manifests generation, so I'd like to be careful (maybe we can spit the changes?) and try to verify as much as possible? Have you done any kind of verification that things are sane after this PR?

aditsachde:
I did a find and replace of `google/metallb` and `danderson/metallb` to `metallb/metallb` and then went back and made sure all the changes were to Github urls (github.com or raw.githubusercontent.com). There are only a couple of exceptions. 

In the `README.md`, there are three shields that depend on the repository url: license, Circle CI, and Go report card. Those still work with the updated urls.

Last place is here:

https://github.com/metallb/metallb/blob/37586c4658df243964fbf2577a333c2f39989a00/tasks.py#L308-L310

`_replace` seems to be a function that updates the references in website from the old version to the new version. Originally, it matched for `google/metallb`, but since those references have also been updated to `metallb/metallb` in the website files, it should still work. However, I am not sure about how to test this to make sure. 

Because it tries to find those references in the website files, I don't think it is possible to update just the website urls without changing the patterns at the same time as then the script will not be able to find the correct references to update. (Maybe a better solution would be to add a second block and keep both `google/metallb` and `metallb/metallb` present in tasks.py?)

rata:
Thanks a lot!  But do you have any script (even if afterward you did manual changes) to achieve this diff?

> then went back and made sure all the changes were to Github urls (github.com or raw.githubusercontent.com). There are only a couple of exceptions.

Exceptions like places were we didn't want to change it? Do you happen to have them handy?

> `_replace` seems to be a function that updates the references in website from the old version to the new version. Originally, it matched for `google/metallb`, but since those references have also been updated to `metallb/metallb` in the website files, it should still work. However, I am not sure about how to test this to make sure.

That is used when creating a release. We can do it locally, no need to actually create a release. But maybe we can keep them and see that they actually didn't change anything and also add the new ones using metallb github org (as the code should have been updated in this PR).

> Because it tries to find those references in the website files, I don't think it is possible to update just the website urls without changing the patterns at the same time as then the script will not be able to find the correct references to update. (Maybe a better solution would be to add a second block and keep both `google/metallb` and `metallb/metallb` present in tasks.py?)

Yeah, I think we should keep them for a few releases, just to be safe :)

These changes are not top prio now, though, and is important to not break the links. So, I'd be slower to review this. Let me know if this is a problem in any way for you.

Not sure if it's worth doing some scripts to verify (like running locally with `hugo` and follow links with a "bot", or something alike. Will sleep on this :)

Thanks again!

aditsachde:
I just used the vscode find and replace, so no script.

The only things that weren't github urls (i.e. the exceptions) were the readme shields, and those were still updated to point to the new url. Using some scripts to check if everything works would be cool, but I don't know if that would really be faster than just having another set of eyes do a sanity check.

I don't really know how to test the release stuff, so ideally someone more experienced with the codebase would check it but I know that everyone is quite busy at the moment. 

As you mentioned, this is quite a low priority task as all the redirects are still working and there are much more important things to spend time on right now.

caboteria:
LGTM.  I can vouch for the fact that this is a "papercut" that's confusing to new users.  Russellb already merged a PR that implemented part of this (the widgets on the readme page) but this PR would be helpful to newbies.

caboteria:
LGTM cc @rata

rata:
@caboteria as I mentioned here and in the community call today. I'd like to verify that this doesn't break the links in the project prior to merge. And as there are several other things have more priority (as important bug fixes), I didn't look at this yet.

If you want to check and share why this doesn't break any links in a way than others can review, that would be great. If you don't want to do it, is okay too. I'll do it when other very important bugs are fixed :)

website/content/release-notes/_index.md
```diff
@@ -445,15 +445,15 @@ New features:
...
+  and [file bugs](https://github.com/metallb/metallb/issues/new)!
 - BGP peers now have
   a
   [node selector]({{% relref "configuration/_index.md" %}}#limiting-peers-to-certain-nodes). You
   can use this to integrate MetalLB into more complex cluster network
   topologies.
 - MetalLB now has
   a
-  [Helm chart](https://github.com/google/metallb/tree/main/helm/metallb). If
+  [Helm chart](https://github.com/metallb/metallb/tree/main/helm/metallb). If
```
johananl, line 456: This gives a 404 (but the old link does, too).

johananl:
Tried to rebase this but looks like the fork has been deleted. I'll open a new PR with the patch.

johananl:
> Tried to rebase this but looks like the fork has been deleted. I'll open a new PR with the patch.

Closing in favor of #688.

Issue #495, opened 2019-11-15
Potential solution for multiple eBGP sessions over same pair of K8S node and TOR


zy1ting:
I'd like to propose a solution to have both calico eBGP and metalLB eBGP over the same K8S and TOR by using an eBGP session between the loopback interface on the K8S node and TOR for an eBGP multi-hop session.

1. bring up a loopback interface on K8S node and TOR

2. Calico peer eBGP session with TOR over directed connected physical link and let both K8S node and TOR to advertise all directly connected ip including the loopback's so there is reachability between two loopback interface.

3. metalLB peer an eBGP session between the loopback interface on K8S node and TOR. 

In this way, we keep the same BGP ASN on both eBGP sessions.

Zyt612:
in fact, the ideal solution can be to add a feature allowing metalLB specify the local ip address/interface from which it will establish BGP sessions.

arnaudgelly:
This is my current setup (seeing your post before would have saved me some time wondering if it was the right way) :


 ![K8S](https://user-images.githubusercontent.com/2733574/75915396-b54ad480-5e56-11ea-9b43-4d229e2969cd.png)


iBGP with Route Reflector on the 2 core routers (I don't see the advantage of eBGP).

MetalLB BGP sessions between loopbacks and Calico BGP sessions between LAN interfaces has you suggested.

Need a static route on the node toward core routers' loopbacks with the IP of the node's loopback IP has source :
`ip route <CORE-ROUTER-A-LOOPBACK-IP> via <GATEWAY> src <NODE-LOOPBACK-IP>`
`ip route <CORE-ROUTER-B-LOOPBACK-IP> via <GATEWAY> src <NODE-LOOPBACK-IP>`

Working well so far.

Adding the capability of specifying the source interface in the metalLB BGP configuration would simplify the node's configuration for sure.

ToR / Spine or VRF are good options but not adapted to my network topology.

Peering metalLB to calico inside the node is a nice idea but that imply some evolution on Calico's and metalLB BGP features (route-maps would be nice).

Looks like loopbacks is the best option for simple topology.

ckoester2:
For what it's worth, I drafted up a quick code change to allow specifying a source interface to MetalLB: https://github.com/ckoester2/metallb/compare/main...ckoester2:issue-495

This needs a bit more testing before being ready for a PR, but in a nutshell, adds an optional `source-interface` configurable to the BGP peer configuration that may contain an interface name (not an IP address). Example:

```
    peers:
[...]

Issue #465, opened 2019-08-07
ARP/NDP not work correctly on machines with multiple interfaces


icefed:
#### Problem
I have three servers and each has two interfaces for Internal & External network (different subnets), MetalLB speaker reply on the wrong interface when it received ExternalIP ARP request from Internal network interface, it cause my client(External network) can not reach ExternalIP.

#### What I want
MetalLB speaker should only reply when senderIP in interface‘s subnet.

Related to https://github.com/danderson/metallb/issues/165

danderson:
Duplicate of #277

icefed:
@danderson Please read this issue's content, it's not the same.

danderson:
MetalLB replies on the interface where it receives the request. If you're seeing replies coming from the wrong interface, it means the ARP request is being delivered to (at least) the wrong interface. This is due to either a network topology issue (e.g. both your subnets are actually sharing an ethernet segment, rather than being separate VLANs), or to the problem described in the linked issue, where composite linux interfaces can cause spurious receives on the wrong interface.

The fix you suggest won't work, because listening on an interface that has no addresses assigned is a valid configuration that's being used by other deployments.

icefed:
I think you misunderstand.
I have two interfaces which have different subnet 192.168.10.x and 192.168.20.x, connect to one switch, so maybe this is a network topology issue, but it should work.

my fix is not `listening all interfaces alse contains that has no addresses assigned `, but just implemented your design: https://github.com/danderson/metallb/issues/165
> 
> What you expected to happen:

> The ARP speaker should just listen for ARP traffic on all interfaces, and respond to relevant queries on the correct interface (i.e. the interface that has an address/cidr compatible with the requested IP).

> So, then, each goroutine just needs to do the same readPacket loop that already exists, but additionally needs to check if the IP in the ARP request is contained within the configured subnet for the current interface (e.g. "is 10.1.2.3 in 10.1.2.0/24 ? Yes, that means I should respond").

I don't know why you not code it, I fixed it (https://github.com/icefed/metallb/commit/f8270d6e1aebe305b1d68e5cd57b87fe6adf899b) and working good in my environment, also solved my problem.

PR #573, opened 2020-04-06
Add source-interface configuration to BGP peer configuration

carstenkoester:
Add an optional "source-interface" parameter to the BGP peer
configuration.

When configured, this must be an interface name, and the BGP
session to the peer is sourced from that interface. This allows
the BGP session to originate from a Dummy interface (equivalent
to a Loopback interface on other BGP routers) rather than the
physical interface facing the BGP peer. This, in turn, can
help avoid conflicts when other processes (eg. Calico) are also
BGP peering off the same node.

Fixed: #495

### DIFF
--- a/internal/bgp/bgp.go
+++ b/internal/bgp/bgp.go
@@ -28,6 +28,7 @@ type Session struct {
 	asn              uint32
 	routerID         net.IP // May be nil, meaning "derive from context"
 	myNode           string
+	srcIntf          string // Source interface; empty string means "use any available"
 	addr             string
 	peerASN          uint32
 	peerFBASNSupport bool
@@ -170,7 +171,7 @@ func (s *Session) connect() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 	deadline, _ := ctx.Deadline()
-	conn, err := dialMD5(ctx, s.addr, s.password)
+	conn, err := dialMD5(ctx, s.srcIntf, s.addr, s.password)
 	if err != nil {
 		return fmt.Errorf("dial %q: %s", s.addr, err)
 	}
@@ -351,8 +352,9 @@ func (s *Session) sendKeepalive() error {
 //
 // The session will immediately try to connect and synchronize its
 // local state with the peer.
-func New(l log.Logger, addr string, asn uint32, routerID net.IP, peerASN uint32, holdTime time.Duration, password string, myNode string) (*Session, error) {
+func New(l log.Logger, srcIntf string, addr string, asn uint32, routerID net.IP, peerASN uint32, holdTime time.Duration, password string, myNode string) (*Session, error) {
 	ret := &Session{
+		srcIntf:     srcIntf,
 		addr:        addr,
 		asn:         asn,
 		routerID:    routerID.To4(),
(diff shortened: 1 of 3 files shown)


ckoester2:
Making this a draft PR as I'd like to solicit feedback on the overall implementation, while a couple of changes are still pending and in progress:

 - Testing with IPv6
 - Documentation update
 - Test cases (if there is any elegant way this can be included in test automation)

champtar:
Hi @ckoester2 
I haven't looked at the code yet, but maybe we should use source IP, so you can just add the IP as secondary on an interface ?
Adding a dummy interface is an extra step that might not make it easier for people.

ckoester2:
> Hi @ckoester2
> I haven't looked at the code yet, but maybe we should use source IP, so you can just add the IP as secondary on an interface ?
> Adding a dummy interface is an extra step that might not make it easier for people.

I thought about that, but I'm not sure  there's a really elegant and easy way to make this configuration input. Specifying an IP address for each node in the MetalLB config map, would considerably inflate the `node-selectors`/`match-labels`/`match-expressions` section in (or require us to create some other per-node configuration entity), and turn out to be lengthy and error prone. Or we could remove this configuration from the configmap altogether, and expect some other environment variable to be injected into the pod - but that would prevent users from running published YAML file unmodified. Ideas on how to inject a per-node variable in an easily usable way would be welcome. In either case, this would likely require users to make two changes in tandem: Add a secondary IP address on the node (using host-level configuration on the node), and then adding that same IP address for the same node in MetalLB config. This also means that every time a node is re-configured, *two* places would need to be updated and kept in sync. Unless we want MetalLB to actually modify host interface configuration and configure that IP address on the host - which would come with its own set of caveats.

On the other hand, creating an interface isn't *that* complicated (see my comment in the issue #495 thread), and provides a single source of truth for the IP address (the specific per-node IP address needs to be configured in only one place). And using a Loopback interface for peering is a *relatively* familiar concept in BGP configuration overall, I think.

champtar:
@ckoester2 you are totally right, the fact that the IP is unique per node make my ask/solution worse
I don't know if someone was working on per node peer (I'm not using bgp)

Elegant996:
Wouldn't this also resolve #465? I just finished reverting a previous setup due to ARP going out over interfaces that used to create VLANs. Only one interface actually has an IP.

johananl:
Thanks a lot for the PR @carstenkoester.

This is very similar to the `interface` mode of the [Calico IP autodetection mechanism](https://docs.projectcalico.org/reference/node/configuration#ip-autodetection-methods).

Since this is similar in essence to https://github.com/metallb/metallb/issues/670 (selecting the source IP using interface name vs. specifying the IP directly) I think we should agree on a design first to see how the two methods for selecting the source IP interact.

In addition, here are some things to consider:

- What if the selected interface has multiple IP addresses of the same address family and the user wants to use a specific one?
- Maybe we should accept an interface *regex* rather than a name to provide more flexibility.

johananl:
I've opened https://github.com/metallb/metallb/issues/689 to discuss things around this solution.

russellb:
Since we have no plan to act on this PR until after #689 is resolved, I think we should just close this out.  It can be reopened if the discussion resulting from #689 warrants it.  Thanks

PR #574, opened 2020-04-06
Add IPv6 BGP support, fixing #7

TheEnbyperor:
Does what is says on the tin really.

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,2 @@
+# Default ignored files
+/workspace.xml
--- a/internal/bgp/bgp.go
+++ b/internal/bgp/bgp.go
@@ -425,13 +425,6 @@ func (s *Session) Set(advs ...*Advertisement) error {
 
 	newAdvs := map[string]*Advertisement{}
 	for _, adv := range advs {
-		if adv.Prefix.IP.To4() == nil {
-			return fmt.Errorf("cannot advertise non-v4 prefix %q", adv.Prefix)
-		}
-
-		if adv.NextHop != nil && adv.NextHop.To4() == nil {
-			return fmt.Errorf("next-hop must be IPv4, got %q", adv.NextHop)
-		}
 		if len(adv.Communities) > 63 {
 			return fmt.Errorf("max supported communities is 63, got %d", len(adv.Communities))
 		}
--- a/internal/bgp/messages.go
+++ b/internal/bgp/messages.go
@@ -301,12 +301,22 @@ func sendUpdate(w io.Writer, asn uint32, ibgp, fbasn bool, defaultNextHop net.IP
 		return err
 	}
 	l := b.Len()
-	if err := encodePathAttrs(&b, asn, ibgp, fbasn, defaultNextHop, adv); err != nil {
-		return err
+	if adv.Prefix.IP.To4() != nil {
+        if err := encodePathAttrs(&b, asn, ibgp, fbasn, defaultNextHop, adv); err != nil {
+            return err
+        }
+        binary.BigEndian.PutUint16(b.Bytes()[21:23], uint16(b.Len()-l))
+        encodePrefixes(&b, []*net.IPNet{adv.Prefix})
+        binary.BigEndian.PutUint16(b.Bytes()[16:18], uint16(b.Len()))
+	} else if adv.Prefix.IP.To16() != nil {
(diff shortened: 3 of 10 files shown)


.idea/.gitignore
```diff
@@ -0,0 +1,2 @@
+# Default ignored files
```
rata, line 1: Please remove all the `.idea` files and dir

internal/bgp/messages.go
```diff
@@ -301,12 +301,22 @@ func sendUpdate(w io.Writer, asn uint32, ibgp, fbasn bool, defaultNextHop net.IP
...
+        encodePrefixes(&b, []*net.IPNet{adv.Prefix})
+        binary.BigEndian.PutUint16(b.Bytes()[16:18], uint16(b.Len()))
+	} else if adv.Prefix.IP.To16() != nil {
+        if err := encodePathAttrsIPv6(&b, asn, ibgp, fbasn, defaultNextHop, adv); err != nil {
+            return err
+        }
+        binary.BigEndian.PutUint16(b.Bytes()[21:23], uint16(b.Len()-l))
+        binary.BigEndian.PutUint16(b.Bytes()[16:18], uint16(b.Len()))
+	} else {
+	    return nil
```
rata, line 318: Why nil here? If `adv.Prefix.IP.To4()` and `adv.Prefix.IP.To16()` fail, don't we want to throw an error?

internal/bgp/messages.go
```diff
@@ -396,29 +406,133 @@ func encodePathAttrs(b *bytes.Buffer, asn uint32, ibgp, fbasn bool, defaultNextH
...
+    } else if prefixes[0].IP.To16() != nil {
+        if err := binary.Write(&b, binary.BigEndian, uint16(0)); err != nil {
+            return err
+        }
+        if err := encodePathAttrsIPv6Withdraw(&b, prefixes); err != nil {
+            return err
+        }
+        binary.BigEndian.PutUint16(b.Bytes()[21:23], uint16(b.Len()-l))
+        binary.BigEndian.PutUint16(b.Bytes()[16:18], uint16(b.Len()))
+    }
```
rata, line 535: If none of the branches are taken, do we want to continue?

TheEnbyperor:
yes this is MP-BGP. I've tested it in the http://as207960.net autonomous system and it successfully pushes routes to the ToR routers. I'll try to get some PCAP files to test it.

internal/bgp/bgp.go
```diff
@@ -425,13 +425,6 @@ func (s *Session) Set(advs ...*Advertisement) error {
 
 	newAdvs := map[string]*Advertisement{}
 	for _, adv := range advs {
-		if adv.Prefix.IP.To4() == nil {
-			return fmt.Errorf("cannot advertise non-v4 prefix %q", adv.Prefix)
-		}
-
-		if adv.NextHop != nil && adv.NextHop.To4() == nil {
-			return fmt.Errorf("next-hop must be IPv4, got %q", adv.NextHop)
-		}
```
rata, line 434: Any reason to remove this check on the next hop?

ydotk:
Hi @TheEnbyperor
Thanks for your PR! I am also very interested in this feature. It looks like we were working on the same feature with you in parallel.
I also implemented this feature. I uploaded my code, maybe it will be useful for you as reference. 
https://github.com/metallb/metallb/pull/575

internal/bgp/bgp.go
TheEnbyperor: Ah, I thought that was just a ipv4 check. Misread what it was doing.

internal/bgp/messages.go
TheEnbyperor: after removing the explicit check for ipv4 some announcements where getting through with an IP of nil. It seemed harmless to discard them.

internal/bgp/messages.go
TheEnbyperor: If i understand it correctly merely a blank bgp message will be sent, which seems within the spec.

russellb:
@TheEnbyperor based on your comment on #590 here: https://github.com/metallb/metallb/pull/590#issuecomment-619622188

Did you get a chance to review that one and see if it meets your needs?  Based on your comment there, do you think we should close this one in favor of getting #590 in?

TheEnbyperor:
Yes I think #590 is overall a better PR that this one

russellb:
Thanks!

PR #575, opened 2020-04-09
Add support for MP BGP encoding for IPv4 and IPv6

ydotk:
Hi,
It looks like we were working on the same feature with @TheEnbyperor in parallel.
TheEnbyperor PR: https://github.com/metallb/metallb/pull/574

I also added IPv6 support for BGP speaker, but I'm late with PR.
In any case, I want to submit my PR so that you can see an alternative point of view on the feature implementation.

My implementation is a little different and provide some additional feature.
1)    MP BGP encoding will be used for IPv4 prefixes announcement if the peer supports it, legacy encoding will be used otherwise.
2)    Don’t announce IPv6 prefixes for peer which don’t support them
3)    You can use IPv4 or IPv6 to establish a BGP session for IPv4 and IPv6 prefixes announcement
4) You can use different aggregation length for IPv6 prefixes
@TheEnbyperor already provide implementation for the feature, so PR is more for reference.
Feel free to close this PR if you don’t need it.

### DIFF
--- a/internal/bgp/bgp.go
+++ b/internal/bgp/bgp.go
@@ -31,6 +31,8 @@ type Session struct {
 	addr             string
 	peerASN          uint32
 	peerFBASNSupport bool
+	mpIPv4Support    bool
+	mpIPv6Support    bool
 	holdTime         time.Duration
 	logger           log.Logger
 	password         string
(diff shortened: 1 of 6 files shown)


TheEnbyperor:
Hey, the code you added for handling aggregation length looks much better than mine, perhaps we should use that

PR #586, opened 2020-04-16
Always do gratuitious advertisement when calling SetBalancer

### LINKED ISSUE #584: L2 mode doesn't resend gratuitous arp when node is diconnected and reconnected

     - Version of MetalLB v0.9.3
     - Version of Kubernetes v1.15

champtar:
The actual call to gratuitous(name) will happen after we release
the lock so we are fine

This fixes #584 

(not tested yet)

### DIFF
--- a/internal/layer2/announcer.go
+++ b/internal/layer2/announcer.go
@@ -184,6 +184,9 @@ func (a *Announce) SetBalancer(name string, ip net.IP) {
 	a.Lock()
 	defer a.Unlock()
 
+	// Always do gratuitous advertisement
+	go a.spam(name)
+
 	// Kubernetes may inform us that we should advertise this address multiple
 	// times, so just no-op any subsequent requests.
 	if _, ok := a.ips[name]; ok {
@@ -203,9 +206,6 @@ func (a *Announce) SetBalancer(name string, ip net.IP) {
 			a.logger.Log("op", "watchMulticastGroup", "error", err, "ip", ip, "msg", "failed to watch NDP multicast group for IP, NDP responder will not respond to requests for this address")
 		}
 	}
-
-	go a.spam(name)
-
 }
 
 // DeleteBalancer deletes an address from the set of addresses we should announce.


champtar:
@liuyuan10 can you test this PR ?

liuyuan10:
sure. will test it tomorrow

liuyuan10:
While it's working as expected, I have a bit of concern. take a example with 100 nodes

nodes: N1-N100
services: S1-S100 (assuming one on each node , s1->n1 ...)

And if we disconnect N1 and reconnect it later, the duration is Td. By saying disconnect, I mean traffic from N1 is blackholed (it's different from unplug N1 where memberlist can notice it's an local error)

1. if Td > max suspicious timeout (check memberlist doc on what is mean, up to 120s with our config), on N1, it fails all the other node.  Once it gets reconnected, the other 99 nodes will join one by one which causes a row of forced syncs so GARPs. I feel there is chance that before N1 gets a full picture of the nodes, it will send GARPs for other services causing disruptions.

2. If min suspicious timeout < Td < max suspicious timeout, N1 will be failed in the rest of N2-N100 and say N2 becomes the new master of S1. Once N1 comes back, both N1 and N2 will be acting as S1 master. There is no node event in N1 because Td < max suspicious timemout and no confirmation is received from other nodes. But I notice k8s client will trigger a resync seconds later which causes GARP sent for 5s. I'm not sure how confident that k8s client will trigger a resync in time but I feel a bit uncomfortable to rely on that.

I notice resync is anyway triggered at random time which means finally it will fix the inconsistent state but there is no expectation on how long that will be. It can be seconds or minutes. Overall I feel the change is helping more. I can't think of other better approaches so this seems what we can do now.

champtar:
@liuyuan10 thanks a lot for all your work
At 100 nodes I would really recommend BGP mode.
Small improvements we can make right now
- ensure we only have 1 "spam" goroutine per IP at a time
Longer term maybe we can fix a bunch of those issues when we implement "sticky" master

liuyuan10:
On Fri, Apr 17, 2020 at 6:09 AM Etienne Champetier <notifications@github.com>
wrote:

> @liuyuan10 <https://github.com/liuyuan10> thanks a lot for all your work
> At 100 nodes I would really recommend BGP mode.
>
Is there more detail on this? I know BGP is better L2 mode in several ways,
but is there guidance on how many nodes L2 mode is fine for? I can even
image it works with 1000 nodes with longer time to converge.  Any previous
test results on how many nodes and services metallb can handle will be
very helpful.

> Small improvements we can make right now
>
>    - ensure we only have 1 "spam" goroutine per IP at a time
>
> Yeah. It's definitely an improvement. We can even use only a single worker
goroutine to handle all the GARPs so that we scale better when handling say
1000 services. But that's more work to send GARP every 1 second  for every
services.

PR #588, opened 2020-04-17
Update github.com/mdlayher/{arp,ethernet,ndp}

### LINKED ISSUE #547: Incorporate upstream fix for infiniband devices.
Hello. I'm running metallb on an on-prem HPC cluster utilizing an infiniband network and IPoIB (IP over Infiniband).

I found an issue with how packets are being checked and differences between the Ethernet and Infiniband networks. It was fixed upstream in mdlayher/arp. Can you please incorporate these fixes into your package? 

https://github.com/mdlayher/arp/issues/18
https://github.com/mdlayher/arp/pull/22

champtar:
go get -u -t github.com/mdlayher/arp github.com/mdlayher/ethernet github.com/mdlayher/ndp

This fixes #547

(not tested yet)


caboteria:
Hi!  This has been a "draft" for a few months now, what's the plan?  Thanks!

champtar:
no plan, @jcatana do you think you can rebuild that and test ?

daxmc99:
@champtar Is there anything we can do to help you build & test this? 
This looks like a reasonable bump

dylex:
As per #779, I checked out this PR (616e84f4), did an `inv build` and `inv push` to a local repo, switched my existing 0.9.5 install to point controller and speaker at these `dev-amd64` images, and re-applied. Things seem to be working, and I no longer see the messages referenced on #779. The only questionable thing I see in the speaker logs:
```
{"caller":"main.go:213","component":"MemberList","msg":"state.go:1196: [WARN] memberlist: Refuting a dead message (from: cluster-k8s-3)","ts":"2020-11-16T01:47:37.629694146Z"}
{"caller":"main.go:174","error ?":null,"msg":"Memberlist join","nb joigned":2,"op":"startup","ts":"2020-11-16T01:47:37.629746358Z"}
{"caller":"main.go:197","msg":"Node event","node addr":"10.2.250.1","node event":"NodeJoin","node name":"cluster-k8s-1","ts":"2020-11-16T01:47:37.629758335Z"}
```
I should note that I am not using metallb to assign IP-over-IB addresses, but do have an IP-over-IB interface on the cluster used for other things.

Anything specific I should try?

champtar:
Just rebased and updated the commits, can you retest what I just pushed ?
Also if you can try using a VIP on the IPoIB interface that would be great

dylex:
Updated to 2cd7e751 and everything looks clean and seems to work.
Our IB fabric is in production, so I'm hesitant to do anything that might be disruptive, but if you think it should be safe, I can try adding a new address pool on the IB network and adding a service on it.

champtar:
"I don't always test my code, but when I do, it's in production" :D

champtar:
Sending random packets should not affect your IB network as this would be a vulnerability for me, but at the same time I've never touched IB, so no idea how good/bad it is. Also no preprod or staging ?

champtar:
BTW testing this PR you are now sending GARP packets on your IB interface, so you already started to test in prod ;)

dylex:
Well, okay then.  (This is an academic HPC cluster and IB hardware is rather expensive, so we don't tend to have fabrics on test infrastructure. IB, being a flat, non-routed, largely proprietary network, is notoriously sensitive to misbehaving nodes.  They often  insist on having the same firmware on all cards, for instance.)

I added a new address pool on the IB network, created a new service (based on an existing working one) and assigned a manual `loadBalancerIP` from the fabric. I can connect to the service from nodes in the k8s cluster, but not from other nodes on the IB fabric. arp reports `(incomplete)`. I tried default and local traffic policies.
```
{"caller":"main.go:279","event":"serviceAnnounced","ip":"10.2.250.17","msg":"service has IP, announcing","pool":"ibtest","protocol":"layer2","service":"default/ibtest","ts":"2020-11-16T15:28:02.525189998Z"}
```
I do see the ARP requests coming in, but no response:
```
07:39:34.021955 ARP, Request who-has 10.2.250.17 tell 10.2.0.12, length 56
```
Happy to poke a bit more, but don't want to do anything too invasive.

Out of curiousity, does metallb send GARPs for VIPs on interfaces with different subnets? Does that serve any purpose? In case there are bridged interfaces or something?

champtar:
MetalLB does send GARP and respond to resquests on all interfaces. You can have multiple subnet in a VLAN, so you could have a private IP / default gateway and a public IP as VIP, as long as the responses go to the right gateway everything works fine.

3 things that you can do:
1) see if the GARP that we send are valid / received by other nodes on the IB network (look at the pcap in wireshark / compare with the real requests responses)
2) try to strace the MetalLB process to see if the ARP requests is received by MetalLB or not (I think we use raw sockets with filters, and maybe the filter is for Ethernet not IB so it doesn't work)
3) add some debugging in the code if we receive the packets (I can help with that later)

dylex:
For 1, the GARPs don't look great. Here's an ARP from metallb and a normal ARP reply:
```
Frame 2: 62 bytes on wire (496 bits), 62 bytes captured (496 bits)
Linux cooked capture
    Packet type: Sent by us (4)
    Link-layer address type: 32
    Link-layer address length: 0
    Unused: 0000000000000000
    Protocol: ARP (0x0806)
    Padding: 0000506b4b0300c414040a02fa11ffffffffffff00000000?
Address Resolution Protocol (reserved)
    Hardware type: Unknown (1327)
    Protocol type: Unknown (0xfe80)
    Hardware size: 0
    Protocol size: 0
    Opcode: reserved (0)
VSS Monitoring Ethernet trailer, Source Port: 64017
    Src Port: 64017

0000  00 04 00 20 00 00 00 00 00 00 00 00 00 00 08 06   ... ............
0010  05 2f fe 80 00 00 00 00 00 00 50 6b 4b 03 00 c4   ./........PkK...
0020  14 04 0a 02 fa 11 ff ff ff ff ff ff 00 00 00 00   ................
0030  00 00 00 00 00 00 00 00 00 00 0a 02 fa 11         ..............

Frame 4: 72 bytes on wire (576 bits), 72 bytes captured (576 bits)
Linux cooked capture
    Packet type: Sent by us (4)
    Link-layer address type: 32
    Link-layer address length: 0
    Unused: 0000000000000000
    Protocol: ARP (0x0806)
Address Resolution Protocol (reply)
    Hardware type: InfiniBand (32)
    Protocol type: IPv4 (0x0800)
    Hardware size: 20
    Protocol size: 4
    Opcode: reply (2)
    Sender hardware address: 0000052ffe80000000000000506b4b0300c41404
    Sender IP address: 10.2.250.1
    Target hardware address: 00000e92fe8000000000000098039b0300034c00
    Target IP address: 10.2.201.2

0000  00 04 00 20 00 00 00 00 00 00 00 00 00 00 08 06   ... ............
0010  00 20 08 00 14 04 00 02 00 00 05 2f fe 80 00 00   . ........./....
0020  00 00 00 00 50 6b 4b 03 00 c4 14 04 0a 02 fa 01   ....PkK.........
0030  00 00 0e 92 fe 80 00 00 00 00 00 00 98 03 9b 03   ................
0040  00 03 4c 00 0a 02 c9 02                           ..L.....
```
Everything in the ARP frame (starting with the hardware type) seems wrong. These packets never get anywhere, it seems.

As for 2, there's a lot of traffic -- do you have an easy way of decoding the raw recv packets to find the right one?

Is there an easier way to trigger to GARP than deleting and-recreating the service?

champtar:
I see "Linux cooked capture", can you capture using "tcpdump -i ib0 arp" and not "tcpdump -i any arp" ?

I think next steps are to create a minimal golang program that send GARP and dump ARP using the same libs.
Thanks for all the testing I'll see when I have time to create that.

dylex:
This was captured with `tcpdump -p -nn -s0 -w file -i ib0`. Because IP-over-IB is a software layer on top of the raw IB, I don't believe it supports other capture types.

jcatana:
I've tested this by applying the fix in the PR on the current main branch and it's working.

russellb:
@champtar thank you for preparing this PR and @jcatana thank you for reporting the issue and confirming that this change resolves the original issue.  I'll merge it and then it'll be included in v0.10

PR #592, opened 2020-04-22
website: improve kustomize installation section

Trojan295:
This PR fixes a few errors in the kustomize installation on the website.

The most important is the `bases` section, because the current installation doesn't work

### DIFF
--- a/website/content/installation/_index.md
+++ b/website/content/installation/_index.md
@@ -82,7 +82,7 @@ on the remote kustomization file:
 namespace: metallb-system
 
 resources:
-  - github.com/danderson/metallb//manifests?ref=v0.9.3
+  - github.com/metallb/metallb//manifests?ref=v0.9.3
   - configmap.yml 
   - secret.yml
 ```
@@ -99,7 +99,7 @@ the configMap, as MetalLB is waiting for a configMap named `config`
 namespace: metallb-system
 
 resources:
-  - github.com/danderson/metallb//manifests?ref=v0.9.3
+  - github.com/metallb/metallb//manifests?ref=v0.9.3
 
 configMapGenerator:
 - name: config
--- a/website/content/installation/_index.md
+++ b/website/content/installation/_index.md
@@ -71,16 +71,18 @@ The `memberlist` secret contains the `secretkey` to encrypt the communication be
 
 ## Installation with kustomize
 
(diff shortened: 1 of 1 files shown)


rata:
Hi!

Thanks a lot for the PR. I'm not familiar with kustomize, so I'll have a look later. Just curious: is this not working currently? Was it broken by a new kustomize release? If you can mention that in the commit msg (and PR description), it would be great :).

Also, how are you running it? I guess you are using `kustomize` instead of `kubectl -k`?

Regarding the version mismatch, I'll create an issue to fix it in the release process (it should be updated automatically).

Trojan295:
Kustomize is now built into kubectl as `kubectl apply -k`. The current instruction isn't working, `bases` should be used to reference an another `kustomization.yaml`. I don't know, if this changed some time ago, but for halt an year I'm using this syntax.

The changes I made here, were the once I needed to deploy MetalLB on an k3s deployment. I think I could add also example `configs/config.yaml` and `configs/secretkey.env` to this PR, which are referenced in this kustomize config.

Trojan295:
OK, I checked now the docs and it looks, that I'm out of date, because `bases` is deprecated. So I think this PR, besides the version bump can be closed.

rata:
@Trojan295 cool! Can you confirm it works fine with `kubectl -k`, then? Also, would you like to update the PR to just contain that fix you mention, if it is the only one relevant? :)

Thanks!

Trojan295:
`kubectl -k` doesn't work, for me, as it included kustomize 2.x.
I'm gonna recheck it with `kustomize` 3.x and updated/close the PR, if needed.

rata:
@Trojan295 thanks! I'm not into the kustomize world to know if people is using it with kubectl or just outside of it.

Do you know? :)

Trojan295:
I personally use `kubectl -k`, cause it's already there with `kubectl` :)

Trojan295:
OK, the current instructions work for kustomize v2.1.0+, but `kubectl` still includes some older one.

Finally I updated the version in the `kustomization.yaml`, added sample configs for the generators and added a note about v2.1.0 compatibility.

website/content/installation/_index.md
```diff
@@ -110,5 +112,19 @@ secretGenerator:
...
+address-pools:
+  - name: default
+    protocol: layer2
+    addresses:
+      - 192.168.0.128/25
+```
+
+```
+# configs/secretkey
+secretkey=SZlmcghfKLpoj+rnGKcxAeKQY+H/lwo0t10GJ8M3dI8=
```
russellb, line 129: I don't think we should include a sample secret here because inevitably someone will just copy/paste this as-is and have an insecure cluster as a result.  I think just providing instructions on how to generate the secret is better.

russellb:
As noted in my last review, I don't think we should merge this as-is.  I really appreciate the doc contribution, though, so please re-open if and when you're ready to update this.  Thanks!

Issue #521, opened 2020-01-28
BGP peers per address-pool


Tatsu-Kishi:
I'm currently setting up our second Kubernetes cluster (to someday supersede the first one) that is supposed to get public LB IPs (but otherwise be in our internal network).
For this we peer with our edgerouters.
Now for internal LB IPs instead of doing L2 mode I'd like to peer with our core routers without them getting the DMZ IP routes or the edgerouters getting the internal routes, but from what I understand this currently isn't possible with metallb.

One idea would be to group peers and add an optional parameter to address-pools to limit which peer group(s) this pool's addresses will be advertised to.

SDBrett:
+1

It would make sense to add the field `peers[].address-pools` to accept an array of strings.

Example config:

```
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: metallb-system
  name: config
data:
  config: |
    peers:
    - peer-address: 10.0.0.1
      peer-asn: 64501
      my-asn: 64500
      address-pools:
      - default
    - peer-address: 172.0.0.1
      peer-asn: 64501
      my-asn: 64500
      address-pools:
      - dmz       
    address-pools:
    - name: default
      protocol: bgp
      addresses:
      - 192.168.10.0/24
    - name: dmz
      protocol: bgp
      addresses:
      - 10.168.10.0/24
```

Tatsu-Kishi:
I like this approach way better than my idea to define the peers in the address-pools. This keeps address-pools protocol agnostic unlike my approach

SDBrett:
> I like this approach way better than my idea to define the peers in the address-pools. This keeps address-pools protocol agnostic unlike my approach

Thanks mate :)

SDBrett:
I'm happy to work on this (and made a start) feature as I've seen similar requirements from clients on networking projects before.

I do need assistance to better understand out the internal process flow for BGP advertisements.

ykulazhenkov:
I am also very interested in this feature. I like the config change that @SDBrett proposed. Has someone already started work on the implementation? If no one working on that, maybe I will try to implement this feature soon.

ykulazhenkov:
I created PR with this feature https://github.com/metallb/metallb/pull/596

johananl:
I'd like to see if we can make this feature generic - see https://github.com/metallb/metallb/issues/630.

Tatsu-Kishi:
I don't think #630 applies here - this isn't about which nodes to announce from, but which pools to announce to which peers.

[...]

PR #596, opened 2020-04-29
Limit announcement of address pools to specific BGP peers

ykulazhenkov:
This patch provides ability to configure exact set of address pools that should be announced to a specific BGP peer, as described in https://github.com/metallb/metallb/issues/521
Example config:
```
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: metallb-system
  name: config
data:
  config: |
    peers:
    - peer-address: 10.0.0.1
      peer-asn: 64501
      my-asn: 64500
      address-pools:
      - default
    - peer-address: 172.0.0.1
      peer-asn: 64501
      my-asn: 64500
      address-pools:
      - dmz       
    address-pools:
    - name: default
      protocol: bgp
      addresses:
      - 192.168.10.0/24
    - name: dmz
      protocol: bgp
      addresses:
      - 10.168.10.0/24
```
image

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -45,6 +45,7 @@ type peer struct {
 	RouterID      string         `yaml:"router-id"`
 	NodeSelectors []nodeSelector `yaml:"node-selectors"`
 	Password      string         `yaml:"password"`
+	AddressPools  []string       `yaml:"address-pools"`
 }
 
 type nodeSelector struct {
@@ -109,11 +110,15 @@ type Peer struct {
 	NodeSelectors []labels.Selector
 	// Authentication password for routers enforcing TCP MD5 authenticated sessions
 	Password string
+	// AddressPools which should be announced to the peer
+	AddressPools map[string]struct{}
 	// TODO: more BGP session settings
 }
 
 // Pool is the configuration of an IP address pool.
 type Pool struct {
+	// Name address pool name
+	Name string
 	// Protocol for this pool.
 	Protocol Proto
 	// The addresses that are part of this pool, expressed as CIDR
@@ -195,13 +200,6 @@ func Parse(bs []byte) (*Config, error) {
 	}
 
 	cfg := &Config{Pools: map[string]*Pool{}}
-	for i, p := range raw.Peers {
-		peer, err := parsePeer(p)
-		if err != nil {
-			return nil, fmt.Errorf("parsing peer #%d: %s", i+1, err)
-		}
-		cfg.Peers = append(cfg.Peers, peer)
-	}
 
 	communities := map[string]uint32{}
 	for n, v := range raw.BGPCommunities {
@@ -211,7 +209,6 @@ func Parse(bs []byte) (*Config, error) {
 		}
 		communities[n] = c
 	}
-
 	var allCIDRs []*net.IPNet
 	for i, p := range raw.Pools {
 		pool, err := parseAddressPool(p, communities)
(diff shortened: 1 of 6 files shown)


rata:
Hi! Thanks for the PR! It seems like a reasonable functionality from a very quick look.

However, do you know if there are more generic ways to enforce policies like this outside of MetalLB? I really really don't know if it might be possible to do it outside if we do some integration with [OPA](https://www.openpolicyagent.org/), for example. Do you know?

If something like this is possible, it would be great. We can solve a more generic problem in MetalLB, and allow flexibility from the user POV without the need of specific patches. But maybe is just wishful thinking, I'm not really sure if OPA will help in this case (and not familiar at all with OPA yet).

ykulazhenkov:
Thanks for the comment, @rata. I read about OPA. I think that It will require doing a lot of refactoring to be able to use a single policy/filtering solution for all use cases(at least to support policy enforcement between node and BGP peer, node and address pool, BGP peer and address pool). My personal preference is to have a set of well-documented filters/mappings in a configuration file instead of adding a new abstraction layer with a policy agent. I think that adding a "policy enforcement layer" can make MetalLB behavior harder to understand.

johananl:
Thank you for the PR @ykulazhenkov !

Since we have multiple enhancement requests around having an "affinity" between pools and nodes, and since I'd like to explore the idea of having a **generic** (namely protocol-independent) way to support this, I'm wondering if it makes sense to reverse the direction of the relationship: instead of specifying a list of pools to advertise under a node, specify a node selector under a pool.

Would something like this address the same use case? See https://github.com/metallb/metallb/issues/630.

skydoctor:
This would be a very useful feature for our telco use-case where exactly as shown in the diagram in the original comment, we typically have every K8s node connected to multiple routers. In this setting, we need a way to control which pools are advertised to which router. Also, I think #630 does not solve this use-case since each node is connected to all the routers.

rata:
@skydoctor hi! Not sure I follow, all nodes are connected to all routers but if you can select which nodes (with a selector) will advertise different IP pools to the routers, what is missing? Sorry, not sure I got that part correctly

skydoctor:
Hi @rata. To expand on our use-case, what we need is that if there are say 2 peers and 2 pools configured, we want all the nodes to advertise pool-1 to peer-1 and all the nodes to advertise pool-2 to peer-2.  There are no differences between the worker nodes or any special per-node label. Pods from a microservice can land on any node, so we don't want to limit the announcement of say pool-1 to come only from some nodes and not others. We use [externalTrafficPolicy=local](https://metallb.universe.tf/usage/#local-traffic-policy-1) to have the peer send packets only to nodes which have a pod of the service type.

rata:
@skydoctor Ohh, thanks! I see now why that wouldn't help. Let's keep this open, then. Will update #630 to point this. Then, we can see all the use cases and see what is the most reasonable way to implement it :)

skydoctor:
Thanks for your efforts on the project @rata!

rata:
Thank you, sorry for the big delay! :)

janeczku:
This seems like the correct approach to support the use case of announcing specific address pools to specific BGP peers. There is no comparable scenario with Layer-2 because there is only ever exactly one receiver of the VIP information per node (the broadcast domain associated with the VIP interface).

The generic approach documented in [630](https://github.com/metallb/metallb/issues/630) addresses the different problem of pool->node-affinity.

fedepaol:
Closing this as superseeded by https://github.com/metallb/metallb/pull/1171

PR #602, opened 2020-05-14
website: Reflect the move to memberlist

russellb:
While reading about metallb via the web site, I noticed some content
that was out of date with the current implementation.  The layer 2
mode page still discussed the use of the Kubernetes API for leader
election.  I've updated the text to reflect that memberlist is now
used for cluster membership in metallb.

### DIFF
--- a/website/content/concepts/layer2.md
+++ b/website/content/concepts/layer2.md
@@ -25,9 +25,10 @@ In that sense, layer 2 does not implement a load-balancer. Rather, it implements
 a failover mechanism so that a different node can take over should the current
 leader node fail for some reason.
 
-If the leader node fails for some reason, failover is automatic: the old
-leader's lease times out after 10 seconds, at which point another node becomes
-the leader and takes over ownership of the service IP.
+If the leader node fails for some reason, failover is automatic: the failed
+node is detected using [memberlist](https://github.com/hashicorp/memberlist),
+at which point new nodes take over ownership of the IP addresses from the
+failed node.
 
 ## Limitations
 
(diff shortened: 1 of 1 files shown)


website/content/concepts/layer2.md
```diff
@@ -77,10 +78,10 @@ Keepalived uses the Virtual Router Redundancy Protocol (VRRP). Instances of
...
 leader and to notice when that leader goes away.
 
-MetalLB on the other hand relies on Kubernetes to know when pods and nodes go up
-and down. It doesn't need to speak a separate protocol to select leaders,
-instead it just lets Kubernetes do most of the work of deciding which pods are
-healthy, and which nodes are ready.
+MetalLB on the other hand relies on
+[memberlist](https://github.com/hashicorp/memberlist) to know when a Node in
+the cluster is no longer reachable and the service IPs from that node should be
+moved elsewhere.
```
champtar, line 84: We still use the K8S API for the pods status, but not sure how to convey that

champtar:
A small nitpick but it can be merged as is also

website/content/concepts/layer2.md
russellb: Sure, how about this.  I'm not even sure if this is right, I'm guessing.  :-)

> MetalLB on the other hand relies on [memberlist](https://github.com/hashicorp/memberlist) to know when a Node in the cluster is no longer reachable or the Kubernetes API to determine if a MetalLB pod (speaker) is unhealthy. This information is used to determine if service IPs from that node should be moved elsewhere.

champtar: I was thinking of the pods that host the services, not the speaker pod.

russellb: Oh, I see what you meant.

Since this text mainly seemed about the failover part, I'm not sure how to mention that.  I think it's ok how it is.

russellb: ... and now that I've actually read the code to see how it works!  How about:

```
MetalLB on the other hand relies on
[memberlist](https://github.com/hashicorp/memberlist) to know when a Node in
the cluster is no longer reachable and the service IPs from that node should be
moved to another Node that hosts an endpoint for each affected Service.
```

rata: @champtar IMO this is fine. But not sure I understood what your concern was? :)

rata:
@russellb awesome, thanks a lot for the PR! When the review comment is resolved, we can merge. Don't hesitate to ping me :)

website/content/concepts/layer2.md
champtar: @russellb last proposal is good

rata: Cool, marking as solved, then! Thanks! :)
(let me know if I'm misunderstanding)

rata:
Updated the website with this change, is live now :)

Issue #302, opened 2018-08-15
Layer 2 controller function usableNodes() should return all nodes for "externalTrafficPolicy: Cluster" service


adidenko:
**Is this a bug report or a feature request?**:
Feature request

**What happened**:
If you run a service with `externalTrafficPolicy: Cluster` MetalLB speaker will run announcer only on nodes with at least one ready endpoint. Which makes perfect sense for `externalTrafficPolicy: Local`, but not required for services with 'externalTrafficPolicy: Cluster`.

**What you expected to happen**:
MetalLB speaker for service with `externalTrafficPolicy: Cluster` is started on any Ready node.

**How to reproduce it (as minimally and precisely as possible)**:
1. On a 2+ node cluster run MetalLB speaker with `nodeSelector` set to a part of nodes.
2. Run the Nginx deployment and service from https://raw.githubusercontent.com/google/metallb/v0.7.2/manifests/tutorial-2.yaml with `nodeSelector` set to the other part of nodes. I.e. make sure Nginx pod is not running on the same node as MetalLB speaker.
3. Speaker is not responding to ARP `who-has` requests for external IP.

**Anything else we need to know?**:
The typical use case is border ingress nodes. Which are supposed to terminate external traffic and load balance / ingress into the k8s cluster. Such nodes may have specific network configuration, etc which is different from the rest or worker nodes.

Making `usableNodes()` function take `externalTrafficPolicy` into account and return all nodes for `Cluster` type should solve this.

**Environment**:
- MetalLB version: v0.7.2
- Kubernetes version: 1.10.3
- BGP router type/version: layer2
- OS (e.g. from /etc/os-release): CentOS Linux release 7.5
- Kernel (e.g. `uname -a`): 3.10.0-862.3.2.el7.x86_64 #1 SMP Mon May 21 23:36:36 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux

adidenko:
It looks like there's no easy fix for this with the current (0.7.x) design. In order to implement this feature properly Speaker needs to either have leader election (per service or global like in 0.6.x) or some other external control mechanism (from Controller?).

So basically, it's not possible to use MetalLB with layer2 protocol on a subset of Kubernetes workers. It will work only if Speaker is running on all the nodes.

kfox1111:
bummer. I was hoping to do this too.

kfox1111:
[...]

PR #613, opened 2020-06-04
MetalLB speakers working on a subset of nodes in layer2.

thielemans:
I ran into issue #302 as our nodes run in two different subnets and I only want the speakers working on nodes with the public facing subnet. 

Validate by adding a nodeSelector on the speaker DaemonSet and label on the desired nodes.

Feel free to improve my code as it is my first time ever with golang.

### DIFF
--- a/speaker/layer2_controller.go
+++ b/speaker/layer2_controller.go
@@ -37,6 +37,16 @@ func (c *layer2Controller) SetConfig(log.Logger, *config.Config) error {
 	return nil
 }
 
+func activeNodes(mList *memberlist.Memberlist) []string {
+	var activeNodes []string
+	if mList != nil {
+		for _, n := range mList.Members() {
+		    activeNodes = append(activeNodes, n.Name)
+		}
+	}
+    return activeNodes
+}
+
 // usableNodes returns all nodes that have at least one fully ready
 // endpoint on them.
 func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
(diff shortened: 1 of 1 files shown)


speaker/layer2_controller.go
```diff
@@ -37,6 +37,16 @@ func (c *layer2Controller) SetConfig(log.Logger, *config.Config) error {
...
 
+func activeNodes(mList *memberlist.Memberlist) []string {
+	var activeNodes []string
+	if mList != nil {
+		for _, n := range mList.Members() {
+		    activeNodes = append(activeNodes, n.Name)
+		}
+	}
+    return activeNodes
+}
```
russellb, line 48: The code below in `usableNodes()` already builds up an `activeNodes` map.  Maybe just update that to return both the active and usable nodes?  You can return more than one thing in golang.

speaker/layer2_controller.go
```diff
@@ -37,6 +37,16 @@ func (c *layer2Controller) SetConfig(log.Logger, *config.Config) error {
 	return nil
 }
 
+func activeNodes(mList *memberlist.Memberlist) []string {
```
johananl, line 40: Could you please run `go fmt` on this file? It has inconsistent indentation.

russellb: I've raised #632 to track adding a CI job to catch this sort of thing automatically

WKBae:
I've faced the same issue and really hope this to be merged.
@thielemans, if you can't continue on, can I take over rest of the work?

rata:
@thielemans thanks! As I was saying [here][link], the use case is quite valid, but this is "by design" currently on MetalLB. See for example this issue comment: #302 (comment). So, this should probably go into a minor release (just to set more clear expectations on when this might be fixed: not in the 0.9 patch releases :)).

As @liuyuan10 was saying in the other PR, it probably makes sense to put this on hold until a solution for the memberlist re-join issue is merged. That will change this code quite a lot and is critical for the next patch release.

When v0.9.4 is out, we can investigate different ways (it might become even simpler) to have this functionality.

I'll add a do not merge label and we can revisit afterwards. Thanks again for the PR! :)

[link]: https://github.com/metallb/metallb/pull/665#issuecomment-662650650

PR #620, opened 2020-06-09
ci: Set up build and unit test CI jobs

russellb:
This change sets up two CI jobs using Github actions.  This
environment seems like a good fit since it's free, included with
github, and capable of running e2e CI jobs based on kind.

In addition to the CI jobs, this change adds a new "unit" target for
invoke so that it's easy to manually run the same commands that are
used in CI.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,66 @@
+name: metallb-ci
+
+on:
+  push:
+  pull_request:
+    branches: [ master ]
+
+env:
+  GO_VERSION: 1.13.4
+
+jobs:
+  build:
+    name: Build
+    runs-on: ubuntu-latest
+    steps:
+    - name: Set up Go
+      uses: actions/setup-go@v1
+      with:
+        go-version: ${{ env.GO_VERSION }}
+      id: go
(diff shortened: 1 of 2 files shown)


russellb:
A sample run is here: https://github.com/russellb/metallb/actions/runs/129294514

champtar:
Thanks for this !
would it be possible to build the containers and have them usable so bug reporter can test in their environment ?
Also bonus points if you can run tests on each separate commits

russellb:
> Thanks for this !
> would it be possible to build the containers and have them usable so bug reporter can test in their environment ?

I think it's possible.  I've filed an issue for the idea: https://github.com/metallb/metallb/issues/623

> Also bonus points if you can run tests on each separate commits

Interesting idea.  I agree that each commit should be functional on its own.  I'll have to think about the best way to do it and still have the results easy to interpret.

russellb:
I'm going to close this because I completely missed that CircleCI was already running against this repo.  CircleCI runs the unit tests, but also publishes container images post-merge.  At some point reporting on PRs stopped working, but I'm going to look at fixing that next.

PR #631, opened 2020-06-12
Out of date insall instructions for kustomize

wfhartford:
I'm not sure if referring to a URL was ever permitted in the resources section, but it is not possible in the kustomize version bundled with recent releases of `kubectl`. Instead, remote `kustomization.yaml` files can be referenced in the `bases:` section. I've also updated the tag to the current release, I'm not sure if there is a better way to ensure that the tag remains up to date.

### DIFF
--- a/website/content/installation/_index.md
+++ b/website/content/installation/_index.md
@@ -79,8 +79,10 @@ on the remote kustomization fle :
 # kustomization.yml
 namespace: metallb-system
 
+bases:
+  - https://github.com/metallb/metallb/manifests?ref=v0.9.3
+
 resources:
-  - github.com/danderson/metallb//manifests?ref=v0.8.2
   - configmap.yml 
   - secret.yml
 ```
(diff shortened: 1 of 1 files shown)


wfhartford:
It looks like I was wrong about this issue. What is out of date is the kustomize functionality built into kubectl. Later versions of kustomize allow remote bases to be listed under resources, and have deprecated the `bases` section all together.

I'm going to leave this PR open since it may be worth adding some explanation of this to the metallb documentation if kustomize is a recommended way to install metallb. I imaging many users, like me, will want to use the `kubectl apply -k` feature rather than install the standalone kustomize binary.

russellb:
@wfhartford PR #592 includes an addition to the docs to clarify the version of kustomize that is required.  Can you take a look at that PR and see if you think that's enough?

wfhartford:
Yes, that wording is great.

PR #633, opened 2020-06-15
Fix small bug in tasks.py error handling

caboteria:
A comma got used where a period was needed.


russellb:
good catch, lgtm.

cc @rata

rata:
@caboteria Thanks for the PR

@russellb thanks you for the review and cc too! :)

PR #634, opened 2020-06-15
circleci: Update branch filters for deploy jobs

russellb:
The config attempted to filter out PRs already, but for some reason it
didn't work.  This updated config just explicitly lists the branches
to run against: main and the v.* release branches.  This makes it skip
PRs for those jobs in my testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -55,7 +55,9 @@ workflows:
       - deploy-controller:
           filters:
             branches:
-              ignore: /pull\/.*/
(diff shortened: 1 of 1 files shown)


russellb:
After merging this, a maintainer will need to go into circleci into the project settings and turn on "Build forked pull requests", which is currently off.  Then we should start getting the test job running automatically against PRs.

rata:
@russellb thanks! I'll do that tomorrow! :)

PR #638, opened 2020-06-17
ci: Split lint out into its own job

russellb:
gometalinter was previously run in the test job which runs "go test".
There should be a slight runtime improvement by moving this into its
own job, but the main reason to split it out is so it shows up as its
own status line item when results are reported back to github.  When
something fails, you will be able to more quickly determine which area
the failure is in (linting vs the tests).

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -11,11 +11,18 @@ jobs:
       - run: sudo apt-get install python-pip
       - run: sudo pip install invoke semver pyyaml
       - run: echo "export GO111MODULE=on" >>$BASH_ENV
-      - run: curl -L https://git.io/vp6lP | sh # gometalinter
       - run: go test -short ./...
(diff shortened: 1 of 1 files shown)


russellb:
It's interesting that CircleCI didn't run on this change.  I guess that makes sense, as we don't want CI running any arbitrary test config change that hasn't been reviewed.  I did run this change against my fork though and it was successful there.

.circleci/config.yml
```diff
@@ -52,6 +59,10 @@ workflows:
           filters:
             tags:
               only: /.*/
+      - lint-1.13:
+          filters:
+            tags:
+              only: /.*/
```
rata, line 65: Got curious and checked why we have that: https://circleci.com/docs/2.0/configuration-reference/#tags

It seems to be to run on tags too :)

PR #640, opened 2020-06-18
Fix the README "widgets"

caboteria:
A trivial fix, but as a metallb newbie I was distracted by the broken "build" widget.  Since metallb's builds are actually passing it's a shame for other first-timers to get the wrong impression.


russellb:
Thanks!  These are also fixed in #570, but that one was held up because the changes were a bit more broad.  This one looks safe to merge right away.

cc @rata

rata:
@caboteria thanks a lot for the PR!
@russellb and thaks for the review!

PR #644, opened 2020-06-20
Add exec permission to controller in docker image

### LINKED ISSUE #643: standard_init_linux.go:211: exec user process caused "permission denied"
I got this error when running a local built metallb-controller in k8s (v1.17.4)

The root cause is that output binary of "go build" is of permission 750. When copied into docker image, the permission is kept but owner is changed to "root".

Steps to reproduce:
inv build -t 1
docker run -it --entrypoint /bin/sh  metallb/controller:1-amd64
ls -l controller
-rwxr-x---    1 root     root      34229033 Jun 18 03:35 controller

We are running controller as non-root user in metallb.yaml so it doesn't have execute permission.  The public metallb docker image doesn't have this issue. So I'm wondering if it's an issue for me only. Should we add a "chmod a+x" into the Dockerfile?

golang version:
go version go1.13.4 linux/amd64
docker version:
Client: Docker Engine - Community
 Version:           19.03.5
 API version:       1.40
 Go version:        go1.12.12
 Git commit:        633a0ea838
 Built:             Wed Nov 13 07:25:58 2019
 OS/Arch:           linux/amd64
 Experimental:      false

Server: Docker Engine - Community
 Engine:
  Version:          19.03.5
  API version:      1.40 (minimum version 1.12)
  Go version:       go1.12.12
  Git commit:       633a0ea838
  Built:            Wed Nov 13 07:24:29 2019
  OS/Arch:          linux/amd64
  Experimental:     false
 containerd:
  Version:          1.2.10
  GitCommit:        b34a5c8af56e510852c35414db4c1f4fa6172339
 runc:
  Version:          1.0.0-rc8+dev
  GitCommit:        3e425f80a8c931f88e6d94a8c831b9d5aa481657
 docker-init:
  Version:          0.18.0
  GitCommit:        fec3683

liuyuan10:
metallb.yaml runs controller as non-root user but in docker image,
controller binary is owned by root. Because docker ADD keeps file
permission, this change makes sure that /controller can be exeucted by
non-root user in the docker image.

fixes #643

   accepted.
   the PR is merged.

### DIFF
--- a/controller/Dockerfile
+++ b/controller/Dockerfile
@@ -1,4 +1,6 @@
 FROM alpine:latest
 
 ADD controller /controller
+# The file is owned by root. Make sure non-root user can execute it
+RUN chmod a+x /controller
 ENTRYPOINT ["/controller"]


rata:
I'd like to understand under which setups this is needed, to see if other things might need fixing too, but seems fine to add this line if it solves a problem for someone :-)

cc @johananl @russellb @daxmc99 WDYT? :)

rata:
LGTM. I'd add a comment mentioning `umask`, though, so we don't forget why we changed it in the first place :)

controller/Dockerfile
```diff
@@ -1,4 +1,6 @@
 FROM alpine:latest
 
 ADD controller /controller
+# The file is owned by root. Make sure non-root user can execute it
```
rata, line 4:
```suggestion
# When running as non root and building in an environment that `umask` masks out '+x' for others, it won't be possible to execute. Make sure all can execute, just in case
```

rata:
If you don't mind, I'll amend that simple comment myself and merge the ammended version :)

controller/Dockerfile
rata: Amended this myself :)

PR #647, opened 2020-06-21
RFC: Layer2: Make nodes re-join memberlist cluster

rata:
This is a quick&dirty patch (probably can use only one channel, channel size is a random number, the mlJoin has an arbitrary number of retries without much thought, need to add and adapt tests and ton's several other things) to make nodes join the memberlist clusters. But I think it is enough to get the gist of what I was trying to say in the other PR.

As this is a quick&dirty PR, it is not the idea to discuss the specifics (like re-join after 5 min vs 1 min, etc.) but to see the general approach. 

I cherry-picked the first 2 commits from https://github.com/metallb/metallb/pull/595, that split the memberlist code in a separate package. I've added this re-join logic on top. If you see only the last commit, you will see my proposed approach. Looking at all the changes, without spiting by commit, will be hard to spot what I say here. Please, consider looking at the last patch only.

With this idea, a node can re-join a cluster when:
 * The k8s node object changes and the `SetNode()` is invoked
 * 5 min have passed and tries to re-join (5 min is a totally arbitrary number that I haven't thought about. It is just to show the idea).

The internal kubernetes watcher already has this logic to notify when the node this is running changes (not other nodes in the cluster, it has a selector to be only about changes to **this node** where this is running). In the Layer2 case, though, it was just a "return nil". Now, it queues the message to process a re-join.

So, if a node losses network connectivity it will leave the memberlist cluster and eventually kubernetes will mark it as not ready. If connection comes back, the node will change to ready state and the `SetNode()` callback will be called in that node only, as the node changed. And then, join the cluster.

In this way, we are using memberlist to detect node failure and k8s to detect when to join again. This is not ideal, but memberlist doesn't seem to support the re-join mechanism. IMHO, we can look for more improvements later (I have several ideas :)).

However, if the node losses network connectivity and is restored before kubernetes marks is as not ready, the `SetNode()` callback will not be invoked. In this case, there is the periodic "try to join memberlist" to catch those errors.

@champtar sorry if I wasn't clear on the other PR. This is what I meant. It is basically just re-joining with the same logic it is today on master, but with some retries, but on node changes and periodically. What do you think? Does it make sense? Am I missing something?

@johananl @russellb @daxmc99 Please have a look at the approach, if you have some time :)

### DIFF
--- a/internal/speakerlist/speakerlist.go
+++ b/internal/speakerlist/speakerlist.go
@@ -1,0 +1,146 @@
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//      http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package speakerlist
+
+import (
+	"crypto/sha256"
+	golog "log"
+	"strconv"
+	"time"
+
+	"go.universe.tf/metallb/internal/k8s"
+
+	gokitlog "github.com/go-kit/kit/log"
+	"github.com/hashicorp/memberlist"
+)
+
+// SpeakerList gives you the list of healthy speaker
+type SpeakerList struct {
+	l         gokitlog.Logger
+	client    *k8s.Client
+	stopCh    chan struct{}
+	namespace string
+	labels    string
+	// following fields are nil when MemberList is disabled
+	mlEventCh chan memberlist.NodeEvent
+	mList     *memberlist.Memberlist
+}
+
+// New creates a new SpeakerList
+func New(logger gokitlog.Logger, nodeName, bindAddr, bindPort, secret, namespace, labels string, stopCh chan struct{}) (*SpeakerList, error) {
+	sl := SpeakerList{
+		l:         logger,
+		stopCh:    stopCh,
+		namespace: namespace,
+		labels:    labels,
+	}
+
+	if namespace == "" || labels == "" || bindAddr == "" {
+		logger.Log("op", "startup", "msg", "Not starting fast dead node detection (MemberList), need ml-bindaddr / ml-labels / ml-namespace config")
+		return &sl, nil
+	}
+
+	mconfig := memberlist.DefaultLANConfig()
+	// mconfig.Name MUST be spec.nodeName, as we will match it against Enpoints nodeName in usableNodes()
+	mconfig.Name = nodeName
+	mconfig.BindAddr = bindAddr
+	if bindPort != "" {
+		mlport, err := strconv.Atoi(bindPort)
+		if err != nil {
+			sl.l.Log("op", "startup", "error", "unable to parse ml-bindport", "msg", err)
+			return nil, err
+		}
+		mconfig.BindPort = mlport
+		mconfig.AdvertisePort = mlport
+	}
+	loggerout := gokitlog.NewStdlibAdapter(gokitlog.With(sl.l, "component", "MemberList"))
+	mconfig.Logger = golog.New(loggerout, "", golog.Lshortfile)
+	if secret == "" {
+		sl.l.Log("op", "startup", "warning", "no ml-secret-key set, MemberList traffic will not be encrypted")
+	} else {
+		sha := sha256.New()
+		mconfig.SecretKey = sha.Sum([]byte(secret))[:16]
+	}
+	// ChannelEventDelegate hint that it should not block, so make mlEventCh 'big'
+	sl.mlEventCh = make(chan memberlist.NodeEvent, 1024)
+	mconfig.Events = &memberlist.ChannelEventDelegate{Ch: sl.mlEventCh}
+	var err error
+	sl.mList, err = memberlist.Create(mconfig)
+	if err != nil {
+		sl.l.Log("op", "startup", "error", err, "msg", "failed to create memberlist")
+		return nil, err
+	}
+	return &sl, nil
+}
+
+// Start starts the needed goroutines
+func (sl *SpeakerList) Start(client *k8s.Client) error {
+	sl.client = client
+
+	if sl.mList == nil {
+		return nil
+	}
+
+	go sl.memberlistWatchEvents()
+
+	iplist, err := sl.client.GetPodsIPs(sl.namespace, sl.labels)
+	if err != nil {
+		sl.l.Log("op", "startup", "error", err, "msg", "failed to get PodsIPs")
+		return err
+	}
+	n, err := sl.mList.Join(iplist)
+	sl.l.Log("op", "startup", "msg", "Memberlist join", "nb joigned", n, "error", err)
+
+	return nil
+}
+
+// UsableSpeakers return a map of usable nodes
+func (sl *SpeakerList) UsableSpeakers() map[string]bool {
+	if sl.mList == nil {
+		return nil
+	}
+	activeNodes := map[string]bool{}
+	for _, n := range sl.mList.Members() {
+		activeNodes[n.Name] = true
+	}
+	return activeNodes
+}
+
+// Stop properly Leave / Shutdown MemberList cluster
+func (sl *SpeakerList) Stop() {
+	if sl.mList == nil {
+		return
+	}
+	sl.l.Log("op", "shutdown", "msg", "leaving MemberList cluster")
+	err := sl.mList.Leave(time.Second)
+	sl.l.Log("op", "shutdown", "msg", "left MemberList cluster", "error", err)
+	err = sl.mList.Shutdown()
+	sl.l.Log("op", "shutdown", "msg", "MemberList shutdown", "error", err)
+}
+
+func event2String(e memberlist.NodeEventType) string {
+	return []string{"NodeJoin", "NodeLeave", "NodeUpdate"}[e]
+}
+
+func (sl *SpeakerList) memberlistWatchEvents() {
+	for {
+		select {
+		case e := <-sl.mlEventCh:
+			sl.l.Log("msg", "Node event", "node addr", e.Node.Addr, "node name", e.Node.Name, "node event", event2String(e.Event))
+			sl.l.Log("msg", "Call Force Sync")
+			sl.client.ForceSync()
+		case <-sl.stopCh:
+			return
+		}
+	}
+}
--- a/internal/speakerlist/speakerlist.go
+++ b/internal/speakerlist/speakerlist.go
@@ -32,8 +32,10 @@ type SpeakerList struct {
 	namespace string
 	labels    string
 	// following fields are nil when MemberList is disabled
-	mlEventCh chan memberlist.NodeEvent
-	mList     *memberlist.Memberlist
+	mlEventCh    chan memberlist.NodeEvent
+	mList        *memberlist.Memberlist
+	mlJoinCh     chan struct{}
+	mlJoinStopCh chan struct{}
 }
 
 // New creates a new SpeakerList
@@ -71,6 +73,12 @@ func New(logger gokitlog.Logger, nodeName, bindAddr, bindPort, secret, namespace
 		sha := sha256.New()
 		mconfig.SecretKey = sha.Sum([]byte(secret))[:16]
 	}
+
+	// Should not block, so make it 'big'
+	sl.mlJoinCh = make(chan struct{}, 1024)
+
+	sl.mlJoinStopCh = make(chan struct{})
+
 	// ChannelEventDelegate hint that it should not block, so make mlEventCh 'big'
 	sl.mlEventCh = make(chan memberlist.NodeEvent, 1024)
 	mconfig.Events = &memberlist.ChannelEventDelegate{Ch: sl.mlEventCh}
@@ -92,16 +100,58 @@ func (sl *SpeakerList) Start(client *k8s.Client) error {
 	}
 
 	go sl.memberlistWatchEvents()
+	go sl.mlJoinLoop()
 
+	return nil
+}
+
+func (sl *SpeakerList) mlJoinLoop() {
+	ticker := time.NewTicker(5 * time.Minute)
+
+	for {
+		select {
+		case <-sl.mlJoinStopCh:
+			return
+		case <-ticker.C:
+			sl.mlJoin()
+		case <-sl.mlJoinCh:
+			sl.mlJoin()
+		}
+	}
+}
+
+func (sl *SpeakerList) mlJoin() (nr int, err error) {
 	iplist, err := sl.client.GetPodsIPs(sl.namespace, sl.labels)
 	if err != nil {
 		sl.l.Log("op", "startup", "error", err, "msg", "failed to get PodsIPs")
-		return err
+		return 0, err
 	}
-	n, err := sl.mList.Join(iplist)
-	sl.l.Log("op", "startup", "msg", "Memberlist join", "nb joigned", n, "error", err)
 
-	return nil
+	for i := 0; i < 6; i++ {
+		nr, err = sl.mList.Join(iplist)
+		if err != nil || len(iplist) != nr {
+			sl.l.Log("op", "Member detection", "msg", "Memberlist join", "nb joigned", nr, "error", err)
+			time.Sleep(10 * time.Second)
+			break
+		}
+
+		sl.l.Log("op", "Member detection", "msg", "Memberlist join succesfully", "number of other nodes", nr)
+		return
+	}
+
+	return
+}
+
+// ReJoin initiates a discovery and join of all cluster speakers.
+func (sl *SpeakerList) ReJoin() {
+	if sl.mList == nil {
+		return
+	}
+
+	sl.l.Log("op", "Member detection", "msg", "Sending message to re-process cluster members")
+	sl.mlJoinCh <- struct{}{}
+	sl.l.Log("op", "Member detection", "msg", "Sent message to re-process cluster members")
+	return
 }
 
 // UsableSpeakers return a map of usable nodes
@@ -140,6 +190,7 @@ func (sl *SpeakerList) memberlistWatchEvents() {
 			sl.l.Log("msg", "Call Force Sync")
 			sl.client.ForceSync()
 		case <-sl.stopCh:
+			close(sl.mlJoinStopCh)
 			return
 		}
 	}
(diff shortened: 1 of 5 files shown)


champtar:
Hello @rata,

Your version will work, but has a ton of limitations:

- Your version doesn't remove the reference to k8s.Client in SpeakerList, this is what make it much smaller.
- SetNode is only called when the local node status change, so if the node is disconnected for just 35s (not ready is 40s) it will timeout on the MemberList side and maybe no try to reconnect for ~ 4min, anything more that 10s is outside of my design goals.
- In your version you can't rejoin if the API is down, mine retry with the latest list available.
I guess you chose 5min to not do too many requests to the API, I'm using a watch to always have instant response with the minimum requests possible
- In mlJoin() you are joining all the ips and not just the missing ones, MemberList.Join() is not a noop for ips that are already members (it takes time), and the join is sequential.
- I also fix a long standing issue that when memberlist is disabled the speaker selection code consider that all nodes are running a speaker.
- You are still not addressing the re spam issue

Maybe instead of focusing on the diff, you should just review the final code, because yes my PR is huge but I cleanup dirty hacks that I introduced in the first version of memberList, and the end result is cleaner, more independent packages.

rata:
> Hello @rata,

Hi, thanks for the comments. Here go some other questions to those, I think understanding the answer might help me to better understand some things that were never mentioned in the PR or comments.

> Your version will work, but has a ton of limitations:

Yes, as I mentioned :)

> 
>     * Your version doesn't remove the reference to k8s.Client in SpeakerList, this is what make it much smaller.

Well, several other things too. Basically that this only tries to fix the re-join issue without major semantic changes.

> 
>     * SetNode is only called when the local node status change, so if the node is disconnected for just 35s (not ready is 40s) it will timeout on the MemberList side and maybe no try to reconnect for ~ 4min, anything more that 10s is outside of my design goals.

Yes, I described that case in the PR description. Can you please elaborate why anything else than 10 seconds wouldn't be useful? Like in real world scenarios. I think knowing why you think that can be really useful and understand the consequences in your cluster(s) would be great :)

>     * In your version you can't rejoin if the API is down, mine retry with the latest list available.
>       I guess you chose 5min to not do too many requests to the API, I'm using a watch to always have instant response with the minimum requests possible

Well, but isn't that watching at services? I don't think it will return anything new when the API server is down. It might be possible to join nodes that were reported before the API server being down, but is only in that race. Am I missing something?

Keep the last list of IPs can work too. I'd like to understand why we need it. Can you please elaborate why you think it is needed in real world scenarios? Or just a nice improvement?

>     * In mlJoin() you are joining all the ips and not just the missing ones, MemberList.Join() is not a noop for ips that are already members (it takes time), and the join is sequential.

This is the kind of things that I'm not worried about at this stage, while very useful to know.

I asked about the general approach. The "diff" between members and all can be calculated if needed.

> 
>     * I also fix a long standing issue that when memberlist is disabled the speaker selection code consider that all nodes are running a speaker.

Can you please explain the bug or point to the fix? Didn't see it clearly by looking at the 

IIUC, this is a design decision of MetalLB and there might be other things in the way. Is a valid use case, and I'd like to support it, but completely unrelated to re-joining memberlist cluster.

But I'd like to keep this of topic for the re-join discussion, unless you think it is needed to discuss on the approaches to do the  re-join :-)

>     * You are still not addressing the re spam issue

No. This is a RFC for the re-join issue.

champtar:
Hi @rata 

> > Hello @rata,
> 
> Hi, thanks for the comments. Here go some other questions to those, I think understanding the answer might help me to better understand some things that were never mentioned in the PR or comments.
> 
> > Your version will work, but has a ton of limitations:
> 
> Yes, as I mentioned :)
> 
> > ```
> > * Your version doesn't remove the reference to k8s.Client in SpeakerList, this is what make it much smaller.
> > ```
> 
> Well, several other things too. Basically that this only tries to fix the re-join issue without major semantic changes.

But are all the other issues worth fixing, I think they are, and I'm not sure you will end up with something drastically simpler if you do

> 
> > ```
> > * SetNode is only called when the local node status change, so if the node is disconnected for just 35s (not ready is 40s) it will timeout on the MemberList side and maybe no try to reconnect for ~ 4min, anything more that 10s is outside of my design goals.
> > ```
> 
> Yes, I described that case in the PR description. Can you please elaborate why anything else than 10 seconds wouldn't be useful? Like in real world scenarios. I think knowing why you think that can be really useful and understand the consequences in your cluster(s) would be great :)

One of the use case of MetalLB for me is LoadBalancer to have a VIP for a request router service (an http server that does 302 redirect to another server) for Live OTT content (HLS/Dash video)
Good video player are retrying on error but not for long, for sure after 10s the end user will get an error and have to retry, which is not a great user experience. We were using ucarp before K8S, so switch time of around 1s, I don't want to try to justify 30s+ outages when I can do 10s-

> 
> > ```
> > * In your version you can't rejoin if the API is down, mine retry with the latest list available.
> >   I guess you chose 5min to not do too many requests to the API, I'm using a watch to always have instant response with the minimum requests possible
> > ```
> 
> Well, but isn't that watching at services? I don't think it will return anything new when the API server is down.

Correct

> It might be possible to join nodes that were reported before the API server being down, but is only in that race. Am I missing something?

Yes, if they are alive we will join them, if not we will fail to, but no harm trying. Also you just need to rejoin the 2 parts of the cluster

> Keep the last list of IPs can work too.

That's what I do, always keep the last list

> I'd like to understand why we need it. Can you please elaborate why you think it is needed in real world scenarios? Or just a nice improvement?

Trying as hard as possible to keep everything running even when the API is down isn't a bad thing IMO

> > ```
> > * In mlJoin() you are joining all the ips and not just the missing ones, MemberList.Join() is not a noop for ips that are already members (it takes time), and the join is sequential.
> > ```
> 
> This is the kind of things that I'm not worried about at this stage, while very useful to know.
> I asked about the general approach. The "diff" between members and all can be calculated if needed.

The more details like that you handle the closer you will be from my code ;)

> > ```
> > * I also fix a long standing issue that when memberlist is disabled the speaker selection code consider that all nodes are running a speaker.
> > ```
> 
> Can you please explain the bug or point to the fix? Didn't see it clearly by looking at the

With MemberList disabled, in current master code, we take the list of Endpoints, take the nodes they are running on, hash them, and pick the first one.
At no point we filter out the nodes that have no speakers deployed.


> IIUC, this is a design decision of MetalLB and there might be other things in the way. Is a valid use case, and I'd like to support it, but completely unrelated to re-joining memberlist cluster.

Yes, but it's so easy to do with my code refactor there is no point in not doing it
SetSpeakers() update the list when needed, so you can use it in UsableSpeakers(), even when MemberList is off

> But I'd like to keep this of topic for the re-join discussion, unless you think it is needed to discuss on the approaches to do the re-join :-)

Maybe consider my re-join PR as a V2 of the initial MemberList PR, I'm refactoring a bunch, but it's for the best
Small targeted patch are nice when you need to backport on many version, we are not doing that for MetalLB

At some point that's open source without dedicated time, if you want to have more confidence, do an RC and ask people to test
I'll be here to fix my bugs

> > ```
> > * You are still not addressing the re spam issue
> > ```
> 
> No. This is a RFC for the re-join issue.

That is still a blocker issue for me, that need to be fixed.

Hope I answered everything

rata:
> > > ```
> > > * SetNode is only called when the local node status change, so if the node is disconnected for just 35s (not ready is 40s) it will timeout on the MemberList side and maybe no try to reconnect for ~ 4min, anything more that 10s is outside of my design goals.
> > > ```
> > 
> > 
> > Yes, I described that case in the PR description. Can you please elaborate why anything else than 10 seconds wouldn't be useful? Like in real world scenarios. I think knowing why you think that can be really useful and understand the consequences in your cluster(s) would be great :)
> 
> One of the use case of MetalLB for me is LoadBalancer to have a VIP for a request router service (an http server that does 302 redirect to another server) for Live OTT content (HLS/Dash video)
> Good video player are retrying on error but not for long, for sure after 10s the end user will get an error and have to retry, which is not a great user experience. We were using ucarp before K8S, so switch time of around 1s, I don't want to try to justify 30s+ outages when I can do 10s-

But dead node detection is as fast as memberlist can do with this config, so you won't have **more** downtime if you have more than one node running MetalLB speakers. Right? Memberlist will detect the dead node in a fast way in any case and remove it So, what is the **real** problem? Do you have only one MetalLB speaker? Hmm, even on that case, the node always has itself on the cluster, so it shouldn't be an issue IIUC.

I really want to understand what is the problem of having a worst case not very fast re-join if the network downtime was in the interval `(<memberlist consider node dead>, <kubernetes consider node not ready>)`. The downtime is the same: as fast as memberlist can detect. This is pretty fast and the new node is doing the announce. The only thing that changes is when that node is added again to the memberlist cluster. What is it so bad about having a worst case to re-join not so fast?

> > It might be possible to join nodes that were reported before the API server being down, but is only in that race. Am I missing something?
> 
> Yes, if they are alive we will join them, if not we will fail to, but no harm trying. Also you just need to rejoin the 2 parts of the cluster

Okay, we can investigate later why this is or not worth and the effects. Not something I'm worried at this stage about the approach in general, as the approach can use that if we want to.

> > > ```
> > > * I also fix a long standing issue that when memberlist is disabled the speaker selection code consider that all nodes are running a speaker.
> > > ```
> > 
> > 
> > Can you please explain the bug or point to the fix? Didn't see it clearly by looking at the
> 
> With MemberList disabled, in current master code, we take the list of Endpoints, take the nodes they are running on, hash them, and pick the first one.
> At no point we filter out the nodes that have no speakers deployed.

Right. That fix is independent of the rejoin, of course. At least if we go with this approach, but unless it is a side effect of re-joining, I'd want different commits for different things.

> > IIUC, this is a design decision of MetalLB and there might be other things in the way. Is a valid use case, and I'd like to support it, but completely unrelated to re-joining memberlist cluster.
> 
> Yes, but it's so easy to do with my code refactor there is no point in not doing it
> SetSpeakers() update the list when needed, so you can use it in UsableSpeakers(), even when MemberList is off

Oh, that is nice. But the changes to monitor ourselves in that PR seem very intrusive and I'd prefer something simpler. I can elaborate on the PR, as it is off topic here.

> Maybe consider my re-join PR as a V2 of the initial MemberList PR, I'm refactoring a bunch, but it's for the best
> Small targeted patch are nice when you need to backport on many version, we are not doing that for MetalLB

And also for patch releases, that should be safe to update without reading much.

It might be worth doing some refactors, like using services for Layer 2 speakers, but doing all together, combined with changing several stuff to use channels, etc. is not something that I'd like to do in a patch release. And to be honest, if the code can't be cleaner, not sure if worth doing. I'd prefer to separate the improvements from bug fixes, so we can fix bugs faster and think improvements clearly.

But this is getting off-topic to understanding the pros/cons of the approach I proposed here. So, enough about that here :)

> At some point that's open source without dedicated time, if you want to have more confidence, do an RC and ask people to test
> I'll be here to fix my bugs

Thanks :)

> > No. This is a RFC for the re-join issue.
> 
> That is still a blocker issue for me, that need to be fixed.

Yes, bugs will be fixed. it is just off-topic here.

champtar:
> > > > ```
> > > > * SetNode is only called when the local node status change, so if the node is disconnected for just 35s (not ready is 40s) it will timeout on the MemberList side and maybe no try to reconnect for ~ 4min, anything more that 10s is outside of my design goals.
> > > > ```
> > > 
> > > 
> > > Yes, I described that case in the PR description. Can you please elaborate why anything else than 10 seconds wouldn't be useful? Like in real world scenarios. I think knowing why you think that can be really useful and understand the consequences in your cluster(s) would be great :)
> > 
> > 
> > One of the use case of MetalLB for me is LoadBalancer to have a VIP for a request router service (an http server that does 302 redirect to another server) for Live OTT content (HLS/Dash video)
> > Good video player are retrying on error but not for long, for sure after 10s the end user will get an error and have to retry, which is not a great user experience. We were using ucarp before K8S, so switch time of around 1s, I don't want to try to justify 30s+ outages when I can do 10s-
> 
> But dead node detection is as fast as memberlist can do with this config, so you won't have **more** downtime if you have more than one node running MetalLB speakers. Right? Memberlist will detect the dead node in a fast way in any case and remove it So, what is the **real** problem? Do you have only one MetalLB speaker? Hmm, even on that case, the node always has itself on the cluster, so it shouldn't be an issue IIUC.

Switches goes down say 1 min and comes back up (bad netadmin manipulation), and the API is also down, in that case every node is main, in your version we will not recover from that before we have the API back

> I really want to understand what is the problem of having a worst case not very fast re-join if the network downtime was in the interval `(<memberlist consider node dead>, <kubernetes consider node not ready>)`. The downtime is the same: as fast as memberlist can detect. This is pretty fast and the new node is doing the announce. The only thing that changes is when that node is added again to the memberlist cluster. What is it so bad about having a worst case to re-join not so fast?

MemberList does the dead node detection, we have to implement the recovery from split brain. If the VIP flaps between multiple nodes you will get timeouts or tcp reset

> 
> > > It might be possible to join nodes that were reported before the API server being down, but is only in that race. Am I missing something?
> > 
> > 
> > Yes, if they are alive we will join them, if not we will fail to, but no harm trying. Also you just need to rejoin the 2 parts of the cluster
> 
> Okay, we can investigate later why this is or not worth and the effects. Not something I'm worried at this stage about the approach in general, as the approach can use that if we want to.

This is part of the re-join problem, I don't see why you want to think about it later. Trying really hard to work while the API is down doesn't cost anything more


> 
> > > > ```
> > > > * I also fix a long standing issue that when memberlist is disabled the speaker selection code consider that all nodes are running a speaker.
> > > > ```
> > > 
> > > 
> > > Can you please explain the bug or point to the fix? Didn't see it clearly by looking at the
> > 
> > 
> > With MemberList disabled, in current master code, we take the list of Endpoints, take the nodes they are running on, hash them, and pick the first one.
> > At no point we filter out the nodes that have no speakers deployed.
> 
> Right. That fix is independent of the rejoin, of course. At least if we go with this approach, but unless it is a side effect of re-joining, I'd want different commits for different things.

This is a side effect of having SetSpeakers and having the last K8S speaker list, it allow to have UsableSpeakers() response that make sense even without MemberList
I can try to split it out, but as said before, you are reviewing my PR as if it was bug fix only, when it's a V2 that redo everything.

> 
> > > IIUC, this is a design decision of MetalLB and there might be other things in the way. Is a valid use case, and I'd like to support it, but completely unrelated to re-joining memberlist cluster.
> > 
> > 
> > Yes, but it's so easy to do with my code refactor there is no point in not doing it
> > SetSpeakers() update the list when needed, so you can use it in UsableSpeakers(), even when MemberList is off
> 
> Oh, that is nice. But the changes to monitor ourselves in that PR seem very intrusive and I'd prefer something simpler. I can elaborate on the PR, as it is off topic here.

Nop, look at it again, I'm trading the ugly GetPodsIPs hack for SetSpeakers, a good part of the changes are line formatting, then changes in SpeakerList keep the UsableNodes up to date

Again, review SpeakerList in one go


> 
> > Maybe consider my re-join PR as a V2 of the initial MemberList PR, I'm refactoring a bunch, but it's for the best
> > Small targeted patch are nice when you need to backport on many version, we are not doing that for MetalLB
> 
> And also for patch releases, that should be safe to update without reading much.
> 
> It might be worth doing some refactors, like using services for Layer 2 speakers, but doing all together, combined with changing several stuff to use channels, etc. is not something that I'd like to do in a patch release.

You don't need to do a patch release ...

> And to be honest, if the code can't be cleaner, not sure if worth doing. I'd prefer to separate the improvements from bug fixes, so we can fix bugs faster and think improvements clearly.

The code is way cleaner, GetPodsIPs should never have been merged, but now that I'm trying to clean things up it's too much change.

I don't see anyone working on all the issues that I fix, if you don't want my code fine, but in this case step up and fix things, don't write RFC that handle 1/4 of the problems, and then justify it by saying it's an RFC, or that it'll be addressed in a separate PR, no it will not, stop lying to yourself, you don't have the bandwidth now, and you will not have it later.

> 
> But this is getting off-topic to understanding the pros/cons of the approach I proposed here. So, enough about that here :)
> 
> > At some point that's open source without dedicated time, if you want to have more confidence, do an RC and ask people to test
> > I'll be here to fix my bugs
> 
> Thanks :)
> 
> > > No. This is a RFC for the re-join issue.
> > 
> > 
> > That is still a blocker issue for me, that need to be fixed.
> 
> Yes, bugs will be fixed. it is just off-topic here.

johananl:
>I don't see anyone working on all the issues that I fix, if you don't want my code fine, but in this case step up and fix things, don't write RFC that handle 1/4 of the problems, and then justify it by saying it's an RFC, or that it'll be addressed in a separate PR, no it will not, stop lying to yourself, you don't have the bandwidth now, and you will not have it later.

@champtar we do appreciate your contributions to the project. We also do our best to review issues and PRs in a timely manner despite having, like yourself, a limited time to dedicate to the project. That said, I'd like to ask you to stay respectful and keep the communication constructive even if you aren't satisfied with the way the maintainers handle something.

champtar:
I'll try to be more constructive, even if after 220 messages exchanged and 2 months, I feel we haven't made much progress.
I don't view my previous message as disrespectful, pretty direct yes, but I'm not a native speaker, so sorry if I offended you @rata  

My biggest problem is this insistence on doing a patch release, I genuinely don't see the benefit for the users, they will just have a new version with less fixes and it'll just take longer to have all fixes

johananl:
> My biggest problem is this insistence on doing a patch release, I genuinely don't see the benefit for the users, they will just have a new version with less fixes and it'll just take longer to have all fixes

As maintainers we have a responsibility to the users to keep MetalLB as reliable as possible as well as improving the project by adding new functionality. These two things naturally contradict each other: when you add new code, you risk lowering the reliability.
In addition to reducing convergence time in layer 2 mode, the memberlist feature as originally merged had the side effect of reducing the reliability of MetalLB in layer 2 mode (very possibly due to not enough scrutiny on behalf of the maintainers, not trying to point any fingers at you personally). In order to correct that we have to be careful with the changes we introduce as bug fixes. @rata and I have discussed this and decided that making a patch release is the best course of action because it should allow us to address outstanding bugs without radically changing the implementation of the feature. Since merging *all* of the changes you've proposed will likely take time seeing that the proposed code is controversial (like you said - over 200 messages...), we would like to issue a patch release as soon as possible to address the most critical bugs (e.g. memberlist rejoin) and release any bigger changes in a minor release later. We'd be happy to have you on board.

champtar:
See #657, let's merge small bits at a time

rata:
> Switches goes down say 1 min and comes back up (bad netadmin manipulation), and the API is also down, in that case every node is main, in your version we will not recover from that before we have the API back

Oh, good point. I'll try to cache/store the IPs then.

> I'll try to be more constructive, even if after 220 messages exchanged and 2 months, I feel we haven't made much progress.
> I don't view my previous message as disrespectful, pretty direct yes, but I'm not a native speaker, so sorry if I offended you @rata

Ok, no hard feelings

Will open another PR more polished and with the cached IPs, closing this one.

Issue #489, opened 2019-10-30
Fix release script for post-helm world.


danderson:
After removing helm and adding kustomize, the release script is missing a couple of renaming points for version numbers:

- `manifests/metallb.yaml` needs to be updated (used to be done by updating helm chart and regenerating from there).
- Kustomize install options burn the version into a query parameter, which the regexes don't know to look for.

russellb:
commit 9094482fa fixed the manifest version number update.

It doesn't look like anything fixes the version number in the kustomize install docs, though.

rata:
@russellb thanks! I missed this issue and created another one for the second item: https://github.com/metallb/metallb/issues/594.

Will close that and just keep this

PR #650, opened 2020-06-27
tasks.py: Update kustomize version on release

rata:
While we are there, update the github link to use the metallb org.

Fixes: #489

I've tested this locally, running the `inv release --skip-release-notes 0.9.4`. Make sure this commit is on `main` _and_ `v0.9` branch to test, though. Iit worked just fine here.

To undo the release created locally, you can run: `git checkout  v0.9 && git tag -d v0.9.4 && git reset --hard HEAD^`

There are some PRs open to update the kustomize instructions, so hopefully this might be removed soon. For now, it's better to have the proper version on new releases :)


caboteria:
As long as you're updating website/content/installation/_index.html you might want to change "on the remote kustomization fle" to "to the remote kustomization file" on line 90.

rata:
@daxmc99 thanks for the review!

@caboteria There are open PRs about kustomize. I think we might need to remove that completely. Not familiar with kustomize, but what we have in the repo seems to be for kustomize 2.x and docs for 3.x (or something like that). So, I'd let that to the PR cleaning it up :)

PR #651, opened 2020-06-27
Add command-line flags to run processes outside of k8s

caboteria:
When you're developing, running components at the command line and
having them attach to a cluster might be more convenient than
redeploying them to a cluster over and over.

For the controller, -kubeconfig and -config-ns are needed.  Speakers
need those and -node-name.

For example:

 metallb$ go run ./controller/main.go ./controller/service.go -config-ns metallb-system -kubeconfig $KUBECONFIG

 metallb$ go run ./speaker/main.go ./speaker/*controller.go -config-ns metallb-system -kubeconfig $KUBECONFIG -node-name node0

   accepted.
   the PR is merged.

### DIFF
--- a/website/content/community/_index.md
+++ b/website/content/community/_index.md
@@ -116,7 +116,7 @@ To develop MetalLB, you'll need a couple of pieces of software:
 >versions may not work since there have been breaking changes between minor
 >versions.
 
-## Building the code
+## Building and running the code
 
 Start by fetching the MetalLB repository, with `git clone
 https://github.com/danderson/metallb`.
@@ -126,6 +126,19 @@ registries, and so forth. `inv -l` lists the available tasks.
 
 To build and deploy MetalLB to a local development environment using a kind
 cluster, run `inv dev-env`.
+
+When you're developing, running components at the command line and
+having them attach to a cluster might be more convenient than
+redeploying them to a cluster over and over.
+
+For the controller, the `-kubeconfig` and `-config-ns` command-line flags
+are needed.  Speakers need those and `-node-name`.
+
+For example:
+
+ metallb$ go run ./controller/main.go ./controller/service.go -config-ns metallb-system -kubeconfig $KUBECONFIG
+
+ metallb$ go run ./speaker/main.go ./speaker/*controller.go -config-ns metallb-system -kubeconfig $KUBECONFIG -node-name node0
 
 For development, fork
 the [github repository](https://github.com/google/metallb), and add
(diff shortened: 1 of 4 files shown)


rata:
Hi!

Sorry for the quick answer, I'm on my phone.

It's not very well documented here, https://metallb.universe.tf/community/, but basically you you install those things you can deploy locally. You can run all invoke commands, for local development inv dev-env (or alike, don't remember and I'm on my phone so not trivial to check) will do what you want automatically and quite easily (first time will create a cluster, next time you run it will recompile and update metalLB).

Just curious, have you tried that?

caboteria:
Yup, dev-env is a useful tool and it works well.

The changes I'm proposing add another option that benefits developers in a few ways:
 * speed: running "dev-env" with an update takes about 15s because it builds docker images, pushes them into the cluster, undeploys everything in the cluster, and then redeploys everything in the cluster.  Running "go run" externally takes about 2s for the controller to run.
 * flexibility: these changes let you attach a locally-running controller or speaker to *any* cluster that you've got a kubeconfig file for.  This is great for testing setups with special networking config or complex apps.
 * debug: because the local process is running on the host and not inside a docker container inside kind it should be easier to attach a debugger to it (although I haven't tested that yet).

russellb:
This seems like a useful addition to help an alternate dev flow.  I'd use this, as I'm more likely to do dev against a different cluster than the dev env cluster.

Can you add some docs about it to the page @rata mentioned, though?

caboteria:
No problem!

website/content/community/_index.md
```diff
@@ -127,6 +127,19 @@ registries, and so forth. `inv -l` lists the available tasks.
...
+When you're developing, running components at the command line and
+having them attach to a cluster might be more convenient than
+redeploying them to a cluster over and over.
+
+For the controller, the `-kubeconfig` and `-config-ns` command-line flags
+are needed.  Speakers need those and `-node-name`.
+
+For example:
+
+ metallb$ go run ./controller/main.go ./controller/service.go -config-ns metallb-system -kubeconfig $KUBECONFIG
```
russellb, line 139: As a future idea, maybe a tasks.py addition of a "run-controller" command would be a little easier.  same with "run-speaker".

russellb:
lgtm, pretty simple change, doesn't get in the way if you don't need it, and I would use this in my personal dev workflow.

russellb:
cc @rata

Issue #653, opened 2020-07-01
Re-implement helm chart


gclawes:
I have started a re-implementation of the metallb helm chart based on a combination of the previous chart in this repo, the [stable/metallb](https://github.com/helm/charts/tree/master/stable/metallb) chart in the soon-to-be-deprecated stable repo, and the [bitnami/metallb](https://github.com/bitnami/charts/tree/master/bitnami/metallb) chart.

My feature goals for this are:
* Use modern helm chart base from Helm v3
* Support new configuration options related to MemberList features
* Improve RBAC
* Switch from ServiceMonitor to PodMonitor, removing the need to create services for metrics

Maxpain177:
@gclawes Nice work! Any ETA?

gclawes:
Remaining tasks (in no particular order):

- Helm tests (pods to run in kind during CI).
- Add JSONSchema validation for values (and possibly ConfigMaps provided in the values)
- Chart release GitHub Action

I'm hoping to be able to get these done by next week.  See #658 for progress.

PR #658, opened 2020-07-07
Re-add helm chart under charts/metallb

gclawes:
Create a new helm chart integrating features of the previous chart, as
well as stable/metallb and bitnami/metallb.

Features:
* Support new environment variables in speaker
* Use PodMonitor instead of ServiceMonitor (no need to create Services)
* Create PrometheusRule to detect stale config and config-not-loaded
* Configurable PrometheusRule alerts for address pool exhaustion
* Support MetalLB controller creating memberlist secret
* Standardize labels using template helper
* Create config-watcher and pod-lister roles
* OPA/rego based chart output validation
* JSON Schema values validation

Fixes metallb/metallb#653

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,6 +23,26 @@ jobs:
       - run: sudo apt-get install python3-pip
       - run: sudo pip3 install invoke semver pyyaml
       - run: inv checkpatch
+  # FIXME: this is pinned to v3.0.0 because subsequent versions of ct pull in new versions of helm, which are subject
+  #        to this bug: https://github.com/helm/helm/issues/8835, specifically for rbac object names with ':' in them
+  helm-lint:
+    working_directory: /repo
+    docker:
+      - image: quay.io/helmpack/chart-testing:v3.0.0
+    steps:
+      - checkout
+      - run: ct lint
+  # FIXME: instrumenta/helm-conftest does not yet support helm v3, v2 causes tests to fail.
+  #        Switch to it once https://github.com/instrumenta/helm-conftest/pull/8 is merged
+  helm-conftest:
+    working_directory: /repo
+    docker:
+      - image: alpine/helm:3.2.4
+    steps:
+      - run: apk add --update --no-cache git curl bash
+      - run: helm plugin install --debug https://github.com/instrumenta/helm-conftest
+      - checkout
+      - run: helm conftest charts/metallb/ -p charts/metallb/policy/ --fail-on-warn
   publish-images:
     docker:
       - image: cimg/go:1.13
@@ -62,6 +82,14 @@ workflows:
             tags:
               only: /.*/
       - lint-1.13:
+          filters:
+            tags:
+              only: /.*/
+      - helm-lint:
+          filters:
+            tags:
+              only: /.*/
+      - helm-conftest:
           filters:
             tags:
               only: /.*/
--- a/charts/metallb/Chart.yaml
+++ b/charts/metallb/Chart.yaml
@@ -1,0 +1,27 @@
+apiVersion: v2
+name: metallb
+description: A network load-balancer implementation for Kubernetes using standard routing protocols
+home: https://metallb.universe.tf
+sources:
+  - https://github.com/metallb/metallb
+icon: https://metallb.universe.tf/images/logo/metallb-white.png
+
+# A chart can be either an 'application' or a 'library' chart.
+#
+# Application charts are a collection of templates that can be packaged into versioned archives
+# to be deployed.
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.9.6
+
+# This is the version number of the application being deployed. This version number should be
+# incremented each time you make changes to the application. Versions are not expected to
+# follow Semantic Versioning. They should reflect the version the application is using.
+appVersion: v0.9.6
--- a/charts/metallb/README.md
+++ b/charts/metallb/README.md
@@ -1,0 +1,75 @@
+metallb
+=======
+A network load-balancer implementation for Kubernetes using standard routing protocols
+
+Current chart version is `0.9.6`
+
+Source code can be found [here](https://metallb.universe.tf)
+
+
+
+## Chart Values
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| configInline | object | `{}` |  |
+| controller.affinity | object | `{}` |  |
+| controller.image.pullPolicy | string | `"IfNotPresent"` |  |
+| controller.image.repository | string | `"metallb/controller"` |  |
+| controller.image.tag | string | `nil` |  |
+| controller.livenessProbe.enabled | bool | `true` |  |
+| controller.livenessProbe.failureThreshold | int | `3` |  |
+| controller.livenessProbe.initialDelaySeconds | int | `10` |  |
+| controller.livenessProbe.periodSeconds | int | `10` |  |
+| controller.livenessProbe.successThreshold | int | `1` |  |
+| controller.livenessProbe.timeoutSeconds | int | `1` |  |
+| controller.nodeSelector | object | `{}` |  |
+| controller.podAnnotations | object | `{}` |  |
+| controller.readinessProbe.enabled | bool | `true` |  |
+| controller.readinessProbe.failureThreshold | int | `3` |  |
+| controller.readinessProbe.initialDelaySeconds | int | `10` |  |
+| controller.readinessProbe.periodSeconds | int | `10` |  |
+| controller.readinessProbe.successThreshold | int | `1` |  |
+| controller.readinessProbe.timeoutSeconds | int | `1` |  |
+| controller.resources | object | `{}` |  |
+| controller.serviceAccount.annotations | object | `{}` |  |
+| controller.serviceAccount.create | bool | `true` |  |
+| controller.serviceAccount.name | string | `""` |  |
+| controller.tolerations | list | `[]` |  |
+| existingConfigMap | string | `"metallb-config"` |  |
+| fullnameOverride | string | `""` |  |
+| imagePullSecrets | list | `[]` |  |
+| nameOverride | string | `""` |  |
+| prometheus.podMonitor.enabled | bool | `false` |  |
+| prometheus.podMonitor.interval | string | `""` |  |
+| prometheus.podMonitor.jobLabel | string | `"metallb"` |  |
+| prometheus.podMonitor.metricRelabelings | list | `[]` |  |
+| prometheus.podMonitor.relabelings | list | `[]` |  |
+| prometheus.prometheusRule.enabled | bool | `false` |  |
+| prometheus.scrapeAnnotations | bool | `false` |  |
+| psp.create | bool | `true` |  |
+| rbac.create | bool | `true` |  |
+| speaker.affinity | object | `{}` |  |
+| speaker.image.pullPolicy | string | `"IfNotPresent"` |  |
+| speaker.image.repository | string | `"metallb/speaker"` |  |
+| speaker.image.tag | string | `nil` |  |
+| speaker.livenessProbe.enabled | bool | `true` |  |
+| speaker.livenessProbe.failureThreshold | int | `3` |  |
+| speaker.livenessProbe.initialDelaySeconds | int | `10` |  |
+| speaker.livenessProbe.periodSeconds | int | `10` |  |
+| speaker.livenessProbe.successThreshold | int | `1` |  |
+| speaker.livenessProbe.timeoutSeconds | int | `1` |  |
+| speaker.nodeSelector | object | `{}` |  |
+| speaker.podAnnotations | object | `{}` |  |
+| speaker.readinessProbe.enabled | bool | `true` |  |
+| speaker.readinessProbe.failureThreshold | int | `3` |  |
+| speaker.readinessProbe.initialDelaySeconds | int | `10` |  |
+| speaker.readinessProbe.periodSeconds | int | `10` |  |
+| speaker.readinessProbe.successThreshold | int | `1` |  |
+| speaker.readinessProbe.timeoutSeconds | int | `1` |  |
+| speaker.resources | object | `{}` |  |
+| speaker.serviceAccount.annotations | object | `{}` |  |
+| speaker.serviceAccount.create | bool | `true` |  |
+| speaker.serviceAccount.name | string | `""` |  |
+| speaker.tolerateMaster | bool | `true` |  |
+| speaker.tolerations | list | `[]` |  |
--- a/charts/metallb/templates/NOTES.txt
+++ b/charts/metallb/templates/NOTES.txt
@@ -1,0 +1,14 @@
+MetalLB is now running in the cluster.
+{{- if .Values.configInline }}
+LoadBalancer Services in your cluster are now available on the IPs you
+defined in MetalLB's configuration:
+
+config:
+{{ toYaml .Values.configInline | indent 2 }}
+
+To see IP assignments, try `kubectl get services`.
+{{- else }}
+WARNING: you specified a ConfigMap that isn't managed by
+Helm. LoadBalancer services will not function until you add that
+ConfigMap to your cluster yourself.
+{{- end }}
--- a/charts/metallb/templates/controller.yaml
+++ b/charts/metallb/templates/controller.yaml
@@ -1,0 +1,102 @@
+{{- if .Values.controller.enabled }}
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: {{ template "metallb.fullname" . }}-controller
+  labels:
+    {{- include "metallb.labels" . | nindent 4 }}
+    app.kubernetes.io/component: controller
+spec:
+  selector:
+    matchLabels:
+      {{- include "metallb.selectorLabels" . | nindent 6 }}
+      app.kubernetes.io/component: controller
+  template:
+    metadata:
+      {{- if or .Values.prometheus.scrapeAnnotations .Values.controller.podAnnotations }}
+      annotations:
+        {{- if .Values.prometheus.scrapeAnnotations }}
+        prometheus.io/scrape: "true"
+        prometheus.io/port: "7472"
+        {{- end }}
+        {{- with .Values.podAnnotations }}
+          {{- toYaml . | nindent 8 }}
+        {{- end }}
+      {{- end }}
+      labels:
+        {{- include "metallb.selectorLabels" . | nindent 8 }}
+        app.kubernetes.io/component: controller
+    spec:
+      {{- with .Values.imagePullSecrets }}
+      imagePullSecrets:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      serviceAccountName: {{ template "metallb.controller.serviceAccountName" . }}
+      terminationGracePeriodSeconds: 0
+      securityContext:
+        runAsNonRoot: true
+        runAsUser: 65534 # nobody
+      containers:
+      - name: controller
+        image: {{ .Values.controller.image.repository }}:{{ .Values.controller.image.tag | default .Chart.AppVersion }}
+        {{- if .Values.controller.image.pullPolicy }}
+        imagePullPolicy: {{ .Values.controller.image.pullPolicy }}
+        {{- end }}
+        args:
+        - --config={{ template "metallb.configMapName" . }}
+        env:
+        {{- if and .Values.speaker.enabled .Values.speaker.memberlist.enabled }}
+        - name: METALLB_ML_SECRET_NAME
+          value: {{ include "metallb.secretName" . }}
+        - name: METALLB_DEPLOYMENT
+          value: {{ template "metallb.fullname" . }}-controller
+        {{- end }}
+        ports:
+        - name: metrics
+          containerPort: 7472
+        {{- if .Values.controller.livenessProbe.enabled }}
+        livenessProbe:
+          httpGet:
+            path: /metrics
+            port: metrics
+          initialDelaySeconds: {{ .Values.controller.livenessProbe.initialDelaySeconds }}
+          periodSeconds: {{ .Values.controller.livenessProbe.periodSeconds }}
+          timeoutSeconds: {{ .Values.controller.livenessProbe.timeoutSeconds }}
+          successThreshold: {{ .Values.controller.livenessProbe.successThreshold }}
+          failureThreshold: {{ .Values.controller.livenessProbe.failureThreshold }}
+        {{- end }}
+        {{- if .Values.controller.readinessProbe.enabled }}
+        readinessProbe:
+          httpGet:
+            path: /metrics
+            port: metrics
+          initialDelaySeconds: {{ .Values.controller.readinessProbe.initialDelaySeconds }}
+          periodSeconds: {{ .Values.controller.readinessProbe.periodSeconds }}
+          timeoutSeconds: {{ .Values.controller.readinessProbe.timeoutSeconds }}
+          successThreshold: {{ .Values.controller.readinessProbe.successThreshold }}
+          failureThreshold: {{ .Values.controller.readinessProbe.failureThreshold }}
+        {{- end }}
+        {{- with .Values.controller.resources }}
+        resources:
+          {{- toYaml . | nindent 12 }}
+        {{- end }}
+        securityContext:
+          allowPrivilegeEscalation: false
+          readOnlyRootFilesystem: true
+          capabilities:
+            drop:
+            - ALL
+      nodeSelector:
+        "kubernetes.io/os": linux
+        {{- with .Values.controller.nodeSelector }}
+          {{- toYaml . | nindent 8 }}
+        {{- end }}
+      {{- with .Values.controller.affinity }}
+      affinity:
+        {{- toYaml . | nindent 8 }}
+      {{- end }}
+      {{- with .Values.controller.tolerations }}
+      tolerations:
+        {{- toYaml . | nindent 6 }}
+      {{- end }}
+{{- end }}
--- a/charts/metallb/templates/psp.yaml
+++ b/charts/metallb/templates/psp.yaml
@@ -1,0 +1,95 @@
+{{- if .Values.psp.create -}}
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: {{ template "metallb.fullname" . }}-controller
+  labels:
+    {{- include "metallb.labels" . | nindent 4 }}
+spec:
+  privileged: false
+  # Host namespaces
+  hostPID: false
+  hostIPC: false
+  hostNetwork: false
+  hostPorts:
+  - min: 7472
+    max: 7472
+  # Volumes and file systems
+  volumes:
+  - configMap
+  - secret
+  - emptyDir
+  allowedHostPaths: []
+  readOnlyRootFilesystem: true
+  # Users and groups
+  runAsUser:
+    rule: MustRunAsNonRoot
+  supplementalGroups:
+    rule: MustRunAs
+    ranges:
+    - min: 1
+      max: 65535
+  fsGroup:
+    rule: MustRunAs
+    ranges:
+    - min: 1
+      max: 65535
+  # Privilege Escalation
+  allowPrivilegeEscalation: false
+  defaultAllowPrivilegeEscalation: false
+  # Capabilities
+  allowedCapabilities: []
+  defaultAddCapabilities: []
+  requiredDropCapabilities:
+  - ALL
+  # SELinux
+  seLinux:
+    rule: RunAsAny
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: {{ template "metallb.fullname" . }}-speaker
+  labels:
+    {{- include "metallb.labels" . | nindent 4 }}
+spec:
+  privileged: true
+  # Host namespaces
+  hostPID: false
+  hostIPC: false
+  hostNetwork: true
+  hostPorts:
+  - min: 7472
+    max: 7472
+  {{- if .Values.speaker.memberlist.enabled }}
+  - max: {{ .Values.speaker.memberlist.mlBindPort }}
+    min: {{ .Values.speaker.memberlist.mlBindPort }}
+  {{- end }}
+  # Volumes and file systems
+  volumes:
+  - configMap
+  - secret
+  - emptyDir
+  allowedHostPaths: []
+  readOnlyRootFilesystem: true
+  # Users and groups
+  runAsUser:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  fsGroup:
+    rule: RunAsAny
+  # Privilege Escalation
+  allowPrivilegeEscalation: false
+  defaultAllowPrivilegeEscalation: false
+  # Capabilities
+  allowedCapabilities:
+  - NET_RAW
+  defaultAddCapabilities: []
+  requiredDropCapabilities:
+  - ALL
+  # SELinux
+  seLinux:
+    rule: RunAsAny
+{{- end -}}
--- a/charts/metallb/templates/rbac.yaml
+++ b/charts/metallb/templates/rbac.yaml
@@ -1,0 +1,165 @@
+{{- if .Values.rbac.create -}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "metallb.fullname" . }}:controller
+  labels:
+    {{- include "metallb.labels" . | nindent 4 }}
+rules:
+- apiGroups: [""]
+  resources: ["services"]
+  verbs: ["get", "list", "watch", "update"]
+- apiGroups: [""]
+  resources: ["services/status"]
+  verbs: ["update"]
+- apiGroups: [""]
+  resources: ["events"]
+  verbs: ["create", "patch"]
+{{- if .Values.psp.create }}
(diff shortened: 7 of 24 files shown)


russellb:
I don't know helm, so won't be much help in reviewing this, but one question:

Do you have any thoughts about testing this in CI?  If this is included with metallb, I feel like we should have a CI job that validates the helm charts so we're sure they work with each release that gets tagged.  An alternative would be to put it in its own repo so it can be tagged and released independently and documented for which releases it has been tested against.

Also, I know this is still marked as a draft, but could you also include some docs?

gclawes:
Yeah, I can come up with some tests to validate that it produces valid k8s configs.  I also intend to add JSONSchema validations for the values file, once I'm confident nothing else needs to change for that.  Helm also has test that can be run against a running cluster (if we can get minikube working in CI that would be sufficient): https://helm.sh/docs/topics/chart_tests/.

In the long run I think it makes sense to release chart updates with metallb releases, since changes to how metallb is configured must be reflected in the chart.  In my experience maintaining a separate repo for a chart can become difficult.

Docs are my next step, just wanted to get the draft out there so people could see the template code.

russellb:
There's been some recent discussion about getting a functional CI job based on `kind` working, so we could probably do a variant of that for the helm based install later

gclawes:
I'm working on documentation.  What are your thoughts on hosting the helm chart on metallb.universe.tf?  What should the path for the chart repo be?  The other option would be github release artifacts/pages, helm has a tool and associated GitHub action to do this: https://github.com/helm/chart-releaser https://github.com/helm/chart-releaser-action

daxmc99:
I like this change and I believe this should be nested under the project. 
Thanks for the change! Let us know if you need help or have questions!

billimek:
Like @gclawes suggests, there is an established and demonstrated pattern of leveraging the helm chart testing action as well as the helm chart releaser action to make the automated linting, testing (via `kind`), and releasing charts very simple:

* [Chat Testing Action](https://github.com/helm/chart-testing-action)
* [Chart Releaser Action](https://github.com/helm/chart-releaser-action)
* [Example repo with it all tied-together](https://github.com/helm/charts-repo-actions-demo)

gclawes:
I've updated a GitHub Actions workflow that runs the chart-testing lint job.  I'll add the chart test job once I get tests written.

gclawes:
This is ready for review.

Here's a gist with the templated output of the chart with the default values: https://gist.github.com/gclawes/4b2f4a1bfd64e53bedc1e3700c2e56d6

Once the 0.9.4 release is made I'll rebase this branch and squash most of the commits together.

gclawes:
The one remaining concern I have is related to the `secretgen` Job to generate the memberlist secret.  As it is right now, it will not work in air-gapped clusters because it runs a standard `alpine` image and installs `openssl`/`kubectl` on the fly:
```
            set -ex

            # install openssl and kubectl
            echo "http://dl-cdn.alpinelinux.org/alpine/edge/testing" >> /etc/apk/repositories
            apk --no-cache add openssl kubectl

            # generate a random string with no newlines
            # store in file secret does not leak to logs
            openssl rand -base64 128 > /tmp/secret

            # create secret
            kubectl \
              create secret generic metallb-memberlist \
              --from-file=secretkey=/tmp/secret
```

I can overcome this by adding a new image called `metallb/secretgen`, but that'd be more invasive to the `tasks.py`, it would involve creating new tasks for building an image without a metallb binary in it (all the current tasks assume a go binary is built with each image).

Ultimately I think we want the secret generation to be handled by the controller, but we may still want a Job to clean up the secret on uninstall of the chart, so an extra image may be warranted regardless.

johananl:
@gclawes `v0.9.4` is now released :slightly_smiling_face: - FYI.

shaneutt:
> @gclawes `v0.9.4` is now released slightly_smiling_face - FYI.

Sorry to bother you @johananl and @gclawes but I simply wanted to check something:

This appears to be targeting the `v0.10.0` milestone, but that milestone has no due date associated with it.

My organization has been monitoring this PR as we're intending to switch to this helm chart from the soon to be deprecated one at `helm/charts`, do you have any kind of even very rough estimate you can provide on when you think this chart might be merged and ready for general usage so that I can set some rough expectations back at my org?

Thank you for your time!

gclawes:
@johananl working on the rebase this week.

@shaneutt I do not have a date for v0.10.0, though we do want to get it out as soon as possible (I would guess on the order of weeks, not months).  Others may have a better estimate.

champtar:
@gclawes have you tried `helm upgrade` from the bitnami chart ? not a blocker but nice to have

gclawes:
@champtar I have not.  Upgrading from the bitnami chart will likely not work, as the values and internal mechanics of this chart are sufficiently different.  I guess in theory it might work, but I'm unsure.

charts/metallb/templates/speaker.yaml
```diff
@@ -0,0 +1,131 @@
...
+              fieldPath: metadata.namespace
+        - name: METALLB_NODE_NAME
+          valueFrom:
+            fieldRef:
+              fieldPath: spec.nodeName
+        - name: METALLB_HOST
+          valueFrom:
+            fieldRef:
+              fieldPath: status.hostIP
+        - name: METALLB_ML_BIND_ADDR
```
champtar, line 57: We need a "enable memberlist" boolean, set to true by default, that if set to false remove all the METALLB_ML*
We've had pretty good results with memberlist but BGP only user don't need it, and it's nice to have it for debug

charts/metallb/templates/rbac.yaml
```diff
@@ -0,0 +1,161 @@
...
+- apiGroups: [""]
+  resources: ["services/status"]
+  verbs: ["update"]
+- apiGroups: [""]
+  resources: ["events"]
+  verbs: ["create", "patch"]
+{{- if .Values.psp.create }}
+- apiGroups: ["policy"]
+  resources: ["podsecuritypolicies"]
+  resourceNames: [{{ printf "%s-controller" (include "metallb.fullname" .) | quote}}]
```
champtar, line 21: I see you use printf here and template later, I prefer template, it's easier to read

champtar: `["{{ template "metallb.fullname" . }}-controller"]`

charts/metallb/templates/NOTES.txt
```diff
@@ -0,0 +1,11 @@
+
+MetalLB is now running in the cluster.
+{{- if .Values.configInline }}
+LoadBalancer Services in your cluster are now available on the IPs you
+defined in MetalLB's configuration. To see IP assignments,
+try `kubectl get services`.
```
champtar, line 6: maybe print the inline config ?

Also if someone changed the ConfigMap directly and the next helm upgrade "revert" the config, is there a way to print the old and new config ?

It's more a limitation of Helm but having to type the repo and version just to change the current config without upgrading is not user friendly at all, so not a fan of configInline when `kubectl edit` is all we need ;)

champtar (commented):
Some small things, still need to test it

charts/metallb/Chart.yaml
```diff
@@ -0,0 +1,27 @@
+apiVersion: v2
+name: metallb
+description: A network load-balancer implementation for Kubernetes using standard routing protocols
+home: https://metallb.universe.tf
+sources:
+  - https://github.com/metallb/metallba
```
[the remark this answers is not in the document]
gclawes: Good eye, I'll fix this.

charts/metallb/README.md
```diff
@@ -0,0 +1,75 @@
+metallb
+=======
+A network load-balancer implementation for Kubernetes using standard routing protocols
+
+Current chart version is `0.9.3`
```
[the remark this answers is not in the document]
gclawes: It can be automated in the version bump task, I'll add that

charts/metallb/values.yaml
```diff
@@ -0,0 +1,196 @@
...
+    extraAlerts: []
+
+# controller contains configuration specific to the MetalLB cluster
+# controller.
+controller:
+  enabled: true
+  image:
+    repository: metallb/controller
+    tag:
+    pullPolicy: IfNotPresent
```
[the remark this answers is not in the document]
gclawes: When it is empty, it defaults to the value of `appVersion` in `Chart.yaml`.  This is done by this line:
```
image: {{ .Values.controller.image.repository }}:{{ .Values.controller.image.tag | default .Chart.AppVersion }}
```

Which was part of the default skeleton template provided by `helm create` in helm v3.

charts/metallb/templates/speaker.yaml
gclawes: I'll add that, thanks.  Will memberlist just not be enabled if the `METALLB_ML*` variables aren't present?

The switch will be in `.Values.metallb.speaker.memberlist.enabled` (boolean).

charts/metallb/templates/rbac.yaml
gclawes: Thanks for spotting the inconsistency, I'll fix it.

charts/metallb/templates/NOTES.txt
gclawes: I can display the inline config.  I don't think there's a way to get diffs between releases/upgrades.

charts/metallb/values.yaml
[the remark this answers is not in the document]
champtar: I was talking about the pullPolicy, not the tag ;)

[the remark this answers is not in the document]
gclawes: Ahh, gotcha.  Yeah, I can make that empty.  I think it came as part of the `helm create` boilerplate.

charts/metallb/templates/NOTES.txt
ilhaan: So [helm-diff](https://github.com/databus23/helm-diff) is a plugin that allows you to check diffs between releases/upgrades. There is talk of this plugin being merged into helm core [here](https://github.com/databus23/helm-diff/issues/211). 

However, I don't think this chart has to worry about manual changes being reverted upon a helm upgrade. I think it is well understood that manual changes won't persist if helm is part of your workflow and you have not updated helm upgrade params.

charts/metallb/templates/speaker.yaml
champtar: > Will memberlist just not be enabled if the METALLB_ML* variables aren't present?

yes
https://github.com/metallb/metallb/blob/main/internal/speakerlist/speakerlist.go#L54

charts/metallb/templates/NOTES.txt
champtar: > I think it is well understood that manual changes won't persist if helm is part of your workflow and you have not updated helm upgrade params.

you assume people know how things are working and are not following blindly procedures and first google reponse :P

I agree this is not the job of the chart, but it would have been nice. I'm thinking about the message displayed by apt when default config changes, do you want to keep local or maintainer version, see a diff, edit, ..., this is really a good UX IMO.

gclawes: I think helm diff is a reasonable solution, especially when combined with tools like [Praqma/helmsman](https://github.com/Praqma/helmsman).

charts/metallb/README.md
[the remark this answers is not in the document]
gclawes: Ah, forgot I already did this:
```
run("perl -pi -e 's,^Current chart version is: .*,Current chart version is: `{}`,g' charts/metallb/README.md".format(version), echo=True)
```

charts/metallb/values.yaml
```diff
@@ -0,0 +1,196 @@
...
+    # Specifies whether a ServiceAccount should be created
+    create: true
+    # The name of the ServiceAccount to use.  If not set and create is
+    # true, a name is generated using the fullname template
+    name: ""
+    annotations: {}
+  resources: {}
+    # limits:
+      # cpu: 100m
+      # memory: 100Mi
```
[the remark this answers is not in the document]
gclawes: Agreed, this should be up to cluster operators, we can't make assumptions on their behalf.

charts/metallb/values.yaml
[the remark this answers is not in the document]
gclawes: Made optional and updated values.schema.json to allow null values

charts/metallb/templates/rbac.yaml
gclawes: Fixed.

gclawes: Fixed

charts/metallb/templates/NOTES.txt
gclawes: Updated NOTES.txt template

gclawes:
I've squashed the commit down to 5 commits.  If anyone things it should be more compact I can do that.

champtar:
Seems good to me, you are just missing `.Values.metallb.speaker.memberlist.enabled`

charts/metallb/templates/psp.yaml
```diff
@@ -0,0 +1,93 @@
...
+  fsGroup:
+    rule: RunAsAny
+  # Privilege Escalation
+  allowPrivilegeEscalation: false
+  defaultAllowPrivilegeEscalation: false
+  # Capabilities
+  allowedCapabilities:
+  - NET_ADMIN
+  - NET_RAW
+  - SYS_ADMIN
```
champtar, line 86: See #764, `SYS_ADMIN` and `NET_ADMIN` must be remove IMO

champtar:
Just tried upgrade from bitnami chart:
```
Error: UPGRADE FAILED: error validating "": error validating data: [ValidationError(Role.rules[0]): unknown field "resources:" in io.k8s.api.rbac.v1.PolicyRule, ValidationError(Role.rules[1]): unknown field "resources:" in io.k8s.api.rbac.v1.PolicyRule]
```

charts/metallb/templates/rbac.yaml
```diff
@@ -0,0 +1,161 @@
...
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ include "metallb.fullname" . }}-controller
+  namespace: {{ .Release.Namespace }}
+  labels: {{- include "metallb.labels" . | nindent 4 }}
+rules:
+- apiGroups: [""]
+  resources:: ["secrets"]
```
champtar, line 76: typo

charts/metallb/templates/rbac.yaml
```diff
@@ -0,0 +1,161 @@
...
+  verbs: ["create"]
+- apiGroups: [""]
+  resources:: ["secrets"]
```
champtar, line 79: typo

champtar (changes requested):

champtar:
Upgrade from bitnami to your chart seems to work \o/
```
# helm install metallb --namespace metallb-system bitnami/metallb --version 0.1.28 --set existingConfigMap=config
...
# helm upgrade metallb --namespace metallb-system metallb --version 0.9.4 --set existingConfigMap=config
Release "metallb" has been upgraded. Happy Helming!
NAME: metallb
LAST DEPLOYED: Sat Oct 24 21:11:26 2020
NAMESPACE: metallb-system
STATUS: deployed
REVISION: 2
```

charts/metallb/templates/rbac.yaml
gclawes: good eye!

gclawes:
I wonder if it will work by transplanting the contents of `configInLine` instead of referencing an existing config map.

charts/metallb/templates/psp.yaml
gclawes: I'll wait for input on #764 before updating this.  If this PR merges before that one, we can update it there.

charts/metallb/templates/rbac.yaml
gclawes: Fixed.

charts/metallb/templates/rbac.yaml
gclawes: Fixed.

charts/metallb/templates/speaker.yaml
gclawes: Added switch to disable these vars:

speaker - METALLB_ML_BIND_ADDR METALLB_ML_LABELS METALLB_ML_BIND_PORT METALLB_ML_SECRET_KEY
controller - METALLB_ML_SECRET_NAME

lexfrei:
Is it ready to test with 0.9.5 on arm64?

gclawes:
@lexfrei you can test with a few caveats:

- It won't be deployed to the github pages helm repo until this PR is merged, so you will have to install from the git repository.
- This depends on #747 to create the memberlist secret, until that is merged and 0.10.0 is released, you will have to follow the procedure to create the secret manually here: https://metallb.universe.tf/installation/#installation-by-manifest

lexfrei:
Can you sync your config map syntax with [bitnami's one](https://github.com/bitnami/charts/blob/master/bitnami/metallb/templates/configmap.yaml)?
It can simplify the migration for bitnami's users.

gclawes:
What specifically is preventing migration now?

lexfrei:
@gclawes hm. Can't reproduce the unmarshal error now. Looks like my own typo.

Now I can see this `null` for my config, but all works fine, and configmap created properly.

```
NOTES:
MetalLB is now running in the cluster.
LoadBalancer Services in your cluster are now available on the IPs you
defined in MetalLB's configuration:

null

To see IP assignments, try `kubectl get services`.
```

gclawes:
I see that in mine, I'll figure out what's going on and fix it:
```
2020-11-15 11:48:55 NOTICE: Install release [ metallb ] version [ 0.9.5 ] in namespace [ metallb-system ]
2020-11-15 11:48:59 NOTICE: Release "metallb" does not exist. Installing it now.
NAME: metallb
LAST DEPLOYED: Sun Nov 15 11:48:56 2020
NAMESPACE: metallb-system
STATUS: deployed
REVISION: 1
TEST SUITE: None
NOTES:
MetalLB is now running in the cluster.
LoadBalancer Services in your cluster are now available on the IPs you
defined in MetalLB's configuration:

null

To see IP assignments, try `kubectl get services`.

2020-11-15 11:48:59 NOTICE: Finished: Install release [ metallb ] version [ 0.9.5 ] in namespace [ metallb-system ]
```

gclawes:
Fixed rendering the config:
```
NOTES:
MetalLB is now running in the cluster.
LoadBalancer Services in your cluster are now available on the IPs you
defined in MetalLB's configuration:

config:
  address-pools:
  - addresses:
    - 10.102.0.0/16
    name: default
    protocol: bgp
  peers:
  - my-asn: 65003
    peer-address: 127.0.0.1
    peer-asn: 65002

To see IP assignments, try `kubectl get services`.
```

daxmc99 (approved):

johananl:
Thanks a lot for the effort @gclawes! I'm currently reviewing and will update as soon as I can.

.circleci/config.yml
```diff
@@ -17,6 +17,26 @@ jobs:
       - checkout
       - run: curl -L https://git.io/vp6lP | sh # gometalinter
       - run: PATH=./bin:$PATH gometalinter --deadline=5m --disable-all --enable=gofmt --enable=vet --vendor ./...
+  # FIXME: this is pinned to v3.0.0 because subsequent versions of ct pull in new versions of helm, which are subject
+  #        to this bug: https://github.com/helm/helm/issues/8596, specifically for rbac object names with ':' in them
```
daxmc99, line 21: This appears to be fixed 
https://github.com/helm/helm/issues/8596

charts/metallb/templates/rbac.yaml
```diff
@@ -0,0 +1,165 @@
+{{- if .Values.rbac.create -}}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "metallb.fullname" . }}:controller
```
daxmc99, line 5: I get a `helm lint` error here 
`"test-release-metallb:controller": invalid metadata name, must match regex ^[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*$ and the length must not be longer than 253`

charts/metallb/templates/rbac.yaml
```diff
@@ -0,0 +1,165 @@
...
+- apiGroups: ["policy"]
+  resources: ["podsecuritypolicies"]
+  resourceNames: ["{{ template "metallb.fullname" . }}-controller"]
+  verbs: ["use"]
+{{- end }}
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "metallb.fullname" . }}:speaker
```
daxmc99, line 28: Same lint error

charts/metallb/templates/rbac.yaml
```diff
@@ -0,0 +1,165 @@
...
+  namespace: {{ .Release.Namespace }}
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: {{ template "metallb.fullname" . }}:controller
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: {{ template "metallb.fullname" . }}:speaker
```
daxmc99, line 107: Same helm lint error

daxmc99 (changes requested):
Thank you so much again @gclawes! 
I'm sorry this has been open for so long. I left some comments in the PR.

charts/metallb/templates/rbac.yaml
gclawes: Interesting, ClusterRole names with `:` are definitely valid:
```
$ kubectl get clusterrole | grep system | head -1
system:discovery                                                       2021-01-01T17:10:36Z
```

gclawes: Ah, I left this note in `.circleci/config.yml`:
``` 20   # FIXME: this is pinned to v3.0.0 because subsequent versions of ct pull in new versions of helm, which are subject
 21   #        to this bug: https://github.com/helm/helm/issues/8596, specifically for rbac object names with ':' in them
 22   helm-lint:
 23     working_directory: /repo
 24     docker:
 25       - image: quay.io/helmpack/chart-testing:v3.0.0
```

.circleci/config.yml
gclawes: Still appears to be a problem: https://github.com/helm/helm/issues/8835

daxmc99: Ahh okay, can we update the comment?

charts/metallb/templates/rbac.yaml
daxmc99: I am! This clear that up

gclawes: It appears they fixed the issue in helm/helm#8596 about deprecated API versions, but not the `:` issue.  I'll update the note in `.circleci/config.yml`

.circleci/config.yml
gclawes: Yeah, I'll do that.

charts/metallb/templates/rbac.yaml
gclawes: Fixed, adjusted the comment about this in `.circleci/config.yml` to point to the new upstream issue for `:` in name.

gclawes:
@daxmc99 do these changes look good?

daxmc99 (approved):
LGTM! 
These changes look good to me. Also thank you for updating the release process to update the chart version.

.circleci/config.yml
```diff
@@ -17,6 +17,26 @@ jobs:
...
+  #        Switch to it once https://github.com/instrumenta/helm-conftest/pull/8 is merged
+  helm-conftest:
+    working_directory: /repo
+    docker:
+      - image: alpine/helm:3.2.4
+    steps:
+      - run: apk add --update --no-cache git curl bash
+      - run: helm plugin install --debug https://github.com/instrumenta/helm-conftest
+      - checkout
+      - run: helm conftest charts/metallb/ -p charts/metallb/policy/ --fail-on-warn
```
daxmc99, line 39: Very small nit: This PR seems to mix CircleCI additions and Github Actions. 
Is there a reason we cannot use only CircleCI or Github Actions?

gclawes: Sorry for the late response.  The GitHub Action is a convenient pre-made way to publish the Helm chart in GitHub Pages.  The only reason to use it is to avoid porting that same logic to CircleCI.

charts/metallb/templates/controller.yaml
```diff
@@ -0,0 +1,107 @@
...
+        runAsNonRoot: true
+        runAsUser: 65534 # nobody
+      containers:
+      - name: controller
+        image: {{ .Values.controller.image.repository }}:{{ .Values.controller.image.tag | default .Chart.AppVersion }}
+        {{- if .Values.controller.image.pullPolicy }}
+        imagePullPolicy: {{ .Values.controller.image.pullPolicy }}
+        {{- end }}
+        args:
+        - --port=7472
```
champtar, line 46: not needed, this is already the default

charts/metallb/templates/controller.yaml
```diff
@@ -0,0 +1,107 @@
...
+        - --config={{ template "metallb.configMapName" . }}
+        env:
+        - name: METALLB_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
```
champtar, line 52: Not required anymore (remove it)

champtar, line 52: Not required anymore (remove it)

charts/metallb/templates/controller.yaml
```diff
@@ -0,0 +1,107 @@
...
+        {{- if .Values.speaker.memberlist.enabled }}
+        - name: METALLB_ML_SECRET_NAME
+          value: {{ include "metallb.secretName" . }}
+        {{- end }}
+        - name: METALLB_DEPLOYMENT
+          value: {{ template "metallb.fullname" . }}-controller
```
champtar, line 58: `METALLB_DEPLOYMENT` is not needed when memberlist is disabled, can you move the if/end accordingly ?

charts/metallb/templates/psp.yaml
champtar: #764 is merged now

charts/metallb/templates/psp.yaml
```diff
@@ -0,0 +1,93 @@
...
+    {{- include "metallb.labels" . | nindent 4 }}
+spec:
+  privileged: true
+  # Host namespaces
+  hostPID: false
+  hostIPC: false
+  hostNetwork: true
+  hostPorts:
+  - min: 7472
+    max: 7472
```
champtar, line 64: see https://github.com/metallb/metallb/pull/770/files, memberlist port has been added

charts/metallb/templates/speaker.yaml
```diff
@@ -0,0 +1,134 @@
...
+        {{- end }}
+        - name: METALLB_ML_SECRET_KEY
+          valueFrom:
+            secretKeyRef:
+              name: {{ include "metallb.secretName" . }}
+              key: secretkey
+        {{- end }}
+        ports:
+        - name: metrics
+          containerPort: 7472
```
champtar, line 79: see https://github.com/metallb/metallb/pull/770/files, memberlist port has been added

charts/metallb/templates/speaker.yaml
```diff
@@ -0,0 +1,134 @@
...
+        imagePullPolicy: {{ .Values.speaker.image.pullPolicy }}
+        {{- end }}
+        args:
+        - --port=7472
+        - --config={{ template "metallb.configMapName" . }}
+        env:
+        - name: METALLB_NAMESPACE
+          valueFrom:
+            fieldRef:
+              fieldPath: metadata.namespace
```
champtar, line 50: to be removed

champtar (changes requested):
If you can do a diff with what is in master

charts/metallb/Chart.yaml
```diff
@@ -0,0 +1,27 @@
...
+#
+# Library charts provide useful utilities or functions for the chart developer. They're included as
+# a dependency of application charts to inject those utilities and functions into the rendering
+# pipeline. Library charts do not define any templates and therefore cannot be deployed.
+type: application
+
+# This is the chart version. This version number should be incremented each time you make changes
+# to the chart and its templates, including the app version.
+# Versions are expected to follow Semantic Versioning (https://semver.org/)
+version: 0.9.5
```
champtar, line 22: 0.9.6 or 0.10.0 ?

charts/metallb/README.md
champtar, line 5: 0.9.6 or 0.10.0 ?

charts/metallb/templates/psp.yaml
```diff
@@ -0,0 +1,95 @@
...
+  {{- if .Values.speaker.memberlist.enabled }}
+  - max: {{ .Values.speaker.mlBindPort }}
+    min: {{ .Values.speaker.mlBindPort }}
```
champtar, line 67: you need to set the default value for `mlBindPort` else this is going to fail

charts/metallb/templates/speaker.yaml
```diff
@@ -0,0 +1,137 @@
...
+            fieldRef:
+              fieldPath: status.hostIP
+        {{- if .Values.speaker.memberlist.enabled }}
+        - name: METALLB_ML_BIND_ADDR
+          valueFrom:
+            fieldRef:
+              fieldPath: status.podIP
+        - name: METALLB_ML_LABELS
+          value: "app.kubernetes.io/name={{ include "metallb.name" . }},app.kubernetes.io/component=speaker"
+        {{ if .Values.speaker.mlBindPort }}
```
champtar, line 61: As you are going to have a default `mlBindPort` let's make that `if .Values.speaker.memberlist.enabled`

charts/metallb/templates/psp.yaml
champtar: Also what about `Values.speaker.memberlist.mlBindPort` ?

champtar (changes requested):
Fix needed for `mlBindPort`

charts/metallb/templates/speaker.yaml
champtar: just saw it's in a `if .Values.speaker.memberlist.enabled`, so just remove it

charts/metallb/Chart.yaml
gclawes: It should be whatever the latest release tag is.  It'll get updated by the pipeline when a new tag is cut.

gclawes: I'll update it to 0.9.6

charts/metallb/README.md
gclawes: 0.9.6

charts/metallb/templates/speaker.yaml
gclawes: Will remove.

charts/metallb/templates/speaker.yaml
```diff
@@ -0,0 +1,135 @@
...
+        {{- if .Values.speaker.memberlist.enabled }}
+        - name: METALLB_ML_BIND_ADDR
+          valueFrom:
+            fieldRef:
+              fieldPath: status.podIP
+        - name: METALLB_ML_LABELS
+          value: "app.kubernetes.io/name={{ include "metallb.name" . }},app.kubernetes.io/component=speaker"
+        # needed when another software is also using memberlist / port 7946
+        - name: METALLB_ML_BIND_PORT
+          value: {{ .Values.speaker.memberlist.mlBindPort }}
```
champtar, line 63:
```suggestion
        - name: METALLB_ML_BIND_PORT
          value: "{{ .Values.speaker.memberlist.mlBindPort }}"
```

charts/metallb/templates/controller.yaml
```diff
@@ -0,0 +1,102 @@
...
+      containers:
+      - name: controller
+        image: {{ .Values.controller.image.repository }}:{{ .Values.controller.image.tag | default .Chart.AppVersion }}
+        {{- if .Values.controller.image.pullPolicy }}
+        imagePullPolicy: {{ .Values.controller.image.pullPolicy }}
+        {{- end }}
+        args:
+        - --config={{ template "metallb.configMapName" . }}
+        env:
+        {{- if .Values.speaker.memberlist.enabled }}
```
champtar, line 48: should the memberlist option be global ? it's a bit weird to have a speaker option impact the controller, even more so if we only want to deploy the controller (Calico BGP peering + MetalLB controller)

champtar (changes requested):
I think those are the last 2 comments

champtar:
Migration from bitnami to our chart works in my simple test
```
helm repo add bitnami https://charts.bitnami.com/bitnami

kubectl create namespace metallb-system

kubectl apply -f - <<'EOM'
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: metallb-system
  name: config
data:
  config: |
    address-pools:
    - name: ingress-pool
      protocol: layer2
      addresses:
      - 10.10.52.150/32
      auto-assign: false
EOM

helm install metallb --namespace metallb-system bitnami/metallb --version 2.3.5 --set existingConfigMap=config

helm upgrade metallb --namespace metallb-system /root/metallb/charts/metallb --version 0.9.6 --values ~/metallbval.yaml
```

```
# cat ~/metallbval.yaml

existingConfigMap: config

controller:
  image:
    tag: main
    pullPolicy: Always

speaker:
  image:
    tag: main
    pullPolicy: Always
```

charts/metallb/templates/controller.yaml
champtar: We have now switched to `{{- if and .Values.speaker.enabled .Values.speaker.memberlist.enabled }}`
@russellb except if you think we should move memberlist config to be at the top level just to be safe (if we find more memberlist config that are global) you can resolve this one

russellb: I'll leave this to you both to decide.  I think it makes sense to think of memberlist as a speaker feature.  That's where it operates, right?  It's also probably harmless to set these env vars even if memberlist is not enabled.  Then you won't be looking at the setting from the controller stuff here

champtar:
This magicaly triggered a metallb-0.9.6 "release"
image

champtar:
this is latest commit from main, not v0.9.6

PR #659, opened 2020-07-07
Publish images for each open pull request

### LINKED ISSUE #623: [CI] Push a container image for each PR
While building out CI, we should consider pushing a container image for each PR to a public registry to make it easy for people to test out a PR.

russellb:
CircleCI was previously only configured to publish images for pushes
to branches or tags.  While CI runs against open PRs, the publish jobs
that published images was excluded.

This PR updates the deploy jobs to detect when they are running
against a pull request.  When run against a PR, the images are pushed
to the `metallbdev` org instead of `metallb` and the image tag is
"pr-<ID>".

The dev docs on the web site are also updated to note how to pull an
image built for a PR.

Fixes #623

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -51,6 +51,57 @@ jobs:
       - run: wget -O ./bin/manifest-tool https://github.com/estesp/manifest-tool/releases/download/v1.0.0/manifest-tool-linux-amd64 && chmod +x ./bin/manifest-tool
       - run: echo "export GO111MODULE=on" >>$BASH_ENV
       - run: PATH=./bin:$PATH inv push-multiarch --binaries=speaker --tag=${CIRCLE_BRANCH:-${CIRCLE_TAG}} --docker-user=metallb
+  deploy-controller-pr:
+    working_directory: /go/src/go.universe.tf/metallb
+    docker:
+      - image: circleci/golang:1.13
+    steps:
+      - checkout
+      - setup_remote_docker
+      - run: sudo apt-get install python-pip
+      - run: sudo pip install invoke semver pyyaml
+      - run: docker login -u $DOCKER_USER -p $DOCKER_PASSWORD
+      - run: mkdir -p ./bin
+      - run: wget -O ./bin/manifest-tool https://github.com/estesp/manifest-tool/releases/download/v1.0.0/manifest-tool-linux-amd64 && chmod +x ./bin/manifest-tool
+      - run: echo "export GO111MODULE=on" >>$BASH_ENV
+      - run: echo ${CIRCLE_PULL_REQUEST}
+      - run: echo ${CIRCLE_BRANCH}
+      - run: echo ${CIRCLE_TAG}
+      # CIRCLE_PULL_REQUEST is a URL in the form: https://github.com/metallb/metallb/pull/1
+      # We use sed to get the pull request ID out of the URL. CIRCLE_PR_NUMBER would be
+      # more convenient, but unfortunately it is not always available.
+      # https://circleci.com/docs/2.0/env-vars/
+      #
+      # If the job is being run against a PR, use "pr-<ID>" as the image tag and push the
+      # result to metallbdev/controller. Otherwise this job is not expected to run so we
+      # just fail.
+      - run: if [ -n "$CIRCLE_PULL_REQUEST" ]; then PATH=./bin:$PATH inv push-multiarch --binaries=controller --tag=pr-$(basename "$CIRCLE_PULL_REQUEST") --docker-user=metallbdev ; else exit 1 ; fi
+  deploy-speaker-pr:
+    working_directory: /go/src/go.universe.tf/metallb
+    docker:
+      - image: circleci/golang:1.13
+    steps:
+      - checkout
+      - setup_remote_docker
+      - run: sudo apt-get install python-pip
+      - run: sudo pip install invoke semver pyyaml
+      - run: docker login -u $DOCKER_USER -p $DOCKER_PASSWORD
+      - run: mkdir -p ./bin
+      - run: wget -O ./bin/manifest-tool https://github.com/estesp/manifest-tool/releases/download/v1.0.0/manifest-tool-linux-amd64 && chmod +x ./bin/manifest-tool
+      - run: echo "export GO111MODULE=on" >>$BASH_ENV
+      - run: echo ${CIRCLE_PULL_REQUEST}
+      - run: echo ${CIRCLE_BRANCH}
+      - run: echo ${CIRCLE_TAG}
+      # CIRCLE_PULL_REQUEST is a URL in the form: https://github.com/metallb/metallb/pull/1
+      # We use sed to get the pull request ID out of the URL. CIRCLE_PR_NUMBER would be
+      # more convenient, but unfortunately it is not always available.
+      # https://circleci.com/docs/2.0/env-vars/
+      #
+      # If the job is being run against a PR, use "pr-<ID>" as the image tag and push the
+      # result to metallbdev/speaker. Otherwise this job is not expected to run so we
+      # just fail.
+      - run: if [ -n "$CIRCLE_PULL_REQUEST" ]; then PATH=./bin:$PATH inv push-multiarch --binaries=speaker --tag=pr-$(basename "$CIRCLE_PULL_REQUEST") --docker-user=metallbdev ; else exit 1 ; fi
+
 workflows:
   version: 2
   test-and-deploy:
@@ -82,6 +133,29 @@ workflows:
                 - /v.*/
             tags:
               only: /.*/
+          requires:
+            - test-1.13
+            - lint-1.13
+      # The *-pr jobs should only run against pull requests.
+      - deploy-controller-pr:
+          filters:
+            branches:
+              ignore:
+                - main
+                - /v.*/
+            tags:
+              ignore: /.*/
+          requires:
+            - test-1.13
+            - lint-1.13
+      - deploy-speaker-pr:
+          filters:
+            branches:
+              ignore:
+                - main
+                - /v.*/
+            tags:
+              ignore: /.*/
           requires:
             - test-1.13
             - lint-1.13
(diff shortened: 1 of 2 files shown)


russellb:
Test runs of these changes were done here: https://github.com/russellb/metallb/pull/2

Before merging this, I think all I need to do is know what DOCKER_USER is set to in circleci so I can grant that user access to the metallbdev org.

.circleci/config.yml
```diff
@@ -50,7 +61,18 @@ jobs:
...
+      - run: echo ${CIRCLE_TAG}
+      # CIRCLE_PULL_REQUEST is a URL in the form: https://github.com/metallb/metallb/pull/1
+      # We use sed to get the pull request ID out of the URL.  CIRCLE_PR_NUMBER would be
+      # more convenient, but unfortunately it is not always available.
+      # https://circleci.com/docs/2.0/env-vars/
+      #
+      # If the job is being run against a PR, use "pr-<ID>" as the image tag and push the
+      # result to metallbdev/speaker.  Otherwise we use the branch or tag name as the
+      # image tag and push to the main metallb/ image.
+      - run: if [ -n $CIRCLE_PULL_REQUEST ]; then PATH=./bin:$PATH inv push-multiarch --binaries=controller --tag=pr-$(echo $CIRCLE_PULL_REQUEST | sed -e 's/.*pull\/\(.*\)/\1/') --docker-user=metallbdev ; else PATH=./bin:$PATH inv push-multiarch --binaries=controller --tag=${CIRCLE_BRANCH:-${CIRCLE_TAG}} --docker-user=metallb ; fi
```
daxmc99, line 75: Is there where the `--docker-use=metallb` user needs to be replaced with docker user?

russellb: No, earlier there is a `docker login` command that logs in as whatever `DOCKER_USER` is set to in the circle CI configuration.  I need to grant that user access to the new `metallbdev` organization on docker hub.

rata, line 75: Sorry for being so paranoid. But what do you think of instead of adding this to `deploy_speaker` we create a `deploy_pr` that just pushed to that docker org?

That way the filters here: https://github.com/metallb/metallb/pull/659/files#diff-1d37e48f9ceff6d8030570cd36286a61L67-L73 can stay and there is not trick (maybe create a PR that sets `CIRCLE_PULL_REQUEST`? or something smarter, that will fail) that anyone can do to abuse this?

russellb: Yeah, I think that makes sense.  That does seem like an improvement.  Thanks!

.circleci/config.yml
```diff
@@ -50,7 +61,18 @@ jobs:
       - run: mkdir -p ./bin
       - run: wget -O ./bin/manifest-tool https://github.com/estesp/manifest-tool/releases/download/v1.0.0/manifest-tool-linux-amd64 && chmod +x ./bin/manifest-tool
       - run: echo "export GO111MODULE=on" >>$BASH_ENV
-      - run: PATH=./bin:$PATH inv push-multiarch --binaries=speaker --tag=${CIRCLE_BRANCH:-${CIRCLE_TAG}} --docker-user=metallb
+      - run: echo ${CIRCLE_PULL_REQUEST}
```
rata, line 64: just curious. Would it make sense to instead of `pr-$CIRCLE_PULL_REQUEST` use something like: `pr-$CIRCLE_PULL_REQUEST-<git hash>`?

I mean, if a PR is updated the pods might need to pull the image again. The manifest default [is to always pull][link] but I wouldn't be surprised if someone changed that.

Or push `pr-$CIRCLE_PULL_REQUEST-<git hash>` *and* `pr-$CIRCLE_PULL_REQUEST-latest`?

[link]: https://github.com/metallb/metallb/blob/b60c6a110d81d7f9cc4b6a2eeb36a6b583a2aaa5/manifests/metallb.yaml#L317

russellb: My thought was that testing the current version is what is most interesting.  Good idea on updating the docs about the image pull policy.

russellb: I updated the docs

.circleci/config.yml
russellb: done!

russellb:
@rata I believe I addressed your feedback on this one.

.circleci/config.yml
rata: Thanks!

.circleci/config.yml
```diff
@@ -51,6 +51,57 @@ jobs:
...
+      - run: sudo pip install invoke semver pyyaml
+      - run: docker login -u $DOCKER_USER -p $DOCKER_PASSWORD
+      - run: mkdir -p ./bin
+      - run: wget -O ./bin/manifest-tool https://github.com/estesp/manifest-tool/releases/download/v1.0.0/manifest-tool-linux-amd64 && chmod +x ./bin/manifest-tool
+      - run: echo "export GO111MODULE=on" >>$BASH_ENV
+      - run: echo ${CIRCLE_PULL_REQUEST}
+      - run: echo ${CIRCLE_BRANCH}
+      - run: echo ${CIRCLE_TAG}
+      # CIRCLE_PULL_REQUEST is a URL in the form: https://github.com/metallb/metallb/pull/1
+      # We use sed to get the pull request ID out of the URL.  CIRCLE_PR_NUMBER would be
```
johananl, line 71:
```suggestion
      # We use sed to get the pull request ID out of the URL. CIRCLE_PR_NUMBER would be
```

.circleci/config.yml
```diff
@@ -51,6 +51,57 @@ jobs:
...
+      # more convenient, but unfortunately it is not always available.
+      # https://circleci.com/docs/2.0/env-vars/
+      #
+      # If the job is being run against a PR, use "pr-<ID>" as the image tag and push the
+      # result to metallbdev/controller.  Otherwise this job is not expected to run so we
```
johananl, line 76:
```suggestion
      # result to metallbdev/controller. Otherwise this job is not expected to run so we
```

.circleci/config.yml
```diff
@@ -51,6 +51,57 @@ jobs:
...
+      - run: echo "export GO111MODULE=on" >>$BASH_ENV
+      - run: echo ${CIRCLE_PULL_REQUEST}
+      - run: echo ${CIRCLE_BRANCH}
+      - run: echo ${CIRCLE_TAG}
+      # CIRCLE_PULL_REQUEST is a URL in the form: https://github.com/metallb/metallb/pull/1
+      # We use sed to get the pull request ID out of the URL.  CIRCLE_PR_NUMBER would be
+      # more convenient, but unfortunately it is not always available.
+      # https://circleci.com/docs/2.0/env-vars/
+      #
+      # If the job is being run against a PR, use "pr-<ID>" as the image tag and push the
```
johananl, line 75: I guess CircleCI doesn't support triggering only on PRs (which would make this check unnecessary). I see we use branch filters but we still need this check in case someone (e.g. me :grimacing:) pushes to a branch in this repo (right?).

website/content/community/_index.md
```diff
@@ -145,6 +145,18 @@ the [github repository](https://github.com/google/metallb), and add
 your fork as a remote in `$GOPATH/src/go.universe.tf/metallb`, with
 `git remote add fork git@github.com:<your-github-user>/metallb.git`.
 
+## Testing a Pull Request
+
+MetalLB uses CI to publish images of the latest version of code from each open
```
johananl, line 150:
```suggestion
MetalLB has a CI process which publishes Docker images from the latest code revision of each open
```

I find this phrasing better.

website/content/community/_index.md
```diff
@@ -145,6 +145,18 @@ the [github repository](https://github.com/google/metallb), and add
...
+pull request.  If you would like to test out a prebuilt image of a PR, make
```
johananl, line 151:
```suggestion
pull request. If you would like to test out a prebuilt image of a PR, make
```

website/content/community/_index.md
```diff
@@ -145,6 +145,18 @@ the [github repository](https://github.com/google/metallb), and add
...
+sure the `deploy-speaker` or `deploy-controller` jobs have succeeded on the PR,
+and then you may pull `metallbdev/speaker:pr-<ID>` or
+`metallbdev/controller:pr-<ID>`.
+
+If a new revision of a pull request is posted, the image will be rebuilt.  If
```
johananl, line 156:
```suggestion
If a new revision of a pull request is posted, the image will be rebuilt. If
```

Looks like your editor consistently inserts double spaces after a period :thinking: Could you please fix all occurrences?

russellb: My thumb does it.  ;-)

I'll change them.

website/content/community/_index.md
russellb: ok, done.

.circleci/config.yml
russellb: Right ... it may be possible, but I haven't figured out how.  I had to change a previous attempt to filter out pull requests here: https://github.com/metallb/metallb/commit/8b5c60f55a7799cec01e15b94cf6b7bc3dc8ad34

.circleci/config.yml
```diff
@@ -51,6 +51,57 @@ jobs:
...
+      - run: echo ${CIRCLE_TAG}
+      # CIRCLE_PULL_REQUEST is a URL in the form: https://github.com/metallb/metallb/pull/1
+      # We use sed to get the pull request ID out of the URL. CIRCLE_PR_NUMBER would be
+      # more convenient, but unfortunately it is not always available.
+      # https://circleci.com/docs/2.0/env-vars/
+      #
+      # If the job is being run against a PR, use "pr-<ID>" as the image tag and push the
+      # result to metallbdev/controller. Otherwise this job is not expected to run so we
+      # just fail.
+      - run: if [ -n "$CIRCLE_PULL_REQUEST" ]; then PATH=./bin:$PATH inv push-multiarch --binaries=controller --tag=pr-$(echo $CIRCLE_PULL_REQUEST | sed -e 's/.*pull\/\(.*\)/\1/') --docker-user=metallbdev ; else exit 1 ; fi
```
rata, line 78: using something like: `$(basename "$CIRCLE_PULL_REQUEST")` wouldn't be cleaner to read rather than `sed`?

russellb: good suggestion, thanks!

Issue #646, opened 2020-06-21
Announce route for nodeless endpoints


calfox:
LB Services with non-pod endpoints defined are assigned an IP from the pool in Layer2 mode but this is not currently announced as they can't be linked to a particular node.

## Testing

This service will get assigned an IP but this will not be accessible outside of the cluster as no node takes ownership of ARP replies:

```
apiVersion: v1
kind: Service
metadata:
  name: test-svc
spec:
  type: LoadBalancer
  ports:
    - port: 80
      protocol: TCP
---
apiVersion: v1
kind: Endpoints
metadata:
  name: test-svc
subsets:
  - addresses:
      - ip: 192.168.0.10
    ports:
      - port: 80
```

## Possible Solutions

Update the `usableNodes()` function in `speaker/layer2_controller.go` to support EndpointAddress structs that only return an IP (as above), listing all nodes as usable if none are linkable but an external IP exists:

```
func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
	var activeNodes map[string]bool
	if mList != nil {
		activeNodes = map[string]bool{}
		for _, n := range mList.Members() {
			activeNodes[n.Name] = true
		}
	}

	usable := map[string]bool{}
	extEp := false
	for _, subset := range eps.Subsets {
		for _, ep := range subset.Addresses {
			if ep.NodeName == nil {
				extEp = true
				continue
			}
			if activeNodes != nil {
				if _, ok := activeNodes[*ep.NodeName]; !ok {
					continue
				}
			}
			if _, ok := usable[*ep.NodeName]; !ok {
				usable[*ep.NodeName] = true
			}
		}
	}

	if extEp && len(usable) == 0 {
		usable = activeNodes
	}

	var ret []string
	for node, ok := range usable {
		if ok {
			ret = append(ret, node)
		}
	}

	return ret
}
```

A simpler solution may also be possible where all nodes are returned as usable on the first external IP, assuming node-based and external Endpoints cannot be mixed, wasn't sure on that:

```
usable := map[string]bool{}
SubsetLoop:
	for _, subset := range eps.Subsets {
		for _, ep := range subset.Addresses {
			if ep.NodeName == nil {
				usable = activeNodes
				break SubsetLoop
			}
			if activeNodes != nil {
				if _, ok := activeNodes[*ep.NodeName]; !ok {
					continue
				}
			}
			if _, ok := usable[*ep.NodeName]; !ok {
				usable[*ep.NodeName] = true
[...]

PR #665, opened 2020-07-13
layer2: fallback speaker election if no endpoints are in nodes with speaker daemonset

### LINKED ISSUE #302: Layer 2 controller function usableNodes() should return all nodes for "externalTrafficPolicy: Cluster" service

WKBae:
Fixes #302.
As the issue states, even if service specifies `externalTrafficPolicy: Cluster`, MetalLB will not announce if no endpoint reside in the node with speakers. This can occur if speakers are deployed only on a subset of nodes in the cluster.

This commit makes use of `activeNodes` as a fallback if no usable nodes are available. If memberlist is not enabled, it will act same as if the fallback is disabled.
The idea is basically same as #613, with minor code implementation differences.

### DIFF
--- a/speaker/layer2_controller.go
+++ b/speaker/layer2_controller.go
@@ -39,7 +39,7 @@ func (c *layer2Controller) SetConfig(log.Logger, *config.Config) error {
 
 // usableNodes returns all nodes that have at least one fully ready
 // endpoint on them.
-func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
+func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist, fallbackToActive bool) []string {
 	var activeNodes map[string]bool
 	if mList != nil {
 		activeNodes = map[string]bool{}
@@ -71,12 +71,18 @@ func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
 			ret = append(ret, node)
 		}
 	}
+	if len(ret) == 0 && fallbackToActive {
+		for node := range activeNodes {
+			ret = append(ret, node)
+		}
+	}
 
 	return ret
 }
 
 func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, svc *v1.Service, eps *v1.Endpoints) string {
-	nodes := usableNodes(eps, c.mList)
+	isClusterSvc := svc.Spec.ExternalTrafficPolicy == "Cluster"
+	nodes := usableNodes(eps, c.mList, isClusterSvc)
 	// Sort the slice by the hash of node + service name. This
 	// produces an ordering of ready nodes that is unique to this
 	// service.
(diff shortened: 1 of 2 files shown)


russellb:
Could you consider adding a test case that covers this?  Thanks!

liuyuan10:
@russellb given what we've discussed in #562 shall we drop this? note that this change also doesn't solve the issue if memberlist is not enabled.

rata:
Hi @WKBae ! Thanks a lot for the PR. The use case is quite valid, but this is "by design" currently on MetalLB. See for example this issue comment: https://github.com/metallb/metallb/issues/302#issuecomment-422950439. So, this should probably go into a minor release (just to set more clear expectations on when this might be fixed: not in the 0.9 patch releases :)).

As @liuyuan10 was saying, it probably makes sense to put this on hold until a solution for the memberlist re-join issue is merged. That will change this code quite a lot and is critical for the next patch release.

When v0.9.4 is out, we can investigate different ways (it might become even simpler) to have this functionality.

I'll add a do not merge label and we can revisit afterwards. Thanks again for the PR! :)

WKBae:
Thanks for checking this out! I've read through some issues and seen somthing bigger is going on. I'm looking forward to those PRs being merged.
Feel free to mention me or close this PR once they're settled :)

rata:
@WKBae will do, thanks again! :)

WeirdMachine:
This would also fix https://github.com/metallb/metallb/issues/646

Issue #666, opened 2020-07-13
Add Copyright Year and Name to the License


sha224:
Right now, the license is the default template that GitHub automatically provides. It needs to be modified [here](https://github.com/metallb/metallb/blob/main/LICENSE#L190) to reflect the copyright year and name. I stumbled upon this while filling out a Software Approval Form at my job which requires me to provide the copyright holder's name and the copyright year.

russellb:
I do not believe that changing the LICENSE file is correct.  That is the literal text of the license: https://www.apache.org/licenses/LICENSE-2.0.txt and is just providing general instructions about noting the license within another file

sha224:
My bad. I misunderstood how licensing a project works. Thanks for the clarification.

PR #667, opened 2020-07-13
Add Copyright Year and Name to the License

sha224:
Fix Issue #666

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -187,7 +187,7 @@
       same "printed page" as the copyright notice for easier
       identification within third-party archives.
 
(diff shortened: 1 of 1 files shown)


champtar:
I'm not a lawyer, but each contributor has copyright on his own contribution I think, that is why it's pretty hard to change the license of an Open Source project

russellb:
That line is part of the license itself and is part of the instructions for how to note the license within a source file.  This file should not be changed.  https://www.apache.org/licenses/LICENSE-2.0.txt

Issue #669, opened 2020-07-14
In layer2 mode, metallb don't find address-pools other than first one


dtourde:
In layer2 mode (not tested with BGP yet), services created with other than first address-pool fails to find pool by name.

configmap.yml
```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: metallb-system
  name: config
data:
  config: |
    address-pools:
    - name: default
      protocol: layer2
      addresses:
      - 10.0.11.120-10.0.11.159
    - name: core-services
      protocol: layer2
      addresses:
      - 10.0.11.100-10.0.11.119
      auto-assign: false
    - name: public
      protocol: layer2
      addresses:
        - 10.0.11.254
      auto-assign: false
```

mynginx-svc-lb.yml
```yaml
apiVersion: v1
kind: Service
metadata:
  name: mynginx-svc-lb
  namespace: mynginx
  annotations:
    metallb.universe.tf/address-pool: public
spec:
  selector:
    app: mynginx
  ports:
    - port: 80
      targetPort: 80
  type: LoadBalancer
```

service mynginx-svc-lb:
```shell
Name:                     mynginx-svc-lb
Namespace:                mynginx
Labels:                   <none>
Annotations:              metallb.universe.tf/address-pool: public
Selector:                 app=mynginx
Type:                     LoadBalancer
IP:                       10.43.236.132
Port:                     <unset>  80/TCP
TargetPort:               80/TCP
NodePort:                 <unset>  30077/TCP
Endpoints:                <none>
Session Affinity:         None
External Traffic Policy:  Cluster
Events:
  Type     Reason            Age   From                Message
  ----     ------            ----  ----                -------
  Warning  AllocationFailed  3s    metallb-controller  Failed to allocate IP for "mynginx/mynginx-svc-lb": unknown pool "public"
```

kubectl version:
```shell
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.3", GitCommit:"2e7996e3e2712684bc73f0dec0200d64eec7fe40", GitTreeState:"archive", BuildDate:"2020-06-02T11:18:12Z", GoVersion:"go1.14.3", Compiler:"gc", Platform:"linux/amd64"}
[...]

PR #671, opened 2020-07-19
Add a note to the website about single-IP address pool

dtourde:
This PR concerns issue #669 I filled about configuration problem.
And the solution was to use a CIDR notation even for a single IP.

There is my proposal to the documentation.

### DIFF
--- a/website/content/configuration/_index.md
+++ b/website/content/configuration/_index.md
@@ -265,6 +265,11 @@ Addresses can still be specifically allocated from the "expensive"
 pool with the methods described in
 the [usage](/usage/#requesting-specific-ips) section.
 
+{{% notice note %}}
+Please note that if you have a single address pool, you must use `/32`
+CIDR notation (e.g. `42.176.25.64/32`).
+{{% /notice %}}
+
 ### Handling buggy networks
 
 Some old consumer network equipment mistakenly blocks IP addresses


caboteria:
LGTM @rata

website/content/configuration/_index.md
```diff
@@ -265,6 +265,24 @@ Addresses can still be specifically allocated from the "expensive"
 pool with the methods described in
 the [usage](/usage/#requesting-specific-ips) section.
 
+### Single-IP address pool
```
johananl, line 268: I think having a dedicated section for this "gotcha" is a bit of an overkill. I suggest we add a `{{% notice note %}}` instead in some place where it makes sense.

website/content/configuration/_index.md
```diff
@@ -265,6 +265,24 @@ Addresses can still be specifically allocated from the "expensive"
...
+
+As the implementation uses `net.ParseCIDR`, you need to provide a /32 mask for a single IP pool.
```
johananl, line 270: I wouldn't mention implementation details in the user-facing docs, but rather simply state what's supported without the "why".
I would also omit the YAML snippet since all we have to do is mention that entries under `addresses` should always contain a mask, even when the CIDR is `/32`.

dtourde: You're right @johananl , whole section might be overkill for this.

As a single IP pool make sense only when dealing with multiple address pool & controlling automatic address allocation, I choose this section to update.

I added another commit rather than squashing, so approver still have the possibility to revert to the previous proposition.

russellb:
The text looks good, though @johananl had suggested to use the syntax to have this rendered as a note.  See this for an example:

https://learn.netlify.app/en/shortcodes/notice/

johananl:
LGTM. @dtourde could you please squash the 3 commits into a single commit?

dtourde:
> LGTM. @dtourde could you please squash the 3 commits into a single commit?

Done ! :)

johananl:
Thanks a lot @dtourde.

PR #672, opened 2020-07-20
Make BGP source address configurable

### LINKED ISSUE #670: support selecting the source address when speaker sets up a bgp session
When the speaker sets up a bgp session, because this is low-level networking, it sometimes is important to set the source address for the socket without necessarily relying on the kernel's routing table. 

This [line in the speaker](https://github.com/metallb/metallb/blob/81dc21aa9ae9aa7478bb93523fb4664ce7f3ecf1/internal/bgp/bgp.go#L523) explicitly just takes the default:

```go
func dialMD5(ctx context.Context, addr, password string) (net.Conn, error) {
	laddr, err := net.ResolveTCPAddr("tcp", "[::]:0")
```

We already have the ability to set the `router-id` in the config yaml. For example, if I have 2 addresses on my default interface, say private `10.10.10.10` and public `100.200.100.200`, with the default route set to via the public, then if in a certain config I use the router-id as 10.10.10.10, or even 10.100.100.100, I will want to do one of two things:

* use the router-id as my source address - this was suggested in #422 with a related PR that was closed for reasons unknown
* explicitly set the source address

Either of these would work. The UX on the config would look something like:

```yaml
    peers:
    - my-asn: 65000
      peer-asn: 65530
      peer-address: 1.2.3.4
      router-id: 10.10.10.10
      router-id-as-src: true
```

or 

```yaml
    peers:
    - my-asn: 65000
      peer-asn: 65530
      peer-address: 1.2.3.4
      router-id: 10.10.10.10
      src: 10.10.10.10
```

The second seems a bit redundant, but is flexible in certain cases.

FWIW, much bgp-aware routing software has the possibility to do this, bird being the most prominent example.

In our use case, metallb flat out doesn't work without being able to control the packets. We could mess around with routing tables on the host, but there are 2 major shortcomings to it:

1. We don't always have access to the host; the speaker _is_ the agent that runs on the host.
1. More importantly, I don't always want packets targeted at 1.2.3.4 to be sourced from 10.10.10.10. It is just for certain low-level things (like bgp from the metallb speaker).

I am more than happy for this to be an option, rather than the default, but some way to make it happen is important.

cc @johananl @rata

johananl:
## General

Fixes #670.

This PR adds the `src-address` knob to the BGP peer configuration.

The implementation falls back to the existing behavior (let the kernel figure out the source address) in case the specified address doesn't exist or if there is any other problem with the specified address.

## Testing

To test this PR, deploy MetalLB using `quay.io/jlieb/metallb-speaker:bgp-src` as the speaker image and apply the following sample config while modifying the IPs and ASNs as necessary:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  name: config
  namespace: metallb-system
data:
  config: |
    address-pools:
    - name: default
      protocol: bgp
      addresses:
      - 1.2.3.4/32
    peers:
    - peer-address: 169.254.10.20
      peer-asn: 64501
      my-asn: 64500
      src-address: 10.1.2.3
```

## TODO

- [ ] Update docs
- [ ] Add more tests

### DIFF
--- a/internal/bgp/bgp.go
+++ b/internal/bgp/bgp.go
@@ -29,6 +29,7 @@ type Session struct {
 	routerID         net.IP // May be nil, meaning "derive from context"
 	myNode           string
 	addr             string
+	srcAddr          string
 	peerASN          uint32
 	peerFBASNSupport bool
 	holdTime         time.Duration
@@ -170,7 +171,7 @@ func (s *Session) connect() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 	deadline, _ := ctx.Deadline()
-	conn, err := dialMD5(ctx, s.addr, s.password)
+	conn, err := dialMD5(ctx, s.addr, s.srcAddr, s.password)
 	if err != nil {
 		return fmt.Errorf("dial %q: %s", s.addr, err)
 	}
@@ -351,9 +352,10 @@ func (s *Session) sendKeepalive() error {
 //
 // The session will immediately try to connect and synchronize its
 // local state with the peer.
-func New(l log.Logger, addr string, asn uint32, routerID net.IP, peerASN uint32, holdTime time.Duration, password string, myNode string) (*Session, error) {
+func New(l log.Logger, addr string, srcAddr string, asn uint32, routerID net.IP, peerASN uint32, holdTime time.Duration, password string, myNode string) (*Session, error) {
 	ret := &Session{
 		addr:        addr,
+		srcAddr:     srcAddr,
 		asn:         asn,
 		routerID:    routerID.To4(),
 		myNode:      myNode,
@@ -519,8 +521,20 @@ type tcpmd5sig struct {
 // proper TCP MD5 options when the password is not empty. Works by manupulating
 // the low level FD's, skipping the net.Conn API as it has not hooks to set
 // the neccessary sockopts for TCP MD5.
-func dialMD5(ctx context.Context, addr, password string) (net.Conn, error) {
-	laddr, err := net.ResolveTCPAddr("tcp", "[::]:0")
+func dialMD5(ctx context.Context, addr string, srcAddr string, password string) (net.Conn, error) {
+	// If srcAddr exists on any of the local network interfaces, use it as the
+	// source address of the TCP socket. Otherwise, use the IPv6 unspecified
+	// address ("::") to let the kernel figure out the source address.
+	// NOTE: On Linux, "::" also includes "0.0.0.0" (all IPv4 addresses).
+	//
+	// When the src-address configuration option isn't set by the user, srcAddr
+	// becomes "<nil>", which makes localAddressExists() return false.
+	a := "[::]"
+	if localAddressExists(srcAddr) {
+		a = srcAddr
+	}
+
+	laddr, err := net.ResolveTCPAddr("tcp", fmt.Sprintf("%s:0", a))
 	if err != nil {
 		return nil, fmt.Errorf("Error resolving local address: %s ", err)
 	}
@@ -664,4 +678,34 @@ func buildTCPMD5Sig(addr net.IP, key string) tcpmd5sig {
 	copy(t.key[0:], []byte(key))
 
 	return t
+}
+
+// localAddressExists returns true if the provided address exists locally on
+// any of the host's network interfaces. The function doesn't return errors and
+// instead returns false if an error has occurred while looking for the
+// address. This function is safe to call if addr is nil or malformed.
+func localAddressExists(addr string) bool {
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		return false
+	}
+
+	for _, i := range interfaces {
+		addresses, err := i.Addrs()
+		if err != nil {
+			return false
+		}
+
+		for _, a := range addresses {
+			ip, ok := a.(*net.IPNet)
+			if !ok {
+				return false
+			}
+			if ip.IP.Equal(net.ParseIP(addr)) {
+				return true
+			}
+		}
+	}
+
+	return false
 }
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -40,6 +40,7 @@ type peer struct {
 	MyASN         uint32         `yaml:"my-asn"`
 	ASN           uint32         `yaml:"peer-asn"`
 	Addr          string         `yaml:"peer-address"`
+	SrcAddr       string         `yaml:"src-address"`
 	Port          uint16         `yaml:"peer-port"`
 	HoldTime      string         `yaml:"hold-time"`
 	RouterID      string         `yaml:"router-id"`
@@ -98,6 +99,8 @@ type Peer struct {
 	ASN uint32
 	// Address to dial when establishing the session.
 	Addr net.IP
+	// Source address to use when establishing the session.
+	SrcAddr net.IP
 	// Port to dial when establishing the session.
 	Port uint16
 	// Requested BGP hold time, per RFC4271.
@@ -269,6 +272,7 @@ func parsePeer(p peer) (*Peer, error) {
 			return nil, fmt.Errorf("invalid router ID %q", p.RouterID)
 		}
 	}
+	src := net.ParseIP(p.SrcAddr)
 
 	// We use a non-pointer in the raw json object, so that if the
 	// user doesn't provide a node selector, we end up with an empty,
@@ -294,6 +298,7 @@ func parsePeer(p peer) (*Peer, error) {
 		MyASN:         p.MyASN,
 		ASN:           p.ASN,
 		Addr:          ip,
+		SrcAddr:       src,
 		Port:          port,
 		HoldTime:      holdTime,
 		RouterID:      routerID,
--- a/speaker/bgp_controller_test.go
+++ b/speaker/bgp_controller_test.go
@@ -95,7 +95,7 @@ type fakeBGP struct {
 	gotAds map[string][]*bgp.Advertisement
 }
 
-func (f *fakeBGP) New(_ log.Logger, addr string, _ uint32, _ net.IP, _ uint32, _ time.Duration, _, _ string) (session, error) {
+func (f *fakeBGP) New(_ log.Logger, addr string, _ string, _ uint32, _ net.IP, _ uint32, _ time.Duration, _, _ string) (session, error) {
 	f.Lock()
 	defer f.Unlock()
 
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -54,6 +54,7 @@ peers:
   peer-port: 1179
   hold-time: 180s
   router-id: 10.20.30.40
+  src-address: 10.20.30.40
 - my-asn: 100
   peer-asn: 200
   peer-address: 2.3.4.5
@@ -98,6 +99,7 @@ address-pools:
 						MyASN:         42,
 						ASN:           142,
 						Addr:          net.ParseIP("1.2.3.4"),
+						SrcAddr:       net.ParseIP("10.20.30.40"),
 						Port:          1179,
 						HoldTime:      180 * time.Second,
 						RouterID:      net.ParseIP("10.20.30.40"),
--- a/speaker/bgp_controller.go
+++ b/speaker/bgp_controller.go
@@ -184,7 +184,7 @@ func (c *bgpController) syncPeers(l log.Logger) error {
 			if p.cfg.RouterID != nil {
 				routerID = p.cfg.RouterID
 			}
-			s, err := newBGP(c.logger, net.JoinHostPort(p.cfg.Addr.String(), strconv.Itoa(int(p.cfg.Port))), p.cfg.MyASN, routerID, p.cfg.ASN, p.cfg.HoldTime, p.cfg.Password, c.myNode)
+			s, err := newBGP(c.logger, net.JoinHostPort(p.cfg.Addr.String(), strconv.Itoa(int(p.cfg.Port))), p.cfg.SrcAddr.String(), p.cfg.MyASN, routerID, p.cfg.ASN, p.cfg.HoldTime, p.cfg.Password, c.myNode)
 			if err != nil {
 				l.Log("op", "syncPeers", "error", err, "peer", p.cfg.Addr, "msg", "failed to create BGP session")
 				errs++
@@ -286,6 +286,6 @@ func (c *bgpController) SetNode(l log.Logger, node *v1.Node) error {
 	return c.syncPeers(l)
 }
 
-var newBGP = func(logger log.Logger, addr string, myASN uint32, routerID net.IP, asn uint32, hold time.Duration, password string, myNode string) (session, error) {
-	return bgp.New(logger, addr, myASN, routerID, asn, hold, password, myNode)
+var newBGP = func(logger log.Logger, addr string, srcAddr string, myASN uint32, routerID net.IP, asn uint32, hold time.Duration, password string, myNode string) (session, error) {
+	return bgp.New(logger, addr, srcAddr, myASN, routerID, asn, hold, password, myNode)
 }


johananl:
@deitch - FYI.

internal/bgp/bgp.go
```diff
@@ -665,3 +679,30 @@ func buildTCPMD5Sig(addr net.IP, key string) tcpmd5sig {
...
+// localAddressExists returns true if the provided address exists locally on
+// any of the host's network interfaces. The function doesn't return errors and
+// instead returns false if an error has occurred while looking for the
+// address. This function is safe to call if addr is nil or malformed.
+func localAddressExists(addr string) bool {
+	interfaces, err := net.Interfaces()
+	if err != nil {
+		return false
+	}
+
```
rata, line 692: You probably want to check that the interface is up. See this function, for example: https://github.com/metallb/metallb/blob/80bfe1cf45bf4338e6eced11ab7a88d3d2356663/internal/layer2/announcer.go#L46

internal/bgp/bgp.go
```diff
@@ -665,3 +679,30 @@ func buildTCPMD5Sig(addr net.IP, key string) tcpmd5sig {
...
+	for _, i := range interfaces {
+		addresses, err := i.Addrs()
+		if err != nil {
+			return false
+		}
+
+		for _, a := range addresses {
+			ip := strings.Split(a.String(), "/")[0]
+			if net.ParseIP(ip).Equal(net.ParseIP(addr)) {
```
rata, line 701: You can probably simplify this by using what we are doing for layer2. Just "inspire" from this: https://github.com/metallb/metallb/blob/80bfe1cf45bf4338e6eced11ab7a88d3d2356663/internal/layer2/announcer.go#L85-L89

So, you can just check if it is type IPNet and if it is, just `net.ParseIP()` with the `.IP` field.

The strings container with `/` seems more fragile and inconsistent with other parts of the code :)

internal/config/config.go
```diff
@@ -269,6 +272,7 @@ func parsePeer(p peer) (*Peer, error) {
 			return nil, fmt.Errorf("invalid router ID %q", p.RouterID)
 		}
 	}
+	src := net.ParseIP(p.SrcAddr)
```
rata, line 275: if this returns nil, I guess the code handles it correctly. Right?

internal/bgp/bgp.go
```diff
@@ -519,8 +522,19 @@ type tcpmd5sig struct {
...
+	// If srcAddr exists on any of the local network interfaces, use it as the
+	// source address of the TCP socket. Otherwise, fall back to the IPv6
+	// unspecified address ("[::]") to let the kernel figure out the source
+	// address.
+	var a string
+	if localAddressExists(srcAddr) {
+		a = srcAddr
+	} else {
+		a = "[::]"
+	}
```
rata, line 535: do this instead, to simplify?
```suggestion

	a := "[::]"
	if localAddressExists(srcAddr) {
		a = srcAddr
	}
```

internal/bgp/bgp.go
```diff
@@ -519,8 +522,19 @@ type tcpmd5sig struct {
 // proper TCP MD5 options when the password is not empty. Works by manupulating
 // the low level FD's, skipping the net.Conn API as it has not hooks to set
 // the neccessary sockopts for TCP MD5.
-func dialMD5(ctx context.Context, addr, password string) (net.Conn, error) {
-	laddr, err := net.ResolveTCPAddr("tcp", "[::]:0")
+func dialMD5(ctx context.Context, addr string, srcAddr string, password string) (net.Conn, error) {
+	// If srcAddr exists on any of the local network interfaces, use it as the
+	// source address of the TCP socket. Otherwise, fall back to the IPv6
+	// unspecified address ("[::]") to let the kernel figure out the source
+	// address.
```
rata, line 529: This seems confusing, as "[::]" will of course use an IPv4 address if that is okay too. The text seems confusing in that regard.

I'd also say that this will set the src-address only if the field is set in the configuration (is optional)

internal/bgp/bgp.go
```diff
@@ -665,3 +679,30 @@ func buildTCPMD5Sig(addr net.IP, key string) tcpmd5sig {
 
 	return t
 }
+
+// localAddressExists returns true if the provided address exists locally on
+// any of the host's network interfaces. The function doesn't return errors and
+// instead returns false if an error has occurred while looking for the
+// address. This function is safe to call if addr is nil or malformed.
+func localAddressExists(addr string) bool {
```
rata, line 687: This function is missing logs. It will be hard to diagnose any problems otherwise

internal/bgp/bgp.go
```diff
@@ -665,3 +679,30 @@ func buildTCPMD5Sig(addr net.IP, key string) tcpmd5sig {
...
+	interfaces, err := net.Interfaces()
```
rata, line 688: I'd add a check for:
```
if addr == nil {
return false
}
```

So this is clearly a no-op when the string wasn't set and after that you can log. If you are doing something after, it is worth logging as *some* address was specified and the user expects a behavior. So, doing that will make logging just easy too :-)

internal/bgp/bgp.go
rata, line 687: Just curious: why use this function to see if the srcIP is present instead of trying to use it directly? The error is not clear if we just use it? Or it may allow sending some weird packages over the network to a user that can just modify metalLB config?

Just to understand the reasoning. Seems legit.

adamdunstan:
Perhaps there is something I dont understand.  It appears there is a single address, (not a list) therefore isn't this attempting to use the same source address on every speaker?  Wouldn't only one speaker match the address and the rest of the speakers would use their host address?  I'm not suggesting this should be a list, that would be a configuration nightmare.  (Hence the reason I suggested match logic as the solution to this problem). Perhaps a variant is to use an address mask, but that would have limited utility as the configuration is cluster wide and different address blocks could be used by many speakers.  Unless I am missing something?

speaker/bgp_controller_test.go
```diff
@@ -95,7 +95,7 @@ type fakeBGP struct {
 	gotAds map[string][]*bgp.Advertisement
 }
 
-func (f *fakeBGP) New(_ log.Logger, addr string, _ uint32, _ net.IP, _ uint32, _ time.Duration, _, _ string) (session, error) {
+func (f *fakeBGP) New(_ log.Logger, addr string, srcAddr string, _ uint32, _ net.IP, _ uint32, _ time.Duration, _, _ string) (session, error) {
```
johananl, line 98: TODO: change `srcAddr` to `_` since this argument isn't used.

deitch:
@adamdunstan wrote:

> It appears there is a single address, (not a list) therefore isn't this attempting to use the same source address on every speaker?

I think so, yes. That is acceptable in many scenarios, especially with the powerful use of  `node-selectors` in the metallb config. When I do it, I actually have a unique config for each and every node, since BGP peers may vary by speaker as well.

I agree it is suboptimal (and based on my conversations with @rata and @johananl, they pretty much agree as well). A better formulation would be one of the following:

* the annotations work that is in #593 - if you want specific settings, just use the annotations. I am a big fan of this one.
* a "discovery" or "logic" rule to determine what the source address would be. A regex would work (your "match logic"), a "can-I-reach" would work (wherein speaker would try to create a session with the peer from each discovered address until it succeeds), so would simple rules (e.g. "private IPv4 on first interface" or similar)

The virtue of the approach in this PR is that it is easy (therefore) fast to implement, it will work on specific nodes if you use the `node-selectors` correctly, it is completely optional, and it doesn't break anything. Further, it aligns really nicely with the annotations approach in #593 , which would make the "per-node" config an outside-of-metallb thing by setting annotations (however someone chooses to do that), which scales well.

The annotations and/or logic could be added as future features, but doesn't contradict them.

johananl:
@adamdunstan you are 100% correct, this new functionality only makes sense to use in cases where you have per-speaker BGP peers. This use case seems to be quite popular among MetalLB users and this is one reason I'm trying to push #593.

I agree this solution is suboptimal. I'd rather have a more complete solution which can cover more use cases. However, since this solution is quick and simple and since it has built-in fallback to the current, unconfigurable behavior, I'm OK with getting this merged to address at least **one** use case (@deitch 's) which is an improvement over the current situation. We'll add a note in the docs that this knob is intended to be used with per-speaker peers.

I'm very open to expanding the solution in the near future and making it more robust.

internal/bgp/bgp.go
johananl: You can use an address that's configured on an interface that is down as the source address :smile: Weird, but it works on my testing, at least for the outbound packets. I have no idea if a BGP session will actually be established in case the user uses an address of a downed interface, however I don't think we should actively prevent users from trying to do so.

internal/bgp/bgp.go
johananl: Nice, thanks. I wasn't satisfied with the `/` thing.

deitch:
@johananl I have not tested yet. I saw all of the comments and your pushes. Kindly flag me when you are at another milestone, and I will try it.

internal/config/config.go
johananl: Yes, a nil value in `src` will eventually make `localAddressExists()` return `false`.

internal/bgp/bgp.go
johananl: Thought about this but opted for clear rather than clever here. Anyway, I'm fine with your suggestion, too. Fixed.

internal/bgp/bgp.go
johananl: True, and thanks for noticing. For some reason I thought the entire package doesn't have logging and didn't want to expand the scope of this PR by adding logging. Now I see that I can easily add logging to this function.

adamdunstan:
A load balancer that isnt a load balancer, a router that isnt a router?  I think this a PR is a great example of why a bi-weekly call is important.  I do not agree, I think this add a configuration option that will cause confusion.  To make the configuration valid for both k8s cluster and routing infra, the user will need to configure a complex set of node selectors and match those selectors, node addresses, reachability and bgp peer address (unless using bgp listen ranges which I doubt anyone with this level of static configuration would use).  Incorrect configuration would result in silent failure, failure modes would be complex.  If this level of static configuration is acceptable, I dont see why another address is needed, unless the node address is being used by another bgp router on the node (eg CNI),  and if that was the case, further static configuration can be added to the BGP router to work around that problem.    I guess this is why I have asked for a specific use case that this solves as this looks like a PR addressing a symptom, not a cause.

internal/bgp/bgp.go
johananl: I've used the term "IPv6 unspecified address" because that's the official name of `::`. Anyway, I added a note which makes it clear that `::` also includes IPv4 on Linux.

johananl: >I'd also say that this will set the src-address only if the field is set in the configuration (is optional)

Done.

deitch:
Hi @adamdunstan , as I am one of the end-users who has a need for this (and opened #670 ), going to address which of your notes that I can.

> the user will need to configure a complex set of node selectors and match those selectors, node addresses, reachability and bgp peer address (unless using bgp listen ranges which I doubt anyone with this level of static configuration would use). Incorrect configuration would result in silent failure, failure modes would be complex. 

Isn't this already the case? There are plenty of installations where there isn't one BGP peer, config, etc. for the entire region/zone, but one for every node. A custom source IP fits this already-existing model exactly. Using complex set of node selectors is exactly what we already have to do. 

I do not contend that this is perfect; I do contend that it fits with the existing pattern and capabilities as it is today.

To be really honest, the metallb-style node-selector config is the _best_ solution I have seen to that mess. The fact that I can configure a single file (the ConfigMap) with sufficient label-driven rules to pick the right thing is far simpler than anything I have had to do in this space until now.

Given the real world, what would be ideal? Not the algorithms, much as I like them. They actually would make debugging harder, as some of the config would be statically defined (peer IP, e.g.) and some algorithmically defined (source IP).

I think ideal would be #593, and include source-ip as an annotation. That would let metallb say, "you want per-node config? You want funny algorithms? No problem. _Leave me out of it_. Set the right annotations on a node and I will consume them. How you set them is your problem." The whole thing would be massively cleaner from metallb's perspective. But that one has been in process for ~3 months and shows no sign of going in soon.

Given the above - not ideal, but "not ideal" in the same way as the current static config is; ideal solution seems to be a while out; supporting selected source-ip for a network connection is not only standard but almost required for all network apps and utilities - how would you suggest we move it ahead? 

I am just trying to solve a very real world problem.

johananl:
> A load balancer that isnt a load balancer, a router that isnt a router? I think this a PR is a great example of why a bi-weekly call is important.

We have one scheduled @adamdunstan and you're even on the invite :wink:

>I do not agree, I think this add a configuration option that will cause confusion. To make the configuration valid for both k8s cluster and routing infra, the user will need to configure a complex set of node selectors and match those selectors, node addresses, reachability and bgp peer address (unless using bgp listen ranges which I doubt anyone with this level of static configuration would use).

The reality is that MetalLB doesn't allow you to specify the source address for BGP connections right now, and at least some of the users need this functionality. I agree the use case is rather specific, but given that we are able to make this change "unintrusive", namely without causing any trouble to users who don't care about changing the source address, I don't see a strong reason not to move forward with this.

How would you solve the problem? Do you have a practical alternative you'd like to suggest? I'm happy to consider alternatives.

>Incorrect configuration would result in silent failure, failure modes would be complex.

That's a good point. I'm currently adding logging to the new `localAddressExists()` function which AFAICT is the only complex part of the changeset.

>If this level of static configuration is acceptable, I dont see why another address is needed, unless the node address is being used by another bgp router on the node (eg CNI), and if that was the case, further static configuration can be added to the BGP router to work around that problem. I guess this is why I have asked for a specific use case that this solves as this looks like a PR addressing a symptom, not a cause.

I don't think we're addressing a symptom. As I said above, IMO it is clear that MetalLB doesn't allow you to control the source IP address. Other BGP stacks allow this, e.g. the `update-source` flag on Cisco routers or `local-address` on Juniper. I don't see a reason why BGP source address shouldn't be configurable.

Looks like @deitch has already elaborated on the use case (thanks!).

As a note/question to both @deitch and @adamdunstan, if #593 were already merged and included support for controlling the BGP source using annotations, would it make this PR unnecessary?

deitch:
> As a note/question to both @deitch and @adamdunstan, if #593 were already merged and included support for controlling the BGP source using annotations, would it make this PR unnecessary?

Yes! 100% yes. The whole discussion about, "how should metallb pick its source and peer, and what algorithms, and..." just go away. I would take responsibility for setting these annotations _outside_ of metallb and leave you alone. :-)

internal/bgp/bgp.go
rata: What you mean with "you can use"? What will happen? The same as with any address not configured on the server?

internal/bgp/bgp.go
rata: With your comment [here][link], I doubt if we want to check something at all.  I'm sure you add it for a reason, but can you please share it? :)


[link]: https://github.com/metallb/metallb/pull/672#discussion_r458050524

adamdunstan:
@deitch @johananl   I understand the construct of clusters of speakers using node selectors, however I'm fascinated by how other components of system design have resulted in the requirement for a single speaker with a static IP address, probably beyond the scope of the discussion, I would like to learn if you guys want to continue the conversation elsewhere.  Expanding the problem with the solution a bit.....  One of the elegant components in metallb bgp is that it cleanly uses an address block, allowing the block to be simply put side in IPAM.   Speakers can be run and peered to bgp routers without any more knowledge than the peer address.  The design allows the infrastructure IPAM to do its thing, either using DHCP or some other configuration based scheme during installation.  This PR breaks that isolation as  for the intended behavior a configmap must be updated with the correct address, which I assume is coming from an IPAM?  The solution that retains separation between infrastructure IPAM and metallb is using NIC masks, which will apply uniformly to all nodes where the speaker runs (assuming uniform udev rules).    However, I have had my say......

deitch:
> I would like to learn if you guys want to continue the conversation elsewhere

I would be happy to. Ping me on Kubernetes slack or CNCF slack or Docker slack or.... 😄

> One of the elegant components in metallb bgp is that it cleanly uses an address block, allowing the block to be simply put side in IPAM
> Speakers can be run and peered to bgp routers without any more knowledge than the peer address. The design allows the infrastructure IPAM to do its thing, either using DHCP or some other configuration based scheme during installation. This PR breaks that isolation as for the intended behavior a configmap must be updated with the correct address, which I assume is coming from an IPAM?

When you say, "infrastructure IPAM", I assume you mean on the host? Since metallb speakers run with `network=host` anyways?

To some degree, it never could be fully isolated, since it needs the BGP peer address. Doesn't that, in essence, mean it is (at least somewhat) aware of network structure and addresses? It doesn't discover the peer (thus breaking the isolation) but is provided it by the config. We would do the same for source-ip. metallb can remain ignorant (isolated?) from all of that, just accepting its config as is.

> The solution that retains separation between infrastructure IPAM and metallb is using NIC masks

How would that work?

caboteria:
Sorry I missed this until now.  I don't understand something (and apologies in advance if this is an ELI5) - speakers are managed by a DaemonSet so they run on all nodes in the cluster, but there's only one "src-address".  How does this work in a multi-node cluster?  Wouldn't we need an address per node (i.e., map instead of single address)?

johananl:
> Sorry I missed this until now. I don't understand something (and apologies in advance if this is an ELI5) - speakers are managed by a DaemonSet so they run on all nodes in the cluster, but there's only one "src-address". How does this work in a multi-node cluster? Wouldn't we need an address per node (i.e., map instead of single address)?

This feature only makes sense to use in cases where you use the "peer per speaker" topology. You are right that in the current state specifying `src-address` when a peer applies to multiple speakers is meaningless. See https://github.com/metallb/metallb/pull/672#issuecomment-661349756 and https://github.com/metallb/metallb/pull/672#issuecomment-661801084.

Personally I'd rather solve this problem "properly" as part of #593, however doing so will take a bit longer since #593 is a bigger changeset. Since we have our first community meeting tomorrow, how about we discuss it then?

johananl:
#593 now allows specifying the source address for node peers. I'll close this PR once #593 is merged.

rata:
@johananl any reason to leave it open, if the other PR will supersede this? I mean, is not bad either, just if you are *sure* this won't ever switch to "Ready for review" and no more comments here will help this... why not close? It can be re-opened if needed :)

johananl:
> @johananl any reason to leave it open, if the other PR will supersede this? I mean, is not bad either, just if you are _sure_ this won't ever switch to "Ready for review" and no more comments here will help this... why not close? It can be re-opened if needed :)

Closed!

rata:
@johananl In the future, please don't use branches in the MetalLB repo to open a PR. Use a branch in a fork, instead. Thanks!

johananl:
> @johananl In the future, please don't use branches in the MetalLB repo to open a PR. Use a branch in a fork, instead. Thanks!

No problem, but why? :-)

rata:
That is the standard in most communities and it might be confusing to users to see those branches, IMHO.

russellb:
> That is the standard in most communities and it might be confusing to users to see those branches, IMHO.

agree, fwiw

PR #677, opened 2020-07-22
Use distroless static images

### LINKED ISSUE #676: Use distroless static images
It is a common practise to use `SRATCH` or `distroless/static` images for Go executables, thus no need to use `alpine` image.

I did not find any use case where an existing shell is needed, but please add your comment here.

Either `SRATCH` or `distroless/static` image does not contain shell which reduces attack surface and size, too.

Image size changes:
| controller                | Size       |
|------------------- | -------- |
| `alpine`                   | 74.8MB |
| `distroless/static`  | 36.4MB |

| speaker                   | Size       |
|------------------- | -------- |
| `alpine`                   | 83.6MB |
| `distroless/static`  | 40.9MB |

alex1989hu:
Reduce final image size and attack surface

Note: to keep executable permission we must keep
alpine as intermediate container beacuse not all
the developers use the latest Docker which supports
chmod flag for COPY/ADD command.

Image size changes:
 - controller: 74.8MB -> 36.4MB
 - speaker: 83.6MB -> 40.9MB

Fixed #676

Docker xref: https://github.com/moby/buildkit/pull/1492

Signed-off-by: Alex Szakaly <alex.szakaly@gmail.com>

### DIFF
--- a/controller/Dockerfile
+++ b/controller/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:latest
+FROM alpine:latest as intermediate
 
 ADD controller /controller
 
@@ -6,5 +6,9 @@ ADD controller /controller
 # '+x' for others, it won't be possible to execute. Make sure all can execute,
 # just in case
 RUN chmod a+x /controller
+
+# Production image
+FROM gcr.io/distroless/static:latest
+COPY --from=intermediate /controller /controller
 
 ENTRYPOINT ["/controller"]
--- a/speaker/Dockerfile
+++ b/speaker/Dockerfile
@@ -1,4 +1,4 @@
-FROM alpine:latest
+FROM alpine:latest as intermediate
 
 ADD speaker /speaker
 
@@ -6,5 +6,9 @@ ADD speaker /speaker
 # '+x' for others, it won't be possible to execute. Make sure all can execute,
 # just in case
 RUN chmod a+x /speaker
+
+# Production image
+FROM gcr.io/distroless/static:latest
+COPY --from=intermediate /speaker /speaker
 
 ENTRYPOINT ["/speaker"]


controller/Dockerfile
```diff
@@ -7,4 +7,10 @@ ADD controller /controller
 # just in case
 RUN chmod a+x /controller
 
+# Production image
+FROM gcr.io/distroless/static:latest
+COPY --from=intermediate /controller /controller
+
+USER 65534
```
rata, line 14: Any reason to use the `USER` directive instead of using the k8s pod spec?

speaker/Dockerfile
```diff
@@ -7,4 +7,8 @@ ADD speaker /speaker
 # just in case
 RUN chmod a+x /speaker
 
+# Production image
+FROM gcr.io/distroless/static:latest
+COPY --from=intermediate /speaker /speaker
+
```
rata, line 13: Any reason the `USER` directive was there but not here?

controller/Dockerfile
alex1989hu: I wanted to clearly indicate that the image runs as non-root which is the default behaviour. On the other hand, the gain here is that you do not specify the `runAsUser` to actually run as `65534` - I know it is already specified in the k8s manifest.

speaker/Dockerfile
alex1989hu: The image runs as root due to the required capabilities. I was thinking about do I need to indicate the intention here as well I did for `controller` or not - I mean the directive `USER 0`?

russellb:
Note that I started some discussion in the corresponding issue ... #676

speaker/Dockerfile
rata: Well, user 0 is not really required but the capabilities. I think not specifying it here makes sense, then. Thanks! :)

controller/Dockerfile
rata: Well, you can have a PSP in your cluster that mutates the pod spec to not run as root too. So, no need to specify it manually on the pod spec.

I think this is a different concern and I'd leave it for another PR. I think I'd lean towards not specifying the `USER` in the docker image, in fact. I think specifying that on Kubernetes makes more sense (there is little we can know about which user to use here in the MetalLB Dockerfile to choose the user), it can be done automatically via PSPs (no need to manually modify the pod spec) and if the container runtime support user namespaces (we at Kinvolk are working to add that to Kubernetes upstream) it has little sense to have some default at the Dockerfile layer, I think.

In any case, we can discuss and see what is best for the `USER` in another issue/PR and just keep this for distroless. What do you think?

Thanks!

rata:
@russellb thanks! Adding the do-not-merge label to avoid mistakes, until the issue figures out a way forward :)

controller/Dockerfile
alex1989hu: Agreed 👌 will rewrite the commit without `USER`

russellb:
> @russellb thanks! Adding the do-not-merge label to avoid mistakes, until the issue figures out a way forward :)

feel free to drop the label, i don't feel strongly enough to block this

johananl:
Added to the discussion at https://github.com/metallb/metallb/issues/676#issuecomment-665151979. TL;DR: I don't think we should get rid of the shell before ephemeral containers are available on most k8s distros.

daxmc99:
@alex1989hu
Thanks for your contribution and this is valuable but I don't think we plan to merge this right now. 

Your work here serves as a helpful example on how for others to do it if they need smaller images and increased security.

PR #687, opened 2020-08-04
dev-env: Add BGP support

russellb:
Expand the dev-env setup with an option to configure MetalLB in BGP
mode.  It will run frr in a container along side the kind cluster and
each of the speakers will peer with it.

Running `inv dev-env` will stand up a cluster with MetalLB
unconfigured as before.  If you run `inv dev-env -p bgp`, the BGP
configuration will be applied.

To keep the environment completely isolated, no routing is done
outside of the frr container.  You can get a shell in the container to
test connectivity to LoadBalancer IPs.  You can also use the vtysh
console of the router to see the router's status information.  A
README.md file is included in dev-env/bgp/ to demonstrate how to use
this bgp dev env.


dev-env/bgp/frr/bgpd.conf
```diff
@@ -0,0 +1,39 @@
...
+password zebra
+!enable password please-set-at-here
+!
+router bgp 64512
+ neighbor 172.18.0.2 remote-as 64512
+ neighbor 172.18.0.2 next-hop-self
+ neighbor 172.18.0.3 remote-as 64512
+ neighbor 172.18.0.3 next-hop-self
+ neighbor 172.18.0.4 remote-as 64512
+ neighbor 172.18.0.4 next-hop-self
```
aojea, line 17: bear in mind that the ips provided to the kind nodes are assigned by docker, you are not guarantee to have the same ips or in the same order

dev-env/bgp/frr/bgpd.conf
```diff
@@ -0,0 +1,39 @@
...
+!
+! BGPd sample configuration file
+!
+! $Id: bgpd.conf.sample,v 1.1 2002/12/13 20:15:29 paul Exp $
+!
+hostname bgpd
+password zebra
+!enable password please-set-at-here
+!
+router bgp 64512
```
aojea, line 11: if you use the same bgp AS number for both peers it means you run iBGP (internal BGP), in scenarios I think is recommended to run eBGP (external BGP) because you have two different network realms or Autonomous Systems. Also, a "route learnt from an IBGP peer will not be advertised back to another iBGP neighbor by default",

dev-env/bgp/frr/bgpd.conf
russellb: yeah ... I thought about making this smarter, but it has been the same every time I've run it, so then I got lazy :)

dev-env/bgp/frr/bgpd.conf
russellb: In this setup, I don't think propagating routes is necessary.  I can change one of the ASNs though if that makes more sense, though.

aojea: is fine, just important to take into account if you want to grow it adding more peers

dev-env/bgp/frr/bgpd.conf
aojea: if you use for automated testing is risky, if is an example, maybe adding a comment explaining how to obtain those ips?

russellb: Yeah, good point.  I think it should be automated before relying on it for automated testing.

russellb: I'll just do it now.

russellb: OK, this is done now

tasks.py
```diff
@@ -231,6 +233,58 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None):
...
+    with open("%s/frr/bgpd.conf.tmpl" % dev_env_dir, 'r') as f:
+        bgpd_config = "! THIS FILE IS AUTOGENERATED\n" + f.read()
+    bgpd_config = bgpd_config.replace("NODE1_IP", node1_ip.stdout.strip())
+    bgpd_config = bgpd_config.replace("NODE2_IP", node2_ip.stdout.strip())
+    bgpd_config = bgpd_config.replace("NODE3_IP", node3_ip.stdout.strip())
+    with open("%s/frr/bgpd.conf" % dev_env_dir, 'w') as f:
+        f.write(bgpd_config)
+
+    # Run a BGP router in a container for all of the speakers to peer with.
+    run("docker rm etcfrr || true", echo=True)
```
aojea, line 268: maybe it would be useful to use `-f` flag to delete it, you are using it some lines below

tasks.py
```diff
@@ -231,6 +233,58 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None):
...
+    run("docker create -v /etc/frr --name etcfrr alpine:3.4 /bin/true",
+            echo=True)
+    run("docker cp \"%s/dev-env/bgp/frr/.\" etcfrr:/etc/frr" % os.getcwd(),
+            echo=True)
+    run("docker rm -f frr || true", echo=True)
+    run("docker run -d --privileged --network kind --rm --name frr "
+            "--volumes-from etcfrr frrouting/frr:latest", echo=True)
+
+    peer_address = run('docker inspect -f "{{ '
```
aojea, line 277: add a TODO for IPv6 later?

tasks.py
russellb: The `etcfrr` container is never started, though

tasks.py
russellb: good point.  I'll update the README to note this has only been tested with IPv4, as well.

I think the dev-env in general needs some updates to support ipv6

aojea: that's the cool part ;)

aojea:
/lgtm
very nice

tasks.py
```diff
@@ -71,7 +71,7 @@ def build(ctx, binaries, architectures, tag="dev", docker_user="metallb"):
     binaries = _check_binaries(binaries)
     architectures = _check_architectures(architectures)
     _make_build_dirs()
-    
+
```
russellb, line 74: My editor is set up to strip trailing whitespace and I got tired of removing these changes, so I left them.  Normally I would leave unnecessary formatting changes out completely, but since it was only whitespace-only lines changed, I figured it wasn't too distracting this time.

rata: No problem :)

tasks.py
```diff
@@ -231,6 +233,60 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None):
...
+    node1_ip = run('docker inspect -f "{{ '
+            '.NetworkSettings.Networks.kind.IPAddress }}" kind-control-plane',
+            echo=True)
+    node2_ip = run('docker inspect -f "{{ '
+            '.NetworkSettings.Networks.kind.IPAddress }}" kind-worker',
+            echo=True)
+    node3_ip = run('docker inspect -f "{{ '
+            '.NetworkSettings.Networks.kind.IPAddress }}" kind-worker2',
+            echo=True)
```
rata, line 259: All of this fail locally here :-(

I'm using kind: 
```
$ kind version
kind v0.7.0 go1.13.6 linux/amd64
```
Are you using a different version of kind?


The output to help to see why it fails here is:
```
$  docker inspect  kind-control-plane
[
   ....
    {
        "NetworkSettings": {
            "Bridge": "",
            "SandboxID": ...,
            "HairpinMode": false,
            "LinkLocalIPv6Address": "",
            "LinkLocalIPv6PrefixLen": 0,
            "Ports": {
                "6443/tcp": [
                    {
                        "HostIp": "127.0.0.1",
                        "HostPort": "32768"
                    }
                ]
            },
            "SandboxKey": "/var/run/docker/netns/bf9ddfad6a78",
            "SecondaryIPAddresses": null,
            "SecondaryIPv6Addresses": null,
            "EndpointID": ...
            "Gateway": ...
            "GlobalIPv6Address": "",
            "GlobalIPv6PrefixLen": 0,
            "IPAddress": ...
            "IPPrefixLen": 16,
            "IPv6Gateway": "",
            "MacAddress":  ....
            "Networks": {
                "bridge": {
                    "IPAMConfig": null,
                    "Links": null,
                    "Aliases": null,
                    "NetworkID": "596f61bc70fe5d0960da560b55154965274a784dce236a17bd740e8cc4ca6b39",
                    "EndpointID": "e80728eb8b5ce97a017a07912bcf2dfe6181c63e11f16bcc0d3982587ce7c39e",
                    "Gateway": "172.17.0.1",
                    "IPAddress": "172.17.0.4",
                    "IPPrefixLen": 16,
                    "IPv6Gateway": "",
                    "GlobalIPv6Address": "",
                    "GlobalIPv6PrefixLen": 0,
                    "MacAddress": "02:42:ac:11:00:04",
                    "DriverOpts": null
                }
            }
        }
    }
]
$ docker inspect -f "{{ .NetworkSettings.Networks.kind.IPAddress }}" kind-control-plane
<no value>
$ kubectl get nodes
NAME                 STATUS   ROLES    AGE     VERSION
kind-control-plane   Ready    master   2m19s   v1.17.0
kind-worker          Ready    <none>   103s    v1.17.0
kind-worker2         Ready    <none>   103s    v1.17.0
(metallb-venv) rodrigo@lindsay: ~/src/kinvolk/metallb :bgptest$ docker inspect -f "{{ .NetworkSettings.Networks.kind.IPAddress }}" kind-control-plane
<no value>

$ kubectl get nodes -o wide
NAME                 STATUS   ROLES    AGE   VERSION   INTERNAL-IP   EXTERNAL-IP   OS-IMAGE       KERNEL-VERSION   CONTAINER-RUNTIME
kind-control-plane   Ready    master   14m   v1.17.0   172.17.0.4    <none>        Ubuntu 19.10   5.7.0-2-amd64    containerd://1.3.2
kind-worker          Ready    <none>   13m   v1.17.0   172.17.0.2    <none>        Ubuntu 19.10   5.7.0-2-amd64    containerd://1.3.2
kind-worker2         Ready    <none>   13m   v1.17.0   172.17.0.3    <none>        Ubuntu 19.10   5.7.0-2-amd64    containerd://1.3.2

```

Maybe we can use kubectl internal-ip field to get the IPs?

tasks.py
```diff
@@ -231,6 +233,60 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None):
...
+# router in a container and configure MetalLB to peer with it.
+# See dev-env/bgp/README.md for some more information.
+def bgp_dev_env():
+    dev_env_dir = os.getcwd() + "/dev-env/bgp"
+
+    # TODO -- The IP address handling will need updates to add support for IPv6
+
+    node1_ip = run('docker inspect -f "{{ '
+            '.NetworkSettings.Networks.kind.IPAddress }}" kind-control-plane',
+            echo=True)
```
rata, line 253: Haven't checked how to do it in invoke, but will be nice to fail if these steps fail

tasks.py
```diff
@@ -231,6 +233,60 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None):
...
+        f.write(bgpd_config)
+
+    # Run a BGP router in a container for all of the speakers to peer with.
+    run("docker rm etcfrr || true", echo=True)
+    run("docker create -v /etc/frr --name etcfrr alpine:3.4 /bin/true",
+            echo=True)
+    run("docker cp \"%s/dev-env/bgp/frr/.\" etcfrr:/etc/frr" % os.getcwd(),
+            echo=True)
+    run("docker rm -f frr || true", echo=True)
+    run("docker run -d --privileged --network kind --rm --name frr "
```
rata, line 276: Why is privileged needed? To use the kind network?

tasks.py
```diff
@@ -231,6 +233,60 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None):
...
+        bgpd_config = "! THIS FILE IS AUTOGENERATED\n" + f.read()
+    bgpd_config = bgpd_config.replace("NODE1_IP", node1_ip.stdout.strip())
+    bgpd_config = bgpd_config.replace("NODE2_IP", node2_ip.stdout.strip())
+    bgpd_config = bgpd_config.replace("NODE3_IP", node3_ip.stdout.strip())
+    with open("%s/frr/bgpd.conf" % dev_env_dir, 'w') as f:
+        f.write(bgpd_config)
+
+    # Run a BGP router in a container for all of the speakers to peer with.
+    run("docker rm etcfrr || true", echo=True)
+    run("docker create -v /etc/frr --name etcfrr alpine:3.4 /bin/true",
```
rata, line 271: Not sure I follow, the `etcfrr` container is created but never run later?

I guess this is needed for the config to be used [here][link]? Just curious, any reason to create this container and use `--volume-from` and not create a directory and use that as volume?

[link]: https://github.com/metallb/metallb/pull/687/files#diff-dc31105b816206de63c47bf42ec2a4d7R277

tasks.py
```diff
@@ -231,6 +233,60 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None):
...
+        print("Configuring MetalLB with a BGP test environment")
+        bgp_dev_env()
+    else:
+        print("Leaving MetalLB unconfigured")
+
+
+# Configure MetalLB in the dev-env for BGP testing.  Start an frr based BGP
+# router in a container and configure MetalLB to peer with it.
+# See dev-env/bgp/README.md for some more information.
+def bgp_dev_env():
```
rata, line 246: Now we are not leaving just a kind cluster, but a kind cluster and some containers (some not even started). Would it make sense to add a inv task to clean that up?

tasks.py
aojea: kind since 0.8.0+ changed the network model and uses a custom bridge named kind

aojea: iterating over the range works for both versions, this snippet is from the kind code and works for ipv4 and ipv6, but just remove the family that you don't want :)
```
	// retrieve the IP address of the node using docker inspect
	cmd := exec.Command("docker", "inspect",
		"-f", "{{range .NetworkSettings.Networks}}{{.IPAddress}},{{.GlobalIPv6Address}}{{end}}",
		n.name, // ... against the "node" container
	)
```

rata: Thanks!

I _hope_ kubectl might be simpler but both will work :)

russellb: `kubectl` works fine for getting the Node IPs, but I still have to use `docker inspect` to get the IP of the `frr` container (BGP router container)

tasks.py
russellb: I think it needs CAP_NET_ADMIN.  It doesn't work without `--privileged`.  We could add only the capabilities needed, but I wasn't feeling too concerned since this is a dev env.

tasks.py
russellb: I'm dropping `etcfrr` in the next revision

tasks.py
russellb: added

tasks.py
russellb: All of these `run()` commands will raise an exception if they fail

tasks.py
russellb: I've addressed this using `kubectl` for Node IPs, and an updated `docker inspect` for the `frr` container.  It still works for me with kind 0.8.1, but hopefully someone else can test with older kind.

russellb: ... if we actually care about older kind?  We could also just add a version check on kind so we don't have to deal with supporting differences between versions.  Let's see if it works first, though.

tasks.py
rata: Ohh, permissions! Okay, that seems like a good reason. Maybe we can add a comment for that so it is obvious for other readers?

tasks.py
rata: Oh, right, the net stuff. Makes sense, thanks!

tasks.py
rata: Ups, sorry for the noise, then! My invoke foo is not strong :)

tasks.py
russellb: I dropped it already in favor of a temporary dir with the rendered config files.  that works ok

tasks.py
rata: No, sorry, of course it's totally okay to upgrade kind. I just wanted to avoid "it only works on Rusell's setup" thing :-)

Will update kind and try again :)

tasks.py
rata: Thanks!

tasks.py
```diff
@@ -231,6 +233,91 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None):
...
+        # This will fail if there's no cluster.
+        pass
+
+    run('([ "$(docker ps -a | grep frr)" ] && docker rm -f frr) || /bin/true')
+
+    dev_env_dir = os.getcwd() + "/dev-env/bgp"
+    frr_volume_dir = dev_env_dir + "/frr-volume"
+
+    # sudo because past docker runs will have changed ownership of this dir
+    sudo('rm -rf "%s"' % frr_volume_dir)
```
rata, line 318: @russellb This doesn't work here. It needs to ask for a password and immediately fails. From http://docs.pyinvoke.org/en/stable/api/context.html#invoke.context.Context.sudo I see that:

> If you intend to respond to sudo’s password prompt by hand, just use run("sudo command") instead! The autoresponding features in this method will just get in your way.

I have configure to ask for a password here. Switching it to `run('sudo ...')` work fine here. Does it work there too?

tasks.py
```diff
@@ -231,6 +233,91 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None):
...
+    dev_env_dir = os.getcwd() + "/dev-env/bgp"
+    frr_volume_dir = dev_env_dir + "/frr-volume"
+
+    # TODO -- The IP address handling will need updates to add support for IPv6
+
+    # We need the IPs for each Node in the cluster to place them in the BGP
+    # router configuration file (bgpd.conf).  Each Node will peer with this
+    # router.
+    node_ips = run("kubectl get nodes -o jsonpath='{.items[*].status.addresses"
+            "[?(@.type==\"InternalIP\")].address}'", echo=True)
```
rata, line 256: I see the jsonpath foo is strong on you :)

rata, line 256: I'm not sure about this suggestion, the code I'm adding is horrible but without the ending new line the next command output ends up in the same line.

```suggestion
            "[?(@.type==\"InternalIP\")].address}{\"\\n\"}'", echo=True)
```

What do you think? Take it or leave it. I wanted that until I wrote it and saw how ugly it was :joy:

tasks.py
```diff
@@ -231,6 +233,91 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None):
...
+    # Node IPs.
+    with open("%s/frr/bgpd.conf.tmpl" % dev_env_dir, 'r') as f:
+        bgpd_config = "! THIS FILE IS AUTOGENERATED\n" + f.read()
+    for n in range(0, len(node_ips)):
+        bgpd_config = bgpd_config.replace("NODE%d_IP" % n, node_ips[n])
+    with open("%s/bgpd.conf" % frr_volume_dir, 'w') as f:
+        f.write(bgpd_config)
+
+    # Run a BGP router in a container for all of the speakers to peer with.
+    run('([ "$(docker ps -a | grep frr)" ] && docker rm -f frr) || /bin/true')
```
rata, line 288: The only wierd thing about this, for me, is that it prints this output and is not obvious if it is correct or what (when adding the "\n" as I mentioned in the previous command):

```
Configuring MetalLB with a BGP test environment
kubectl get nodes -o jsonpath='{.items[*].status.addresses[?(@.type=="InternalIP")].address}{"\n"}'
172.18.0.3 172.18.0.2 172.18.0.4
frr
```

Maybe an `echo=True` here makes it more clear? I'm fine either way, just pointing out what got my attention when running it :)

rata (commented):
@russellb Thanks for the PR, it really helps to improve MetalLB test story! :)

I treid with kind 0.8.1 and it works now, mostly LGTM. The only important thing is about `sudo()` that doesn't work here, proposed a workaround (hopefully it works there too :-D. Who said that containers would remove the "works on my machine" thing :joy: )

dev-env/bgp/README.md
```diff
@@ -0,0 +1,95 @@
+# BGP test env
+
+This assumes a 3-node `kind` based test cluster as set up by running `inv
+dev-env -p bgp`, which sets up the development environment with some
+configuration and a BGP router for development and testing purposes.
+
+Along with the cluster, there will be a container named `frr` which is the BGP
+router.  Note that this only works in an IPv4 configuration for now.
```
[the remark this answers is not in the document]
russellb: Nope.  It's my finger and the space bar.  It's an old habit that I refuse to let go of.  :-D

tasks.py
russellb: Ah, good catch!  yes that works for me too

tasks.py
rata, line 288: We probably want something like: `docker ps -a -f name=frr` instead of the grep. If someone has been playing with frr (like me :D) the grep might return something but the docker rm later will fail

tasks.py
russellb: sure, happy to make the output a little prettier.  This command is already pretty ugly!

daxmc99 (approved):
Looks good! LGTM after @rata's sudo change request

tasks.py
russellb: added `echo=True`

russellb: Good point, I will adjust

russellb: done

dev-env/bgp/README.md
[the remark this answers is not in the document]
rata: haha, okay. Can I ask why? It seems to be rendered fine in github, so not a big deal :)

rata (approved):
LGTM.

@russellb I see you didn't request a new review in github, not sure if you had some other change in the pipeline. Let me know if something is missing or we can just merge this :)

dev-env/bgp/README.md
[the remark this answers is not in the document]
russellb: I guess I'm just old?  A long time ago I was taught to do that and I haven't let it go.  If you want to go down an internet rabbit hole, search for "2 spaces after a period" and see how strongly people feel about this one way or the other!  There's even been research on the topic.  It's quite divisive.  :-)  Example: https://www.theatlantic.com/science/archive/2018/05/two-spaces-after-a-period/559304/

It's a code comment, and I don't think it's rendered any differently as you note ... if you care enough to block merging on it, then I'll change it.  ;-)  In that case we should start a guidelines doc that calls this out among any other project-specific style guidelines that come up.

russellb:
> LGTM.
> 
> @russellb I see you didn't request a new review in github, not sure if you had some other change in the pipeline. Let me know if something is missing or we can just merge this :)

I don't have any more changes coming.  I don't know how to request another review, I guess?

dev-env/bgp/README.md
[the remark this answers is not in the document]
johananl: I've never seen two spaces after a period used as a standard in any project I collaborated on :confused: I'll have to believe you, however it does seem way less common than a single space and a single space is also what the MetalLB codebase already uses. I suggest you align with that @russellb :-)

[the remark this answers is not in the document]
russellb: I'm not suggesting a standard at all - rather the opposite, that it doesn't matter.

You both have commented enough on this topic across PRs that it's clear enough that you care, so I'll update this and try for the future (but the habit may still sneak in sometimes).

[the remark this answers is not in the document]
russellb: updated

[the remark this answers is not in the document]
rata: I'm fine either way, as I said: not a big deal to me :-)

Will read that over the weekend, seems like a rabbit hole I was totally unaware of :)

tasks.py
```diff
@@ -285,8 +285,9 @@ def bgp_dev_env():
         f.write(bgpd_config)
 
     # Run a BGP router in a container for all of the speakers to peer with.
-    run('([ "$(docker ps -a | grep frr)" ] && docker rm -f frr) || /bin/true',
-            echo=True)
+    run('for frr in $(docker ps -a -f name=frr --format {{.Names}}) ; do '
```
rata, line 288: If you want a follow-up PR, I think something simpler will do it, maybe like: `docker rm $(docker ps -aq -name=frr)`. Even if there are more than one, docker rm takes more than one container id as argument (and ps -aq should print exactly what docke rm needs)

Issue #554, opened 2020-03-22
Update website to work with new hugo versions (>= 0.67.1)


rata:
When trying to run `hugo serve` with hugo `0.67.1` (also tested `0.68.1` released 12hs ago) it fails with:

```
metallb/website/content/_header.md:1:1": plain HTML documents not supported
```

This error message seems to be introduced here (from a quick google search): https://github.com/gohugoio/hugo/pull/7034

Using older hugo versions build the site but, after a binary search, only up to hugo version 0.62.2 showes the header correctly.

PS: I've checked and it seems no security release was made for hugo, so I guess it is not risky to point it out here (also, site is static, as long as we review PRs to the website, it seems okay :))

russellb:
I fixed this in #822

PR #688, opened 2020-08-05
Update repository URLs

### LINKED ISSUE #567: Replace danderson/metallb with metallb/metallb

johananl:
Supersedes #570. Fixes #567.

Opening a new PR as we needed to rebase and I couldn't push to the original branch, likely because the fork no longer exists.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -396,9 +396,9 @@ def release(ctx, version, skip_release_notes=False):
         newpat = pattern.format("v{}").format(version)
         run("perl -pi -e 's#{}#{}#g' website/content/*.md website/content/*/*.md".format(oldpat, newpat),
             echo=True)
-    _replace("/google/metallb/{}")
-    _replace("/google/metallb/tree/{}")
-    _replace("/google/metallb/blob/{}")
+    _replace("/metallb/metallb/{}")
+    _replace("/metallb/metallb/tree/{}")
+    _replace("/metallb/metallb/blob/{}")
 
     # Update the version listed on the website sidebar
     run("perl -pi -e 's/MetalLB .*/MetalLB v{}/g' website/content/_header.md".format(version), echo=True)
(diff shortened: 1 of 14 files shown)


johananl:
I've tested more or less all the links manually and they seem to work. The only thing I couldn't test is the `_replace()` function in tasks.py:

https://github.com/metallb/metallb/pull/688/files#diff-dc31105b816206de63c47bf42ec2a4d7R302-R304

If we can test this without crafting a release, let's do it and the merge (can we @rata?). If we can't, let's merge as is and fix any problems which may arise as we encounter them.

russellb:
lgtm

johananl:
@rata do you still have any remaining concerns? If not, I'll go ahead and merge.

tasks.py
```diff
@@ -299,9 +299,9 @@ def _replace(pattern):
         newpat = pattern.format("v{}").format(version)
         run("perl -pi -e 's#{}#{}#g' website/content/*.md website/content/*/*.md".format(oldpat, newpat),
             echo=True)
-    _replace("/google/metallb/{}")
-    _replace("/google/metallb/tree/{}")
-    _replace("/google/metallb/blob/{}")
+    _replace("/metallb/metallb/{}")
+    _replace("/metallb/metallb/tree/{}")
+    _replace("/metallb/metallb/blob/{}")
```
rata, line 304: Have you created a release locally to test this? I mean, test that this replace works as expected and there is no missing URL that isn't replaced?

johananl:
> Have you tested locally with `hugo-serve` just in case? Be aware of: #554, though

Yes, I always do on doc changes.

tasks.py
johananl: I wasn't sure it was safe to run the release logic. I'll give it a try.

johananl: This turned out to be tricky to check/fix. I keep getting unexpected results. Will update once I manage to figure it out.

johananl: Looks like this works fine.

johananl:
@rata I'd like to add this to `v0.9.5` since without this change we aren't updating the raw manifests anyway.

rata (approved):
Yes, LGTM.

PR #694, opened 2020-08-13
Test to see if CircleCI runs changed config on a PR

russellb:


russellb:
Interesting that it just didn't run the jobs that got changed

champtar:
You need to rebase to have the right config maybe ? Also if you have circleci on your fork I've seen weird behavior

PR #695, opened 2020-08-13
Revert "Publish images for each open pull request"

russellb:
This reverts commit 2b900accf30354ed8935fdaf3e1bf41396b4e665.

After merging this and testing it, we realized that we need to enable
an option that turns on passing environment variables to jobs run
against PRs from forked repos.  There is more discussion about using
CircleCI for an open source project here:

https://circleci.com/docs/2.0/oss/

An important quote from the docs is this one:

    "If you are comfortable sharing secrets with anyone who forks your
    project and opens a PR, you can enable the Pass secrets to builds
    from forked pull requests option. In the Project Settings>Advanced
    of your project, set the Pass secrets to builds from forked pull
    requests option to On."

Since this is the security stance of CircleCI, we must not turn this
on.  We should revert the change and instead look at an alternative
for publishing images for PRs.

Reopens #623


rata:
@russellb Thanks a lot!

PR #697, opened 2020-08-14
Added `controller-name` service annotation

### LINKED ISSUE #685: Ignore a service of type=LoadBalancer
Is there a way to have metallb ignore a `Service` of `type=LoadBalancer`? I couldn't find it in the docs or a brief search in the code.

From a UX perspective, I could see three ways to do this:

1. Annotation on the service, e.g. `metallb.universe.tf/ignore=true` or similar
1. configmap, where I can explicitly list services to ignore
1. My favourite, configmap, where I can provide annotations that, if set, would cause metallb to ignore the service.

The last is my favourite, as it is is very similar to the standard taints/tolerations. Short of that, I would do service annotation, followed by configmap.

But if it already exists, I will take whatever it has.

juanrgm:
With this annotation it could supports multiple metallb controllers in the same cluster or simply to ignore services handled by other load balancers.

#685

### DIFF
--- a/controller/service.go
+++ b/controller/service.go
@@ -37,6 +37,15 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 		return true
 	}
 
+	// Service not handled by this controller
+	if c.config.ControllerName != "" {
+		controllerName := svc.Annotations["metallb.universe.tf/controller-name"]
+		if controllerName != c.config.ControllerName {
+			c.clearServiceState(key, svc)
+			return true
+		}
+	}
+
 	// If the ClusterIP is malformed or not set we can't determine the
 	// ipFamily to use.
 	clusterIP := net.ParseIP(svc.Spec.ClusterIP)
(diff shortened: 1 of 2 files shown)


rata:
As mentioned here https://github.com/metallb/metallb/issues/685#issuecomment-674095803, please open a new issue and explain the use cases for several MetalLB deployments on a single cluster and explain with more detail how your proposal address them.

russellb:
> As mentioned here [#685 (comment)](https://github.com/metallb/metallb/issues/685#issuecomment-674095803), please open a new issue and explain the use cases for several MetalLB deployments on a single cluster and explain with more detail how your proposal address them.
> 
> Thanks!

What if we just ignore the "multiple metallb controllers" part?  The "only pay attention to a subset of Services" seems like a reasonable enough use case?

Maybe it would help to link to some other LoadBalancer implementations and how they would do the same thing (only act on a subset of Services).

rata:
@russellb oh, sure. I didn't understand this solves the "ignore other services" problem completely (not sure why now, though :see_no_evil:).

I'd read more about the issue to see which kind of semantics we want: "catch all except the ones that say to ignore them on metallb" or "ignore all except the one that ask for MetalLB". I'd gather more context before looking at the PR itself :)

russellb:
I'm going to close this out for now. I would suggest a design doc as a next step if you'd like to pursue this further.  Thanks!

PR #706, opened 2020-08-26
Disallow duplicate peers in config

johananl:
Configuring the same peer twice is an error. Configuring two identical peers results in a rapidly-flapping BGP session. This PR disallows identical peers in the configuration. In the future we may want to perform smarter checks, however at the very least this fix will catch copy-paste mistakes.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"reflect"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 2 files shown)


russellb:
Thanks for including the test case!  This looks like a reasonable interim fix, at least.  As you suggest, we probably want to consider some smarter validation, but what you have is an improvement over what's there now, so +1 from me.

lgtm to merge

johananl:
Thanks Russell. I'll wait for a quick approval by @rata or @daxmc99.

PR #707, opened 2020-09-01
inv: Add task to run unit tests

russellb:
Since inv is used to run other dev workflows, add another task to run
the unit tests.  Also update CircleCI config to use inv to run the
tests so that we keep the local dev commands in sync with what runs in
CI.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -444,3 +444,10 @@ def release(ctx, version, skip_release_notes=False):
     run("git commit -a -m 'Automated update for release v{}'".format(version), echo=True)
     run("git tag v{} -m 'See the release notes for details:\n\nhttps://metallb.universe.tf/release-notes/#version-{}-{}-{}'".format(version, version.major, version.minor, version.patch), echo=True)
     run("git checkout main", echo=True)
+
+
+@task
+def test(ctx):
+    """Run unit tests."""
+    run("go test -short ./...")
+    run("go test -short -race ./...")
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -7,8 +7,10 @@ jobs:
       - image: cimg/go:1.13
     steps:
       - checkout
-      - run: go test -short ./...
-      - run: go test -short -race ./...
+      - run: sudo apt-get update
+      - run: sudo apt-get install python3-pip
+      - run: sudo pip3 install invoke semver pyyaml
+      - run: inv test
       - run: cp manifests/metallb.yaml manifests/metallb.yaml.prev
   lint-1.13:
     docker:


tasks.py
```diff
@@ -324,3 +324,10 @@ def _replace(pattern):
     run("git commit -a -m 'Automated update for release v{}'".format(version), echo=True)
     run("git tag v{} -m 'See the release notes for details:\n\nhttps://metallb.universe.tf/release-notes/#version-{}-{}-{}'".format(version, version.major, version.minor, version.patch), echo=True)
     run("git checkout main", echo=True)
+
+
+@task
+def unit(ctx):
+    """Run unit tests"""
```
johananl, line 331:
```suggestion
    """Run unit tests."""
```

tasks.py
```diff
@@ -324,3 +324,10 @@ def _replace(pattern):
     run("git commit -a -m 'Automated update for release v{}'".format(version), echo=True)
     run("git tag v{} -m 'See the release notes for details:\n\nhttps://metallb.universe.tf/release-notes/#version-{}-{}-{}'".format(version, version.major, version.minor, version.patch), echo=True)
     run("git checkout main", echo=True)
+
```
johananl, line 327: We have only one empty line between functions in other places.

tasks.py
```diff
@@ -324,3 +324,10 @@ def _replace(pattern):
...
+
+@task
+def unit(ctx):
```
johananl, line 330: Why not `test`? Seems more intuitive to me (and very common).

tasks.py
```diff
@@ -324,3 +324,10 @@ def _replace(pattern):
...
+    """Run unit tests"""
+    run("go test -short ./...")
+    run("go test -short -race ./...")
```
johananl, line 333: Couldn't we just run once with `-race`? Is there an advantage to the two separate runs? One downside is longer execution time (given that this task will likely be run very often by devs).

johananl (commented):
Thanks! Left a few comments.

rata (approved):
LGTM, thanks!

tasks.py
rata: I think is fine having it be the same as it is today, `go test` will print any test failures and exit. This should be quite fast (2.7s) and if that works, a more heavy run is done for the `-race`.

russellb: I honestly did this just because I was keeping the same behavior that existed in CI today.  I didn't have a good explanation for why it was run twice, but your explanation makes sense.  :-)

tasks.py
russellb: All of the other places are wrong, lol.

https://www.python.org/dev/peps/pep-0008/#blank-lines

... I could do a follow-up to fix up formatting and enforce it with a CI job.

tasks.py
russellb: done

tasks.py
russellb: I have no strong opinion.  I guess I was envisioning a future `e2e` target or something similar for other types of tests.  This is always easy to change, though.  I've changed it to `test`.

russellb:
@rata you approved this months ago, but I had to rebase it.  I also had to re-introduce the commands to install `inv`.  It added about 26 seconds to this CI job run, which doesn't seem so bad to me.  I figure I should let someone ack it one more time before merging, though.

cc @johananl since you had removed the commands I'm re-introducing in commit 66698c23b352c2fac03eac0e87e10214b0527564

johananl (approved):
LGTM

PR #709, opened 2020-09-02
layer2: Send GARPs on cluster membership changes

### LINKED ISSUE #584: L2 mode doesn't resend gratuitous arp when node is diconnected and reconnected

russellb:
Speaker ownership of IPs is recalculated each time membership of the
cluster changes.  When a new speaker is chosen, that speaker will send
a set of gratuitous ARP or NDP messages to inform the network of the
new location for that IP address.

A problem existed where the cluster could end up in an inconsistent
state if a speaker lost ownership of an IP but didn't realize it.
In the meantime, a new speaker will have taken over the IP.  When the
original owner rejoins the cluster, it will fail to reclaim the IP
with gratuitous ARP / NDP if it thinks it never lost it in the first
place.

One way to look at this problem is that it's a failure to enforce
quorum before assuming ownership of an IP address.  If this was
guaranteed, we know that a Node on the losing side of a cluster
partition would not continue to think it owns any IPs.  It seems that
memberlist, the cluster membership library in use by metallb, does not
provide quorum enforcement, so we must address this problem a
different way.

When the speaker is run with memberlist enabled, we already force a
sync of all Services when a cluster membership occurs: either with a
Node joins or leaves a cluster.  This patch builds on that by keeping
track of the last known state of cluster membership each time we sync
a Service.  When a change is detected, gratuitous ARP or NDP messages
are sent.  This will inform the network of the current correct
ownership in case the change is the result of recovering from a
network partition.

The cluster membership tracking is done on a per-IP basis.  We can not
do this globally, because the first time we see that cluster
membership has changed, we likely have not finished syncing all
services to figure out who is the new owner of each IP.  By tracking
changes on a per-IP basis, we ensure we send the messages only after
confirming we are still the owner of that IP after the cluster
membership change.

One downside of this change is that it creates points in time where we
send a larger amount of ARP or NDP traffic without any attempted rate
limiting in place.  It is our guess that this will not be high enough
in volume to be problematic.  We may have to revisit this in the
future if sending gratuitous ARP or NDP messages for all load balancer
IPs in the cluster breaks something.

Closes #584

### DIFF
--- a/internal/layer2/announcer.go
+++ b/internal/layer2/announcer.go
@@ -131,6 +131,15 @@ func (a *Announce) updateInterfaces() {
 	return
 }
 
+// Force a proactive announcement of all IP addresses.
+//
+// Proactively send another round of gratuitious ARP / NDP messages. This may be
+// necessary after recovering from a split in cluster membership to ensure that
+// the current intended owner of the IP address is respected by the network.
+func (a *Announce) ReAnnounce(name string) {
+	go a.spam(name)
+}
+
 func (a *Announce) spam(name string) {
 	// TODO: should abort if we lose control of the IP mid-spam.
 	start := time.Now()
@@ -209,13 +218,15 @@ func (a *Announce) SetBalancer(name string, ip net.IP) {
 }
 
 // DeleteBalancer deletes an address from the set of addresses we should announce.
-func (a *Announce) DeleteBalancer(name string) {
+//
+// Returns the IP if it is no longer in use.
+func (a *Announce) DeleteBalancer(name string) net.IP {
 	a.Lock()
 	defer a.Unlock()
 
 	ip, ok := a.ips[name]
 	if !ok {
-		return
+		return nil
 	}
 	delete(a.ips, name)
 
@@ -223,7 +234,7 @@ func (a *Announce) DeleteBalancer(name string) {
 	if a.ipRefcnt[ip.String()] > 0 {
 		// Another service is still using this IP, don't touch any
 		// more things.
-		return
+		return nil
 	}
 
 	for _, client := range a.ndps {
@@ -232,6 +243,7 @@ func (a *Announce) DeleteBalancer(name string) {
 		}
 	}
 
+	return ip
 }
 
 // AnnounceName returns true when we have an announcement under name.
--- a/speaker/layer2_controller.go
+++ b/speaker/layer2_controller.go
@@ -18,6 +18,7 @@ import (
 	"bytes"
 	"crypto/sha256"
 	"net"
+	"reflect"
 	"sort"
 
 	"github.com/go-kit/kit/log"
@@ -31,6 +32,13 @@ type layer2Controller struct {
 	announcer *layer2.Announce
 	myNode    string
 	mList     *memberlist.Memberlist
+	// Last known value of the active nodes known via memberlist, per service IP.
+	//
+	// This is used to detect a cluster membership change since the last time we
+	// looked at this service. The outer key is the service IP since we need to
+	// detect changes on a per IP basis. The value is a map of the current
+	// active nodes in the same format as the return of getActiveNodes().
+	activeNodes map[string]map[string]bool
 }
 
 func (c *layer2Controller) SetConfig(log.Logger, *config.Config) error {
@@ -39,15 +47,7 @@ func (c *layer2Controller) SetConfig(log.Logger, *config.Config) error {
 
 // usableNodes returns all nodes that have at least one fully ready
 // endpoint on them.
-func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
-	var activeNodes map[string]bool
-	if mList != nil {
-		activeNodes = map[string]bool{}
-		for _, n := range mList.Members() {
-			activeNodes[n.Name] = true
-		}
-	}
-
+func usableNodes(activeNodes map[string]bool, eps *v1.Endpoints) []string {
 	usable := map[string]bool{}
 	for _, subset := range eps.Subsets {
 		for _, ep := range subset.Addresses {
@@ -75,8 +75,21 @@ func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
 	return ret
 }
 
+func (c *layer2Controller) getActiveNodes() map[string]bool {
+	var activeNodes map[string]bool
+	if c.mList != nil {
+		activeNodes = map[string]bool{}
+		for _, n := range c.mList.Members() {
+			activeNodes[n.Name] = true
+		}
+	}
+
+	return activeNodes
+}
+
 func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, svc *v1.Service, eps *v1.Endpoints) string {
-	nodes := usableNodes(eps, c.mList)
+	activeNodes := c.getActiveNodes()
+	nodes := usableNodes(activeNodes, eps)
 	// Sort the slice by the hash of node + service name. This
 	// produces an ordering of ready nodes that is unique to this
 	// service.
@@ -97,15 +110,40 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, svc *v1.Ser
 }
 
 func (c *layer2Controller) SetBalancer(l log.Logger, name string, lbIP net.IP, pool *config.Pool) error {
+	// Determine if the list of active nodes as determined by memberlist has
+	// changed since the last time we checked this service's IP. If so, we kick off
+	// sending gratuitous ARP / NDP for this service IP. This will handle the
+	// case where the cluster membership change is recovering from a network
+	// partition and we need to establish who the real current owner is of each
+	// IP address. This happens asynchronously.
+	var newService bool
+	activeNodes := c.getActiveNodes()
+	if _, ok := c.activeNodes[lbIP.String()]; !ok {
+		// We hadn't seen this service yet, so we shouldn't re-announce
+		newService = true
+		c.activeNodes[lbIP.String()] = activeNodes
+		// We've seen this service before and its active nodes have changed.
+	}
+	if !newService && !reflect.DeepEqual(activeNodes, c.activeNodes[lbIP.String()]) {
+		// We've seen this service before and its active nodes have changed.
+		l.Log("op", "SetBalancer", "msg", "Detected cluster membership change, re-announcing IP",
+			"newActiveNodes", activeNodes, "oldActiveNodes", c.activeNodes[lbIP.String()])
+		c.announcer.ReAnnounce(name)
+		c.activeNodes[lbIP.String()] = activeNodes
+	}
+
 	c.announcer.SetBalancer(name, lbIP)
 	return nil
 }
 
 func (c *layer2Controller) DeleteBalancer(l log.Logger, name, reason string) error {
 	if !c.announcer.AnnounceName(name) {
+		// No longer tracked by the announcer, so nothing else to do.
 		return nil
 	}
-	c.announcer.DeleteBalancer(name)
+	if unusedIP := c.announcer.DeleteBalancer(name); unusedIP != nil {
+		delete(c.activeNodes, unusedIP.String())
+	}
 	return nil
 }
 
--- a/speaker/layer2_controller_test.go
+++ b/speaker/layer2_controller_test.go
@@ -112,7 +112,7 @@ func TestUsableNodes(t *testing.T) {
 	}
 
 	for _, test := range tests {
-		response := usableNodes(test.eps, nil)
+		response := usableNodes(nil, test.eps)
 		sort.Strings(response)
 		if !compareUseableNodesReturnedValue(response, test.cExpectedResult) {
 			t.Errorf("%q: shouldAnnounce for controller returned incorrect result, expected '%s', but received '%s'", test.desc, test.cExpectedResult, response)
--- a/speaker/main.go
+++ b/speaker/main.go
@@ -244,9 +244,10 @@ func newController(cfg controllerConfig) (*controller, error) {
 			return nil, fmt.Errorf("making layer2 announcer: %s", err)
 		}
 		protocols[config.Layer2] = &layer2Controller{
-			announcer: a,
-			myNode:    cfg.MyNode,
-			mList:     cfg.MList,
+			announcer:   a,
+			myNode:      cfg.MyNode,
+			mList:       cfg.MList,
+			activeNodes: make(map[string]map[string]bool),
 		}
 	}
 


russellb:
So far I've been testing this manually using the kind based dev-env.  I create a bunch of Services and then just `docker pause` one of the kind Node's containers, which will cause the other speakers to think that one disappeared.  Unpausing that Node's container will let it rejoin.  I can see the correct GARPs going out at each Node join/leave event.

(I can provide more specific configs / manifests, if someone else wants to replicate)

speaker/layer2_controller.go
```diff
@@ -33,21 +34,25 @@ type layer2Controller struct {
 	mList     *memberlist.Memberlist
 }
 
+// Last known value of the active nodes known via memberlist, per Service IP.
+//
+// This is used to detect a cluster membership change since the last time we
+// looked at this Service. The outer key is the service IP since we need to
+// detect changes on a per IP basis.  The value is a map of the current
+// active Nodes in the same format as the return of getActiveNodes().
+var lastActiveNodes map[string]map[string]bool
```
liuyuan10, line 43: why keeping activenode per service? active node is cluster wide from memberlist, not per service. is it simpler to respam all services when active node changes?

russellb: > The cluster membership tracking is done on a per-IP basis. We can not
do this globally, because the first time we see that cluster
membership has changed, we likely have not finished syncing all
services to figure out who is the new owner of each IP. By tracking
changes on a per-IP basis, we ensure we send the messages only after
confirming we are still the owner of that IP after the cluster
membership change.

speaker/layer2_controller.go
```diff
@@ -97,15 +115,42 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, svc *v1.Ser
...
+	var newService bool
+	activeNodes = c.getActiveNodes()
+	if _, ok := lastActiveNodes[lbIP.String()]; !ok {
+		// If it wasn't in the lastActiveNodes map, we hadn't seen this Service yet,
+		// so we shouldn't trigger try triggering re-announcement.
+		newService = true
+		lastActiveNodes[lbIP.String()] = activeNodes
+	}
+	if !newService && !reflect.DeepEqual(activeNodes, lastActiveNodes[lbIP.String()]) {
+		l.Log("op", "SetBalancer", "msg", "Detected cluster membership change, re-announcing IP",
```
johananl, line 134:
```suggestion
		// We've seen this service before and its active nodes have changed.
		l.Log("op", "SetBalancer", "msg", "Detected cluster membership change, re-announcing IP",
```

speaker/layer2_controller.go
```diff
@@ -97,15 +115,42 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, svc *v1.Ser
...
 
 func (c *layer2Controller) DeleteBalancer(l log.Logger, name, reason string) error {
 	if !c.announcer.AnnounceName(name) {
+		// No longer tracked by the announcer, so nothing else to do.
 		return nil
 	}
-	c.announcer.DeleteBalancer(name)
+	var unusedIP net.IP
+	unusedIP = c.announcer.DeleteBalancer(name)
+	if unusedIP != nil {
```
johananl, line 151:
```suggestion
	if unusedIP := c.announcer.DeleteBalancer(name); unusedIP != nil {
```

speaker/layer2_controller.go
```diff
@@ -18,13 +18,14 @@ import (
...
 	"net"
+	"reflect"
 	"sort"
 
 	"github.com/go-kit/kit/log"
 	"github.com/hashicorp/memberlist"
 	"go.universe.tf/metallb/internal/config"
 	"go.universe.tf/metallb/internal/layer2"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
johananl, line 28: Your editor is doing this. I had this with VS Code and was able to disable this behavior, however I don't remember what I did :-/

internal/layer2/announcer.go
```diff
@@ -209,21 +218,23 @@ func (a *Announce) SetBalancer(name string, ip net.IP) {
 }
 
 // DeleteBalancer deletes an address from the set of addresses we should announce.
-func (a *Announce) DeleteBalancer(name string) {
+//
+// Return the IP if it is no longer in use.
```
johananl, line 222:
```suggestion
// Returns the IP if it is no longer in use.
```

I think the normal tense for doc comments is simple present rather than imperative.

speaker/layer2_controller.go
```diff
@@ -97,15 +115,42 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, svc *v1.Ser
...
 func (c *layer2Controller) DeleteBalancer(l log.Logger, name, reason string) error {
 	if !c.announcer.AnnounceName(name) {
+		// No longer tracked by the announcer, so nothing else to do.
 		return nil
 	}
-	c.announcer.DeleteBalancer(name)
+	var unusedIP net.IP
+	unusedIP = c.announcer.DeleteBalancer(name)
+	if unusedIP != nil {
+		delete(lastActiveNodes, unusedIP.String())
```
johananl, line 152: I'm a bit confused by this. In which cases exactly does `announcer.DeleteBalancer()` return an IP, and which IP is it?

johananl (commented):
Thanks a lot for the effort @russellb. I've left some comments.

internal/layer2/announcer.go
```diff
@@ -131,6 +131,15 @@ func (a *Announce) updateInterfaces() {
 	return
 }
 
+// Force a proactive announcement of all IP addresses
+//
+// Proactively send another round of gratuitious ARP / NDP messages.  This may be
```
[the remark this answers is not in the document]
russellb: Yes.  Btw, #710 will make it so you don't have to nag me manually on this.  :-)

I also used it to help me find them:

```
$ inv checkpatch
+// Proactively send another round of gratuitious ARP / NDP messages.  This may be
+// detect changes on a per IP basis.  The value is a map of the current
+       // changed since the last time we checked this Service's IP.  If so, we kick off
+       // sending gratuitous ARP / NDP for this Service IP.  This will handle the
+       // IP address.  This happens asynchronously.
ERROR: Found changed lines with 2 spaces after a period.

```

speaker/layer2_controller.go
```diff
@@ -33,21 +34,25 @@ type layer2Controller struct {
...
-func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
-	var activeNodes map[string]bool
-	if mList != nil {
-		activeNodes = map[string]bool{}
-		for _, n := range mList.Members() {
-			activeNodes[n.Name] = true
-		}
-	}
-
+func usableNodes(activeNodes map[string]bool, eps *v1.Endpoints) []string {
```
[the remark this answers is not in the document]
russellb: Let's just merge the other PR first and I'll figure out how to rebase on it cleanly.

speaker/layer2_controller.go
```diff
@@ -75,8 +80,21 @@ func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
...
+		for _, n := range c.mList.Members() {
+			activeNodes[n.Name] = true
+		}
+	}
+	return activeNodes
+}
+
 func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, svc *v1.Service, eps *v1.Endpoints) string {
-	nodes := usableNodes(eps, c.mList)
+	var activeNodes map[string]bool
```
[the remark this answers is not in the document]
russellb: just a style choice.  will change it

speaker/layer2_controller.go
```diff
@@ -97,15 +115,42 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, svc *v1.Ser
 }
 
 func (c *layer2Controller) SetBalancer(l log.Logger, name string, lbIP net.IP, pool *config.Pool) error {
+	// Determine if the list of active Nodes as determined by memberlist has
```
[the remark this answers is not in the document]
russellb: That's correct.  ok ...

speaker/layer2_controller.go
```diff
@@ -75,8 +80,21 @@ func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
...
 
+func (c *layer2Controller) getActiveNodes() map[string]bool {
+	var activeNodes map[string]bool
+	if c.mList != nil {
+		activeNodes = map[string]bool{}
+		for _, n := range c.mList.Members() {
+			activeNodes[n.Name] = true
+		}
+	}
+	return activeNodes
```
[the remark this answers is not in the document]
russellb: Yes, I strongly prefer introducing linting on trivial style choices that the project intends to enforce to save energy in review for more important things.  I'll fix this one though.

speaker/layer2_controller.go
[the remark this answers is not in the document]
russellb: Yeah, the use of a global here was a pretty lazy choice on my part.  I didn't see any concurrency problems with it for now, but you're right that I should just add it to the controller.  Thanks for calling me on it.  :-)

speaker/layer2_controller.go
russellb: yeah, some golang vim plugin(s) in my case ...

russellb:
Thanks for the review, @johananl.  I think I have addressed all of your feedback.

champtar:
I see there is still ```// TODO: should abort if we lose control of the IP mid-spam.``` so without testing or looking deeper I prefer my version, as with this patch you can end up with 2 nodes spamming for the same IP

champtar:
Sorry, now I remember, the loop continue but we don't spam, ignore my comment

speaker/layer2_controller.go
```diff
@@ -28,9 +29,16 @@ import (
 )
 
 type layer2Controller struct {
-	announcer *layer2.Announce
-	myNode    string
-	mList     *memberlist.Memberlist
+	announcer   *layer2.Announce
+	myNode      string
+	mList       *memberlist.Memberlist
+  // Last known value of the active nodes known via memberlist, per service IP.
```
champtar, line 35: indentation is broken in many spots (tab vs space I guess)

speaker/layer2_controller.go
```diff
@@ -75,8 +75,21 @@ func usableNodes(eps *v1.Endpoints, mList *memberlist.Memberlist) []string {
 	return ret
 }
 
+func (c *layer2Controller) getActiveNodes() map[string]bool {
+	var activeNodes map[string]bool
+	if c.mList != nil {
+		activeNodes = map[string]bool{}
+		for _, n := range c.mList.Members() {
```
champtar, line 82: This is an eventually consistent system and you are trying to keep a state, so you will have race conditions everywhere.
Let's say we have node1 and node2, node1 has the IP. A short networks partition occurs, so node2 start announcing the IP, the resync is also running on node1 but slower, and at this time node2 join again, node2 sent some GARP, but node1 when calling `c.mList.Members()` see node1 & node2 so node1 never send GARP.
See some previous discussions
https://github.com/metallb/metallb/pull/595/commits/d9922aeab35ef568ffdb264facc4479d74cf2e7c

speaker/layer2_controller.go
russellb: argh!!  I just did some major changes to my vim setup which messed this up.  :-(

I see the lint job flagged it too.  I'll clean this up ...

speaker/layer2_controller.go
russellb: Thanks for the feedback and especially for providing an example of the race condition.

I'm just trying to recap the scenario you provided here for my own understanding ...

1. starting state: node1 and node2.  node1 has the IP.

2. network partition.

3. node2 sees the membership change, now decides it owns the IP.  Send GARP.

4. node1 also saw the membership change, thinks it still owns the IP.

5. partition ends, nodes rejoin.

6. node1 had not yet finished processing its resync, and by the time it is resyncing the service, based on its last stored state, there's no change, so no new GARP sent.

7. End result: inconsistent network state, last GARP was from node2

Does that all sound right?

I have an idea to play with but its based on the assumption that every time a memberlist change occurs, every node sees the change (though with a different view of it).  Do you think that's a fair assumption?  Or do you think we have to be prepared for something like: node2 saw itself leave and re-join, but node1 never saw anything?

speaker/layer2_controller.go
russellb: fixed ...

speaker/layer2_controller.go
champtar: The recap sound right.
I think it's possible for a node to see a leave/join and not another one, the protocol is full of timers and other node checking so even if it's really hard to hit I don't think that memberlist gives any guarantee at all.

My approach was to always spam for 5sec from the last resync, for anything more clever I really think we need to put the state in the API (which would allow many other things like sticky announcer, allow to move the announcer, ...)

russellb: OK.  My idea was just a tweak to this patch.  I thought I could do something to be even more sure that once a node join/leave event is received, make sure we always do the updated GARPs, even if the memberlist view doesn't look different by the time the resync happens.  I think that would address the race condition you pointed out, but it only works if we can assume that every speaker sees node join/leave events.

I'm tempted to just throw my hands up and say this is not completely solvable without anything more drastic: state in the API could be one thing, or I was thinking moving away from memberlist completely since it honestly doesn't seem like the best fit for this use case.

I think we may need to settle on "good enough" for whatever we merge as a next step on this.

champtar: I have a full rework of memberlist since May, including a fix for this bug that is the good enough approach (don't try to keep a state that might not stay in sync, just spam as few as possible but as much as needed)
I'm waiting for #662 to be merged to restart the discussion / rebase, but I fear there is not enough maintainer bandwidth :(

russellb: I'm sorry I haven't followed all of the history here, nor have I reviewed all of the past PRs on the memberlist work.  I definitely did not intend to redo any work unnecessarily, so I apologize if I've overlapped.  This was just raised as a todo item for v0.9.4, so I tried to put up a PR for an approach as described by @rata.

This PR conflicts with #662 as well, so let's sort it out after that one merges.  I'm definitely happy to drop this one in favor of your work if we decide that's the right thing to do.  I'll have to go back and read your PR properly, too ...

champtar: No problem, I'm not blaming anyone, it's not merged also because I failed to explain all my choices, and my changes are not minimal at all. I'll try to attend the dev meeting after #662 is merged if that can help.

rata: thanks @champtar for the race scenario. Adn you too @russellb !

This is not what I had in mind, though, as I thought on doing it at `internal/layer2/announcer.go` layer. There the announce is done if some criteria is met (like IP not already announced from this node, if not not point in doing it again) and thought this criteria can be added there. I think, however, if the `usableSpeakers()` used there and in `speaker/layer2_controller.go` (in the`ShouldAnnounce()` function), it will have issues. Not sure if using a snapshot will remove all the issues, though. But maybe it is good enough, will have to think about it a little more. Let me know what you think :)

I think that scenario is wrong, however. Let me know if I'm missing something.

At step 5, if the partition split is solved and a node rejoins, _then_ a new event is sent to the resync channel. There are two cases:

(a) (we are in this case) The resync channel has room to receive one message without blocking, because we have already processed that and we are in the middle of the force resync. We can't be in the middle of the resync without having it removed the element from the channel :-)

This will queue a new resync to be done later (which will fix the inconsistency). So, a missing step 8 which will fix it is missing, IIUC.

(b) (this is not the case the scenario is showing) The resync channel doesn't have room to queue one message, because there is a queued message and is full. In that case no new event will be queued, but the resync will see the nodes rejoin and the desired state will be reached.

Am I missing something?

I'm fine using a different approach, though.

russellb: > thanks @champtar for the race scenario. And you too @russellb !
> 
> This is not what I had in mind, though, as I thought on doing it at `internal/layer2/announcer.go` layer. There the announce is done if some criteria is met (like IP not already announced from this node, etc.) and thought this criteria can be added there. I think, however, if the `usableSpeakers()` used there and in `speaker/layer2_controller.go` (in the`ShouldAnnounce()` function) do not contain the same speakers, it may not always rok. Not sure if using a consistent snapshot of the speakers on both place will remove all the issues, though. But maybe it is good enough, will have to think about it a little more. Let me know what you think :)

I did put some thought into where to put each piece.  I decided on minimal changes under `internal/` because that layer doesn't know about cluster membership or other nodes at all and it seemed nicer to keep it that way.

> I think that scenario is wrong, however. Let me know if I'm missing something.
> 
> At step 5, if the partition split is solved and a node rejoins, _then_ a new event is sent to the resync channel. There are two cases:
> 
> (a) (we are in this case) The resync channel has room to receive one message without blocking, because we have already processed that and we are in the middle of the force resync. We can't be in the middle of the resync without having it removed the element from the channel :-)
> 
> This will queue a new resync to be done later (which will fix the inconsistency). So, a missing step 8 which will fix it is missing, IIUC.

I don't think so.  Even if there is another resync, it still won't send another GARP.  It'll just be another resync where it sees the same "node A and node B" membership, so it won't think anything changed.

> 
> (b) (this is not the case the scenario is showing) The resync channel doesn't have room to queue one message, because there is a queued message and is full. In that case no new event will be queued, but the resync will see the nodes rejoin and the desired state will be reached.
> 
> Am I missing something?
> 
> I'm fine using a different approach, though.

The problem with this race condition is a problem with this implementation, but not a fundamental limitation of the approach.  I think you're probably thinking along the lines of how it *should* work, but in the current revision, it's possible that a membership change happens but no GARP refresh is done.  I can fix that with a tweak to this patch.

I suggest that we discuss further after I propose an update that would at least address the one race condition identified.

russellb: See #730 for another possible take on this.  It's drastically simpler than this PR.

internal/layer2/announcer.go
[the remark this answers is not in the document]
johananl: LOL. Thank you.

speaker/layer2_controller.go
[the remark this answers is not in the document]
johananl: Makes sense. Opened https://github.com/metallb/metallb/issues/735.

speaker/layer2_controller.go
johananl: @russellb is #730 supposed to address the race condition discussed above?

champtar: No race condition but an unbounded amount of spam (we could have 20+ spam goroutine in // for the same service)

johananl: Thanks @champtar, I've seen your comments about the GARP spam. I'm asking if #730 contains the race condition described in https://github.com/metallb/metallb/pull/709#discussion_r490961070, too.

russellb: No, #730 does not have the race condition.

These two PRs offer: most correct (but also sending way more GARP traffic than necessary) which is #730 , or more controlled limiting of GARP, but technically open to the race condition described in this thread

I'm not sure we'll solve this completely without some coordinated agreement on IP ownership, so I tend to think we could settle for something that's at least a net improvement.  

but maybe @champtar 's approach has this solved better.  i haven't had time to look at this more in a bit

johananl: Thanks for the summary. I have to admit I'm leaning towards merging *something* already to drive the `v0.9.4` patch release to a closure. Then we can consider more radical changes in a minor release.

I think that as long as a PR is moving things even slightly forward towards a better state than we have in `v0.9.3`, we probably want to merge it.

I'm trying to wrap my head around the latest status of all the relevant PRs and hope we can figure out the next logical step.

champtar: @johananl once #662 is merged I can rebase my version and do PR with it (just the spam part)

johananl: @champtar please see https://github.com/metallb/metallb/issues/584#issuecomment-702899515. We're leaning towards merging this PR (#709) after verifying the likelihood/frequency of the race condition you've mentioned as well as its impact.

I wouldn't go with another new implementation attempt right now. After we're done with `v0.9.4` we can re-evaluate and possibly refactor things in `v0.10.0` or possibly a later release. Thanks again for taking part in the discussion here.

champtar: It's not a new implementation attempt, it's one that is written for months now ...

champtar:
#736 was merged instead

russellb:
closed in favor of #736

PR #710, opened 2020-09-04
ci: Check for 2 spaces after a period

russellb:
I've received comments on a couple of PRs recently about putting 2
spaces after a period.  This patch will get CI to flag this
automatically.  It also puts the check in an easy command to run
locally, "inv checkpatch", which can be extended for other patch style
checking.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -11,7 +11,7 @@ except ImportError:
     from urllib2 import urlopen
 
 from invoke import run, task
-from invoke.exceptions import Exit
+from invoke.exceptions import Exit, UnexpectedExit
 
 all_binaries = set(["controller",
                     "speaker",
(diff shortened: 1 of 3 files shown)


russellb:
Job runs and fails as expected, because of the grep command that was added that is used to check added lines

tasks.py
```diff
@@ -324,3 +324,21 @@ def _replace(pattern):
...
+        lines = run("git diff $(diff -u <(git rev-list --first-parent HEAD) "
+                                " <(git rev-list --first-parent origin/main) "
+                                " | sed -ne 's/^ //p' | head -1)..HEAD | "
+                                " grep '+.*\.\  '")
+
+        if len(lines.stdout.strip()) > 0:
+            raise Exit(message="ERROR: Found changed lines with 2 spaces "
+                               "after a period.")
+    except UnexpectedExit:
+        # Will exit non-zero if no double-space-after-period lines are found
```
johananl, line 343: Nit: Missing period :-)

johananl (approved):
LGTM (left one nit).
@russellb if you could rebase and verify this still works as intended, I think we can merge. Thanks!

tasks.py
russellb: fixed

russellb:
LOL, once I got the job running properly again, it pointed out how I added lines with 2 spaces after a period in a previous PR I already merged.

russellb:
ok, this works for me locally and CI is sorted out now, so I'm going to merge and we'll see if it behaves unexpectedly on any other PRs.

PR #723, opened 2020-09-21
go.mod: use k8s.io/client-go@v0.20.0

### LINKED ISSUE #722: go.mod: an unspecified version of the k8s api is used
In `go.mod` the k8s.io/api version is unspecified which normally means something like "what happens to be the case when go.mod was updated";
```
module go.universe.tf/metallb

go 1.12

require (
...
        k8s.io/api v0.0.0-20190620084959-7cf5895f2711
        k8s.io/apimachinery v0.0.0-20190612205821-1799e75a0719
        k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
        k8s.io/klog v0.3.1
)
```
While this usually works it is not very good for maintenance.

As an example metallb can't be built with the current k8s.io/api (v0.19.0). I bumbed immediately into this while working on new features in the controller for https://github.com/metallb/metallb/issues/698.

To select a k8s.io/api version do;
```
go get k8s.io/client-go@v0.19.0
go mod tidy
```
For available versions please see;
https://github.com/kubernetes/client-go#compatibility-matrix

uablrek:
Use a defined version of k8s.io/client-go.

Fixes #722


russellb (changes requested):
Check the CI results.  This seems to break the CI jobs.

uablrek:
```
/go/pkg/mod/go.universe.tf/virtuakube@v0.0.0-20190708182722-512c11153571/cluster.go:194:47: not enough arguments in call to c.client.CoreV1().Nodes().List have ("k8s.io/apimachinery/pkg/apis/meta/v1".ListOptions)
```
I don't know what's tested here;
```
find . -name cluster.go
```
in the metallb directory finds nothing.

But it is the same fault I try to fix, there are new parameters in the current k8s api. But metallb _must_ compile with the current k8s-api version or further development is not possible. I have built `controller` and `speaker` manually with this PR applied and it works.

russellb:
virtuakube is a dependency that's failing to build.  It's actually only a dependency of some e2etests that are not even functional right now anyway.  We still need to somehow not break the CI jobs, though.  We could try ignoring `e2etest/` in the CI jobs for the moment?

uablrek:
I find only https://github.com/danderson/virtuakube which is archived. I don't know what to do. The dependency seems to be in `e2etest/` as you say. If this is not used, can it simply be removed?

russellb:
> I find only https://github.com/danderson/virtuakube which is archived. I don't know what to do. The dependency seems to be in `e2etest/` as you say. If this is not used, can it simply be removed?

Yeah, this may be the final straw of bit-rot where we should just remove the `e2etest` directory.  Someone was working on it recently (#192) but he's not working on it anymore.  We can always pull it back out of git history later if needed.  I can submit another PR for this and then we can rebase this PR once it's cleaned up.

russellb:
I proposed dropping the old code getting in your way in #724

uablrek:
@russellb Thanks a lot. I'll wait and rebase :+1:

rata:
@uablrek thanks! This is something I wanted to do for a while. I'll merge after: 0.9.4 (don't want to update this on a patch release for no good reason) and after the other linked PR is merged and this is rebased :)

groundhog2k:
I think this PR could get really important to fix Issue #794 and make metallb compatible with Kubernetes >=1.20.0 again.

See a similar bug and PR/fix for the nfs-external-provisioner => https://github.com/kubernetes-sigs/nfs-subdir-external-provisioner/pull/26

champtar:
@uablrek can you make that v0.19.6 ?

uablrek:
@champtar Took v0.20.0.

champtar:
v0.20.1 ?

champtar:
Also can you squash your commits ? and maybe `go mod tidy`

uablrek:
Squashed. Comment updated. Tidy was done already.

invidian:
I created #802 where CI is passing.

PR #639, opened 2020-06-18
Work with python invoke-powered build process

caboteria:
main_test.go still assumed the old make.go process.  This doesn't get us to working e2e tests but it's a step in the right direction.

PR #724, opened 2020-09-21
Drop e2etest directory

### LINKED ISSUE #191: Add Romana to the e2e test matrix
I want e2e tests that verify against a matrix of the popular network addons for k8s: Flannel, Calico, Romana, and Weave (and kube-router if/when we can get compatibility).

This is currently a little hard with Romana, because the manifests require some customization when using a non-default service CIDR, and those modifications are described in long-form prose rather than something I can search/replace like with Calico and Flannel.

That said, I still want to test with Romana, not least because it has custom integration instructions that need to be exercised. Filing this bug to remind myself to make that work, once I have some of the rest of the test matrix up and working.

### LINKED ISSUE #192: Add e2e tests running on Virtuakube

### LINKED ISSUE #629: e2etest/main_test.go still uses make.go
Hi!

I'm learning how MetalLB development works so I'm trying to run the e2etests and stubbed my toe.  Using MetalLB a9a6f319 running the tests looks like:
```
$ go test ./e2etest
stat make.go: no such file or directory
Failed to build MetalLB images: exit status 1FAIL	go.universe.tf/metallb/e2etest	0.007s
FAIL
```
The problem seems to be e2etest/main_test.go.  It still depends on make.go but that was replaced by Python invoke.  Maybe I missed something and am running the test incorrectly, corrections appreciated!

I hacked main_test.go to work with inv instead of make.go so please LMK if you'd like me to submit a PR to fix it, or if that file is no longer used I can submit a PR to delete it so other newbies don't stub their toes on it.

Thanks!

russellb:
These tests no longer work.  They rely on the now archived virtuakube,
which doesn't build against kube 1.19, which ends up blocking fixing
metallb for kube 1.19.  See #723 for discussion of where it gets in the
way.

At this point we should just drop this directory and re-introduce
e2etest code once someone is able to work on it.  We can always pull
this back out of git history if it helps.

The commit contents are the result of roughly:

 - git rm -r e2etest
 - go mod tidy
 - git add go.mod go.sum

Closes #192 
Closes #629
Closes #639


go.mod
```diff
@@ -22,11 +21,9 @@ require (
 	github.com/osrg/gobgp v2.0.0+incompatible
 	github.com/prometheus/client_golang v1.0.0
 	github.com/satori/go.uuid v1.2.0 // indirect
-	github.com/spf13/cobra v0.0.5
-	github.com/spf13/pflag v1.0.3
+	github.com/spf13/viper v1.3.2 // indirect
```
rata, line 24: Just curious, what did you run exactly to update this?

rata (approved):
LGTM, thanks!

go.mod
russellb: `go mod tidy`

daxmc99 (approved):

go.mod
rata: Oh, perfect. Thanks! :)

champtar:
@russellb can you fix the conflict ?

rata:
This removes the mirror-server yaml, that is used in `inv dev-env`, so it breaks that :-(

rata:
I plan to merge this fixing the conflict, as it makes it way easier to upgrade client-go. Will update soon

PR #590, opened 2020-04-20
Add support for MP BGP encoding for IPv4 and IPv6

ykulazhenkov:
Another implementation of IPv6 support for BGP speaker.
Implementation based on rfc4760. 
MP BGP encoding will be used for IPv4 and IPv6 prefixes if the peer supports it. If peer don't announce support for MP BGP for IPv4, then legacy BGP encoding will be used.
It is possible to announce IPv4 prefixes over IPv6 and vice versa.
Additional configuration option was added to support aggregation for IPv6 prefixes.

The patch was tested with Dell OS 10 switches.
PCAP files:
https://github.com/ykulazhenkov/metallb-pcap

PR #727, opened 2020-09-22
controller; Add support for dual-stack services (k8s dual-stack phase 3)

### LINKED ISSUE #698: Enhancement; Controller support for the updated K8s dual-stack API (v1.20)
### Background

Metallb controller supports the current dual-stack API in K8s where separate services for ipv4 and ipv6 must be defined, example;
```
apiVersion: v1
kind: Service
metadata:
  name: mserver-ipv6
spec:
  ipFamily: IPv6
  selector:
    app: mserver
  ports:
  - port: 5001
  type: LoadBalancer
  loadBalancerIP: "1000::"
```
The update in K8s v1.20 will allow definition of a dual-stack service, example;
```
apiVersion: v1
kind: Service
metadata:
  name: svc-preferdual-lb
spec:
  ipFamilyPolicy: PreferDualStack
  selector:
    app: mconnect-daemonset
  ports:
  - port: 5001
  type: LoadBalancer
```
Here `PreferDualStack` means that on a dual-stack cluster the service shall have both ipv4 and ipv6 addresses.

### Controller updates

The controller already handles pools containing both address families. When a loadBalancerIP is assigned the controller checks the family of the `clusterIP` in the service and assigns a loadBalancerIP of the same family. Since the controller never examines the `ipFamily` field (which will disappear in v1.20) the controller is **still compatible** with the new API if two single-stack services are used.

What needs to be updated is that the controller must check the new `clusterIPs` (plural) field if it exists and assign loadBalancerIPs of both families if appropriate;
```
$ kubectl get svc svc-preferdual-lb -o json | jq .spec
{
  "clusterIP": "12.0.34.137",
  "clusterIPs": [
    "12.0.34.137",
    "fd00:4000::2e7a"
  ],
  "externalTrafficPolicy": "Cluster",
  "ipFamilies": [
    "IPv4",
    "IPv6"
  ],
  "ipFamilyPolicy": "PreferDualStack",
  "ports": [
    {
      "nodePort": 32278,
      "port": 5001,
      "protocol": "TCP",
      "targetPort": 5001
    }
  ],
  "selector": {
    "app": "mconnect-daemonset"
  },
  "sessionAffinity": "None",
  "type": "LoadBalancer"
}
```

### Backward compatibility

When an updated controller (for the new API) is deployed in an old cluster the `clusterIPs` will not be set (empty or nil) and the controller shall resort to examine the `clusterIP` field.

### spec.loadBalancerIP

The `spec.loadBalancerIP` field is not (yet?) updated by the new API. For a dual-stack service the user most likely want to define load-balancer IPs for both families. A `spec.loadBalancerIPs` array is proposed; https://github.com/kubernetes/enhancements/pull/1992.

### Implementation

Implementation _can_ start right away if @khenidak development branch is used as base but is a bit difficult. It is simpler to wait until it is merged.

I plan to create a PR for this a.s.a.p and my gut feeling is that it will not be too much updates in controller code. Test is another matter, it *will* take some time. Unit-tests must use the new API and must be executed with K8s `master` of >=v1.20,

### Speakers

As when dual-stack was first supported it is best to separate controller and speaker updates. The L2 speaker may (again) work out of the box if it handles all enties in the `status.loadBalancer.ingress` array.

### References

* https://github.com/kubernetes/kubernetes/pull/91824
* https://github.com/kubernetes/enhancements/blob/master/keps/sig-network/20180612-ipv4-ipv6-dual-stack.md

uablrek:
#  Draft PR

Fixes #698 

### Build

To build metallb with this PR you must use a local version of the K8s API. At the moment;
```
git clone --depth 1 -b dualstack-phase-3 https://github.com/khenidak/kubernetes.git
```
later when the PR https://github.com/kubernetes/kubernetes/pull/91824 has been merged the `master` branch for K8s shall be used. A "replace" section must be added in `go.mod` to point out the local clone;
```
replace (
        k8s.io/api => /home/uablrek/go/src/k8s.io/kubernetes/staging/src/k8s.io/api
        k8s.io/apimachinery => /home/uablrek/go/src/k8s.io/kubernetes/staging/src/k8s.io/apimachinery
        k8s.io/client-go => /home/uablrek/go/src/k8s.io/kubernetes/staging/src/k8s.io/client-go
)
```
You must use your own path of course.

### Progress

- [x] Build towards a local K8s API
- [x] Detect the presence of a dual-stack service by examining the length of `clusterIPs`
- [x] Assign both ipv4 and ipv6 load-balancer addresses to dual-stack services
- [x] A "metallb.universe.tf/load-balancer-ips" annotation for dual-stack
- [x] Check that metrics works with dual-stack services
- [ ] Test re-start of the controller (state is updated from K8s)
- [x] Unit tests for `allocator` updates
- [ ] Many more tests...

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -45,16 +45,16 @@ func ipnet(s string) *net.IPNet {
 	return n
 }
 
-func statusAssigned(ip string) v1.ServiceStatus {
-	return v1.ServiceStatus{
+func statusAssigned(ips ...string) v1.ServiceStatus {
+	s := v1.ServiceStatus{
 		LoadBalancer: v1.LoadBalancerStatus{
-			Ingress: []v1.LoadBalancerIngress{
-				{
-					IP: ip,
-				},
-			},
+			Ingress: []v1.LoadBalancerIngress{},
 		},
 	}
+	for _, ip := range ips {
+		s.LoadBalancer.Ingress = append(s.LoadBalancer.Ingress, v1.LoadBalancerIngress{IP: ip})
+	}
+	return s
 }
 
 // testK8S implements service by recording what the controller wants
@@ -131,6 +131,11 @@ func TestControllerMutation(t *testing.T) {
 				Protocol:   config.Layer2,
 				AutoAssign: false,
 				CIDR:       []*net.IPNet{ipnet("2000::1/128")},
+			},
+			"pool5": {
+				Protocol:   config.Layer2,
+				AutoAssign: false,
+				CIDR:       []*net.IPNet{ipnet("3010::/127"), ipnet("3.1.0.0/31")},
 			},
 		},
 	}
@@ -478,6 +483,100 @@ func TestControllerMutation(t *testing.T) {
 					ClusterIP: "3000::1",
 				},
 				Status: statusAssigned("1000::"),
+			},
+		},
+
+		{
+			desc: "Dual-stack basic",
+			in: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						"metallb.universe.tf/address-pool": "pool5",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					Type:       "LoadBalancer",
+					ClusterIP:  "3000::1",
+					ClusterIPs: []string{"3000::1", "30.0.0.0"},
+				},
+				Status: statusAssigned("1.2.3.0", "2000::1"),
+			},
+			want: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						"metallb.universe.tf/address-pool": "pool5",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					Type:       "LoadBalancer",
+					ClusterIP:  "3000::1",
+					ClusterIPs: []string{"3000::1", "30.0.0.0"},
+				},
+				Status: statusAssigned("3010::", "3.1.0.0"),
+			},
+		},
+
+		{
+			desc: "Dual-stack but LoadBalancerIP requested",
+			in: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						"metallb.universe.tf/address-pool": "pool5",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					Type:           "LoadBalancer",
+					LoadBalancerIP: "3.1.0.0",
+					ClusterIP:      "30.0.0.0",
+					ClusterIPs:     []string{"30.0.0.0", "3000::1"},
+				},
+				Status: statusAssigned("1.2.3.0", "2000::1"),
+			},
+			want: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						"metallb.universe.tf/address-pool": "pool5",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					Type:           "LoadBalancer",
+					LoadBalancerIP: "3.1.0.0",
+					ClusterIP:      "30.0.0.0",
+					ClusterIPs:     []string{"30.0.0.0", "3000::1"},
+				},
+				Status: statusAssigned("3.1.0.0"),
+			},
+		},
+
+		{
+			desc: "Dual-stack requested addresses with annotation",
+			in: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						"metallb.universe.tf/address-pool":      "pool5",
+						"metallb.universe.tf/load-balancer-ips": "3.1.0.1, 3010::1",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					Type:       "LoadBalancer",
+					ClusterIP:  "30.0.0.0",
+					ClusterIPs: []string{"30.0.0.0", "3000::1"},
+				},
+				Status: statusAssigned("1.2.3.0", "2000::1"),
+			},
+			want: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						"metallb.universe.tf/address-pool":      "pool5",
+						"metallb.universe.tf/load-balancer-ips": "3.1.0.1, 3010::1",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					Type:       "LoadBalancer",
+					ClusterIP:  "30.0.0.0",
+					ClusterIPs: []string{"30.0.0.0", "3000::1"},
+				},
+				Status: statusAssigned("3.1.0.1", "3010::1"),
 			},
 		},
 	}
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"math/rand"
 	"net"
+	"strings"
 	"testing"
 
 	"go.universe.tf/metallb/internal/allocator"
@@ -699,5 +700,44 @@ func TestDeleteRecyclesIP(t *testing.T) {
 	}
 	if len(gotSvc.Status.LoadBalancer.Ingress) == 0 || gotSvc.Status.LoadBalancer.Ingress[0].IP != "1.2.3.0" {
 		t.Fatal("svc2 didn't get an IP")
+	}
+}
+
+func TestRequestedIPs(t *testing.T) {
+	var tests = []struct {
+		testname     string
+		requestedIPs string
+		lbIP, lbIP2  net.IP
+		errContains  string
+	}{
+		{"Empty string", "", nil, nil, ""},
+		{"Ipv4 first", "10.0.0.0,  1000::", net.ParseIP("10.0.0.0"), net.ParseIP("1000::"), ""},
+		{"Ipv6 first", "1000::,10.0.0.0", net.ParseIP("1000::"), net.ParseIP("10.0.0.0"), ""},
+
+		{"Just one", "10.0.0.0", nil, nil, "Must be two addresses"},
+		{"More than two", "10.0.0.0, 1000::, 2000::", nil, nil, "Must be two addresses"},
+		{"Invalid 1", "10.0.0.400, 1000::", nil, nil, "Invalid address"},
+		{"Invalid 2", "10.0.0.0, 1000::2::2", nil, nil, "Invalid address"},
+		{"Same family", "1000::1, 1000::2", nil, nil, "Same family"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.testname, func(t *testing.T) {
+			lbIP, lbIP2, err := parseRequestedIPs(tt.requestedIPs)
+			if (err == nil) != (tt.errContains == "") {
+				t.Errorf("err: Unexpected %v", err)
+			}
+			if err != nil && tt.errContains != "" {
+				if !strings.Contains(err.Error(), tt.errContains) {
+					t.Errorf("err: got %v, should contain %s", err, tt.errContains)
+				}
+			}
+			if !lbIP.Equal(tt.lbIP) {
+				t.Errorf("lbIP: got %v, want %v", lbIP, tt.lbIP)
+			}
+			if !lbIP2.Equal(tt.lbIP2) {
+				t.Errorf("lbIP2: got %v, want %v", lbIP2, tt.lbIP2)
+			}
+		})
 	}
 }
--- a/controller/service.go
+++ b/controller/service.go
@@ -186,7 +186,7 @@ func (c *controller) allocateIP(key string, svc *v1.Service) (net.IP, error) {
 	return c.ips.Allocate(key, isIPv6, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc))
 }
 
-// Dual-stack
+// Dual-stack.
 
 func (c *controller) convergeBalancerDual(l log.Logger, key string, svc *v1.Service) bool {
 	var lbIP, lbIP2 net.IP
--- a/controller/service.go
+++ b/controller/service.go
@@ -17,6 +17,7 @@ package main
 import (
 	"fmt"
 	"net"
+	"strings"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
@@ -45,6 +46,10 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 		level.Info(l).Log("event", "clearAssignment", "reason", "noClusterIP", "msg", "No ClusterIP")
 		c.clearServiceState(key, svc)
 		return true
+	}
+
+	if svc.Spec.ClusterIPs != nil && len(svc.Spec.ClusterIPs) > 1 {
+		return c.convergeBalancerDual(l, key, svc)
 	}
 
 	// The assigned LB IP is the end state of convergence. If there's
@@ -179,4 +184,133 @@ func (c *controller) allocateIP(key string, svc *v1.Service) (net.IP, error) {
 
 	// Okay, in that case just bruteforce across all pools.
 	return c.ips.Allocate(key, isIPv6, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc))
+}
+
+// Dual-stack
+
+func (c *controller) convergeBalancerDual(l log.Logger, key string, svc *v1.Service) bool {
+	var lbIP, lbIP2 net.IP
+
+	// The assigned LB IP is the end state of convergence. If there's
+	// none or a malformed one, nuke all controlled state so that we
+	// start converging from a clean slate.
+	if len(svc.Status.LoadBalancer.Ingress) > 1 {
+		lbIP = net.ParseIP(svc.Status.LoadBalancer.Ingress[0].IP)
+		lbIP2 = net.ParseIP(svc.Status.LoadBalancer.Ingress[1].IP)
+	}
+
+	// It's possible the config mutated and the IP we have no longer
+	// makes sense. If so, clear it out and give the rest of the logic
+	// a chance to allocate again.
+	if lbIP != nil && lbIP2 != nil {
+		// This assign is idempotent if the config is consistent,
+		// otherwise it'll fail and tell us why.
+		if err := c.ips.AssignDual(key, lbIP, lbIP2, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc)); err != nil {
+			l.Log("event", "clearAssignment", "reason", "notAllowedByConfig", "msg", "current IP not allowed by config, clearing")
+			c.clearServiceState(key, svc)
+			lbIP = nil
+		}
+
+		if lbIP != nil {
+			// The user might also have changed the pool annotation, and
+			// requested a different pool than the one that is currently
+			// allocated.
+			desiredPool := svc.Annotations["metallb.universe.tf/address-pool"]
+			if desiredPool != "" && c.ips.Pool(key) != desiredPool {
+				l.Log("event", "clearAssignment", "reason", "differentPoolRequested", "msg", "user requested a different pool than the one currently assigned")
+				c.clearServiceState(key, svc)
+				lbIP = nil
+			}
+		}
+	} else {
+		c.clearServiceState(key, svc)
+		lbIP = nil
+	}
+
+	// The (singular) svc.Spec.LoadBalancerIP is ignored for dual-stack
+	if svc.Spec.LoadBalancerIP != "" {
+		l.Log("event", "loadBalancerIP", "reason", "N/A", "msg", "loadBalancerIP ignored for dual-stack")
+	}
+
+	if requestedIPs := svc.Annotations["metallb.universe.tf/load-balancer-ips"]; requestedIPs != "" {
+		// Until a svc.Spec.LoadBalancerIPs exists we use an annotation.
+		// requestedIPs must be a comma-separated list of 2 addresses, one from each family.
+		ips := strings.Split(requestedIPs, ",")
+		if len(ips) != 2 {
+			l.Log("op", "allocateIP", "load-balancer-ips", len(ips), "msg", "Must be two addresses")
+			return true
+		}
+		if lbIP = net.ParseIP(strings.TrimSpace(ips[0])); lbIP == nil {
+			l.Log("op", "allocateIP", "load-balancer-ips", ips[0], "msg", "Invalid addresses")
+			return true
+		}
+		if lbIP2 = net.ParseIP(strings.TrimSpace(ips[1])); lbIP2 == nil {
+			l.Log("op", "allocateIP", "load-balancer-ips", ips[1], "msg", "Invalid addresses")
+			return true
+		}
+		if (lbIP.To4() == nil) == (lbIP2.To4() == nil) {
+			l.Log("op", "allocateIP", "load-balancer-ips", requestedIPs, "msg", "Same family")
+		}
+
+		// Try to assign the requested IPs
+		if err := c.ips.AssignDual(key, lbIP, lbIP2, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc)); err != nil {
+			l.Log("op", "allocateIP", "error", err, "msg", "Can't assign requested IPs")
+			return true
+		}
+	}
+
+	// If lbIP's is still nil at this point, try to allocate.
+	if lbIP == nil {
+		if !c.synced {
+			l.Log("op", "allocateIP", "error", "controller not synced", "msg", "controller not synced yet, cannot allocate IP; will retry after sync")
+			return false
+		}
+		ip, ip2, err := c.allocateIPDual(key, svc)
+		if err != nil {
+			l.Log("op", "allocateIP", "error", err, "msg", "IP allocation failed")
+			c.client.Errorf(svc, "AllocationFailed", "Failed to allocate IP for %q: %s", key, err)
+			// The outer controller loop will retry converging this
+			// service when another service gets deleted, so there's
+			// nothing to do here but wait to get called again later.
+			return true
+		}
+		lbIP = ip
+		lbIP2 = ip2
+		l.Log("event", "ipAllocated", "ip", lbIP, "ip2", lbIP2, "msg", "IP address assigned by controller")
+		c.client.Infof(svc, "IPAllocated", "Assigned IP %q %q", lbIP, lbIP2)
+	}
+
+	if lbIP == nil || lbIP2 == nil {
+		l.Log("bug", "true", "msg", "internal error: failed to allocate an IP, but did not exit convergeService early!")
+		c.client.Errorf(svc, "InternalError", "didn't allocate an IP but also did not fail")
+		c.clearServiceState(key, svc)
+		return true
+	}
+
+	pool := c.ips.Pool(key)
+	if pool == "" || c.config.Pools[pool] == nil {
+		l.Log("bug", "true", "ip", lbIP, "msg", "internal error: allocated IP has no matching address pool")
+		c.client.Errorf(svc, "InternalError", "allocated an IP that has no pool")
+		c.clearServiceState(key, svc)
+		return true
+	}
+
+	// At this point, we have an IP selected somehow, all that remains
+	// is to program the data plane.
+	svc.Status.LoadBalancer.Ingress = []v1.LoadBalancerIngress{{IP: lbIP.String()}, {IP: lbIP2.String()}}
+	return true
+}
+
+func (c *controller) allocateIPDual(key string, svc *v1.Service) (net.IP, net.IP, error) {
+	desiredPool := svc.Annotations["metallb.universe.tf/address-pool"]
+	if desiredPool != "" {
+		ip, ip2, err := c.ips.AllocateFromPoolDual(key, desiredPool, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc))
+		if err != nil {
+			return nil, nil, err
+		}
+		return ip, ip2, nil
+	}
+
+	// Okay, in that case just bruteforce across all pools.
+	return c.ips.AllocateDual(key, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc))
 }
--- a/controller/service.go
+++ b/controller/service.go
@@ -26,6 +26,11 @@ import (
 	"go.universe.tf/metallb/internal/allocator/k8salloc"
 )
 
+const (
+	annotationAddressPool = "metallb.universe.tf/address-pool"
+	annotationLoadBalancerIPs = "metallb.universe.tf/load-balancer-ips"
+)
+
 func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service) bool {
 	var lbIP net.IP
 
@@ -84,7 +89,7 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 		// The user might also have changed the pool annotation, and
 		// requested a different pool than the one that is currently
 		// allocated.
-		desiredPool := svc.Annotations["metallb.universe.tf/address-pool"]
+		desiredPool := svc.Annotations[annotationAddressPool]
 		if lbIP != nil && desiredPool != "" && c.ips.Pool(key) != desiredPool {
 			level.Info(l).Log("event", "clearAssignment", "reason", "differentPoolRequested", "msg", "user requested a different pool than the one currently assigned")
 			c.clearServiceState(key, svc)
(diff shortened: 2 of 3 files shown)


uablrek:
First draft only detects a dual-stack svc and logs a warning;
```
{"Only one address assigned":"(MISSING)","caller":"service.go:50","event":"dualStack","reason":"NYI","service":"default/mconnect-dual","ts":"2020-09-22T10:41:39.811094373Z"}
```
But build instructions are in place and this PR will remain in "draft" mode until K8s v1.20.0 has been released and serve as base for reviews and progress report.

uablrek:
@russellb @rata Can you please check if the ci/circleci can be turned off for PR drafts (as this one). CI will not work for a long time on this one :smile:

rata:
@uablrek I'll try to check later. But no problem, test can fail here :)

russellb:
Yeah, I'm not sure if we can turn it off, but I'm also not bothered with it failing.

uablrek:
@russellb @rata The last (force-pushed) commit implements the base function. Please take a look if you have time. No real review but so you get an idea where this is heading and hit the beaks if you think it's necessary :smiley: 

I have tried quite hard to keep the dual-stack code separated from the old code. New functions rather than alter the parameters and interleave new code with the old.

Below is logs from assignment of dual-stack addresses. I see immediately that I have forgot to remove some debug printouts but I will do that later.
```
{"caller":"main.go:49","event":"startUpdate","msg":"start of service update","service":"default/mconnect-dual","ts":"2020-09-23T14:33:40.077198203Z"}
AllocateFromPoolDual default/mconnect-dual default
{"caller":"service.go:250","event":"ipAllocated","ip":"10.0.0.2","ip2":"1000::1","msg":"IP address assigned by controller","service":"default/mconnect-dual","ts":"2020-09-23T14:33:40.077345574Z"}
```

uablrek:
I have added a constraint;
**For dual-stack services both address must be taken from the same pool**

I think this is a reasonable constraint that will be accepted by users. I personally would not like it any other way.

But the reason is that the `allocator.alloc` struct contains a the pool name. To accept different pools would mess up the code beyond recognition.

uablrek:
Oh, and no unit-tests are added since I am still probing the way. But old unit-tests runs without problems for regression tests.

uablrek:
I checked the speaker code and unfortunately it handle only one loadBalancerIP. To extend it to handle two (or more?) loadBalancerIP's does not require any new API. The `service.status.loadBalancer.ingress` field is an array already so work can start immadiately.

Shall I write an issue? I will probably not work on the speaker myself since we can't use it, but an issue to explain the problem.

russellb:
> I checked the speaker code and unfortunately it handle only one loadBalancerIP. To extend it to handle two (or more?) loadBalancerIP's does not require any new API. The `service.status.loadBalancer.ingress` field is an array already so work can start immadiately.
> 
> Shall I write an issue? 

Yes, please!

> I will probably not work on the speaker myself since we can't use it, but an issue to explain the problem.

Just to understand your use case, you're using the metallb controller but have your own speaker?

uablrek:
> Just to understand your use case, you're using the metallb controller but have your own speaker?

Yes, we need more BGP functions than the metallb speaker offers, e.g BPF and ipv6.

uablrek:
Last commit adds an annotation as a work-around for the missing `loadBalancerIPs` array;
```
apiVersion: v1
kind: Service
metadata:
  name: dual-fixed-adr
  annotations:
    metallb.universe.tf/load-balancer-ips: "10.0.0.20,1000::20"
spec:
  ipFamilyPolicy: RequireDualStack
  selector:
    app: mconnect
  ports:
  - port: 5001
    name: mconnect
    targetPort: 5001
  type: LoadBalancer
```
Tested manually;
```
$ kubectl get svc dual-fixed-adr
NAME             TYPE           CLUSTER-IP    EXTERNAL-IP          PORT(S)          AGE
dual-fixed-adr   LoadBalancer   12.0.214.12   10.0.0.20,1000::20   5001:30667/TCP   38s
```
The load-balancer-ips must be a comma-separated list of 2 addresses, one from each family. The addresses must be free and belong to the same pool.

uablrek:
@russellb @rata I feel pretty satisfied with the implementation now. Can you please make a "pre-review" before I move on with unit-tests. Unit-tests will probably take far more time than the implementation itself and I would like a "go" so I don't waste time.

I have tested all(?) normal cases on a cluster with K8s dual-stack phase.3. I have tested just a few fault-cases and then most by mistake, but corner cases are probably better to test in unit-tests anyway.

uablrek:
@russellb @rata Dual-stack phase.3 is now merged to the K8s master and will be included in K8s v1.20.0.

Can you please review this PR? I have re-based and tested all basic cases. I will not write unit-tests until you have made a (pre-)review on the functional updates.

controller/service.go
```diff
@@ -179,3 +184,132 @@ func (c *controller) allocateIP(key string, svc *v1.Service) (net.IP, error) {
...
+
+func (c *controller) convergeBalancerDual(l log.Logger, key string, svc *v1.Service) bool {
+	var lbIP, lbIP2 net.IP
+
+	// The assigned LB IP is the end state of convergence. If there's
+	// none or a malformed one, nuke all controlled state so that we
+	// start converging from a clean slate.
+	if len(svc.Status.LoadBalancer.Ingress) > 1 {
+		lbIP = net.ParseIP(svc.Status.LoadBalancer.Ingress[0].IP)
+		lbIP2 = net.ParseIP(svc.Status.LoadBalancer.Ingress[1].IP)
```
rata, line 198: These values are nil if it is malformed (we do the same check on single stack). If we share the logic, those things won't have to be duplicated :)

rata (commented):
@uablrek thanks a lot for keep this going and sorry for the late response!

My biggest concern on this is that it seems a lot of c&p is done for the allocator and in the controller. Basically, single stack and dual stack are a lot similar, but little code is shared. I explain in a little bit more detail on the code comments, but this is what worries me. It seems to increase significantly the tech-debt.

Have you tried sharing code and it was not good way for some reason?

rata:
@uablrek also, how do you test this locally? I've read what you wrote, but to test locally, it of course doesn't work with our current scripts as don't use k8s 1.20 (like `inv dev-env`, that is a script in the repo that creates a k8s cluster with kind).

Just a k8s cluster with `hack/local-up...` or that is not cool for some reason (I guess it would work, 1 node is fine)? Or can you share how you do it? It might save some time if that option won't work, for example (it would have saved me time already, as I already tried with `inv-dev` and the realized the kubernetes deployed there was the culprit :))

controller/service.go
```diff
@@ -179,3 +183,105 @@ func (c *controller) allocateIP(key string, svc *v1.Service) (net.IP, error) {
 	// Okay, in that case just bruteforce across all pools.
 	return c.ips.Allocate(key, isIPv6, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc))
 }
+
+// Dual-stack
+
+func (c *controller) convergeBalancerDual(l log.Logger, key string, svc *v1.Service) bool {
```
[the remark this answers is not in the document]
uablrek: From comments above;

> I have tried quite hard to keep the dual-stack code separated from the old code. New functions rather than alter the parameters and interleave new code with the old.

From experience this is what maintainers wants. The main concern is to not break anything old that works. But if you have trust in your regression tests, I can certainly expore the possibility. But be aware that "if single-stack; then ... else" all over the place is _not_ an improvment or easier to maintain, so if it goes that way I will back out, or suggest a larger re-factoring.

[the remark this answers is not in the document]
uablrek: Another reason for keeping single and dual stack code separated is that singe-stack cluster will become obsolete when dual-stack becomed GA. That is, you can't "turn-off" dual-stack even if only one family is actually used. Then it is nice to remove obsolete code from metallb rather than maintain it forever.

uablrek:
To test locally use K8s v1.20.0-rc.0, it includes dual-stack phase.3. Install with dual-stack enabled and use a CNI-plugin also supporting and configuref for dual-stack, e.g Calico.

Build the controller as described with the modified `go.mod` and build an image. Use a pool that contains both address families.

```
address-pools:
- name: default
  protocol: layer2
  addresses:
  - 10.0.0.0/24
  - "1000::/120"
```

Then create a service with `ipFamilyPolicy` with "RequireDualStack" or "PreferDualStack";

```
apiVersion: v1
kind: Service
metadata:
  name: test-dual
spec:
  ipFamilyPolicy: RequireDualStack
  selector:
    app: someapp
  ports:
  - port: 5001
  type: LoadBalancer
```
It should get both addresses;
```
test-dual               LoadBalancer   12.0.212.245      10.0.0.0,1000::1      5001:32123/TCP   30s
```

controller/service.go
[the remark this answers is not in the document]
rata: > From experience this is what maintainers wants. The main concern is to not break anything old that works. 

I prefer clean code and pay the price to handle multiple ips when adding the dual stack seems reasonable IMHO.

> But if you have trust in your regression tests, I can certainly expore the possibility. But be aware that "if single-stack; then ... else" all over the place is _not_ an improvment or easier to maintain, so if it goes that way I will back out, or suggest a larger re-factoring.

Yes, definitely, I agree. I don't want an "if dual ... then ..." either :+1: 

> Another reason for keeping single and dual stack code separated is that singe-stack cluster will become obsolete when dual-stack becomed GA. That is, you can't "turn-off" dual-stack even if only one family is actually used. Then it is nice to remove obsolete code from metallb rather than maintain it forever

True, that is a good point. The single stack can be removed in a year or two, it is still quite some time. This PR, though, is kind of hacky with "ip" and "ip2" instead of a `ips` slice or something. And if we write using something using a slice to handle dual stack, won't it be very close (if not just "working") to handle also the single stack? What do you think?

internal/k8s/k8s.go
```diff
@@ -260,8 +261,8 @@ func New(cfg *Config) (*Client, error) {
 }
 
 // PodIPs returns the IPs of all the pods matched by the labels string.
-func (c *Client) PodIPs(namespace, labels string) ([]string, error) {
-	pl, err := c.client.CoreV1().Pods(namespace).List(metav1.ListOptions{LabelSelector: labels})
+func (c *Client) PodsIPs(namespace, labels string) ([]string, error) {
```
aojea, line 264: the description commet doesn't  match now

go.mod
```diff
@@ -23,18 +20,19 @@ require (
...
-	gopkg.in/inf.v0 v0.9.1 // indirect
-	gopkg.in/yaml.v2 v2.2.2
-	k8s.io/api v0.0.0-20190620084959-7cf5895f2711
-	k8s.io/apimachinery v0.0.0-20190612205821-1799e75a0719
-	k8s.io/client-go v0.0.0-20190620085101-78d2af792bab
+	golang.org/x/sys v0.0.0-20200622214017-ed371f2e16b4
+	gopkg.in/yaml.v2 v2.2.8
+	k8s.io/api v0.19.0
+	k8s.io/apimachinery v0.19.0
+	k8s.io/client-go v0.19.0
```
aojea, line 30: you need 0.20.0 for the new dual stack api fieds

controller/service.go
```diff
@@ -46,6 +47,10 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 		return true
 	}
 
+	if svc.Spec.ClusterIPs != nil && len(svc.Spec.ClusterIPs) > 1 {
```
aojea, line 50: I think that `len(svc.Spec.ClusterIPs) > 1` will be enough, no?

go.mod
uablrek: Done!

uablrek:
@rata About local test https://github.com/metallb/metallb/pull/727#issuecomment-738936943

K8s v1.20.0 including APIs are released so the `go.mod` is updated (commented by @aojea). No local modifications are needed. I build with;
```
CGO_ENABLED=0 GOOS=linux go install -ldflags "-extldflags '-static' -X go.universe.tf/metallb/internal/version.version=$(date +%F:%T)" ./controller
```

uablrek:
Hi, I spent most of yesterday to interleave dual-stack code. "allocator.go" is about finished. Personally I think it's terrible but I let you have a look where it's heading before I go on with "service.go".

Return values are changed from net.IP to []net.IP and a new enum `IPType` is added in allocator which enforced a dependency in "service.go".

rata:
@uablrek thanks! I'll try to have a look, but I expect to be busy with the public holidays now :(

starcraft66:
Hey, are there any chances this could be looked at further sometime soon? The introduction of this feature would be very helpful to me.

uablrek:
@starcraft66 If you just want to test dual-stack use;

https://github.com/Nordix/assign-lb-ip

and setup routes manually (or with scripts). That's what I do.

starcraft66:
Thanks for the input, I'm just sticking with assigning two LoadBalancers to every service (one for v4 and one for v6) because I don't want to have to do any additional configuration outside of the Kubernetes apiserver.

mmack:
Same here. I have a dual-stack cluster ready and waiting on this here :) If you guys need tester ping me...

uablrek:
@mmack If you build the controller yourself just apply the PR first with;
```
curl -L https://github.com/metallb/metallb/pull/727.patch | patch -p1
```
Use the annotation (https://github.com/metallb/metallb/pull/727#issuecomment-698273081) if you like and remember that the pool used must contain both ipv4 and ipv6 addresses.

mmack:
> @mmack If you build the controller yourself just apply the PR first with;
> 
> ```
> curl -L https://github.com/metallb/metallb/pull/727.patch | patch -p1
> ```
> 
> Use the annotation ([#727 (comment)](https://github.com/metallb/metallb/pull/727#issuecomment-698273081)) if you like and remember that the pool used must contain both ipv4 and ipv6 addresses.

@uablrek  Thanks, i'll try it out! But how's the peer's config then in that case? I have 2 peers - one ipv4 only - and one ipv6 only.
currently i see this in my logs b/c the speaker want's to advertise ipv6 ips to the ipv4 bgp peer.
```
{"caller":"main.go:260","error":"cannot advertise non-v4 prefix \"\u003cnil\u003e\"","ip":"2XXX:8cc2::1","msg":"failed to announce service"
```
Does that change with this PR?

uablrek:
No, update of the speaker is a different issue https://github.com/metallb/metallb/issues/731

No PR yet and since we can't use the metallb speaker I will not create one. I am uncertain if the BGP speaker supports ipv6 yet. I know it is worked on but not the status. The L2 speaker should not be hard to update for dual-stack though and it already supports ipv6.

Jean-Daniel:
BGP speaker IPv6 support was never merged but is ready and validated by many users. See #590

uablrek:
@Jean-Daniel The `status.loadBalancer.ingress` has been an array since _way back_, long before dual-stack. The speaker only uses the first item (index 0).

If the speaker handles all (or at leas 2) item in this array in the same way as now done only for the firsts it will become dual-stack compatible and still be backward compatible.

taxilian:
I just pulled down the branch, merged 0.10.2 (maybe even correctly -- I don't know go, so I may have made mistakes) and brought it up on my node -- things seem to be working great!

https://github.com/taxilian/metallb/tree/dual-stack-3 is where I pushed my changes

I don't need the bgp advertiser because I use calico for bgp and it can advertise it for me, so I just need the controller to allocate addresses -- which this now does! thanks for this

starcraft66:
> I don't need the bgp advertiser because I use calico for bgp and it can advertise it for me, so I just need the controller to allocate addresses -- which this now does! thanks for this

@taxilian Great to hear, I have the exact same setup as you do, calico announcing loadBalancerIPs, no speaker. I think I will set up CI to build MetalLB controller docker images with this pull request merged.

uablrek:
I ditched the modify-all-over-the-pace approach and restored the original add-on approach with minor checks in existing code and call dedicated dual-stack functions.

If someone want to try to the other approach I attach the diff I made before. It can't be applied without rejects, but you get a hint [metallb-interleave.diff.txt](https://github.com/metallb/metallb/files/7172471/metallb-interleave.diff.txt).

Apply in the metallb clone with;
```
patch -p1 < metallb-interleave.diff.txt
```

uablrek:
I attach the manifests I use for local testing; 
[metallb-dual3.yaml.txt](https://github.com/metallb/metallb/files/7172561/metallb-dual3.yaml.txt)
[metallb-dual3-svc.yaml.txt](https://github.com/metallb/metallb/files/7172563/metallb-dual3-svc.yaml.txt)

uablrek:
I have uploaded my test image so _for testing only_, you may try to just replace the controller image with;
```
      containers:
      - name: controller
        image: registry.nordix.org/cloud-native/metallb-dual3:latest
        imagePullPolicy: IfNotPresent
```

internal/allocator/allocator.go
```diff
@@ -381,3 +372,168 @@ func ipConfusesBuggyFirmwares(ip net.IP) bool {
...
+// permissible by sharingKey and backendKey.
+func (a *Allocator) AssignDual(svc string, ip, ip2 net.IP, ports []Port, sharingKey, backendKey string) error {
+	if ipIsIPv6(ip) == ipIsIPv6(ip2) {
+		return fmt.Errorf("%q %q is has the same family", ip, ip2)
+	}
+	pool := poolFor(a.pools, ip)
+	if pool == "" {
+		return fmt.Errorf("%q is not allowed in config", ip)
+	}
+	if poolFor(a.pools, ip2) != pool {
```
msherif1234, line 388: so for this to work we need to have only one address-pool with both addresses something like 
```
addresses:
  - 2620:52:0:4974:4da7:3546:2a72:3fdf - 2620:52:0:4974:4da7:3546:2a72:3fdf
  - 10.73.116.54 - 10.73.116.54
```
having address-pool for v4 and another for v6 won't work right ?

internal/allocator/allocator.go
```diff
@@ -381,3 +372,168 @@ func ipConfusesBuggyFirmwares(ip net.IP) bool {
...
+	// unconditionally safe to do.
+	alloc := &alloc{
+		pool:  pool,
+		ip:    ip,
+		ip2:   ip2,
+		ports: make([]Port, len(ports)),
+		key:   *sk,
+	}
+	for i, port := range ports {
+		port := port
```
msherif1234, line 417: why u need another port variable here ?

internal/allocator/allocator.go
```diff
@@ -381,3 +372,168 @@ func ipConfusesBuggyFirmwares(ip net.IP) bool {
...
+// AllocateFromPoolDual assigns available IP's from pool to service.
+func (a *Allocator) AllocateFromPoolDual(svc string, poolName string, ports []Port, sharingKey, backendKey string) (net.IP, net.IP, error) {
+	if alloc := a.allocated[svc]; alloc != nil {
+		if err := a.AssignDual(svc, alloc.ip, alloc.ip2, ports, sharingKey, backendKey); err != nil {
+			return nil, nil, err
+		}
+		return alloc.ip, alloc.ip2, nil
+	}
+
+	fmt.Println("AllocateFromPoolDual", svc, poolName)
```
msherif1234, line 433: we need better logging or just remove it from this layer and add it to service

internal/allocator/allocator.go
uablrek: No. If you did not have that restriction you must basically rewrite the whole allocator. Beside IMO it is a natural restriction.

internal/allocator/allocator.go
```diff
@@ -381,3 +372,168 @@ func ipConfusesBuggyFirmwares(ip net.IP) bool {
...
+				}
+				if a.checkSharing(svc, pos.IP.String(), ports, sk) != nil {
+					continue
+				}
+				ip = pos.IP
+				break
+			}
+		} else {
+			if cidrIsIPv6(cidr) == ipIsIPv6(ip) {
+				continue
```
msherif1234, line 460: is there a special order assumed here I am bit confused with v6 check here and skip the allocation ?

internal/allocator/allocator.go
uablrek: Seems like a misstake. Will fix it. (strange that it even compiles)

uablrek: But it it a copy/paste thing. The exact same code is in the Assign() function further up. I'll fix that too.

internal/allocator/allocator.go
uablrek: Seem like a lingering debug printout. I'll remove it.

internal/allocator/allocator.go
uablrek: No, any order will do, ipv4,ipv6 or ipv6,ipv4. The `cidrIsIPv6(cidr)` == ipIsIPv6(ip)` ensures that we pick ip2 from a cidr of the "other" family, whatever that may be.

internal/allocator/allocator.go
msherif1234: ok lets be very specific and clear on this can you layout example of what will work and what won't to be 100% sure we are clear ?

msherif1234:
Any chance of adding just few basic test cases for allocator changes ?

controller/service.go
```diff
@@ -180,3 +185,132 @@ func (c *controller) allocateIP(key string, svc *v1.Service) (net.IP, error) {
...
+
+// Dual-stack.
+
+func (c *controller) convergeBalancerDual(l log.Logger, key string, svc *v1.Service) bool {
+	var lbIP, lbIP2 net.IP
+
+	// The assigned LB IP is the end state of convergence. If there's
+	// none or a malformed one, nuke all controlled state so that we
+	// start converging from a clean slate.
+	if len(svc.Status.LoadBalancer.Ingress) > 1 {
```
msherif1234, line 197: do we need to check for an  exact 2 ? we can't do anything with more than that

controller/service.go
```diff
@@ -180,3 +185,132 @@ func (c *controller) allocateIP(key string, svc *v1.Service) (net.IP, error) {
...
+	}
+
+	// It's possible the config mutated and the IP we have no longer
+	// makes sense. If so, clear it out and give the rest of the logic
+	// a chance to allocate again.
+	if lbIP != nil && lbIP2 != nil {
+		// This assign is idempotent if the config is consistent,
+		// otherwise it'll fail and tell us why.
+		if err := c.ips.AssignDual(key, lbIP, lbIP2, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc)); err != nil {
+			l.Log("event", "clearAssignment", "reason", "notAllowedByConfig", "msg", "current IP not allowed by config, clearing")
```
msherif1234, line 209: adding the error to the log might be useful too

controller/service.go
```diff
@@ -180,3 +185,132 @@ func (c *controller) allocateIP(key string, svc *v1.Service) (net.IP, error) {
...
+			c.clearServiceState(key, svc)
+			lbIP = nil
+		}
+
+		if lbIP != nil {
+			// The user might also have changed the pool annotation, and
+			// requested a different pool than the one that is currently
+			// allocated.
+			desiredPool := svc.Annotations["metallb.universe.tf/address-pool"]
```
msherif1234, line 218: can you pls define consts for annotation strings

controller/service.go
```diff
@@ -180,3 +185,132 @@ func (c *controller) allocateIP(key string, svc *v1.Service) (net.IP, error) {
...
+		if lbIP = net.ParseIP(strings.TrimSpace(ips[0])); lbIP == nil {
+			l.Log("op", "allocateIP", "load-balancer-ips", ips[0], "msg", "Invalid addresses")
+			return true
+		}
+		if lbIP2 = net.ParseIP(strings.TrimSpace(ips[1])); lbIP2 == nil {
+			l.Log("op", "allocateIP", "load-balancer-ips", ips[1], "msg", "Invalid addresses")
+			return true
+		}
+		if (lbIP.To4() == nil) == (lbIP2.To4() == nil) {
+			l.Log("op", "allocateIP", "load-balancer-ips", requestedIPs, "msg", "Same family")
```
msherif1234, line 252: do we need to return here if both from the same af ?

internal/allocator/allocator.go
```diff
@@ -381,3 +371,166 @@ func ipConfusesBuggyFirmwares(ip net.IP) bool {
...
+	}
+
+	if ip2 == nil {
+		// Woops, run out of IPs :( Fail.
+		return nil, nil, fmt.Errorf("no available IPs for in pool %q", poolName)
+	}
+
+	err := a.AssignDual(svc, ip, ip2, ports, sharingKey, backendKey)
+	if err != nil {
+		return nil, nil, err
```
msherif1234, line 483: do we need to free up whatever we have allocated earlier ?

controller/service.go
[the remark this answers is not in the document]
fedepaol: +1 on considering to unify the logic. The two functions do not seem to far if we had allocate always returning a slice. A lot of logic is done on the first element anyway.

controller/service.go
```diff
@@ -180,3 +185,132 @@ func (c *controller) allocateIP(key string, svc *v1.Service) (net.IP, error) {
...
+				lbIP = nil
+			}
+		}
+	} else {
+		c.clearServiceState(key, svc)
+		lbIP = nil
+	}
+
+	// The (singular) svc.Spec.LoadBalancerIP is ignored for dual-stack
+	if svc.Spec.LoadBalancerIP != "" {
```
fedepaol, line 231: Is this arbitrary? Given that we can't have two ips, here we are choosing to go against the user's will and assign two ips instead of the one he chose.

controller/service.go
```diff
@@ -180,3 +185,132 @@ func (c *controller) allocateIP(key string, svc *v1.Service) (net.IP, error) {
...
+		l.Log("event", "loadBalancerIP", "reason", "N/A", "msg", "loadBalancerIP ignored for dual-stack")
+	}
+
+	if requestedIPs := svc.Annotations["metallb.universe.tf/load-balancer-ips"]; requestedIPs != "" {
+		// Until a svc.Spec.LoadBalancerIPs exists we use an annotation.
+		// requestedIPs must be a comma-separated list of 2 addresses, one from each family.
+		ips := strings.Split(requestedIPs, ",")
```
fedepaol, line 238: This is a long (and not interesting) section of code. Would it make sense to wrap it in a function?

controller/service.go
```diff
@@ -180,3 +185,132 @@ func (c *controller) allocateIP(key string, svc *v1.Service) (net.IP, error) {
...
+		// Try to assign the requested IPs
+		if err := c.ips.AssignDual(key, lbIP, lbIP2, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc)); err != nil {
+			l.Log("op", "allocateIP", "error", err, "msg", "Can't assign requested IPs")
+			return true
+		}
+	}
+
+	// If lbIP's is still nil at this point, try to allocate.
+	if lbIP == nil {
+		if !c.synced {
```
fedepaol, line 264: this is probably taken from the other one, but I wonder why it can't be done from the beginning or even before invoking `SetBalancer`

internal/allocator/allocator.go
```diff
@@ -381,3 +371,166 @@ func ipConfusesBuggyFirmwares(ip net.IP) bool {
...
+	}
+
+	// Find IP's from both families
+	var ip, ip2 net.IP
+	sk := &key{
+		sharing: sharingKey,
+		backend: backendKey,
+	}
+	for _, cidr := range pool.CIDR {
+		if ip == nil {
```
fedepaol, line 443: Can you refactor the two blocks to avoid duplicating the code inside? 
Either adding ifs to check if we need to assign to ip or ip2 or wrapping in a function would work

fedepaol:
Left some comment. I think the major part of the discussion goes around what was started in https://github.com/metallb/metallb/pull/727#discussion_r536069953 , deciding if we want to duplicate code vs refactor the existing one to cover both.

fedepaol:
Other comment: since the speaker is not ready to advertise both ips (see https://github.com/metallb/metallb/blob/5d236989e61683daf1eaad00777b81ce7dd98c5d/speaker/main.go#L230 ), it might be useful to hide the dual stack allocation logic behind a feature flag or an env variable so we don't allocate ips that can't be advertised.

controller/service.go
uablrek: What should be done for > 2? Reject everything? If you have > 1 for single-stack it just pick the first, the current condition is consistent with that.

controller/service.go
uablrek: Does not `l.Log()` do that? What more is needed? Is there an example on another place in the code?

controller/service.go
uablrek: Done

controller/service.go
uablrek: We can reject of course. Is that preferable?

controller/service.go
uablrek: Yes, that's a bug. Thanks :+1:

controller/service.go
uablrek: Yes, makes sense. To allow unit-test if nothing else.

controller/service.go
uablrek: I am unsure, so I keept Dave Andersons code rather than risking to mess something up.

internal/allocator/allocator.go
uablrek: Done

controller/service.go
uablrek: Now covered in unit-test

controller/service.go
msherif1234: pls use consistent log method like in the rest of the file every where in the dual stack function
`
level.Error(l).Log("event", "clearAssignment", "reason", "notAllowedByConfig", "msg", "current IP not allowed by config, clearing")
`

controller/controller_test.go
```diff
@@ -701,3 +702,42 @@ func TestDeleteRecyclesIP(t *testing.T) {
...
+	}{
+		{"Empty string", "", nil, nil, ""},
+		{"Ipv4 first", "10.0.0.0,  1000::", net.ParseIP("10.0.0.0"), net.ParseIP("1000::"), ""},
+		{"Ipv6 first", "1000::,10.0.0.0", net.ParseIP("1000::"), net.ParseIP("10.0.0.0"), ""},
+
+		{"Just one", "10.0.0.0", nil, nil, "Must be two addresses"},
+		{"More than two", "10.0.0.0, 1000::, 2000::", nil, nil, "Must be two addresses"},
+		{"Invalid 1", "10.0.0.400, 1000::", nil, nil, "Invalid address"},
+		{"Invalid 2", "10.0.0.0, 1000::2::2", nil, nil, "Invalid address"},
+		{"Same family", "1000::1, 1000::2", nil, nil, "Same family"},
```
msherif1234, line 721: is having mcast ipv4 or ipv6 is valid IP address ? the validation check doesn't fail mcast 
```
{"Invalid 3", "10.0.0.0, ff00::2", nil, nil, "Invalid address"},
{"Invalid 4", "224.0.0.40, 1000::", nil, nil, "Invalid address"},
```

msherif1234: can u add dual-stack service test case to `TestControllerMutation`

muhlba91:
are there any updates on this?
we are waiting for dual-stack support and mixed protocol support.

fedepaol:
Hi @uablrek , I went over the PR again and I am still concerned about the code duplication. Here https://github.com/metallb/metallb/pull/727#discussion_r536758008 you expressed the concerns, but now we have enough CI in place that should help finding regression, and I'd prefer to have a cleaner codebase rather than having to apply the same fixes in multiple places.

I had a look at the codebase, and I think (oversimplifying) that we could get away with the `if ipfamily` only at allocation stage, making all the rest able to handle a slice of ips.

This PR is being dragged for over a year now, so I'd understand you not willing to continue nor to refactor, but since you started this effort I thought it was fair to ask you first.

Thanks!

uablrek:
Hi @fedepaol,

> This PR is being dragged for over a year now, so I'd understand you not willing to continue nor to refactor, but since you started this effort I thought it was fair to ask you first.

Thanks for asking. And no, I don't want to refactor this PR.

When I submitted the PR we were working on dual-stack and to work with this was a natural assignment for me. As time went by we stayed with the fall-back to have 2 services. Now I have other priorities and to refactor the code (and add unit and e2e tests) is a large job so I am probably not allowed to spent the necessary time on it.

To refactor the code, types and apis must be updated in very many places. I think that kind of major update is better done by metallb maintainers.

If you decide to abandon the requirement that both ipv4 and ipv6 addresses assigned to a service must belong to the same pool, you must basically re-write the whole address management. But if you keep it, there is some metrics issues like, how to present the number of free addresses if you have 1 ipv4 and a billion ipv6 addresses in a pool.

Feel free to close this PR.

Best Regards,
L Ekman

fedepaol:
> Hi @fedepaol,
> 
> > This PR is being dragged for over a year now, so I'd understand you not willing to continue nor to refactor, but since you started this effort I thought it was fair to ask you first.
> 
> Thanks for asking. And no, I don't want to refactor this PR.
> 
> When I submitted the PR we were working on dual-stack and to work with this was a natural assignment for me. As time went by we stayed with the fall-back to have 2 services. Now I have other priorities and to refactor the code (and add unit and e2e tests) is a large job so I am probably not allowed to spent the necessary time on it.
> 
> To refactor the code, types and apis must be updated in very many places. I think that kind of major update is better done by metallb maintainers.
> 
> If you decide to abandon the requirement that both ipv4 and ipv6 addresses assigned to a service must belong to the same pool, you must basically re-write the whole address management. But if you keep it, there is some metrics issues like, how to present the number of free addresses if you have 1 ipv4 and a billion ipv6 addresses in a pool.
> 

Thanks for the hints. Especially the metrics one, it's something I did not think about. First thing comes into my mind is to extend the metrics to export the # of v4 / v6 addresses. Will think about it.

> Feel free to close this PR.
> 
> Best Regards, L Ekman

Thanks for taking the time to reply!

PR #730, opened 2020-09-23
layer2: Send GARPs on cluster membership changes

### LINKED ISSUE #584: L2 mode doesn't resend gratuitous arp when node is diconnected and reconnected

russellb:
There have been multiple patches looking at alternatives for addressing
this issue, most recently #709.

Speaker ownership of IPs is recalculated each time membership of the
cluster changes. When a new speaker is chosen, that speaker will send
a set of gratuitous ARP or NDP messages to inform the network of the
new location for that IP address.

A problem existed where the cluster could end up in an inconsistent
state if a speaker lost ownership of an IP but didn't realize it.
In the meantime, a new speaker will have taken over the IP. When the
original owner rejoins the cluster, it will fail to reclaim the IP
with gratuitous ARP / NDP if it thinks it never lost it in the first
place.

We already force a resync of all services any time we detect a cluster
membership change via memberlist.  The only time this would result in a
new GARP being sent out is if the node saw this as previously not owning
the IP, but now owning.

This patch is a very simple option for improving this situation.
Instead of only sending a GARP the first time we see an IP in use, send
it every time we resync the Service.

The tradeoff taken with this approach is to increase the consistency of
the network at the cost of more freqently sending GARPs to the network.
It seems possible that there is such thing as "too much" and that some
sort of rate limiting should be done, though that is done here.

Closes #584

### DIFF
--- a/internal/layer2/announcer.go
+++ b/internal/layer2/announcer.go
@@ -192,6 +192,15 @@ func (a *Announce) SetBalancer(name string, ip net.IP) {
 	a.ips[name] = ip
 
 	a.ipRefcnt[ip.String()]++
+
+	// Send a round of gratuitous ARP/NDP, even if the IP was already in use on
+	// this node. We will end up sending these more often than strictly necessary,
+	// but trying to be too clever could leave more opportunity for an inconsistent
+	// state of the network. Note that SetBalancer() is not only called for a new
+	// Service, it's also called any time key changes are detected, either changes
+	// to the Service or changes in this node's view of cluster membership.
+	go a.spam(name)
+
 	if a.ipRefcnt[ip.String()] > 1 {
 		// Multiple services are using this IP, so there's nothing
 		// else to do right now.
@@ -203,9 +212,6 @@ func (a *Announce) SetBalancer(name string, ip net.IP) {
 			a.logger.Log("op", "watchMulticastGroup", "error", err, "ip", ip, "msg", "failed to watch NDP multicast group for IP, NDP responder will not respond to requests for this address")
 		}
 	}
-
-	go a.spam(name)
-
 }
 
 // DeleteBalancer deletes an address from the set of addresses we should announce.


champtar:
You can have a lot of resync in a row, so it's simple but dangerous IMO

russellb:
> You can have a lot of resync in a row, so it's simple but dangerous IMO

Yep, but I'd like to understand what volume is actually considered dangerous.  That requires some estimates on:

1) Number of IPs a node may have in use (in layer2 mode)
2) Frequency of resyncs

and I honestly don't have a good feeling for how frequent we're talking about.  My guess is that if it's frequent enough to be a problem, there's another problem ... but that's a total guess.  I haven't tried to go quantify this in a test setup.

champtar:
I've downed network with fairly low level of multicast or IGMP, ARP is handled by the switch CPU, not the asic, so we need to be careful.
Also the more the network is unstable, the more we spam, that can be a nice snow ball effect.

My implementation is here: https://github.com/metallb/metallb/pull/595/commits/d9922aeab35ef568ffdb264facc4479d74cf2e7c
=> spam 5 sec from last SetBalancer call, with only 1 goroutine so we don't have 20 GARP for 1 IP at the same time. If called again extend the time.
It's more then 2 lines changes but not too complicated either, and more importantly, I don't think we can spam less without the state in k8s api, but yet we spam enough

russellb:
> I've downed network with fairly low level of multicast or IGMP, ARP is handled by the switch CPU, not the asic, so we need to be careful.
> Also the more the network is unstable, the more we spam, that can be a nice snow ball effect.
> 
> My implementation is here: [d9922ae](https://github.com/metallb/metallb/commit/d9922aeab35ef568ffdb264facc4479d74cf2e7c)
> => spam 5 sec from last SetBalancer call, with only 1 goroutine so we don't have 20 GARP for 1 IP at the same time. If called again extend the time.
> It's more then 2 lines changes but not too complicated either, and more importantly, I don't think we can spam less without the state in k8s api, but yet we spam enough

OK, yeah, good point about spamming N times for the same IP during a resync.  Only doing it once per IP in a resync was the primary benefit of my other PR -- #709 

I'll leave this open as a demonstration of one of our choices (obviously with its downsides) and I'll try to do a proper review of your implementation, too.

johananl:
Thanks for taking another stab at this @russellb :+1: 

@rata and I think we shouldn't move forward with this implementation - see [here](https://github.com/metallb/metallb/issues/584#issuecomment-702899515) for the rationale and more context.

russellb:
> Thanks for taking another stab at this @russellb +1
> 
> @rata and I think we shouldn't move forward with this implementation - see [here](https://github.com/metallb/metallb/issues/584#issuecomment-702899515) for the rationale and more context.

sounds good.  It was obviously a trivial PR to post, but wanted to help demonstrate the range of options we have.  :+1:

PR #736, opened 2020-10-02
Rework how gratuitous advertisements (ARP/NDP) work

### LINKED ISSUE #584: L2 mode doesn't resend gratuitous arp when node is diconnected and reconnected

champtar:
With our current approach, we only do gratuitous advertisements on the first
SetBalancer() call, and we don't resend any gratuitous advertisements on the next
calls to reduce the amount of "spam" in the network.
This was working pretty well when we were using K8S API to do all the decisions.
Now that we are also using MemberList status, our decisions are based on eventually consistent
information, and there are at least 2 cases where we need to resend gratuitous advertisements
even if the information that we have makes us think there were no changes in ownership:

1) Split brain with no ownership changes:
3 nodes A B C, A owns the LoadBalancer IP I, cluster is clean.
Now for some reason C can't talk to A and B anymore, and our algorithm
in ShouldAnnounce() continues to pick A as the owner of I.
As there were no changes for A, A doesn't send any gratuitous advertisement.
As C thinks it is alone, it thinks it owns I and sends gratuitous advertisements.
Some seconds later C rejoins A & B, C stops sending gratuitous advertisements,
but A continues to be the owner and doesn't send any gratuitous advertisement.
Depending on the switches' inner working, traffic might continue to go to C for a long time.

2) Race condition on ForceSync:
3 nodes A B C, A owns the LoadBalancer IP I, cluster is clean.
A becomes really slow (cpu limits or ...) and memberlist on B and C decides that A is not part of the memberlist cluster
anymore. B and C each start a ForceSync(), one of B or C becomes the owner of I and starts gratuitous advertisements for I.
A starts to respond again to memberlist and rejoins the cluster, while doing its first ForceSync().
A thinks it was always the owner of I and doesn't send any gratuitous advertisement.

The idea of this patch is to send gratuitous advertissements for 5 seconds from the last SetBalancer() call,
instead of the last time we think we became the owner.
To ensure there is only 1 sender for each IP, we use only one goroutine for all gratuitous advertisement calls.
As gratuitous() was using Lock() (ie exclusive lock), we were sending at most 1 gratuitous advertisement at a time,
so we know that this is fine performance wise, but it might be burstier than before.

This fixes #584

### DIFF
--- a/internal/layer2/announcer.go
+++ b/internal/layer2/announcer.go
@@ -20,6 +20,10 @@ type Announce struct {
 	ndps     map[int]*ndpResponder
 	ips      map[string]net.IP // svcName -> IP
 	ipRefcnt map[string]int    // ip.String() -> number of uses
+
+	// This channel can block - do not write to it while holding the mutex
+	// to avoid deadlocking.
+	spamCh chan net.IP
 }
 
 // New returns an initialized Announce.
@@ -30,8 +34,10 @@ func New(l log.Logger) (*Announce, error) {
 		ndps:     map[int]*ndpResponder{},
 		ips:      map[string]net.IP{},
 		ipRefcnt: map[string]int{},
+		spamCh:   make(chan net.IP, 1024),
 	}
 	go ret.interfaceScan()
+	go ret.spamLoop()
 
 	return ret, nil
 }
@@ -131,24 +137,52 @@ func (a *Announce) updateInterfaces() {
 	return
 }
 
-func (a *Announce) spam(name string) {
-	// TODO: should abort if we lose control of the IP mid-spam.
-	start := time.Now()
-	for time.Since(start) < 5*time.Second {
-		if err := a.gratuitous(name); err != nil {
-			a.logger.Log("op", "gratuitousAnnounce", "error", err, "service", name, "msg", "failed to make gratuitous IP announcement")
+func (a *Announce) spamLoop() {
+	// Map IP to spam stop time.
+	m := map[string]time.Time{}
+	// See https://github.com/metallb/metallb/issues/172 for the 1100 choice.
+	ticker := time.NewTicker(1100 * time.Millisecond)
+	for {
+		select {
+		case ip := <-a.spamCh:
+			ipStr := ip.String()
+			_, ok := m[ipStr]
+			// Set spam stop time to 5 seconds from now.
+			m[ipStr] = time.Now().Add(5 * time.Second)
+			if !ok {
+				// Spam right away to avoid waiting up to 1100 milliseconds even if
+				// it means we call spam() twice in a row in a short amount of time.
+				a.spam(ip)
+			}
+		case now := <-ticker.C:
+			for ipStr, until := range m {
+				if now.After(until) {
+					// We have spammed enough - remove the IP from the map.
+					delete(m, ipStr)
+				} else {
+					a.spam(net.ParseIP(ipStr))
+				}
+			}
 		}
-		time.Sleep(1100 * time.Millisecond)
 	}
 }
 
-func (a *Announce) gratuitous(name string) error {
-	a.Lock()
-	defer a.Unlock()
+func (a *Announce) doSpam(ip net.IP) {
+	a.spamCh <- ip
+}
 
-	ip, ok := a.ips[name]
-	if !ok {
-		// No IP means we've lost control of the IP, someone else is
+func (a *Announce) spam(ip net.IP) {
+	if err := a.gratuitous(ip); err != nil {
+		a.logger.Log("op", "gratuitousAnnounce", "error", err, "ip", ip, "msg", "failed to make gratuitous IP announcement")
+	}
+}
+
+func (a *Announce) gratuitous(ip net.IP) error {
+	a.RLock()
+	defer a.RUnlock()
+
+	if a.ipRefcnt[ip.String()] <= 0 {
+		// We've lost control of the IP, someone else is
 		// doing announcements.
 		return nil
 	}
@@ -181,6 +215,8 @@ func (a *Announce) shouldAnnounce(ip net.IP) dropReason {
 
 // SetBalancer adds ip to the set of announced addresses.
 func (a *Announce) SetBalancer(name string, ip net.IP) {
+	// Call doSpam at the end of the function without holding the lock
+	defer a.doSpam(ip)
 	a.Lock()
 	defer a.Unlock()
 
@@ -203,9 +239,6 @@ func (a *Announce) SetBalancer(name string, ip net.IP) {
 			a.logger.Log("op", "watchMulticastGroup", "error", err, "ip", ip, "msg", "failed to watch NDP multicast group for IP, NDP responder will not respond to requests for this address")
 		}
 	}
-
-	go a.spam(name)
-
 }
 
 // DeleteBalancer deletes an address from the set of addresses we should announce.
--- a/internal/layer2/announcer_test.go
+++ b/internal/layer2/announcer_test.go
@@ -9,6 +9,7 @@ func Test_SetBalancer_AddsToAnnouncedServices(t *testing.T) {
 	announce := &Announce{
 		ips:      map[string]net.IP{},
 		ipRefcnt: map[string]int{},
+		spamCh:   make(chan net.IP, 1),
 	}
 
 	services := []struct {
@@ -27,6 +28,8 @@ func Test_SetBalancer_AddsToAnnouncedServices(t *testing.T) {
 
 	for _, service := range services {
 		announce.SetBalancer(service.name, service.ip)
+		// We need to empty spamCh as spamLoop() is not started.
+		<-announce.spamCh
 
 		if !announce.AnnounceName(service.name) {
 			t.Fatalf("service %v is not anounced", service.name)


champtar:
@russellb if you want to have a look

russellb:
> @russellb if you want to have a look

thanks!  Can you explain a bit more in the description how this fixes #584?  The current description and commit message just talks about the lock change.

champtar:
@russellb added way more explanations, tell me if this is enough

internal/layer2/announcer.go
```diff
@@ -181,6 +225,8 @@ func (a *Announce) shouldAnnounce(ip net.IP) dropReason {
 
 // SetBalancer adds ip to the set of announced addresses.
 func (a *Announce) SetBalancer(name string, ip net.IP) {
+	// Call spamStart at the end of the function without holding the lock
+	defer a.spamStart(name)
```
russellb, line 229: It looks like these changes make it so we're sending GARP for each service using the same IP, bypassing the logic that used to skip it when we detected it was the same IP in use by more than one service.  Or if that's handled, can you help with what I missed?

champtar: Good catch, indeed I spam once per name and not per IP, I'll change that

champtar: Done, it even simplified a bit the code in the end

champtar:
Note from my testing:
```
W1008 00:03:37.209528       1 reflector.go:302] pkg/mod/k8s.io/client-go@v0.0.0-20190620085101-78d2af792bab/tools/cache/reflector.go:98: watch of *v1.ConfigMap ended with: too old resource version: 4356503 (4358042)
{"caller":"main.go:387","configmap":"metallb-system/config","event":"startUpdate","msg":"start of config update","ts":"2020-10-08T00:03:38.213385661Z"}
{"caller":"main.go:411","configmap":"metallb-system/config","event":"endUpdate","msg":"end of config update","ts":"2020-10-08T00:03:38.213495356Z"}
{"caller":"k8s.go:421","configmap":"metallb-system/config","event":"configLoaded","msg":"config (re)loaded","ts":"2020-10-08T00:03:38.213521216Z"}
{"caller":"main.go:268","event":"startUpdate","msg":"start of service update"
...
```
On each config watch expiry, we do a resync, so we spam

russellb:
> Note from my testing:
> 
> ```
> W1008 00:03:37.209528       1 reflector.go:302] pkg/mod/k8s.io/client-go@v0.0.0-20190620085101-78d2af792bab/tools/cache/reflector.go:98: watch of *v1.ConfigMap ended with: too old resource version: 4356503 (4358042)
> {"caller":"main.go:387","configmap":"metallb-system/config","event":"startUpdate","msg":"start of config update","ts":"2020-10-08T00:03:38.213385661Z"}
> {"caller":"main.go:411","configmap":"metallb-system/config","event":"endUpdate","msg":"end of config update","ts":"2020-10-08T00:03:38.213495356Z"}
> {"caller":"k8s.go:421","configmap":"metallb-system/config","event":"configLoaded","msg":"config (re)loaded","ts":"2020-10-08T00:03:38.213521216Z"}
> {"caller":"main.go:268","event":"startUpdate","msg":"start of service update"
> ...
> ```
> 
> On each config watch expiry, we do a resync, so we spam

how often does that happen?

internal/layer2/announcer.go
```diff
@@ -131,24 +137,49 @@ func (a *Announce) updateInterfaces() {
...
-func (a *Announce) spam(name string) {
-	// TODO: should abort if we lose control of the IP mid-spam.
-	start := time.Now()
-	for time.Since(start) < 5*time.Second {
-		if err := a.gratuitous(name); err != nil {
-			a.logger.Log("op", "gratuitousAnnounce", "error", err, "service", name, "msg", "failed to make gratuitous IP announcement")
+func (a *Announce) spamLoop() {
+	m := map[string]time.Time{}
+	// see https://github.com/metallb/metallb/issues/172 for the 1100 choice
+	ticker := time.NewTicker(1100 * time.Millisecond)
```
russellb, line 143: It would be nice to turn this off when there's no active spamming happening.  Otherwise when it should be otherwise idle, it's still waking up every 1.1s for no reason.

champtar: might be doable with ticker.Stop() and ticker.Reset(), will have a look

internal/layer2/announcer.go
```diff
@@ -131,24 +137,49 @@ func (a *Announce) updateInterfaces() {
...
+	for {
+		select {
+		case ip := <-a.spamCh:
+			ips := ip.String()
```
johananl, line 147: I find `ips` confusing as it sounds like the var holds multiple IP addresses.

I suggest doing the following:

```go
case e := <-a.spamCh:
    ip := e.String()
```

internal/layer2/announcer.go
```diff
@@ -20,6 +20,10 @@ type Announce struct {
 	ndps     map[int]*ndpResponder
 	ips      map[string]net.IP // svcName -> IP
 	ipRefcnt map[string]int    // ip.String() -> number of uses
+
+	// channel can block so do not produce to spamCh while holding the Mutex
+	// else you risk causing a dead lock
```
johananl, line 25:
```suggestion
	// This channel can block - do not write to it while holding the mutex
	// to avoid deadlocking.
```

- We should format the comments according to the existing style.
- Improve phrasing.

internal/layer2/announcer.go
```diff
@@ -131,24 +137,49 @@ func (a *Announce) updateInterfaces() {
...
 
-func (a *Announce) spam(name string) {
-	// TODO: should abort if we lose control of the IP mid-spam.
-	start := time.Now()
-	for time.Since(start) < 5*time.Second {
-		if err := a.gratuitous(name); err != nil {
-			a.logger.Log("op", "gratuitousAnnounce", "error", err, "service", name, "msg", "failed to make gratuitous IP announcement")
+func (a *Announce) spamLoop() {
+	m := map[string]time.Time{}
+	// see https://github.com/metallb/metallb/issues/172 for the 1100 choice
```
johananl, line 142:
```suggestion
	// See https://github.com/metallb/metallb/issues/172 for the 1100 choice.
```

internal/layer2/announcer.go
```diff
@@ -131,24 +137,49 @@ func (a *Announce) updateInterfaces() {
...
 }
 
-func (a *Announce) spam(name string) {
-	// TODO: should abort if we lose control of the IP mid-spam.
-	start := time.Now()
-	for time.Since(start) < 5*time.Second {
-		if err := a.gratuitous(name); err != nil {
-			a.logger.Log("op", "gratuitousAnnounce", "error", err, "service", name, "msg", "failed to make gratuitous IP announcement")
+func (a *Announce) spamLoop() {
+	m := map[string]time.Time{}
```
johananl, line 141:
```suggestion
	// Map IP to spam stop time.
	m := map[string]time.Time{}
```

internal/layer2/announcer.go
```diff
@@ -131,24 +137,49 @@ func (a *Announce) updateInterfaces() {
...
+	// see https://github.com/metallb/metallb/issues/172 for the 1100 choice
+	ticker := time.NewTicker(1100 * time.Millisecond)
+	for {
+		select {
+		case ip := <-a.spamCh:
+			ips := ip.String()
+			_, ok := m[ips]
+			m[ips] = time.Now().Add(5 * time.Second)
```
johananl, line 149:
```suggestion
			// Set spam stop time to 5 seconds from now.
			m[ips] = time.Now().Add(5 * time.Second)
```

internal/layer2/announcer.go
```diff
@@ -131,24 +137,49 @@ func (a *Announce) updateInterfaces() {
...
+			if !ok {
+				// The first 2 spam can be close
+				// but we don't want to wait 1.1 sec
```
johananl, line 152: - I don't understand this comment. Could you please make it clearer?

internal/layer2/announcer.go
```diff
@@ -131,24 +137,49 @@ func (a *Announce) updateInterfaces() {
...
+				a.spam(ip)
+			}
+		case now := <-ticker.C:
+			for ips, until := range m {
+				if now.After(until) {
+					delete(m, ips)
```
johananl, line 158:
```suggestion
					// We have spammed enough - remove the IP from the map.
					delete(m, ips)
```

internal/layer2/announcer.go
```diff
@@ -181,6 +212,8 @@ func (a *Announce) shouldAnnounce(ip net.IP) dropReason {
 
 // SetBalancer adds ip to the set of announced addresses.
 func (a *Announce) SetBalancer(name string, ip net.IP) {
+	// Call spamStart at the end of the function without holding the lock
+	defer a.spamStart(ip)
```
johananl, line 216: How about the following?

```go
	defer func() {
		a.spamCh <- ip
	}()
```

We can then get rid of `spamStart()`. In any case, if we keep this function I'd rather name it `doSpam()` since having "start" in the name implies that there is also "stop", which isn't the case.

internal/layer2/announcer_test.go
```diff
@@ -27,6 +28,8 @@ func Test_SetBalancer_AddsToAnnouncedServices(t *testing.T) {
 
 	for _, service := range services {
 		announce.SetBalancer(service.name, service.ip)
+		//we need to empty spamCh
```
johananl, line 31:
```suggestion
		// We need to empty spamCh.
```

If you comment here I'd explain *why* we need to empty the channel.

internal/layer2/announcer.go
```diff
@@ -30,8 +34,10 @@ func New(l log.Logger) (*Announce, error) {
 		ndps:     map[int]*ndpResponder{},
 		ips:      map[string]net.IP{},
 		ipRefcnt: map[string]int{},
+		spamCh:   make(chan net.IP, 1024),
```
johananl, line 37: What happens if the channel is full? What is the implication for the user?

johananl (commented):
LGTM overall. The design looks OK to me and things seem to work on my testing.

Added some nitpicks and questions, mainly around comments.

internal/layer2/announcer.go
champtar: ticker.Reset() is golang 1.15, do we want to bump minimum version to latest and greatest ? (even my Fedora only has 1.14)

internal/layer2/announcer.go
champtar: merged

internal/layer2/announcer.go
champtar: I went with `ipStr` because I also use `ip` in `a.spam(ip)`

internal/layer2/announcer.go
champtar: Is it better ?
```
// The first 2 calls to spam() can be msec away,
// but we can't wait 1.1 sec
```

internal/layer2/announcer.go
champtar: I went with `doSpam()`, just a matter of taste, but if other people prefer the anonymous func i'll change it

internal/layer2/announcer_test.go
champtar: Changed:
// We need to empty spamCh as spamLoop() is not started.

internal/layer2/announcer.go
champtar: If `spamCh` is full `SetBalancer` blocks until we are done spamming in `spamLoop` and empty `spamCh` again
I expect people to have less than 100 layer2 VIP per cluster so 1024 should be good. If I'm wrong full resync will block for a bit which is fine.

champtar:
@johananl thanks a lot for the review, see my answers

champtar:
BTW to speed up thing I'll implement an allow/deny list based on interface name, so we don't spam on each container interface when using calico or others

russellb:
> BTW to speed up thing I'll implement an allow/deny list based on interface name, so we don't spam on each container interface when using calico or others. I've tried in the past to detect veth type but it's really not easy

that's not really related to this change though, right?

champtar:
@russellb yes in a separate PR, at some point in the future

internal/layer2/announcer.go
johananl: I still don't understand the technical meaning behind "but we can't wait 1.1 sec" :confused: What are we trying to say here?

Anyway, let's not block on this. I'd just change `msec` to `milliseconds` and add a period at the end for consistency.

johananl (approved):
LGTM.

internal/layer2/announcer.go
champtar: If the ticker is already running (always the case right now), we might call spam() because it's a new IP, and 1msec after call spam() again in the ticker part. This is not ideal, but the worst case, where we wait 1099 ms for the first spam() call is not acceptable.

johananl: I *think* I understand what you mean now. Can I rephrase the comment to the following?

```
				// Spam right away to avoid waiting 1100 milliseconds even if
				// it means we call spam() twice in a row in a short amount of
				// time.
```

champtar: Done
```
// Spam right away to avoid waiting up to 1100 milliseconds even if
// it means we call spam() twice in a row in a short amount of time.
```

internal/layer2/announcer.go
champtar: I'll open a issue as soon as we merge this PR so we don't forget

PR #746, opened 2020-10-15
Always pass namespace as METALLB_NAMESPACE

champtar:
For the speakers we had
- config-ns for the configmap
- ml-namespace for the speaker namespace
the memberlist secret is taken from the speakers namespace (secretKeyRef)

For the controller we also had config-ns

As we will need to have the secret namespace in the controller,
and we don't really need the extra complexity, just keep one namespace
setting 'namespace'/METALLB_NAMESPACE for both the speakers and the controller


gclawes:
@johananl we should consider this for v0.10.0.  I have included these env var changes in the helm chart.

champtar:
Those commits are included in #747

PR #747, opened 2020-10-15
Make the controller create memberlist secret

champtar:
We name the role secret-updater as the controller will
rotate the secret at some point

### DIFF
--- a/controller/main.go
+++ b/controller/main.go
@@ -17,6 +17,7 @@ package main
 import (
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"reflect"
 
@@ -127,12 +128,21 @@ func main() {
 	var (
 		port       = flag.Int("port", 7472, "HTTP listening port for Prometheus metrics")
 		config     = flag.String("config", "config", "Kubernetes ConfigMap containing MetalLB's configuration")
-		configNS   = flag.String("config-ns", "", "config file namespace (only needed when running outside of k8s)")
+		namespace  = flag.String("namespace", os.Getenv("METALLB_NAMESPACE"), "config file namespace")
 		kubeconfig = flag.String("kubeconfig", "", "absolute path to the kubeconfig file (only needed when running outside of k8s)")
 	)
 	flag.Parse()
 
 	logger.Log("version", version.Version(), "commit", version.CommitHash(), "branch", version.Branch(), "msg", "MetalLB controller starting "+version.String())
+
+	if *namespace == "" {
+		bs, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
+		if err != nil {
+			logger.Log("op", "startup", "msg", "Unable to get namespace from pod service account data, please specify --namespace or METALLB_NAMESPACE", "error", err)
+			os.Exit(1)
+		}
+		*namespace = string(bs)
+	}
 
 	c := &controller{
 		ips: allocator.New(),
@@ -141,7 +151,7 @@ func main() {
 	client, err := k8s.New(&k8s.Config{
 		ProcessName:   "metallb-controller",
 		ConfigMapName: *config,
-		ConfigMapNS:   *configNS,
+		ConfigMapNS:   *namespace,
 		MetricsPort:   *port,
 		Logger:        logger,
 		Kubeconfig:    *kubeconfig,
--- a/controller/main.go
+++ b/controller/main.go
@@ -128,8 +128,10 @@ func main() {
 	var (
 		port       = flag.Int("port", 7472, "HTTP listening port for Prometheus metrics")
 		config     = flag.String("config", "config", "Kubernetes ConfigMap containing MetalLB's configuration")
-		namespace  = flag.String("namespace", os.Getenv("METALLB_NAMESPACE"), "config file namespace")
+		namespace  = flag.String("namespace", os.Getenv("METALLB_NAMESPACE"), "config / memberlist secret namespace")
 		kubeconfig = flag.String("kubeconfig", "", "absolute path to the kubeconfig file (only needed when running outside of k8s)")
+		mlSecret   = flag.String("ml-secret-name", os.Getenv("METALLB_ML_SECRET_NAME"), "name of the memberlist secret to create")
+		deployName = flag.String("deployment", os.Getenv("METALLB_DEPLOYMENT"), "name of the MetalLB controller Deployment")
 	)
 	flag.Parse()
 
@@ -163,6 +165,14 @@ func main() {
 	if err != nil {
 		logger.Log("op", "startup", "error", err, "msg", "failed to create k8s client")
 		os.Exit(1)
+	}
+
+	if *mlSecret != "" {
+		err = client.CreateMlSecret(*namespace, *deployName, *mlSecret)
+		if err != nil {
+			logger.Log("op", "startup", "error", err, "msg", "failed to create memberlist secret")
+			os.Exit(1)
+		}
 	}
 
 	c.client = client
--- a/internal/k8s/k8s.go
+++ b/internal/k8s/k8s.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io/ioutil"
 	"net/http"
 
 	"go.universe.tf/metallb/internal/config"
@@ -115,15 +114,6 @@ func New(cfg *Config) (*Client, error) {
 		return nil, fmt.Errorf("creating Kubernetes client: %s", err)
 	}
 
-	namespace := cfg.ConfigMapNS
-	if namespace == "" {
-		bs, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
-		if err != nil {
-			return nil, fmt.Errorf("getting namespace from pod service account data: %s", err)
-		}
-		namespace = string(bs)
-	}
-
 	broadcaster := record.NewBroadcaster()
 	broadcaster.StartRecordingToSink(&v1core.EventSinkImpl{Interface: v1core.New(clientset.CoreV1().RESTClient()).Events("")})
 	recorder := broadcaster.NewRecorder(scheme.Scheme, v1.EventSource{Component: cfg.ProcessName})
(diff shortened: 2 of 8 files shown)


champtar:
```
# kubectl get pod -n metallb-system -w
NAME                          READY   STATUS                       RESTARTS   AGE
controller-85676b7db6-4b6cd   1/1     Running                      0          7s
speaker-44ft6                 0/1     CreateContainerConfigError   0          7s
speaker-5f5lv                 1/1     Running                      0          7s
speaker-cb2pq                 0/1     CreateContainerConfigError   0          7s
speaker-tf2x2                 0/1     CreateContainerConfigError   0          7s
speaker-44ft6                 1/1     Running                      0          15s
speaker-cb2pq                 1/1     Running                      0          18s
speaker-tf2x2                 1/1     Running                      0          19s
```
If the speakers create faster than the controller, you can see a CreateContainerConfigError as the secret doesn't exists yet.
This was already the case I think

gclawes:
Is `METALLB_NAMESPACE` necessary? Can the controller get the current namespace from the API?

champtar:
It's configurable to be able to develop outside k8s mostly. We can get the current namespace from `/var/run/secrets/kubernetes.io/serviceaccount/namespace`, but there is no concept of current namespace with a token, the current namespace is somewhere in conf

champtar:
namespace is not part of `rest.Config`

gclawes:
Is `METALLB_ML_SECRET_KEY` still necessary if the values is hard-coded into the controller?  Or should the controller respect that value?

gclawes:
Ah, my mistake, `METALLB_ML_SECRET_KEY` is the actual cryptographic secret, not the keyname within the Secret object.

gclawes:
Does it make sense to name the new Role/Rolebinding `secret-creator` to describe what it grants, not what it is.  This lines up with the others, `config-watcher` and `pod-lister`.  I can do the same in the helm chart in #658

champtar:
> Does it make sense to name the new Role/Rolebinding `secret-creator` to describe what it grants, not what it is. This lines up with the others, `config-watcher` and `pod-lister`. I can do the same in the helm chart in #658

Problem is that you can't easily delete Role/Rolebinding on upgrade with the manifest, and in one of my PR I remove `pod-lister`, so to not have this problem again I'm went with `controller` name.
I agree it makes more sense so that you can disable it when memberlist is not enable.

The way to delete object is just to have 1 extra yaml, which is not that bad, but it was not to the taste of everyone
https://github.com/metallb/metallb/pull/595/files#diff-bb1f1dd7212eccbf6a50285a0f8562147324a060ff5e3adf10605ed3b9bf45e8R64
https://github.com/metallb/metallb/pull/595/files#diff-7b27766618d6f7c1c3b5b1cc59f8df158609035a08def0c59492f2e340ad9148R11

internal/k8s/k8s.go
```diff
@@ -259,6 +251,58 @@ func New(cfg *Config) (*Client, error) {
...
+
+	// Create the K8S Secret object.
+	_, err = c.client.CoreV1().Secrets(namespace).Create(&v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: secretName,
+			OwnerReferences: []metav1.OwnerReference{{
+				// d.APIVersion is empty
+				APIVersion: "apps/v1",
+				// d.Kind is empty
+				Kind: "Deployment",
```
champtar, line 293: I set the controller deployment as owner because it's the one that creates it, but we could also set the speakers daemonset as owner as they are the user of it

champtar:
rebased but not retested (just had to update the create call following client-go update)

controller/main.go
```diff
@@ -127,21 +127,28 @@ func main() {
 	var (
 		port       = flag.Int("port", 7472, "HTTP listening port for Prometheus metrics")
 		config     = flag.String("config", "config", "Kubernetes ConfigMap containing MetalLB's configuration")
-		configNS   = flag.String("config-ns", "", "config file namespace (only needed when running outside of k8s)")
+		namespace  = flag.String("namespace", os.Getenv("METALLB_NAMESPACE"), "config / memberlist secret namespace")
```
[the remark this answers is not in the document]
champtar: Yes it does, but we never said to just update the image, you must always update the manifest, even more so for a major version.
The failure will be pretty visible so I don't think this is a problem, but I'm open for discussion.

speaker/main.go
```diff
@@ -64,22 +64,26 @@ func main() {
...
+		config     = flag.String("config", "config", "Kubernetes ConfigMap containing MetalLB's configuration")
+		namespace  = flag.String("namespace", os.Getenv("METALLB_NAMESPACE"), "config file and speakers namespace")
+		kubeconfig = flag.String("kubeconfig", "", "absolute path to the kubeconfig file (only needed when running outside of k8s)")
+		host       = flag.String("host", os.Getenv("METALLB_HOST"), "HTTP host address")
+		mlBindAddr = flag.String("ml-bindaddr", os.Getenv("METALLB_ML_BIND_ADDR"), "Bind addr for MemberList (fast dead node detection)")
+		mlBindPort = flag.String("ml-bindport", os.Getenv("METALLB_ML_BIND_PORT"), "Bind port for MemberList (fast dead node detection)")
+		mlLabels   = flag.String("ml-labels", os.Getenv("METALLB_ML_LABELS"), "Labels to match the speakers (for MemberList / fast dead node detection)")
+		mlSecret   = flag.String("ml-secret-key", os.Getenv("METALLB_ML_SECRET_KEY"), "Secret key for MemberList (fast dead node detection)")
+		myNode     = flag.String("node-name", os.Getenv("METALLB_NODE_NAME"), "name of this Kubernetes node (spec.nodeName)")
+		port       = flag.Int("port", 7472, "HTTP listening port")
```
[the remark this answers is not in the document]
champtar: rebase on top of #826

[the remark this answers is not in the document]
champtar: (i'll re-rebase as soon as merged)

controller/main.go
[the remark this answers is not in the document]
russellb: I think it's still worth calling out in the release notes if we're going to change it.  See #826 where I propose we start capturing notes along the way.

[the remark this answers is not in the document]
champtar: rebased and release note added

website/content/release-notes/_index.md
```diff
@@ -12,6 +12,10 @@ Changes in behavior:
   included with MetalLB, but the binary itself previously defaulted to port
   `80`.
 
+- The `config-ns` option of both the `controller` ad the `speaker` and the `ml-namespace`
+  option of the `speaker` are replaced by the `namespace`/`METALLB_NAMESPACE` option.
```
russellb, line 16:
```suggestion
  option of the `speaker` are replaced by the `namespace` option or the `METALLB_NAMESPACE` environment variable.
```

russellb: and this should note that the `METALLB_ML_NAMESPACE` env var was removed.

internal/k8s/k8s.go
```diff
@@ -115,15 +116,6 @@ func New(cfg *Config) (*Client, error) {
...
 
-	namespace := cfg.ConfigMapNS
-	if namespace == "" {
-		bs, err := ioutil.ReadFile("/var/run/secrets/kubernetes.io/serviceaccount/namespace")
-		if err != nil {
-			return nil, fmt.Errorf("getting namespace from pod service account data: %s", err)
-		}
-		namespace = string(bs)
-	}
-
```
russellb, line 126: Why was this removed?  I believe this means that previously, this would work even if no namespace was provided, but now we have a hard requirement on it being provided through the manifest.  Is there any harm in leaving this code in place?

champtar: I can move this code if you prefer, I personally prefer explicit conf in the manifest that's all

champtar: ok moved it, not tested but it should work, please have a new look

russellb: looks ok, but would like to hear test confirmation first.  thanks

russellb: actually this was easy to test with dev-env since you also removed setting the env var from the manifest

manifests/metallb.yaml
```diff
@@ -369,6 +411,11 @@ spec:
       - args:
         - --port=7472
         - --config=config
+        env:
+        - name: METALLB_ML_SECRET_NAME
+          value: memberlist
+        - name: METALLB_DEPLOYMENT
+          value: controller
```
russellb, line 418: I'd rather move these defaults into the code.  Otherwise, these should be documented as more changes in behavior.  I think it's easier to just make the controller do the right thing.

champtar: `METALLB_DEPLOYMENT` is not `controller` when using helm if I remember correctly, so I would say no for this one

empty / missing `METALLB_ML_SECRET_NAME` disable secret creation, so if I remove it it's disabled which makes more sense than adding `METALLB_ML_SECRET_NAME = ""` to disable it

Let's me document them

champtar: see https://github.com/metallb/metallb/pull/658/files#diff-829c7adc7abb37fd2f497867860e411c3fb4d73e15b62ce283544ba2f88f02c6R53-R58

champtar: improved the release notes a bit

russellb: You should be able to discover the pod identity, right?  Then the pod should have an owner reference to its deployment.

Also there is a SetOwnerReference() helper here we could use, but I don't see controller-runtime in use in metallb, so it would be a new dependency: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#SetControllerReference

champtar: I remember now, I could from `metadata.uid` have the replication controller, and from the replication controller have the deployment, so it's easier to just go to the deployment, and allow targeted RBAC (only get deployment named controller)

russellb: I'm not sure I understand.  I think you're saying you found a way to get it without this env var, but you found the env var easier?

champtar: No, I would still need an env var to get `metadata.uid`, so it's trading an env var for another one (and wider RBAC)

russellb (approved):

PR #764, opened 2020-10-24
Remove NET_ADMIN and SYS_ADMIN capabilities

### LINKED ISSUE #762: Test if we need more than CAP_NET_RAW
https://github.com/metallb/metallb/blob/main/manifests/metallb.yaml#L330
NET_RAW is enough to listen to and send arbitrary traffic
NET_ADMIN / SYS_ADMIN give unnecessary capabilities for me

champtar:
Those capabilities are not needed and allow to break out of the container

This was introduced in a18378a3359284da08acdae4cdb134f8376019ed / #382
Fixes #762

### DIFF
--- a/manifests/metallb.yaml
+++ b/manifests/metallb.yaml
@@ -50,9 +50,7 @@ metadata:
 spec:
   allowPrivilegeEscalation: false
   allowedCapabilities:
-  - NET_ADMIN
   - NET_RAW
-  - SYS_ADMIN
   allowedHostPaths: []
   defaultAddCapabilities: []
   defaultAllowPrivilegeEscalation: false
@@ -327,9 +325,7 @@ spec:
           allowPrivilegeEscalation: false
           capabilities:
             add:
-            - NET_ADMIN
             - NET_RAW
-            - SYS_ADMIN
             drop:
             - ALL
           readOnlyRootFilesystem: true


daxmc99 (approved):

daxmc99:
Agree we should remove these permissions as they have been linked to some possible container breakouts

invidian (approved):
Tested in L2 mode on my cluster, things seem to work just fine :+1:

rata:
Can someone verify this is not needed for BGP in any case either? It is of course not cool to merge without that

champtar:
> Can someone verify this is not needed for BGP in any case either? It is of course not cool to merge without that

I confirmed that TCP_MD5SIG doesn't require any special capabilities with a small C program https://github.com/metallb/metallb/issues/762#issuecomment-716082533
Also BGP MD5 is supported since 0.6.0, and those permissions have been added for no good reasons in 0.8.0, so it was working without those

champtar:
ping @rata

johanfleury:
Just a follow up on this PR: I tested running without `CAP_NET_ADMIN` and `CAP_SYS_ADMIN` and I can confirm that BGP keeps working.

I also tried setting up an authenticated session but for some reasons 2 out of 3 speakers couldn’t connect. My guess is that my firewall is filtering these messages, but I cannot confirm from the host’s logs. However, I’ve checked metallb’s logs and saw nothing unusual.

champtar:
If 1 connected that is enough for me ;)
@rata @johananl can we merge this ?

rata:
@champtar @johanfleury thanks!

@johanfleury do you see the same issue connecting without this patch too? Or do you have more insight about the issue?

johanfleury:
So I can confirm that after changing a rule in my firewall, I was able to get TCP MD5 working the same way with and without `SYS_ADMIN` or `NET_ADMIN`.

I was already pretty sure the issue was not within metallb and this confirmed it.

Just for the sake of clarity, I’m not using any fancy CNI plugins like calico or whatever (just bridge and host-local) so I setup masquerading rules by myself. When I made the following change, everything worked as expected:

```diff
-iptables -nat -A POSTROUTING -o eth0 ! -d {cluster CIDR} -j MASQUERADE
+iptables -nat -A POSTROUTING -o eth0 -s {cluster CIDR} ! -d {cluster CIDR} -j MASQUERADE
```

I then changed the PSP and the speaker’s daemonset to add (and then remove) `SYS_ADMIN` and `NET_ADMIN` and session went up both times.

I think the issue is with iptables masquerading. Using `tcpdump -i eth0 -M {password}` I could see `md5 invalid` before the change and `md5 valid` after.

I didn’t dig further as I won’t be using this feature, and it’s not directly related to metallb.

champtar:
masquerade likely changes the tcp header (source port), so the signature becomes invalid, everything working as expected then

champtar:
Multiple people have validated this change, can someone click merge ?

russellb:
Thank you for the successful test reports!  very helpful.

Issue #696, opened 2020-08-13
Move images to quay.io


rata:
Just received an email from dockerhub. More info here: https://www.docker.com/pricing/retentionfaq
>
>    The container image retention policy will apply to the following plans:
>    * Free plans will have a 6 month image retention limit
>    * Pro and Team plans will have unlimited image retention
>
>    What is an “inactive” image?
>    An inactive image is a container image that has not been either pushed or pulled from the image repository in 6 or months.

We might want to consider other alternatives to Dockerhub. Today the paid option in dockerhub seems to be 7 USD per user per month.

cc @daxmc99 @johananl @russellb

johananl:
We've decided to move to quay.io following discussions in the developer meetings.

johananl:
TODO:

- [ ] Add Docker registry argument to `push-multiarch` Invoke [target](https://github.com/metallb/metallb/blob/ca572eca2098bef9005b060228ab431c6d3dee33/tasks.py#L134).
- [ ] Execute `inv push-multiarch` manually to make a test push to quay.io and verify things look sane for all supported architectures. [This](https://github.com/metallb/metallb/blob/ca572eca2098bef9005b060228ab431c6d3dee33/.circleci/config.yml#L39) can help.
- [ ] **Add** two jobs similar to [this](https://github.com/metallb/metallb/blob/ca572eca2098bef9005b060228ab431c6d3dee33/.circleci/config.yml#L26) one (while keeping the existing ones) to automatically push images to quay.io in parallel with Docker Hub.
- [ ]  Monitor things for a while and ensure things look sane on quay.io.
- [ ] Open a PR to point all the manifests at quay.io.
- [ ] Copy all existing images from Docker Hub to quay.io just in case.
- [ ] Retire Docker Hub repositories when we feel safe to do so.

johananl:
I've added the option for specifying a Docker registry. WIP is [here](https://github.com/kinvolk/metallb/tree/johananl/quay).

Got bitten by https://github.com/docker/for-linux/issues/396 when Invoke executes `manifest-tool`. Looking for a fix/workaround that's nicer than asking everybody to do `chmod o+x /etc/docker`.

johananl:
Started copying image from Docker Hub to Quay. We have about 264 images in each of the repositories (`controller` and `speaker`).
[...]

PR #769, opened 2020-10-28
Update CI

johananl:
This PR does the following:

- Add a CI job to publish images to Quay in parallel to the existing Docker Hub job.
- Update the CircleCI images we use for the build environment.
- Clean up old and unused CI code.
- Consolidate controller and speaker publish jobs to a single job to keep things DRY.
- Bump manifest-tool to `v1.0.3`.

Needed for https://github.com/metallb/metallb/issues/696.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,7 +36,7 @@ jobs:
       - run: mkdir -p ./bin
       - run: wget -O ./bin/manifest-tool https://github.com/estesp/manifest-tool/releases/download/v1.0.0/manifest-tool-linux-amd64 && chmod +x ./bin/manifest-tool
       - run: echo "export GO111MODULE=on" >>$BASH_ENV
-      - run: PATH=./bin:$PATH inv push-multiarch --binaries=controller --tag=${CIRCLE_BRANCH:-${CIRCLE_TAG}} --docker-user=metallb
+      - run: PATH=./bin:$PATH inv push-multiarch --binaries=controller --tag=${CIRCLE_BRANCH:-${CIRCLE_TAG}} --repo=metallb
   deploy-speaker:
     working_directory: /go/src/go.universe.tf/metallb
     docker:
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -3,28 +3,21 @@
 version: 2
 jobs:
   test-1.13:
-    working_directory: /go/src/go.universe.tf/metallb
```
rata, line 6: Just curious: if we don't specify the working directory, how does circle CI know from which working directory to run the `go test` command specified below? I guess the default now is to run it from where your code is also copied by circle CI?

.circleci/config.yml
```diff
@@ -3,28 +3,21 @@
...
     docker:
       - image: cimg/go:1.13
     steps:
       - checkout
-      - run: sudo apt-get install python-pip
-      - run: sudo pip install invoke semver pyyaml
-      - run: echo "export GO111MODULE=on" >>$BASH_ENV
```
rata, line 13: Not sure I follow how the `pip` packages and the env var works on the CI without this. How does it work?

rata (approved):
@johananl thanks a lot, LGTM!

Added some questions on some things I'm not sure I followed. I'd like to know, but it works and is simpler, so I'm totally okay. Just want to understand :)

.circleci/config.yml
johananl: I removed it because when changing to the "next gen" Go image from CircleCI the job was broken *with* this line and worked *without* it.

.circleci/config.yml
johananl: We are not using `pip` (or Python in general) in this job so there is no reason to install the packages.
`go test` works fine without `echo "export GO111MODULE=on" >>$BASH_ENV` so looks like we don't need it. Less code == better IMO.

.circleci/config.yml
rata: Cool!

.circleci/config.yml
rata: Oh, I had misunderstood the diff. Sorry, my bad, now it all makes sense. Perfect, thanks! :)

rata:
@johananl LGTM, please feel free to merge yourself :)

rata (approved):

PR #770, opened 2020-10-28
manifests: add memberlist ports

champtar:
Even if not required it's better to properly document what is used as ports

### DIFF
--- a/manifests/metallb.yaml
+++ b/manifests/metallb.yaml
@@ -64,6 +64,8 @@ spec:
   hostPorts:
   - max: 7472
     min: 7472
+  - max: 7946
+    min: 7946
   privileged: true
   readOnlyRootFilesystem: true
   requiredDropCapabilities:
@@ -300,6 +302,8 @@ spec:
             fieldRef:
               fieldPath: status.podIP
         # needed when another software is also using memberlist / port 7946
+        # when changing this default you also need to update the container ports definition
+        # and the PodSecurityPolicy hostPorts definition
         #- name: METALLB_ML_BIND_PORT
         #  value: "7946"
         - name: METALLB_ML_LABELS
@@ -319,6 +323,11 @@ spec:
         ports:
         - containerPort: 7472
           name: monitoring
+        - containerPort: 7946
+          name: memberlist-tcp
+        - containerPort: 7946
+          name: memberlist-udp
+          protocol: UDP
         resources:
           limits:
             cpu: 100m


champtar:
Please merge

manifests/metallb.yaml
```diff
@@ -319,6 +321,11 @@ spec:
         ports:
         - containerPort: 7472
           name: monitoring
+        - containerPort: 7946
```
russellb, line 324: Can you also update the comment about 20 lines above this that shows how to set `METALLB_ML_BIND_PORT` to note that if that port is changed, it should also be changed in these new places you're adding the port number?

Otherwise, this change lgtm

russellb (changes requested):

manifests/metallb.yaml
champtar: Done

russellb:
Thanks for the quick update to my review!

nnz1024:
Please note that it will not work correctly on update of existing MetalLB installation due to known [Kubernetes bug](https://github.com/kubernetes/kubernetes/issues/39188) disallowing to add ports with same numbers and different protocols to existing pod templates.

To correctly set ports, user need to delete speaker DaemonSet and create (apply manifest) again from scratch.

Please consider adding this to release notes.

russellb:
Thanks for catching this @nnz1024 .  If it breaks simple upgrades, I wonder if we should just revert this since it wasn't required anyway.  Thoughts @champtar ?

champtar:
Let me test, here we add 7946/tcp and 7946/udp at the same time so maybe this case is not broken

champtar:
Thanks @nnz1024 for the heads up in any case

champtar:
Before
```
          name: speaker
          ports:
          - containerPort: 7472
            hostPort: 7472
            name: monitoring
            protocol: TCP
```
After
```
          name: speaker
          ports:
          - containerPort: 7472
            hostPort: 7472
            name: monitoring
            protocol: TCP
          - containerPort: 7946
            hostPort: 7946
            name: memberlist-tcp
            protocol: UDP
```

I still believe it's better to have it as it doesn't break upgrade, make new installs clean, and match the Helm Charts.
Also maybe some day this bug will be fixed :D

champtar:
(tested on 1.18.18)

russellb:
I'm fine with this if it doesn't surprisingly break upgrades for someone

PR #771, opened 2020-10-28
Do not log startUpdate/endUpdate for non LoadBalancer services

champtar:
Reduce log spam


champtar:
Can we merge this ?

russellb:
Thanks for the patch.  I agree that this seems like an improvement.

PR #775, opened 2020-11-02
Allow to share same subnet for metallb address pool and clusterIPs

jpm38:
I need to share the same network range for a metallb allocation pool and a kubernetes cluster IP range.
That's what this PR is addressing.
I've added an option in metallb config file:

apiVersion: v1
kind: ConfigMap
metadata:
  namespace: metallb-system
  name: config
data:
  config: |
    address-pools:
    - name: metallb-adress-pool
      allow-cluster-ip: true
      protocol: layer2
      addresses:
     - 10.0.252.0/22

and then when a service of type "LoadBalancer" is registered, the allocator return the clusterIP address of the service.

Note that without the option (allow-cluster-ip: true), the original behavior remains unchanged.

### DIFF
--- a/controller/service.go
+++ b/controller/service.go
@@ -152,8 +152,15 @@ func (c *controller) allocateIP(key string, svc *v1.Service) (net.IP, error) {
 	isIPv6 := clusterIP.To4() == nil
 
 	// If the user asked for a specific IP, try that.
-	if svc.Spec.LoadBalancerIP != "" {
-		ip := net.ParseIP(svc.Spec.LoadBalancerIP)
+	if c.ips.IsAllowClusterIP(clusterIP) || svc.Spec.LoadBalancerIP != "" {
+                var ip net.IP
+		if (svc.Spec.LoadBalancerIP != "") {
+			// The user asked for a specific loadBalancer IP, try that.
+			ip = net.ParseIP(svc.Spec.LoadBalancerIP)
+		} else {
+			// The user asked for the cluster IP, try that.
+			ip = clusterIP
+		}
 		if ip == nil {
 			return nil, fmt.Errorf("invalid spec.loadBalancerIP %q", svc.Spec.LoadBalancerIP)
 		}
--- a/controller/service.go
+++ b/controller/service.go
@@ -153,8 +153,8 @@ func (c *controller) allocateIP(key string, svc *v1.Service) (net.IP, error) {
 
 	// If the user asked for a specific IP, try that.
 	if c.ips.IsAllowClusterIP(clusterIP) || svc.Spec.LoadBalancerIP != "" {
-                var ip net.IP
-		if (svc.Spec.LoadBalancerIP != "") {
+		var ip net.IP
+		if svc.Spec.LoadBalancerIP != "" {
 			// The user asked for a specific loadBalancer IP, try that.
 			ip = net.ParseIP(svc.Spec.LoadBalancerIP)
 		} else {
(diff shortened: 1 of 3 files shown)


champtar:
@jpm38 can't you already do that today ?
```
spec:
  clusterIP: a.b.c.d
  loadBalancerIP: a.b.c.d
  type: LoadBalancer
```

jpm38:
> @jpm38 can't you already do that today ?
> 
> ```
> spec:
>   clusterIP: a.b.c.d
>   loadBalancerIP: a.b.c.d
>   type: LoadBalancer
> ```

jpm38:
@champtar 
Not really, what we need it to assign automatically a loadBalancerIP with the clusterIP, not to do it manually.
The requirement is to be able to add/remove the same service, and use the coreDNS naming resolution to access the service form outside the cluster even if the clusterIP/loadBalancerIP changes.

strus38:
Any update? Is there a blocking point?

rata:
Well, is not the same subnet but the very same IP, right? I really miss what is the point of using the cluster IP (an internal IP) as an **external IP** for the service. It really seems quite wrong to do that. Can you please elaborate on the use case?

jpm38:
The context is a kubernetes cluster on premise with a flat management network.
We have let's say 10 nodes inside the k8s cluster (masters and workers) and thousands of nodes outside the k8s cluster using the k8s services.
The management network is 10.0.0.0/16 and the clusterIP range 10.0.252.0/22
The requirement is to have all nodes belonging to the same network and to be able to use the service discovery of coreDNS from outside the k8s cluster.
So I need to assign the clusterIP allocated by k8s to the external IP.

Note again that I've added an option and therefore the original behavior remains unchanged.

champtar:
@jpm38 having 2 times the same IP for clusterIP and loadBalancerIP is kube-proxy undefined behavior for me, it's currently working in your setup but this is a hack that might break. Also having external IP / LB IP in the cluster IP range might be blocked in the future for security reasons (see latest CVE and all discussions around it). Have you looked at other solutions like calico: https://docs.projectcalico.org/networking/advertise-service-ips ?

jpm38:
@champtar

I maintain this is not a hack, so I do not understand your answer  as we can already do this manually;

spec:
  clusterIP: a.b.c.d
  loadBalancerIP: a.b.c.d
  type: LoadBalancer

And as said before, our cluster is isolated from the internet, so I do not see any security issue here.
This feature is very key and important for our project, could you lease reconsider your conclusion ?

champtar:
It works for you (tm), but LoadBalancerIP are never in the range of clusterIP except in your cluster, so there is 0 tests that guarantee that it will work with all CNI / kube-proxy, and that it'll continue to work (what if some iptables rules rework start to break the tcp connection),
thus this is not something we should merge with the really low bandwidth of the project.
Again Calico (and maybe other CNI) allow to advertise clusterIPs without hack, have you look at it ?

jpm38:
Ok thanks.

russellb:
I understand the use case and I'm not against including it.  I agree with the points made that it wouldn't be the common case, nor do we have any idea where it will work and where it won't, so that worries me a bit.  We are seriously overwhelmed with support-like requests as it is.

I'm not a huge fan of `allow-cluster-ips`.  Something like `use-cluster-ips` would make a little more sense to me.

I think the bigger thing missing is documentation and testing.  Maybe you could reach out to the mailing list / slack / both and see who else might be interested in this feature?  If you could get successful test reports in different configurations, that would make me feel a bit better.

Until then, I'd rather hold off on merging this, at least for now.

russellb:
since this is a NACK for now, I'm going to close this.  Feel free to re-open when more info is available and you think it should be reconsidered for merging.

PR #781, opened 2020-11-19
fix speakers layer2 with no endpoints when traffic policy is cluster

aarnaud:
Case when speakers is deployed on some specific node with node affinity and service endpoints is on other hosts.

with `ExternalTrafficPolicy=Cluster` none of speaker advertise there MAC with layer2 protocol. So this PR fixed this issue
if `ExternalTrafficPolicy=Local` the behaviour still the same.


liuyuan10:
This is similar to #562. One key issue is that the speakerlist  return null list when memberlist is not enabled and your change will cause different behavior in that case. I'm trying to push that we always uses full node list for traffic cluster services to be consistent everywhere. But I'm blocked by another PR to make speakerlist also work in non-memberlist mode.

aarnaud:
I see, but if we avoid to use endpoint node for sharedIPS to chose where IP is deploy, that fix your issue if traffic policy is cluster

aarnaud:
My only workaround is this patch because my service endpoint will never be on same speakers node

PR #784, opened 2020-11-24
Move images to Quay

### LINKED ISSUE #696: Move images to quay.io

johananl:
This PR points the manifests to Quay instead of Docker Hub.

Fixes #696.

### DIFF
--- a/e2etest/manifests/mirror-server.yaml
+++ b/e2etest/manifests/mirror-server.yaml
@@ -17,7 +17,7 @@ spec:
     spec:
       containers:
       - name: mirror
-        image: metallb/mirror-server:main
+        image: quay.io/metallb/mirror-server:main
         env:
         - name: NODE_NAME
           valueFrom:
--- a/manifests/metallb.yaml
+++ b/manifests/metallb.yaml
@@ -313,7 +313,7 @@ spec:
             secretKeyRef:
               name: memberlist
               key: secretkey
-        image: metallb/speaker:main
+        image: quay.io/metallb/speaker:main
         imagePullPolicy: Always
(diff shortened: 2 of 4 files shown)


johananl:
This breaks `inv dev-env`. Fixing.

johananl:
> This breaks `inv dev-env`. Fixing.

Done.

johananl:
NOTE: Setting the default registry to `quay.io` requires changing the image URIs used by `inv dev-env`: Up until now we haven't specified the registry when dealing with "local" images, however "no registry" means "Docker Hub" to Docker :shrug: 

Rather than leaving the default as it is I've opted for specifying `quay.io/` in the URIs used in `kind load docker-image`.

If anybody thinks it's not a good idea to have `quay.io/` in local images (namely images which aren't going to necessarily get pushed) - please shout at me.

rata:
> If anybody thinks it's not a good idea to have `quay.io/` in local images (namely images which aren't going to necessarily get pushed) - please shout at me.

No, it LGTM. Thanks a lot!

rata (approved):

PR #798, opened 2021-01-07
Set imagePullPolicy to IfNotPresent

### LINKED ISSUE #748: Remove `imagePullPolicy: Always` from release manifests
`imagePullPolicy: Always` makes sense for main images, but we should remove it from the release manifests

johananl:
There is no reason to have it set to Always and re-pulling the image every time can make k8s convergence and scaling slower.

Fixes #748.

### DIFF
--- a/manifests/metallb.yaml
+++ b/manifests/metallb.yaml
@@ -314,7 +314,6 @@ spec:
               name: memberlist
               key: secretkey
         image: quay.io/metallb/speaker:main
-        imagePullPolicy: Always
         name: speaker
         ports:
         - containerPort: 7472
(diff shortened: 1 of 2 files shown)


champtar:
I would just remove it so cluster default is used

johananl:
> I would just remove it so cluster default is used

But then we have a problem here:

https://github.com/metallb/metallb/pull/798/files#diff-2e746c9eacad6e2c2fdefdc6a665b5ce4607ea384ebf97a729044b79acc778d6R235

champtar:
> https://github.com/metallb/metallb/pull/798/files#diff-2e746c9eacad6e2c2fdefdc6a665b5ce4607ea384ebf97a729044b79acc778d6R235

just remove the line

johananl:
OK, I've tested and this seems to behave as expected.

daxmc99 (approved):

champtar:
Merge ?

PR #812, opened 2021-03-04
Fix nodeAssigned event on k8s >= 1.20

rata:
As reported in https://github.com/metallb/metallb/issues/794, Kubernetes 1.20 deprecated the selfLink field. For MetalLB that caused the nodeAssigned event to fail to be sent, as it was constructed using an older client-go version that still used that field and the API server rejected the message.

The IPs are advertised and all seems to work fine in my testing, just that event is lost and you can't find it when running `kubectl describe service <my-service>` to easily know which node is announcing one IP. This is particularly useful for Layer2. Also, please note that the work-around mentioned by @brianmay in the original report (#794) probably works fine, but this is something we want to fix ASAP anyways (and the workaround won't work on k8s 1.21).

This PR is just a mix of these PRs: #802 #723 #724.

PR #723 updated client-go but failed the tests. PR #802 is the same, but fixing the test. To do that it forked virtuakube (a dependency not really used anymore) so we can update client-go without issues and tests work fine.

However, we don't really need to fork virtuakube and update to it client-go 1.20, we can just ditch it. That is what PR #724 was doing, but it removed some files used in `inv dev-env`.

So, this PR just takes the commits from there and amends them to: remove e2e directory (except the manifest used for `inv dev-env`) and update client-go but dropping the virtuakube dependency and replaces hacks it added in go.mod.

As PR #724 is already approved and the client-go updates are easy, I'm going to just merge this after checking it indeed passes the tests.

Thanks to @russellb @brianmay @uablrek @champtar @invidian and lot of others that were involved in the issues and PRs!

This closes:
Closes: #794 
Closes: #722
Closes #192
Closes #629
Closes #639
Closes: #723
Closes: #724 
Closes: #802

PR #813, opened 2021-03-04
website: Check for outdated dependencies when creating a release

rata:
It happened in the past that client-go was out of date and that broke
MetalLB on new Kubernetes releases, like seen in this PR:
	https://github.com/metallb/metallb/pull/812

When we created the 0.9 release branch no one really checked, we could
have updated it by then and never have this issue in the first place.
So, I propose to add this to the release-process.

### DIFF
--- a/website/content/community/release-process.md
+++ b/website/content/community/release-process.md
@@ -13,6 +13,19 @@ Are you actually ready to release? Check the milestone on github and
 verify that all its issues are closed. If there are open issues,
 you'll have to either resolve them, or bump to the next version.
 
+### Check for outdated dependencies
+
+This might be a little late to do at release time, but releases are not so
+frequent for now, so it makes sense to do this before creating a release.
+
+Some dependencies are usually best to stay fresh, like client-go and friends.
+But ideally we should be using recent versions of all our dependencies.
+
+If client-go is too old (like 2-3 releases old), from time to time Kubernetes
+may remove some field that breaks MetalLB if it is using an old version. Make
+sure either to use a recent version or create an issue to do a new release,
+after proper testing, to update client-go.
+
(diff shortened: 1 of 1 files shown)


website/content/community/release-process.md
```diff
@@ -13,6 +13,19 @@ Are you actually ready to release? Check the milestone on github and
...
+This might be a little late to do at release time, but releases are not so
+frequent for now, so it makes sense to do this before creating a release.
+
+Some dependencies are usually best to stay fresh, like client-go and friends.
+But ideally we should be using recent versions of all our dependencies.
+
+If client-go is too old (like 2-3 releases old), from time to time Kubernetes
+may remove some field that breaks MetalLB if it is using an old version. Make
+sure either to use a recent version or create an issue to do a new release,
+after proper testing, to update client-go.
```
russellb, line 27: My only concern here is that I don't want to imply we should do big dependency bumps and then release right away.  I think we'd want big jumps to soak for a bit in CI.  I *think* that's what you're saying here by saying to create an issue.

Would you like to be more strict and saying something like "we should not issue a new release with a client-go that is too old." and have that as part of the release checklist and what would potentially make us hold off a release?  It's not clear from the text here if you think a release should be held over this.

champtar: Another way to have more testing that was not used in the past is doing RC to get more testing from the field.
ie before a major release update dependencies that make sense and build a RC to get feedback.
I don't think many people test/run master ;)

russellb: yes RCs for the feature releases are a good idea, too.

fedepaol:
We now keep CI up to date with latest k8s releases, closing this.

PR #815, opened 2021-03-10
Add operator to toleration

### LINKED ISSUE #814: add Exists operator to toleration
The default toleration doesn't specify an operator, so the operator defaults to `Equal`. It also doesn't specify a value, so there's nothing for `Equal` to match. The result is that the speaker Pods are never scheduled to controlplane nodes and never announce services located there (like Traefik in K3s).

The solution is to add `operator: Exists` to the `tolerations` key:

```
      tolerations:
      - effect: NoSchedule
        key: node-role.kubernetes.io/master
        operator: Exists
```

This is supported by [the docs](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/):

> The default value for operator is Equal.
> 
> A toleration "matches" a taint if the keys are the same and the effects are the same, and:
> 
>     the operator is Exists (in which case no value should be specified), or
>     the operator is Equal and the values are equal.

Demonstration of the failure is at 13:04 in my latest video: https://www.youtube.com/watch?v=9PLw1xalcYA&t=784s

oskapt:
Fixes #814

### DIFF
--- a/manifests/metallb.yaml
+++ b/manifests/metallb.yaml
@@ -341,6 +341,7 @@ spec:
       tolerations:
       - effect: NoSchedule
         key: node-role.kubernetes.io/master
+        operator: Exists
 ---
 apiVersion: apps/v1
 kind: Deployment


daxmc99 (approved):
LGTM! Thanks for the PR!
cc @rata @johananl for a +1

champtar:
@oskapt the whole explanation provided in #814 should be in the commit message

oskapt:
does that mean you want me to add some random commit and update the message?

champtar:
Not add another commit, just modify this one
You can do something like
```
git commit --amend
git push -f
```
Your description in #814 is really good, but the only permanent and easily accessible record is the git history, that's why it's always good to write those stuff in the commit message.

oskapt:
ok - done.

PR #816, opened 2021-03-12
Update to Go 1.16

daxmc99:
Update to Go 1.16

( I'm not familiar with our CI system, it looks like we need to upgrade the images as well,   
 found via https://circleci.com/developer/images/image/cimg/go )

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -4,7 +4,7 @@ version: 2
 jobs:
   test-1.13:
     docker:
-      - image: cimg/go:1.13
+      - image: cimg/go:1.16
     steps:
       - checkout
       - run: go test -short ./...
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -4,22 +4,22 @@ version: 2
 jobs:
   test-1.13:
```
champtar, line 5: `test-1.13` -> `test` ?

.circleci/config.yml
```diff
@@ -4,22 +4,22 @@ version: 2
...
     docker:
-      - image: cimg/go:1.13
+      - image: cimg/go:1.16
     steps:
       - checkout
       - run: go test -short ./...
       - run: go test -short -race ./...
       - run: cp manifests/metallb.yaml manifests/metallb.yaml.prev
   lint-1.13:
```
champtar, line 13: `lint-1.13` -> `lint` ?

daxmc99: I still need to find out what these actually do 😓 
Maybe we can move to github actions

champtar:
Can we have this for 0.10.0 ?
I've been bitten this week by a bug from space where client-go get stuck unable to talk to API server, K8S project added health checks but there was many fixes in golang never backported to golang 1.13
https://github.com/kubernetes/kubernetes/issues/87615
https://github.com/kubernetes/kubernetes/pull/100376

champtar:
@daxmc99 do you think you will have time to work on this "soon"
@rata @johananl as you are the ones doing the releases do we need to bump the go version anywhere else ?

russellb:
I merged #839

PR #823, opened 2021-03-23
Remove unused Service Update()

christarazi:
The Update() to the Service object is never triggered because the code
does not modify the object between the deep copy and when Update() is
called.

Previously, commit cdd44aa9 ("Initial commit. Somewhat working
controller, half-written BGP speaker") modified the Spec and Annotations
field of the Service. However, since commit 3e10f6ea ("Clean up processing
to exclusively use status.loadBalancer.ingress.ip"), that is no longer
the case, as the code exclusively only touches the Status field.

Therefore, this commit removes this dead code and revokes the update
privileges from the ClusterRole as well.

Signed-off-by: Chris Tarazi <chris@isovalent.com>

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -65,11 +65,6 @@ type testK8S struct {
 	t                   *testing.T
 }
 
-func (s *testK8S) Update(svc *v1.Service) (*v1.Service, error) {
-	s.updateService = svc
-	return svc, nil
-}
-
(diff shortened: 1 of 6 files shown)


christarazi:
Another positive is that one call to `reflect.DeepEqual` is removed :)

russellb:
Thanks for the PR!  I appreciate the clear commit message, as well.

PR #828, opened 2021-03-31
gitignore: Update .gitignore to remove e2etest reference

markdgray:
We should remove the e2etest references as these were dropped.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,8 +2,3 @@
 build/
 website/public
 .envrc
-e2etest/vmlinuz
-e2etest/initrd.img
-e2etest/*.qcow2
-e2etest/e2etest
-e2etest-cached-universe


.gitignore
```diff
@@ -2,8 +2,4 @@
 build/
 website/public
 .envrc
-e2etest/vmlinuz
-e2etest/initrd.img
-e2etest/*.qcow2
-e2etest/e2etest
-e2etest-cached-universe
+kind.yaml
```
russellb, line 5: Did this get auto-created?  Or is it an optional thing that people might create?  I just noticed I don't have this file.

markdgray: I just realized it was there from a couple of months ago when I was looking at this repo. I ran "dev-env-cleanup" and didn't check if that was cleaned up, and hence part of, the "dev-env" task. I have removed that.

russellb:
Thanks!

PR #830, opened 2021-04-05
design: Create a directory for design proposals

russellb:
To allow for easier review on design proposals, create a directory for
design proposals.  Future PRs with proposed designs should be posted as
a PR to this directory.

While we don't require a specific outline, we encourage the use of the
KEP template for inspiration.


rata (approved):
LGTM, thanks!

Issue #196, opened 2018-03-14
Replace the MetalLB configmap with k8s custom resources


danderson:
MetalLB currently uses a ConfigMap for its configuration. This is easy, but not very kubernetes-native.

We should consider migrating the config to custom resource definitions. Looking at our config today, we would have ~3 custom resources:

- BGP peer, similar to how Calico does it when it's running in "kubernetes datastore" mode
- Address pool
- BGP community aliases (not sure about this one - it's kind of an odd one out in the current config, not sure where to put it)

Major advantage of custom resources is that Kubernetes users already know how to manipulate them, and it gives us a great attachment point for event logs (peer connected, address allocated...). This lets us inject most of what's in MetalLB logs today into the standard kubernetes event stream. It also lets us use webhooks in a more natural way, where we can allow/deny changes to individual subresources of the config.

Major disadvantage is that CRDs are a nightmare to work with in the k8s go-client, we basically have to compile our own fork of go-client to get custom type definitions, unless things have changed for the better.

Thoughts welcome on this one, I'm conflicted about whether we should do this.

miekg:
[ Quoting <notifications@github.com> in "[google/metallb] Replace the MetalL..." ]
>MetalLB currently uses a ConfigMap for its configuration. This is easy, but not very kubernetes-native.

How does this look from the kubectl side? I quite like the simplicity a 
ConfigMap gives me. Is this still true for custom resources?

That "forking the k8s go client" makes this look like a non starter though....

/Miek

--
Miek Gieben

mleklund:
in k8-ipcontroller they are controlled with kubectl

```
apiVersion: ipcontroller.ext/v1
kind: IpClaimPool
metadata:
    name: test-pool
spec:
    cidr: 192.168.0.248/29
    ranges:
        - - 192.168.0.249
          - 192.168.0.250
        - - 192.168.0.252
          - 192.168.0.253
```

danderson:
Kubectl has some automagic support for handling CRDs, so it would probably look something like:

`kubectl edit metallb.universe.tf/AddressPool my-addr-pool`

Which would let you edit something that looks like:

```yaml
apiVersion: metallb.universe.tf/v1
kind: AddressPool
metadata:
  name: my-addr-pool
spec:
  protocol: layer2
  cidr:
  - 192.168.16.240/30
```

Exact content of each CRD is to be figured out still, but that's approximately what things would look like.

[...]

Issue #630, opened 2020-06-12
Look into a generic pool <> node affinity mechanism


johananl:
We have multiple requests around being able to announce specific IPs from a specific subset of speakers:

- https://github.com/metallb/metallb/issues/511
- https://github.com/metallb/metallb/issues/383

We also have a PR which addresses this for BGP only: https://github.com/metallb/metallb/pull/596

I suggest we look into a **generic** (namely protocol-independent) way for adding this functionality. AFAICT, the UX could be something like the following:

```yaml
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: metallb-system
  name: config
data:
  config: |
    address-pools:
    - name: default
      protocol: bgp # Or layer2!
      addresses:
      - 198.51.100.0/24
      # Announce IPs in this pool only from
      # speakers running on matching nodes.
      node-selectors:
      - match-labels:
          rack: frontend
```

We can then check the node selectors and decide whether to announce an address in the layer2/BGP implementation.

I need to think about this more at a low level to see if this can play well with things like the "single active node" nature of layer 2. It is very likely trickier than I'm currently thinking.

liuyuan10:
another issue with this is that metallb only annonce IPs from a node that runs the pod. even if you make an IP annonced from a subnet of nodes, you have to put the same node selector in your pod spec to make sure you have pod running in those nodes.  you need #613 to make it work as expected.

anton-johansson:
@liuyuan10, very good point.

But I do guess that it would be up to the cluster admins to make sure that you're not locking down your services by not having pods on the nodes being eligible for announcing.

As you say, putting the same node selectors on the pods themselves is one way to go.

If you only take "out" one node a time and have atleast two pods up at any given time, it shouldn't be a problem either.

ykulazhenkov:
When BGP used, we can have multiple BGP peers on each node. With "node" selector we can't fully cover use-case(https://github.com/metallb/metallb/pull/596) when one node establishes connections with multiple BGP peers which are in different network segments.

anton-johansson:
Messing around with workarounds for this, I tried adding `nodeSelector` to my Speaker `DaemonSet`. This almost works as I want it, but it made me understand why this issue could be very tricky.

[...]

PR #833, opened 2021-04-07
design: MetalLB CRD controller enhancement

msherif1234:
This design document discusses implementing MetalLB Custom Resource definition"CRD"
as a mechanism to configure MetalLB's layer2 and BGP features instead of using ConfigMap.

ran mdl against the new md file
```
mdl 0001-metallb-crd.md 
```

Fixes https://github.com/metallb/metallb/issues/196.


design/0001-metallb-crd.md
```diff
@@ -0,0 +1,173 @@
...
+<!-- toc -->
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+- [Proposal](#proposal)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+- [Testing](#testing)
+- [Debugging and Troubleshooting](#debugging-and-troubleshooting)
+<!-- /toc -->
```
markdgray, line 11: I didn't include the TOC in my enhancement.

This is a question for the maintainers. We should probably decide on a template rather than loosely basing on KE template otherwise every enhancement will look different.

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,173 @@
...
+
+## Summary
+Today MetalLB allows only configuration of layer2 and BGP protocols using Configmap objects,
```
markdgray, line 14: ConfigMap

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,173 @@
...
+- [Motivation](#motivation)
+  - [Goals](#goals)
+- [Proposal](#proposal)
+  - [Risks and Mitigations](#risks-and-mitigations)
+- [Design Details](#design-details)
+- [Testing](#testing)
+- [Debugging and Troubleshooting](#debugging-and-troubleshooting)
+<!-- /toc -->
+
+## Summary
```
markdgray, line 13: This reads more like a motivation than a summary. Perhaps move this to the "Motivation" section and create a summary describing the proposed changes.

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,173 @@
...
+      address-pools:
+      - name: default
+        protocol: layer2
+        addresses:
+        - 172.18.255.100-172.18.255.200
+    ```
+  - deploy load balancer service application for example NGINX.
+  - verify traffic to the NGINX service.
+  
+    ```bigquery
```
markdgray, line 76: I don't think you need "bigquery" here as this is not a "bigquery" code block.

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,173 @@
...
+    
+    You'll get out put similar to the following:
+    
+    NAME         TYPE           CLUSTER-IP      EXTERNAL-IP    PORT(S)        AGE
+    kubernetes   ClusterIP      10.96.0.1       <none>         443/TCP        11m
+    nginx        LoadBalancer   10.96.148.110   172.19.255.1   80:32326/TCP   9s
+    Notice that the nginx service publishes an EXTERNAL-IP.
+     Run the curl against the EXTERNAL-IP, for example:
+    ```
+    ```bigquery
```
markdgray, line 91: Same comment. I think you can remove "bigquery"

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,173 @@
...
+
+### Goals
+Configuring MetalLB layer2 and BGP using kubernetes custom resource "CRD" instead of ConfigMap.
+
+## Proposal
+Using SIG's [Kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) framework to build new CRD "MetalLB" object 
+and a controller which works as a translator from the user's configuration yaml to a MetalLB's configmap object.
+with this approach there will be no changes required for the MetalLB management plane.
+
+### Risks and Mitigations
```
markdgray, line 31: There is no automatic markdown linting in this repo yet. Perhaps pass this through a linter to resolve any issues that may pop up. For example, I think there should be new lines after title sections.

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,173 @@
...
+Users need to either use CRD or ConfigMap to provision load balancer instance but not both, if user
+mistakenly used both the CRD configuration will overwrite the user defined ConfigMap.
+
+## Design Details
+![Alt text](./metallb_crd.png?raw=true "MetalLB Kubernetes Controller")
```
markdgray, line 36: I have a preference for something like an ASCII diagram (e.g. https://asciiflow.com/ ) instead because it doesn't add a lot of bytes to the repo, it is easier to review and it can be easily updated after merging (even if you don't have the original source). Maybe other reviewers have other ideas about tooling/approached for this.

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,173 @@
...
+## Motivation
+Custom resource definition "CRD" takes advantage of kubernetes extensibility to deliver the automation advantages of cloud services like 
+provisioning, scaling and backup/restore.
+
+### Goals
+Configuring MetalLB layer2 and BGP using kubernetes custom resource "CRD" instead of ConfigMap.
+
+## Proposal
+Using SIG's [Kubebuilder](https://github.com/kubernetes-sigs/kubebuilder) framework to build new CRD "MetalLB" object 
+and a controller which works as a translator from the user's configuration yaml to a MetalLB's configmap object.
```
markdgray, line 28: Why would we continue to use the intermediate ConfigMap?

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,173 @@
...
+with this approach there will be no changes required for the MetalLB management plane.
+
+### Risks and Mitigations
+Users need to either use CRD or ConfigMap to provision load balancer instance but not both, if user
+mistakenly used both the CRD configuration will overwrite the user defined ConfigMap.
+
+## Design Details
+![Alt text](./metallb_crd.png?raw=true "MetalLB Kubernetes Controller")
+
+1- User will supply the yaml configuration using the MetalLB resource.
```
markdgray, line 38: nit: Maybe you could use a Github markdown ordered list here?

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,173 @@
...
+
+2- Configuration will get authenticated and  authorized based on RBAC rules customized by SIG's framework
```
markdgray, line 40: nit: double space between "and" and "authorized"

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,173 @@
...
+3- Supplied configuration will get verified using openAPIV3Schema based on the definition of MetalLB CRD.
+
+4- New objects will be stored in etcd.
+
+5- MetalLB controller monitors API server looking for creation of MetalLB resource.
+
+6- MetalLB controller will use golang based template to convert the configured MetalLB object to ConfigMap object.
+
+7- The generated configmap object will be written to etcd so MetalLB's components can consume it as if the user directly 
+supplied Configmap configuration.
```
markdgray, line 52: ConfigMap

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,173 @@
+# MetalLB Custom Resource Controller
```
markdgray, line 1: It would be good to see a proposed CRD as part of this design.

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,173 @@
...
+      namespace: metallb-system
+    spec:
+      address-pools:
+      - name: default
+        protocol: layer2
+        addresses:
+        - 172.18.255.100-172.18.255.200
+    ```
+  - deploy load balancer service application for example NGINX.
+  - verify traffic to the NGINX service.
```
markdgray, line 74: Are there any standard Kubernetes e2e tests we could use?

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,173 @@
...
+which is simple from configuration perspective and doesn't require cluster-admin rights,
+however it doesn't have validation on the API server level, no support for API versioning and 
+no fine-grained permission model via RBAC.
+
+## Motivation
+Custom resource definition "CRD" takes advantage of kubernetes extensibility to deliver the automation advantages of cloud services like 
+provisioning, scaling and backup/restore.
+
+### Goals
+Configuring MetalLB layer2 and BGP using kubernetes custom resource "CRD" instead of ConfigMap.
```
markdgray, line 24: Are there any other goals/non-goals? 
* Do we have any goals regarding stability of the CRD?
* I presume that we intend to expose all current ConfigMap configuration via the CRD?

markdgray:
Thanks for doing this design document @msherif1234

design/0001-metallb-crd.md
msherif1234: this graph I did using draw.io if there is agreement on tool I will migrate to it

design/0001-metallb-crd.md
msherif1234: the controller convert the CRD configs to configmap intentionally to avoid chaning Metallb provisiong logic with that approach we have no risk at all if users wanted to continue with configmap

design/0001-metallb-crd.md
msherif1234: u want to share the CRD definition or sample config , sample config is already shared in the testing section

design/0001-metallb-crd.md
msherif1234: not that I am aware of, I am planning to extend existing Metallb e2e test to include MetalLB resource.

design/0001-metallb-crd.md
msherif1234: using k8s CRD enable us to support versioning to help with version upgrades , so that support is built in CRD using Conversion web hook.

design/0001-metallb-crd.md
markdgray: It might be good to propose an initial CRD (rather than a sample) here to get an early review as this is a key part of the design. It may save you rework at a later stage.

design/0001-metallb-crd.md
markdgray: I dont think there is a MetalLB e2e test as it was dropped recently.

design/0001-metallb-crd.md
russellb: ha, I'm OK with including a template.  I'm also just happy to have people working on this type of doc, even if they all look different.  :)

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,251 @@
...
+  - [Initial CRD definition](#initial-crd-definition)
+- [Testing](#testing)
+- [Debugging and Troubleshooting](#debugging-and-troubleshooting)
+
+<!-- /toc -->
+
+## Summary
+
+The purpose of this enhancement is to be able to use kubernetes Custom
+Resource Definition "CRD" to configure MetalLB layer2 and BGP.
```
russellb, line 20: Can you clarify that the proposal is to support custom resources as an alternative configuration interface?  I don't think we want to drop the existing ConfigMap interface (at least for now).

Something like ...

The purpose of this enhancement is to propose an alternative configuration interface for MetalLB based on Kubernetes Custom Resource Definitions (CRDs) while also retaining support for the existing ConfigMap interface.

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,251 @@
...
+
+## Motivation
+
+Today MetalLB allows only configuration of layer2 and BGP protocols using
+ConfigMap objects, which is simple from configuration perspective and doesn't
+require cluster-admin rights, however it doesn't have validation on the API
```
russellb, line 26:
```suggestion
require cluster-admin rights. However, it doesn't have validation on the API
```

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,251 @@
...
+server level, no support for API versioning and no fine-grained permission
+model via RBAC.
+Custom resource definition "CRD" takes advantage of kubernetes extensibility
+to deliver the automation advantages of cloud services like provisioning,
+scaling and backup/restore.
```
russellb, line 31: I'm not sure how these automation advantages apply to this change from a ConfigMap to CRDs.  Maybe just say that CRDs will provide the benefits you just listed above like API level validation?

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,251 @@
...
+
+### Goals
+
+Configuring MetalLB layer2 and BGP using kubernetes custom resource "CRD"
```
russellb, line 35: I think throughout this enhancement you can just say "configuring MetalLB" and not include "layer2 and BGP".  I think it can be assumed that all MetalLB configuration should be supported through this interface.

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,251 @@
+# MetalLB Custom Resource Controller
```
russellb, line 1: Somewhere in here can you add a link to #196 ?  That's the issue opened a few years ago to track this idea.

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,251 @@
...
+Today MetalLB allows only configuration of layer2 and BGP protocols using
+ConfigMap objects, which is simple from configuration perspective and doesn't
+require cluster-admin rights, however it doesn't have validation on the API
+server level, no support for API versioning and no fine-grained permission
+model via RBAC.
+Custom resource definition "CRD" takes advantage of kubernetes extensibility
+to deliver the automation advantages of cloud services like provisioning,
+scaling and backup/restore.
+
+### Goals
```
russellb, line 33: Please also include:

 * Retain support of the existing ConfigMap interface for backwards compatibility

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,251 @@
...
+
+Configuring MetalLB layer2 and BGP using kubernetes custom resource "CRD"
+instead of ConfigMap.
+
+## Proposal
+
+Using SIG's [Kubebuilder](https://github.com/kubernetes-sigs/kubebuilder)
```
russellb, line 40: which SIG?  It's in a github org shared by multiple Kubernetes SIGs.  I would specify which SIG owns this, or just refer to kubebuilder on its own.

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,251 @@
...
+framework to build new CRD "MetalLB" object and a controller which works as
+a translator from the user's configuration yaml to a MetalLB's configmap
+object, with this approach there will be no changes required for the
+MetalLB management plane.
+
+### Risks and Mitigations
+
+Users need to either use CRD or ConfigMap to provision load balancer instance
```
russellb, line 48:
```suggestion
Users need to either use CRD or ConfigMap to configure MetalLB
```

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,251 @@
...
+but not both, if user mistakenly used both the CRD configuration will
```
russellb, line 49:
```suggestion
but not both. If an administrator mistakenly uses both, the CRD configuration will
```

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,251 @@
...
+   MetalLB object to ConfigMap object.
+1. The generated ConfigMap object will be written to etcd so MetalLB's
+   components can consume it as if the user directly supplied the ConfigMap
+   configuration.
+
+### Initial CRD definition
+
+In this section will share initial metallB custom resource definition which
+currently covers only layer2, BGP will be added to the same CRD, or we might
+separate it out and use the same controller to manage both resources.
```
russellb, line 73: I think we should either mark this as a work-in-progress and complete the full CRD design before merging, or perhaps split this into 2 PRs.  The first PR could include the high level proposal (use CRDs, use kubebuilder, still support the ConfigMap, ...) but not specify the CRD yet.  The 2nd PR could get into the API details after we reach consensus on persuing CRDs in the first place.

I'm interested in what other people think.  I'm interested in getting consensus on the general approach before working too hard on all of the CRD details.

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,251 @@
...
+
+Note: the current definition includes NodeAffinity configuration which not
+upstream yet, but we have it here for completeness.
```
russellb, line 76: If you're going to leave that in, I would like to have it in a separate section that gives an illustration of the process of extending it to support new configuration.  I think we should specify what we'd expect to implement and merge by itself, but having a real example of how it would be extended would also be helpful.  That will be important to understand.

russellb (changes requested):

design/0001-metallb-crd.md
markdgray: Looks like @russellb agrees and this should be an alternative configuration method.

design/0001-metallb-crd.md
msherif1234: sure

design/0001-metallb-crd.md
msherif1234: this just to explain using CRD enable leverage k8s capabilities let me try to better reword this.

design/0001-metallb-crd.md
msherif1234: ok I kept just the reference of kubebuilder

design/0001-metallb-crd.md
msherif1234: that will be doc in the next PR with the rest of the operator code changes

design/0001-metallb-crd.md
msherif1234: I didn't include the CRD definition originally Mark requested to include, I personally prefer to share it in the next PR with the rest of the code and it will include BGP configuration

design/0001-metallb-crd.md
msherif1234: it added in the description section of this PR

design/0001-metallb-crd.md
markdgray: @russellb I think your suggestion of splitting it makes sense. 1st PR can be a general proposal so we can get consensus on that and the 2nd can be the CRD or a design doc that discusses the CRD.

@msherif1234 My reasoning was that it would be good to agree and merge the CRD as soon as possible and not let development and merging of the full solution hold this up because a) some changes in the CRD may affect the rest of the design and it would avoid any timely rework - although I can't think why that would be the case right now and b) having an agreed and merged CRD may help start other related work earlier if any one is doing any work in parallel.

msherif1234: [WIP CRD PR](https://github.com/metallb/metallb/pull/850), the crd yaml is actually auto generated based on the config data structure which is in sync with what Metallb is using I wanted to make metallb use Metallb config spec so we have only one source .

design/0001-metallb-crd.md
johananl: Another idea is https://excalidraw.com/ which allows generating nice-looking diagrams and exporting PNG with the diagram's "source" embedded in the PNG metadata, allowing you to magically edit a diagram by opening the PNG file in Excalidraw.

johananl (commented):
Thanks a lot for working on this @msherif1234. I understand the generic k8s benefits of using CRDs vs. using a ConfigMap (e.g. validation, RBAC) and agree with them.

My main concern is that this proposal seems to just "lift and shift" the existing configuration scheme from the ConfigMap into one big CRD.

Looking at https://github.com/metallb/metallb/issues/196, Dave's idea was to have a fine-grained CRD for each entity in the MetalLB data model, entities such as peers or address pools. I agree with Dave's approach for the following reasons:

- Using resource-specific events can be very useful for troubleshooting and monitoring. For example, we could log BGP status messages on a `BGPPeer` resource and IP allocation messages on an `AddressPool` resource.
- I was hoping that moving to CRDs could improve MetalLB's flexibility. We have many [issues](https://github.com/metallb/metallb/issues?q=is%3Aopen+is%3Aissue+label%3Aflexibility) which revolve around things like "I want to do X but only using *some* of the Y" or "I want X to have multiple Y's". I was hoping the switch to CRDs could allow us to improve this aspect of MetalLB, although right not I'm not 100% sure there is actually a difference between one CRD and multiple CRDs in that regard. I'd have to think about this more.

We probably don't want to do *two* radical changes to the way MetalLB is configured in a short time period, so I'm not sure it makes sense to first move to one CRD and later split the CRD into multiple CRDs. My current feeling is that we should convert this design proposal to the multiple CRDs approach and try to nail down the data model (e.g. the association between an address pool and a peer etc.).

msherif1234:
> Thanks a lot for working on this @msherif1234. I understand the generic k8s benefits of using CRDs vs. using a ConfigMap (e.g. validation, RBAC) and agree with them.
> 
> My main concern is that this proposal seems to just "lift and shift" the existing configuration scheme from the ConfigMap into one big CRD.
> 
> Looking at #196, Dave's idea was to have a fine-grained CRD for each entity in the MetalLB data model, entities such as peers or address pools. I agree with Dave's approach for the following reasons:
> 
> * Using resource-specific events can be very useful for troubleshooting and monitoring. For example, we could log BGP status messages on a `BGPPeer` resource and IP allocation messages on an `AddressPool` resource.
> * I was hoping that moving to CRDs could improve MetalLB's flexibility. We have many [issues](https://github.com/metallb/metallb/issues?q=is%3Aopen+is%3Aissue+label%3Aflexibility) which revolve around things like "I want to do X but only using _some_ of the Y" or "I want X to have multiple Y's". I was hoping the switch to CRDs could allow us to improve this aspect of MetalLB, although right not I'm not 100% sure there is actually a difference between one CRD and multiple CRDs in that regard. I'd have to think about this more.
> 
> We probably don't want to do _two_ radical changes to the way MetalLB is configured in a short time period, so I'm not sure it makes sense to first move to one CRD and later split the CRD into multiple CRDs. My current feeling is that we should convert this design proposal to the multiple CRDs approach and try to nail down the data model (e.g. the association between an address pool and a peer etc.).

Thanks @johananl for taking a look 
1- you are 100% right about the model we are proposing here since this proposed scheme is an alternative way to ConfigMap we didn't want to do radical changes to Metallb configuration handling as that will force current customers to migrate to using CRD based right away which we didn't want.
2- I did brought the question about multiple CRD one for Address pool and another for BGP Peers vs single CRD but after taking a close look at [MetalLB Configurations](https://metallb.universe.tf/configuration/) some of the BGP attributes are part of Address pool configs so they are kind of coupled in away, in another words you can't have BGP peer w/o address pool configuration 
3- I just wanted to mention we don't interact directly with CRD itself , the CRD definition is auto generated based on the config data structs.
4- there is nothing to prevent us down the road to separate out the CRD to multiple if needed but we need to think how the impact to current ConfigMap user and how we can best refactor the current config so the config objects are decoupled

johananl:
@msherif1234 I'd advocate for doing the proper thinking about the data model and multiple CRDs *now* rather than down the road. I don't see a strong reason to move to one big CRD right now and I do see a strong reason to the contrary: Doing the move towards CRDs in two stages means forcing users to deal with **two** migrations rather than one. I'm not trying to say we should *never* introduce breaking changes for the sake of significant enhancements, however moving from a ConfigMap to a CRD and then moving from one CRD to multiple CRDs sounds like bothering the users unnecessarily to me.

msherif1234:
so the modified proposal is to have AddressPool and Peers CRDs and customer who wanted to configure BGP peers will need to generate manifest with both objects to achieve the desired configuration. is the advantage of using multiple CRDs vs 1
- avoid monolithic/big CRD.
- ability to monitor events from each object.
that will yield to 2 controllers "one for each CRD" and both will be updating the same configmap object to generate the final configuration to Metallb controller.

johananl:
@msherif1234 yes, this is probably a good start. However, as I said above I think we have to invest more thought into the data model. The following are **sample** questions which come to mind (but in no way an exhaustive list):

- Does it make sense to keep specifying the protocol (BGP/layer 2) as an attribute of the address pool?
- How do the two CRDs interact? For example, when we implement things like #521 or #630, do we specify a pool under a peer or a peer under a pool? Or maybe we even need a 3rd type of CRD to couple pools and peers together?
- Do we need an equivalent of the peer CRD for layer 2?

I don't think we can move forward with this proposal until we've thought about the data model thoroughly.

msherif1234:
> @msherif1234 yes, this is probably a good start. However, as I said above I think we have to invest more thought into the data model. The following are **sample** questions which come to mind (but in no way an exhaustive list):
> 
> * Does it make sense to keep specifying the protocol (BGP/layer 2) as an attribute of the address pool?
> * How do the two CRDs interact? For example, when we implement things like #521 or #630, do we specify a pool under a peer or a peer under a pool? Or maybe we even need a 3rd type of CRD to couple pools and peers together?
> * Do we need an equivalent of the peer CRD for layer 2?
> 
> I don't think we can move forward with this proposal until we've thought about the data model thoroughly.

@johananl I did look closely at the issues you shared from configuration POV having multiple CRD objects should provide enough flexibility given we are bounded by the ConfigMap interface to metallb controller, I updated this PR with pointer to proposed CRDs, sample config and how the issues you shared can be handled, PTAL.

danehans:
I suggest considering [Gateway API ](https://gateway-api.sigs.k8s.io/api-types/gateway/) as you develop this config API. See recent blog post for a quick intro:

https://kubernetes.io/blog/2021/04/22/evolving-kubernetes-networking-with-the-gateway-api/

xref: https://github.com/metallb/metallb/issues/847

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,112 @@
...
+   the definition of CRDs.
+1. New objects will be stored in etcd.
+1. Based on the supplied configuration, All the controllers monitor API server looking for creation of
+   AddressPools/Peers/BgpCommunities resource.
+1. The CRD's controller will use golang based template to convert the configuration object to
+   its crossponding part of the final ConfigMap object.
+1. The generated ConfigMap object will be written to etcd so MetalLB's
+   components can consume it as if the adminstrator directly used ConfigMap.
+
+### CRD Extendability
```
johananl, line 86: This needs more work. Being able to extend CRDs isn't the issue here. We are talking about finding a good data model which can be flexible enough for the currently-known use cases and hopefully some future ones, too.

In other words, it's not enough to say "we can add field X to resource Y", we need to actually figure out if it **makes sense** to do.

For example, consider the following two use cases. They might look identical but they aren't.

1. Announce pool P1 only from node group N1. Do not announce pool P1 from any other node groups.
  - Node group N1 may also announce pool P2.
  - Pool P1 may NOT be announced by node group N2.
2. Announce only pool P1 from node group N1. Do not announce any other pools from group N1.
  - Node group N1 may NOT announce pool P2.
  - Pool P1 may be announced by node group N2.

Given the above, what should we do? Add a node/peer selector to pools? Add a pool selector to peers/nodes? Do both? How do the two knobs interact? Have we thought about side effects?

The above is just a **sample** problem. I'm sure there are more we haven't considered. For example, what about cases where node N peers with routers R1 and R2 and wants to advertise a prefix to R1 but not to R2?

I feel like we need to dedicate more thought to this and discuss such topics in the proposal. As discussed before, **flexibility is a major concern** when touching the config surface. It would be a mistake to add some fields to the CRDs just to get the CRD PR out the door. We need to see that we are _actually_ solving existing problems and aren't introducing new ones. I'm aware we can't foresee everything but we can still think about cases that are already known to us e.g. from open issues.

johananl (changes requested):
Thanks for iterating over this @msherif1234. This proposal needs more work - see my inline comments.

I saw that you've opened #850. I appreciate the enthusiasm, however we aren't settled on the design yet... I suggest you wait with any implementation work until the design looks solid and gets a thumbs-up from the maintainers.

design/0001-metallb-crd.md
msherif1234: it sounded like you are looking for Node pool announcement ACL conifg with implicit deny something like, assuimg we use
Ids for pool assignments :- 

nodeSelectorGroup: ng1
  node1
  node2

nodeSelectorGroup: ng2
  node3

nodeSelectorGroup ng1:
   Accept: Pool >= Pool1 && Pool < Pool10  // range [1:10]

nodeSelectorGroup ng2:
    Accept:  Pool != Pool1

which another object abd can be extended to Peers too

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,112 @@
...
+
+### Design Overview
+
+[Metallb CRD Design diagram](https://excalidraw.com/#json=5898667890311168,GWkLmmW6WMvZu6JdisIT0Q)
+
+![Metallb CRD Design diagram](./metallb-crd.png)
+
+This proposal will implement three custom resource objects :-
+
+- [AddressPools](https://github.com/metallb/metallb/pull/850/files#diff-e700ddc5833f14f3e89d361ddfc1b1965022f7f15a5fc46dcdba3185f8cafe7b)
```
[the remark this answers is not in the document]
msherif1234: I disagree it make sense ti group all pools under list object and u still can update specific pool and leave other unchanged if we created resource for every pool/peer entries will endup with large number of resources which will make it hard to maange

design/0001-metallb-crd.md
markdgray: Maybe we could add a commit(s) to this PR to help understand how we would use the CRD and what it would look like. I don't think we will be able to cover every use case but it would at least give us (increasing) confidence on the structure. For example, we could include the following:

1. List of sample uses cases: I don't think each needs to be very detailed (perhaps a single line for each like the example @johananl specified above `Announce pool P1 only from node group N1. Do not announce pool P1 from any other node groups.`). @msherif1234 maybe you can start this? You could use the unit tests for inspiration and then you could get feedback from the rest of the community on them.

2. The proposed CRD or CRDs

3. Examples of how the CRD(s) from (2) could be used to address the use cases in (1). This could show object definitions, for example.

I think if we don't approach it in a structured way, we may miss some corner cases.

markdgray: On a separate point, I think it would be good to outline some kind of plan for how we would version this API? Also, how (and if) it should align to MetalLB releases.

msherif1234: @markdgray I have been thinking about nodeselector option today its under Peers conifg I think if needs to move up in the level so it can work for both addresspools and peers I will share what I have once its ready

design/0001-metallb-crd.md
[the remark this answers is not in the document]
johananl: I'm not sure why individual resources would be harder to manage than one big resource. Care to explain?

As a matter of fact, we can argue the opposite: With one big resource, modifying e.g. one BGP peer would require iterating through an array of peers. With individual resources, you access just the relevant resource by name.

It doesn't make sense to me to manage multiple BGP peers in a single resource just as it doesn't make sense to me to do so for `Service` or `Deployment` objects.

Lastly, you've mentioned fine-grained RBAC as a motivation for the proposal. How would you grant a user, say, access to peer X but not to peer Y if all peers are under one resource?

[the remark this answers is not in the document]
champtar: If we can have multiple AddressPools this is the best of both worlds IMO, example I create an AddressPools for application X that need IPs in 2 different ranges (management and public). For sure having only 1 global AddressPools I need to get / modify / apply and we are no better than the configmap.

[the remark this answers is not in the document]
msherif1234: all valid points I will divide even further

msherif1234:
@johananl @russellb   what do you think about the following objects as the building blocks for Metallb
- node selectors CRD which is a general objects to specify the node(s) of interest
```
node-selectors:
    - match-labels:
    - match-expressions:
```
- protocol can now be layer2 or bgp but with FRR capability we can add ospf/is-is and other protocols 
```
protocol:
```

- address pools object (existing object w/o protocol and bgp advertisments
```
address-pools:
    - name:
      addresses:
      auto-assign:
      avoid-buggy-ips:
```
- BGP peers object (today object w/o node selector)
```
peers:
    - peer-address:  // we can make it also subnet if we decide to support dynamic neighbors ?
     peer-asn:
     my-asn:
```

- bgp communities object (as is)
```
bgp-communities:
   no-advertise: 
```
- bgp advertisements object *used to be part of address pools)
```
bgp-advertisements:
      - aggregation-length: 
        localpref: 
        communities:
        - no-advertise
      - aggregation-length:    
```
for each of the above object we will have a CRD definition , I am not sure if it make sense to bundle peer and other 2 BGP objects in as single object or its more flexible keeping them separate as shown above

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,163 @@
...
+
+- Address pools object (existing object w/o protocol and bgp advertisments)
+
+```
+address-pools:
+    - name:
+      addresses:
+      auto-assign:
+      avoid-buggy-ips:
+```
```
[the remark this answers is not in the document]
msherif1234: oh I meant to have singular as it been ask b4 and I still have it (s) I will update it , but yes the plan to have one instance for each CRD, sorry for that

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,163 @@
...
+
+- BGP advertisements object *used to be part of address pools)
+
+```
+bgp-advertisements:
+      - aggregation-length: 
+        localpref: 
+        communities:
+        - no-advertise
+```
```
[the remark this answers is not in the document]
msherif1234: sure this one way but the concern @johananl might have is having big CRD , in what I have if we have OSPF for example or anything else we will have new CRD 
```
ospf-neighbour:
 -
```
so no impact to already defined CRDs for address pool we just need to another value to protocol enum and then create new CRD or CRDs  that is/are protocol specific

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,163 @@
...
+
+- Node selectors CRD which is a general objects to specify the node(s) of interest
+
+```
+node-selectors:
+    - match-labels:
+    - match-expressions:
+```    
+
+- Protocol can now be layer2 or bgp but with FRR capability we can add ospf/is-is and other protocols
```
SchSeba, line 72: should the operator support different protocols on different nodes?

msherif1234: we have node selector to segment your cluster , but my understanding today customers uses Metallb either in L2 or BGP but not both, others pls. correct me here if I am wrong.

design/0001-metallb-crd.md
russellb: just that it seems more like an attirbute than an independent object to me, and also seems more consistent with other places the selector pattern exists

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,163 @@
...
+     peer-asn:
+     my-asn:
+```
+
+- BGP communities object
+
+```
+bgp-communities:
+   no-advertise: 
+```
```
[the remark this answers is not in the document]
russellb: Right, but `bgp-communities` just defines aliases that you can then refer to under `bgp-advertisements.communities`.  It's the same info, just a place to define aliases off to the side.

I was thinking that if we just had another way to give the numbers a friendly name, the alias definitions off to the side aren't really required.

design/0001-metallb-crd.md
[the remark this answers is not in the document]
russellb: Somehow these BGP advertisement options must be associated with an address pool.  Can you describe an alternative for how you'd have them tied together?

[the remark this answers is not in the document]
markdgray: Would we want to use the same BGPAdvertisement for multiple AddressPools? i.e. have a default one.

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,228 @@
...
+node-selector:
+    - match-labels:
+    - match-expressions:
+```    
+
+- Protocol can now be layer2 or bgp but with FRR capability we can add ospf/is-is and other protocols
+
+```
+protocol:
```
markdgray, line 81: I don' understand how this will work. Why would this not be part of the AddressPools object?

design/0001-metallb-crd.md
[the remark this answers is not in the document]
msherif1234: the way its used todaay its tied up to address-pool definition but your question is valid maybe it doesn't haeve to be 1:1 relationship @johananl can you share ur thoughts here ?

design/0001-metallb-crd.md
```diff
@@ -0,0 +1,163 @@
...
+        localpref: 
+        communities:
+        - no-advertise
+```
+
+### Configuration Flow
+
+1. User will supply the yaml configuration
+1. Configuration will get authenticated and authorized based on RBAC rules.
+1. Supplied configuration will get verified using openAPIV3Schema based on
```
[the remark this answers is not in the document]
SchSeba: are all the validations are simple and can be handle by the crd-validation?

for example duplicate IP pools, duplicate `asn` between the `peers` and `my` etc..

The upgrade webhook is also important, will you be able to add a note about that on this PR?

rata:
IIUC this proposal is superseeded by this one now: https://github.com/metallb/metallb/pull/859. Let me know if I'm missing something and need to review this. If not, can we close this PR? :)

russellb:
> IIUC this proposal is superseeded by this one now: #859. Let me know if I'm missing something and need to review this. If not, can we close this PR? :)

Yeah, I'm happy with considering CRD based configuration a feature of the metallb-operator for now, and then we can revisit potentially making it a core feature in the future.

johananl:
@msherif1234 following the discussion above, would it make sense to you to close this PR for now and move the effort to https://github.com/metallb/metallb-operator?

msherif1234:
> @msherif1234 following the discussion above, would it make sense to you to close this PR for now and move the effort to https://github.com/metallb/metallb-operator?
sure , I can reference discussions and questions when needed

Issue #277, opened 2018-07-17
Add an option to pass a list of interfaces to create announcers on


RostyslavFridman:
At the moment layer2 mode creates ARP and NDP announcers on all interfaces that are up and of broadcast type.
However, it might be better to have an ability to pass a list of interfaces to create announcerson as a configuration option.
Due to the current code limitations it can be done only as an option passed to the metallb executable itself, not as a ConfigMap option.

Thus, I propose to:

- Add an optional argument to the metallb executable with a list of comma-separated interfaces to create ARP and NDP announcers on
- Pass these interfaces to `updateInterfaces()` method
- Make a decision, whether to create announcer with a condition
```
if len(a.bindInterfaces) > 0 && !isBindInterface(ifi.Name, a.bindInterfaces) {
	continue
}
```

RostyslavFridman:
We have a custom configuration with multiple VLAN interfaces on kubernetes nodes. It doesn't seem like a good idea to listen on all interfaces from security perspective.

Moreover, due to implementation specifics openvswitch needs two interfaces for VLANs to work. When metallb sends ARP requests, it sends them from both of these interfaces and node can't decide where to send traffic. This effectively breaks load balancing.

kfox1111:
I'm in the same boat.

danderson:
Thanks for the details. I have some thoughts.

First, you won't gain anything from a security perspective. Kubernetes itself will happily handle traffic for the service IP no matter where the packets come from. The only change MetalLB makes is that well-behaved clients on those other networks can find the service IP, if the network config is compatible (i.e. they're in the same subnet as the service IP. An attacker can still just edit their ARP neighbor tables by hand and successfully contact the service IP. So, there's no security benefit to restricting the listening interfaces.

The setup you have that breaks LB is interesting. Could you give me more info about that, like `ip addr` on the machine that shows the two conflicting interfaces?

Again, I'm trying to figure out if there's a way we can make your situation work without adding another configuration option. This is especially important to me because people will think that it gives them security, when that is 100% not true... So then I'll get even more bugs that the "security options" are broken.

If we _must_ add this option, then I have a strong preference for building it into the config, not adding a commandline flag. This requires a bit of refactoring of the codebase to support dynamically adding/removing listening interfaces, but that means there's only one place for configuration.

kfox1111:
[...]

PR #834, opened 2021-04-08
Allow to include / exclude interfaces via regexp

champtar:


rata:
@champtar Thanks! I think in some issues it is mentioned to not do this because it can give a false sense of security. Not sure if there were more reasons, have you seen them? I don't have them handy, but have a look if you didn't :)

Is there a problem you hit in your setups due to not having this? You mentioned briefly in the meeting today about having lot of interfaces, what is the problem/what you want to improve?

I wonder if we can do the right thing without exposing complexity to the user (something automatic, like ignore some prefix only used for kernel virtual interfaces, or something like that)

champtar:
It's definitly not for security, people thinking that need a refresher in networking and I have no issue making that explicitly with comments in the conf.

When a VIP move to a host we spam all interfaces and there is no point spaming 50 interfaces instead of just 1 or 2, and the spam loop is only one loop
I tried detecting if this interface is the host side of a veth but haven't found a way, so the next best thing is taking advantage of the CNIs using explicit names like `cali*`

Thinking about it more, another way to go about it is to add an option to ignore all interfaces that have no IPv4 and no IPv6 IPs except link-local

champtar:
I looked at using link-netns(id) but it's no reliable
inside the container:
```
# ip a
3: eth0@if10: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether 02:a8:a9:95:ae:e7 brd ff:ff:ff:ff:ff:ff link-netnsid 0
    inet 100.64.90.1/32 brd 100.64.90.1 scope global eth0
       valid_lft forever preferred_lft forever
    inet6 fe80::a8:a9ff:fe95:aee7/64 scope link 
       valid_lft forever preferred_lft forever
```
on the host
```
# ip a
10: califb3eb82ef50@if3: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default 
    link/ether ee:ee:ee:ee:ee:ee brd ff:ff:ff:ff:ff:ff link-netns cni-af6ac730-d5ea-004f-eb5b-cf18374ba143
    inet6 fe80::ecee:eeff:feee:eeee/64 scope link 
       valid_lft forever preferred_lft forever

# ip netns 
cni-af6ac730-d5ea-004f-eb5b-cf18374ba143 (id: 0)
``` 

But with (golang) netlink you only get the netnsid
```
2021/04/15 01:13:45 Name: lo Type: device Index: 1 MasterIndex: 0 ParentIndex: 0 OperState: unknown Slave: <nil> Flags: up|loopback Protinfo: <nil> EncapType: loopback Namespace: <nil> Alias:  Promisc: 0 NetNsID: -1 Vfs: [] RawFlags: 65609
2021/04/15 01:13:45 Name: ens3 Type: device Index: 2 MasterIndex: 0 ParentIndex: 0 OperState: up Slave: <nil> Flags: up|broadcast|multicast Protinfo: <nil> EncapType: ether Namespace: <nil> Alias:  Promisc: 0 NetNsID: -1 Vfs: [] RawFlags: 69699
2021/04/15 01:13:45 Name: kube-ipvs0 Type: dummy Index: 3 MasterIndex: 0 ParentIndex: 0 OperState: down Slave: <nil> Flags: broadcast Protinfo: <nil> EncapType: ether Namespace: <nil> Alias:  Promisc: 0 NetNsID: -1 Vfs: [] RawFlags: 130
2021/04/15 01:13:45 Name: vxlan.calico Type: vxlan Index: 4 MasterIndex: 0 ParentIndex: 0 OperState: unknown Slave: <nil> Flags: up|broadcast|multicast Protinfo: <nil> EncapType: ether Namespace: <nil> Alias:  Promisc: 0 NetNsID: -1 Vfs: [] RawFlags: 69699
2021/04/15 01:13:45 Name: nodelocaldns Type: dummy Index: 9 MasterIndex: 0 ParentIndex: 0 OperState: down Slave: <nil> Flags: broadcast Protinfo: <nil> EncapType: ether Namespace: <nil> Alias:  Promisc: 0 NetNsID: -1 Vfs: [] RawFlags: 130
2021/04/15 01:13:45 Name: califb3eb82ef50 Type: veth Index: 10 MasterIndex: 0 ParentIndex: 3 OperState: up Slave: <nil> Flags: up|broadcast|multicast Protinfo: <nil> EncapType: ether Namespace: <nil> Alias:  Promisc: 0 NetNsID: 0 Vfs: [] RawFlags: 69699
```
In the container
```
2021/04/15 01:18:10 Name: lo Type: device Index: 1 MasterIndex: 0 ParentIndex: 0 OperState: unknown Slave: <nil> Flags: up|loopback Protinfo: <nil> EncapType: loopback Namespace: <nil> Alias:  Promisc: 0 NetNsID: -1 Vfs: [] RawFlags: 65609
2021/04/15 01:18:10 Name: eth0 Type: veth Index: 3 MasterIndex: 0 ParentIndex: 10 OperState: up Slave: <nil> Flags: up|broadcast|multicast Protinfo: <nil> EncapType: ether Namespace: <nil> Alias:  Promisc: 0 NetNsID: 0 Vfs: [] RawFlags: 69699
```

you see that NetNsID is 0 in both case so I don't have an easy heuristic to exclude interfaces

the test code
```
package main

import (
	"log"

	"github.com/vishvananda/netlink"
)

func main() {
	links, _ := netlink.LinkList()
	for _, l := range links {
		log.Print("Name: ", l.Attrs().Name,
			" Type: ", l.Type(),
			" Index: ", l.Attrs().Index,
			" MasterIndex: ", l.Attrs().MasterIndex,
			" ParentIndex: ", l.Attrs().ParentIndex,
			" OperState: ", l.Attrs().OperState,
			" Slave: ", l.Attrs().Slave,
			" Flags: ", l.Attrs().Flags.String(),
			" Protinfo: ", l.Attrs().Protinfo,
			" EncapType: ", l.Attrs().EncapType,
			" Namespace: ", l.Attrs().Namespace,
			" Alias: ", l.Attrs().Alias,
			" Promisc: ", l.Attrs().Promisc,
			" NetNsID: ", l.Attrs().NetNsID,
			" Vfs: ", l.Attrs().Vfs,
			" RawFlags: ", l.Attrs().RawFlags)

		/*
			if veth, ok := l.(*netlink.Veth); ok {
				log.Print("PeerHardwareAddr ", veth.PeerHardwareAddr, " PeerName ", veth.PeerName)
			}
		*/
	}
}
```

rata:
But is there a _problem_ caused by spaming many interfaces? Not sure I got that part.

Another option, if we want to solve that, is to exclude interfaces usually created by calico and others. Like  hardcode in the code the patterns "cali*", etc. To be safe, we can expose a bool knob to enable/disable this, but not allow tuning the interfaces names and clearly document it is not about security.

I'd like to understand why to do this, though. Is it slow to converge otherwise?

champtar:
There is no functional problem, spamming on the veth to the containers is "ok"

This is a performance optimization, if you have 100 containers and 5 VIPs, you have 100 useless goroutines (one per interface), and when spamming you will spend time to send 505 packets instead of 5 every 1.1 seconds

Now if we exclude the interfaces without IPs, we also fix another actual issue that I've seen multiple times:
an interface is connected and up in the same vlan as the real interface (miscommunication, someone though you needed a bond or configured the wrong vlan)
When this happens, depending on the order the advertisement happens, the switch will decide to send packets via the interface without IP, and the host will discard the incoming packets because `rp_filter = 1`, and you waste a long time debugging.
This would improve the just works side of MetalLB ;)

champtar:
Thinking about it more filtering out interfaces without IPs even if a good default need to be configurable because that can be a working usecase. The safest and easiest to understand for everyone is the regexp I think.

> Like hardcode in the code the patterns "cali*", etc.

I really don't like to hardcode this setting as I've asked multiple times on slack (metallb and kubespray) and got almost no answers for the names that are used by other CNIs (so this will be an incomplete list likely for a long time).
Also we should not put arbitrary restrictions in MetalLB just because some people might misuse them even after reading the documentation

champtar:
Another way to do that is to ignore all interfaces of type veth by default, this would need to be configured when you run k8s inside a container (kind, lxc, ...), that would be the most "just works" option

russellb:
related to issue #277

champtar:
I will think of a no config approach

PR #837, opened 2021-04-17
Support endpoint slices

### LINKED ISSUE #811: Enable metallb to use endpoint slices
This is an enhancement request. 

K8s is in the transition to switch from endpoints to endpoint slices since K8s 1.17. 
Starting with 1.19, kube-proxy will by default use endpoint slices. 
To reduce the risk the metallb and kube-proxy perceive a different state of the world, metallb should eventually consume endpoint slices as well. 
Since it might vary when people switch over to using endpoint slices for reasons, it might make sense to have this configurable. 

* https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/
* https://kubernetes.io/docs/tasks/administer-cluster/enabling-endpointslices/

fedepaol:
Endpoint slices are a more scalable way to distribute service endpoints.

Here we check if they are enabled on the cluster, and we set up the
listeners and the logic to use them, otherwise we fallback to use
endpoints.

Fixes #811


russellb:
Thanks for the PR!  How did you test this?  Did you try it in the metallb kind-based dev env?  I see the unit test cases you added, so thank you for that!

fedepaol:
> Thanks for the PR! How did you test this? Did you try it in the metallb kind-based dev env? I see the unit test cases you added, so thank you for that!

This is one of the things I wanted feedback on. I deployed a kind + ovnk cluster (where I knew ep slices were enabled), deployed there and checked that the events were propagated up to the listeners (with extra logs, now removed).
Once there, the other segment of the path is covered by the unit tests. Not sure if there's a better way to do that.

russellb:
> > Thanks for the PR! How did you test this? Did you try it in the metallb kind-based dev env? I see the unit test cases you added, so thank you for that!
> 
> This is one of the things I wanted feedback on. I deployed a kind + ovnk cluster (where I knew ep slices were enabled), deployed there and checked that the events were propagated up to the listeners (with extra logs, now removed).
> Once there, the other segment of the path is covered by the unit tests. Not sure if there's a better way to do that.

Thanks. I was mostly curious. We lack any functional testing in CI right now. I expect that to add it we'll base it on the existing dev-env.

markdgray:
> > > Thanks for the PR! How did you test this? Did you try it in the metallb kind-based dev env? I see the unit test cases you added, so thank you for that!
> > 
> > 
> > This is one of the things I wanted feedback on. I deployed a kind + ovnk cluster (where I knew ep slices were enabled), deployed there and checked that the events were propagated up to the listeners (with extra logs, now removed).
> > Once there, the other segment of the path is covered by the unit tests. Not sure if there's a better way to do that.
> 
> Thanks. I was mostly curious. We lack any functional testing in CI right now. I expect that to add it we'll base it on the existing dev-env.

There is a proposal [here](https://github.com/metallb/metallb/pull/849), @fedepaol, maybe you could have a look and see if this is something that could potentially be used or if there are any limitations for why it could not be?

fedepaol:
> > > > Thanks for the PR! How did you test this? Did you try it in the metallb kind-based dev env? I see the unit test cases you added, so thank you for that!
> > > 
> > > 
> > > This is one of the things I wanted feedback on. I deployed a kind + ovnk cluster (where I knew ep slices were enabled), deployed there and checked that the events were propagated up to the listeners (with extra logs, now removed).
> > > Once there, the other segment of the path is covered by the unit tests. Not sure if there's a better way to do that.
> > 
> > 
> > Thanks. I was mostly curious. We lack any functional testing in CI right now. I expect that to add it we'll base it on the existing dev-env.
> 
> There is a proposal [here](https://github.com/metallb/metallb/pull/849), @fedepaol, maybe you could have a look and see if this is something that could potentially be used or if there are any limitations for why it could not be?

Will try to give it a spin as soon as I have some time, thanks for pointing out

speaker/bgp_controller.go
```diff
@@ -83,22 +84,50 @@ newPeers:
...
+					if ep.Conditions.Ready == nil {
+						continue
+					}
+
+					if _, ok := ready[addr]; !ok && *ep.Conditions.Ready {
+						// Only set true if nothing else has expressed an
+						// opinion. This means that false will take precedence
+						// if there's any unready ports for a given endpoint.
+						ready[addr] = true
+						fmt.Println("Setting true for addr", addr)
```
markdgray, line 123: Should these be "l.Log()"?

speaker/bgp_controller.go
```diff
@@ -83,22 +84,50 @@ newPeers:
 }
 
 // nodeHasHealthyEndpoint return true if this node has at least one healthy endpoint.
-func nodeHasHealthyEndpoint(eps *v1.Endpoints, node string) bool {
+func nodeHasHealthyEndpoint(eps k8s.EpsOrSlices, node string) bool {
```
markdgray, line 87: I'm wondering could these two functions (and the similar ones in the l2 case) could somehow be combined into one function or multiple functions that lived in the k8s package in a way that could hide the details of Eps versus EpSlices to users of the k8s package.

What do you think @fedepaol ?

internal/k8s/endpoint_slices.go
```diff
@@ -0,0 +1,20 @@
...
+type EpsOrSlices struct {
+	Type      EpsOrSliceType
+	EpVal     *v1.Endpoints
+	SlicesVal []*discovery.EndpointSlice
+}
+
+type EpsOrSliceType int
+
+const (
+	Unknown EpsOrSliceType = iota
```
markdgray, line 17: Do we need an "Unknown" case?

internal/k8s/k8s.go
```diff
@@ -68,17 +72,18 @@ const (
...
-	Kubeconfig    string
-
-	ServiceChanged func(log.Logger, string, *v1.Service, *v1.Endpoints) SyncState
+	ProcessName        string
+	ConfigMapName      string
+	ConfigMapNS        string
+	NodeName           string
+	MetricsHost        string
+	MetricsPort        int
+	ReadEndpoints      bool
```
markdgray, line 81: Any idea why we would not "ReadEndpoints"?

internal/k8s/k8s.go
```diff
@@ -68,17 +72,18 @@ const (
...
+	ReadEndpointSlices bool
```
markdgray, line 82: At the moment, this isn't used. Do we need it? Can you think of a good reason to disable this if EndpointSlices were enabled on the cluster?

internal/k8s/k8s.go
```diff
@@ -157,30 +164,74 @@ func New(cfg *Config) (*Client, error) {
...
+				c.epIndexer, c.epInformer = cache.NewIndexerInformer(epWatcher, &v1.Endpoints{}, 0, epHandlers, cache.Indexers{})
+
+				c.syncFuncs = append(c.syncFuncs, c.epInformer.HasSynced)
+			} else {
+				c.logger.Log("op", "New", "msg", "using endpoint slices")
+				slicesHandlers := cache.ResourceEventHandlerFuncs{
+					AddFunc: func(obj interface{}) {
+						slice, ok := obj.(*discovery.EndpointSlice)
+						if !ok {
+							return
```
markdgray, line 198: Should we be logging these errors?

speaker/bgp_controller_test.go
```diff
@@ -222,21 +223,730 @@ func TestBGPSpeaker(t *testing.T) {
...
+				"1.2.3.4:0": {
+					{
+						Prefix: ipnet("10.20.30.1/32"),
+					},
+				},
+			},
+		},
+
+		{
+			desc:     "Endpoint on our node has some unready ports",
```
markdgray, line 402: Why does this not advertise anything? The traffic policy is "local" but pandora does have at least one endpoint available for this service.

markdgray:
> > Thanks for the PR! How did you test this? Did you try it in the metallb kind-based dev env? I see the unit test cases you added, so thank you for that!
> 
> This is one of the things I wanted feedback on. I deployed a kind + ovnk cluster (where I knew ep slices were enabled), deployed there and checked that the events were propagated up to the listeners (with extra logs, now removed).
> Once there, the other segment of the path is covered by the unit tests. Not sure if there's a better way to do that.

Is there an easy way to enable Kind with endpoint slices enabled? Or do you need a network provider that enables them?

speaker/bgp_controller.go
fedepaol: Ooops this is a debug leftover

internal/k8s/k8s.go
fedepaol: Nope, this is a leftover of a previous version where I intended this to be configurable. Now it's driven by the same ReadEndpoints variable. I'll remove this

internal/k8s/k8s.go
fedepaol: That was already there, and the reason is because the same k8s client is used by the controller that doesn't need it:
https://github.com/metallb/metallb/blob/0c83b4ba1164ba27a64084a6b0bde12a38c2d24a/controller/main.go#L153

speaker/bgp_controller.go
fedepaol: Not sure I get the comment right. My understanding is, we can have a `k8s.nodeHasHealthyEndpoint(svc, node)` function that abstracts away the endpoints source.
By doing that, we would introduce a dependency of `shouldAnnounce` to the state of the cache which I am not entirely sure I like.
This would make testing a bit harder, and not sure the extra complexity would add corner cases. 

I like the fact that we fetch everything we need in the k8s package and we explicitly pass them to `shouldAnnounce`. 
Said that, if you think it's better the other way around I can rework this.

internal/k8s/endpoint_slices.go
fedepaol: I think (but it's been some time) I added this when I had some failures in tests because I forgot to change the type. I thought being explicit about the values instead of having a meaningful 0-value was better.

internal/k8s/k8s.go
fedepaol: This is really unlikely to happen (or should always happen if we wire the controller in the wrong way). This is probably the reason why `epHandlers` didn't have it (and I followed the same logic), but I guess extra logging won't hurt.

speaker/bgp_controller_test.go
fedepaol: In case of traffic policy = local, having at least one non working endpoint is enough for not advertising the node:
https://github.com/metallb/metallb/blob/0c83b4ba1164ba27a64084a6b0bde12a38c2d24a/speaker/bgp_controller.go#L118

This is the same in the test I cloned.

fedepaol:
> > > Thanks for the PR! How did you test this? Did you try it in the metallb kind-based dev env? I see the unit test cases you added, so thank you for that!
> > 
> > 
> > This is one of the things I wanted feedback on. I deployed a kind + ovnk cluster (where I knew ep slices were enabled), deployed there and checked that the events were propagated up to the listeners (with extra logs, now removed).
> > Once there, the other segment of the path is covered by the unit tests. Not sure if there's a better way to do that.
> 
> Is there an easy way to enable Kind with endpoint slices enabled? Or do you need a network provider that enables them?

I need to check, I think it's more about the k8s version than the CNI enabling it, but I may be wrong.

markdgray:
> I need to check, I think it's more about the k8s version than the CNI enabling it, but I may be wrong.

You are probably right. It would be good to know how it could be enabled or disabled as I could add it as part of the e2e tests that I am doing.

speaker/bgp_controller.go
markdgray: Sorry for not being clear. There seems to be duplicate code in nodeHasHealthyEndpoints(), healthyEndpointExists(), usableNodes() so I was just wondering if there was a way to combine them somehow. However, looking at them again, its mostly looping over endpoints structures so maybe not. What do you think?

internal/k8s/endpoint_slices.go
markdgray: Ok, so you think it is better to be in?

internal/k8s/k8s.go
markdgray: The only reason I thought so was that I couldn't see an easy way to detect if that failure happened.

speaker/bgp_controller_test.go
markdgray: nodeHasHealthyEndpoint() shoud return true if "At least one fully healthy endpoint on this machine.". This is different to your statement above (having at least one non working endpoint is enough for not advertising the node). The host "Pandora" has one ready endpoint so it should announce? I must be missing something as I see the cloned test case is the same and it does pass.

speaker/bgp_controller.go
fedepaol: Ok, if this is about refactoring them and removing duplication, I think there's room for improvement.

internal/k8s/endpoint_slices.go
fedepaol: Well, now that we have unit tests maybe it's not necessary. It would be better to have it explode than silently skipping an if because we forgot to set the right type somewhere.

I will remove it.

fedepaol: I think I remember (or, the unit test reminded me).
With this extra value, when we create an empty `EpsOrSlices{}` we know both fields are nil.
On the other hand, with the default value being Eps (or slices), we would need to initialize the field (or have not nil checks around the code).

I'd leave it there.

speaker/bgp_controller.go
```diff
@@ -83,48 +84,50 @@ newPeers:
 }
 
 // nodeHasHealthyEndpoint return true if this node has at least one healthy endpoint.
```
markdgray, line 86: Thanks for indulging me. It seems to have reduced the amount of code a little bit. This comment is wrong now.

speaker/bgp_controller_test.go
markdgray: Ok, so one Endpoint 2.3.4.6 is not ready on Pandora and the other is - so overall it is not Ready ... that is subtle!

markdgray:
Overall this looks good to me apart from a small nit in which the comment doesn't match the updated code. The unit tests looks ok but I can't do an E2E test. @fedepaol  if you can think of a way to force endpoint slices in Kind, please let me know and I can add an E2E test for this.

fedepaol:
> Overall this looks good to me apart from a small nit in which the comment doesn't match the updated code. The unit tests looks ok but I can't do an E2E test. @fedepaol if you can think of a way to force endpoint slices in Kind, please let me know and I can add an E2E test for this.

I finally made my head clear about that (thanks to @aojea). So, the state of epslices is:
- beta version is enabled by default in k8s 1.18
- the resource is available in 1.17, but the objects are not getting created unless we enable the feature gate

This means that:
- having a 1.18 version is enough for epslices to be available and used (regardless of the CNI)
- my detection code is actually buggy, because in 1.17 we could have the objects but not the controller producing them

speaker/bgp_controller_test.go
fedepaol: Indeed :-)

fedepaol:
When following another issue, I also incidentally found another flaw in how we handle the state: https://github.com/coredns/coredns/pull/4632
So it turns out nil ready means ready.
Will change that too

speaker/bgp_controller.go
```diff
@@ -83,48 +84,50 @@ newPeers:
...
-			// At least one fully healthy endpoint on this machine.
-			return true
-		}
-	}
-	return false
-}
+	case k8s.Slices:
+		for _, slice := range eps.SlicesVal {
+			for _, ep := range slice.Endpoints {
+				node := ep.Topology["kubernetes.io/hostname"]
```
aojea, line 110: Topology is going to be deprecated, you have to use NodeName https://github.com/kubernetes/kubernetes/commit/9613f79adde3e0080fe25576cb37a74f25262848

speaker/layer2_controller.go
```diff
@@ -41,20 +42,43 @@ func (c *layer2Controller) SetConfig(log.Logger, *config.Config) error {
...
-			if _, ok := usable[*ep.NodeName]; !ok {
-				usable[*ep.NodeName] = true
+		}
+	case k8s.Slices:
+		for _, slice := range eps.SlicesVal {
+			for _, ep := range slice.Endpoints {
+				if ep.Conditions.Ready == nil || !*ep.Conditions.Ready {
+					continue
+				}
+				nodeName := ep.Topology["kubernetes.io/hostname"]
```
aojea, line 70: ditto, topology is going to be removed soon https://github.com/kubernetes/kubernetes/commit/9613f79adde3e0080fe25576cb37a74f25262848 , use `ep.NodeName`

speaker/bgp_controller.go
fedepaol: @aojea if I use NodeName, this will make this instance compatible with 1.22 only right?

aojea: ups, wait, this is featuregated, you are right,

aojea: so ... I think that you have to use topology meanwhile we are in discoveryv1beta1, and add a comment explaining the conversion that has to be done when it moves to discoveryv1 ...
how is metallb compatibility matrix with kubernetes? at some point when discoverv1beta1 is removed from the apiserver the project has to take a decision

fedepaol:
So, CI looks good. From 1.18, epslices are enabled by default.
I tried to run the same commands locally and it's confirmed by the `using endpoint slices` log I added.

@markdgray I think this is good now.

russellb:
just to be explicit, @markdgray since you had been reviewing this, i'm going to wait for your ACK before proceeding

markdgray:
Ack. LGTM.

PR #839, opened 2021-04-17
Bump Golang to 1.16

champtar:
Golang 1.13 has horrible bugs that might have affected MetalLB
as it's using client-go / watches,
see https://github.com/kubernetes/kubernetes/issues/87615
Newer client-go version (that we are now using) have implemented
HTTP/2 health check by default but this is a workaround not a fix.

Also Golang 1.13 is not supported anymore

This replace #816

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,9 +2,9 @@
 
 version: 2
 jobs:
-  test-1.13:
+  test-1.16:
     docker:
-      - image: cimg/go:1.13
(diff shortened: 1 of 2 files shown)


russellb:
I've seen that kube bug in the wild.  That was a painful one.  Moving to latest golang makes sense, and is also in the spirit of #813

daxmc99:
Thanks @champtar!

PR #840, opened 2021-04-17
Log compiler, compiler version and architecture on startup

champtar:


russellb:
works fine for me, seems like a harmless change, thanks

PR #841, opened 2021-04-17
helm lint: disable CheckVersionIncrement

champtar:
The version will be increased when a release is done


champtar:
@gclawes as we don't auto release we can't enforce version bump check

gclawes:
Makes sense.  Looks good to me.

Issue #718, opened 2020-09-17
Add layer2 configuration to dev-env


russellb:
I recently added support for automatically configuring MetalLB in BGP mode in the dev env by running `inv dev-env` with `--protocol bgp` or `-p bgp`.  It would be helpful to add similar support for configuring MetalLB in layer2 mode.  It would be run with `--protocol layer2` or `-p layer2`.

What I would expect this to do is to configure MetalLB with a minimal `ConfigMap` that includes a small pool of usable IP addresses.  The address pool should be generated dynamically by checking the docker network configuration for the `kind` network as the address pool must be a subset of the `kind` network's subnet.  Ideally we can also update the docker network configuration to mark our address pool as reserved so we know the IP addresses will never be assigned to a new container on the same network.

russellb:
I've included the `help wanted` tag on this issue.  This is something I'd like to have and will likely implement eventually, but I'd be even more happy to give to someone else looking to contribute.  I'd happily test and review the patch!

russellb:
resolved by #842

PR #842, opened 2021-04-19
Layer2 config support in dev-env script

### LINKED ISSUE #559: docs: Add requirements.txt to run invoke script
Currently the release script uses invoke (`inv`), but the way to make it work is manually installing it's dependencies (`pyyaml` and `semver`).

We should make this process straight-forward (probably add a requirements file might be enough, but if you consider it is if needed, add docs too).

s1061123:
Fixed #559 #718.


tasks.py
```diff
@@ -180,7 +181,7 @@ def validate_kind_version():
     "architecture": "CPU architecture of the local machine. Default 'amd64'.",
     "name": "name of the kind cluster to use.",
     "protocol": "Pre-configure MetalLB with the specified protocol. "
-                "Unconfigured by default. Supported: 'bgp'",
+                "Unconfigured by default. Supported: 'bgp','layer2'",
```
markdgray, line 184: I see here, it specifies "unconfigured". Maybe update the doc to reflect that.

tasks.py
```diff
@@ -315,7 +358,6 @@ def bgp_dev_env():
     # Apply the MetalLB ConfigMap
     run("kubectl apply -f %s/config.yaml" % dev_env_dir)
 
-
```
markdgray, line 318: I think two lines between function is the PEP8 convention.

tasks.py
```diff
@@ -250,10 +255,48 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None, protocol=None):
...
+            '"{{range .Containers}}{{.IPv6Address}} {{end}}" kind', echo=True).stdout.strip().split(' ')
+    service_range = {}
+    for i in network_subnet_list:
+        network = ipaddress.ip_network(i)
+        if network.version == 6:
+            used_list = sorted(used_ipv6_list)
+        else:
+            used_list = sorted(used_ipv4_list)
+        service_ip_range_start = ipaddress.ip_interface(used_list[-1]) + 1
+        service_ip_range_end = ipaddress.ip_interface(used_list[-1]) + 11
```
markdgray, line 283: It's not clear what is going on here. I think you are getting the end and the start of 'used list' so I would have thought it would have been

`service_ip_range_start = ipaddress.ip_interface(used_list[0])`

Also, I don't understand the '+1' and '+11'

dev-env/layer2/config.yaml.tmpl
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  namespace: metallb-system
+  name: config
+data:
+  config: |
+    address-pools:
+    - name: dev-env-layer2
+      protocol: layer2
+      addresses:
+      - SERVICE_V4_RANGE
```
markdgray, line 12: Is this missing SERVICE_V6_RANGE? It seems to be referenced below.

s1061123: I'm wondering IPv6 support in MetalLB layer2 mode. Is that supported?

tasks.py
s1061123: This trys to get available 10 IP address fromthe next of last used IP address. If kind-master/worker uses following IPs:

- 192.1.1.2
- 192.1.1.5
- 192.1.1.20

Then try to get range '192.168.1.21-192.168.1.31' for the L2 services.

dev-env/README.md
```diff
@@ -1,17 +1,38 @@
 # dev-env
 
 This directory contains supporting files for the included MetalLB development
-environment. The environment is run using the following command:
+environment. The environment is run using the following command from the root of your git clone:
 
 ```
 inv dev-env
```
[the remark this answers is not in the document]
s1061123: This change is to add additional comment that user should execute inv command at the top of repo directory not dev-env/ directory because user tend to execute command from the same path of the document, so it is just additional advice.

Add some text but I could revert if it is not still clear. This is not so important.

tasks.py
s1061123: Update doc.

dev-env/layer2/config.yaml.tmpl
s1061123: As far as I look kind document, default is IPv4, so IPv6 is not supported as default.

https://kind.sigs.k8s.io/docs/user/configuration/

If you want to enable, it should be taken care of additional issue/PR I guess.

russellb: IPv6 is supported in layer2 mode.  It's fine if that's captured as a follow-up issue, though.  Addressing that is bigger than adding automated layer2 configuration, because we have to expose some configuration of the kind environment itself.

tasks.py
markdgray: Ok, I get it now. Maybe add a comment that you are trying to get a range of 10 addresses after the last assigned address in the node subnet. It would help understanding? What do you think?

I don't know if there is a requirement for more IP addresses. If you wanted it to be more dynamic, you could make a range from the last IP address in the used list to the last IP address in the subnet.

tasks.py
```diff
@@ -250,10 +258,48 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None, protocol=None):
...
+        if service_ip_range_start not in network:
+            raise Exit(message='network range %s is not in %s' % (service_ip_range_start, network))
+        if service_ip_range_end not in network:
+            raise Exit(message='network range %s is not in %s' % (service_ip_range_end, network))
```
markdgray, line 290: Rather than "raise", you could try again to grab a range of 9 and, if that failed, you could try 8 and so on. Otherwise, the developer would need to come into the code, figure out what was going on, and then manually reduce the number.

markdgray:
Thanks for this. It is really timely as I am trying to write some automated tests and I should be able to include L2 as well using this.

tasks.py
s1061123: I'm wondering how many service IPs are the minimum. This is a bit dilemma. If developer uses dev-env with assumption that dev-env reserves 10IPs for service, then 9 IPs,  8 IPs or less are not sufficient.

This is just a dev-env script, for developing and testing, so I suppose 10 IPs or die could be simple and reasonable tradeoff.

For example, currently dev-env does similar stuff (i.e. call raise() and quit) if kind cannot the 3nodes.

https://github.com/metallb/metallb/blob/main/tasks.py#L272-L273

tasks.py
s1061123: Add the comment.

tasks.py
markdgray: Probably best not to over-engineer a test script as you suggest. Maybe change the error to state that you are unable add 10 services IPs?

dev-env/README.md
```diff
@@ -1,17 +1,44 @@
...
+```
+inv dev-env --help
+```
+
+## Requirements
+
+* Go 1.15+
+* Python 3
+* [KIND - Kubernetes in Docker](https://kind.sigs.k8s.io/docs/user/quick-start/)
+  * And optionally [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl-linux/)
```
markdgray, line 36: Is this optional? I just happened to be installing from scratch and I noticed that I needed it.

s1061123: changed.

markdgray: Thanks

markdgray:
LGTM

russellb:
Mark, thanks for the reviews!

Tomo, fantastic job.  Thank you for this!  This worked great for me locally, so I'm going to merge it now.

PR #844, opened 2021-04-20
chart: make prometheus port configurable

### LINKED ISSUE #843: helm chart won't respect prometheus port
I wanted to try the newly added helm chart from #658 and ran into an issue with the readinessProbe on both the controller and speaker pods. The readinessProbe checks port `metrics` which is hardcoded to `7472`. The pods start without `--port=7472` so will listen on port 80 by default and the pods never reach ready state.

I tried this with a checkout of 92d6860a67695a176f5d7c1762d188b5d8144883 (0.10.0 pre-release) on kubernetes 1.18.

I was asked to mention @gclawes on Slack.

skoef:
Make sure pods listen on the `metrics` port.

Fixes #843.


gclawes:
`values.yaml`, `values.schema.json`, and `charts/metallb/README.md` all need to be updated to include `.Values.prometheus.metricsPort`

skoef:
You're right, forgot to commit values.yaml and to add it to README.md. I'm not familiar with values.schema.json, but it doesn't seem to match values.yaml already:

- for instance `.Values.prometheus.scrapeAnnotations` is missing entirely from values.schema.json
- `.Values.prometheus.prometheusRule.staleConfig` is listed directly as property of `.Values.prometheus` among others

Am I overlooking something?

gclawes:
`values.schema.json` is based on JSONSchema: https://helm.sh/docs/topics/charts/#schema-files

I'll put in a separate PR to clean up the validation for `.Values.prometheus.*`, it does appear to be out of sync from `values.schema.json`.  None of it is marked as `required` at the moment, so it's not failing.  Once that's merged, you'll be able to rebase and add validation for `.Values.prometheus.metricsPort`

gclawes:
#841 should be merged before this change.

gclawes:
I've created #846 to clean up the `.Values.prometheus.*` validations.  @skoef once that is merged, validations for `.Validations.prometheus.metricsPort` can be added

russellb:
#846 has merged

gclawes:
@skoef #854 has been added, to update the `charts/metallb/README.md` against the current values in the `main` branch.

Once that's merged, can you please re-generate the README.md in this PR with the following command:
`docker run --rm --volume "$(git rev-parse --show-toplevel)/charts:/helm-docs" -u $(id -u) jnorwood/helm-docs:latest`

Sorry for the delays around values/README, we're still smoothing out the rough edges around changes to the chart.

russellb:
I pushed a README update to the PR.  Look good to you @gclawes ?

russellb:
It looks like you dropped most of the README changes introduced by the automated tool?  See https://github.com/metallb/metallb/commit/afb1434413111f1230b6dec8c9ac0ac6cf3f9a49 .  I think we should leave all of the changes in place.  Otherwise we can't validate that the file is up to date in CI.  Thoughts @gclawes

skoef:
I didn't understand where all these additional changes came from, so I only included the relevant bits in the commit. But I now see that I reverted your change by doing this. It should be fine now!

russellb:
No worries.  Maybe @gclawes excluded that on purpose, but I'm thinking it needs to stay so we can start checking it in CI.

@gclawes can you take a look and ack the latest version of the PR?

gclawes:
I generated my changes in #854, which has since been merged.  A rebase + regenerate here should pick those up and add the necessary ones for this PR.

@skoef what command did you use to generate your diff in this PR? I'm concerned that it's adding diffs like this: https://github.com/metallb/metallb/pull/844/files#diff-9128b1f8a30429b7f861859c42d707b1f81d74bb47ab0b7d45e0b8a6bf9a781dR99
```
----------------------------------------------
Autogenerated from chart metadata using [helm-docs v1.5.0](https://github.com/norwoodj/helm-docs/releases/v1.5.0)
```
These are not present when I run it locally.

skoef:
@gclawes Perhaps you have used an older version of the container? Because I literally ran the command @russellb commented earlier.

Anyway, that's why I excluded it from the commit initially.

gclawes:
Lesson learned about cached `:latest` images, I was generating from an old image of `jnorwood/helm-docs:latest`.  We'll be sure to pin a version in CI.  The diffs introduced here for `README.md` look fine.

Otherwise this PR looks good.  @russellb

PR #848, opened 2021-04-23
tasks: Add name parameter to dev-env-cleanup

markdgray:
Signed-off-by: Mark Gray <mark.d.gray@redhat.com>

If dev clusters are created with a different name the dev-env-cleanup task cannot clean them up.


russellb:
Good catch.  I never used the `name` param when creating the dev env.  This seems fine, but I want to test it first before I hit merge.

russellb:
works for me, thanks!

PR #849, opened 2021-04-23
Re-introduce e2e test framework

markdgray:
This PR re-introduces an e2e test framework for MetalLB and integrates with the CI.


markdgray:
This is ready for review

e2etest/bgp_test.go
```diff
@@ -0,0 +1,80 @@
...
+		svcTCPPort := int(svc.Spec.Ports[0].Port)
+		ingressIP := e2eservice.GetIngressPoint(
+			&svc.Status.LoadBalancer.Ingress[0])
+
+		ginkgo.By("checking connectivity to its external VIP")
+
+		hostport := fmt.Sprintf("http://%s:%d/",
+			ingressIP,
+			svcTCPPort)
+		_, err = runCommand(true, "wget", "-O-", "-q", hostport)
```
s1061123, line 76: If `wget` failed to get the contents, then we need 900 second (15min?) for timeout because wget default timeout is 900sec. how about to add `-T <time>` with shorter time?

e2etest/layer2_test.go
```diff
@@ -0,0 +1,80 @@
...
+		svcTCPPort := int(svc.Spec.Ports[0].Port)
+		ingressIP := e2eservice.GetIngressPoint(
+			&svc.Status.LoadBalancer.Ingress[0])
+
+		ginkgo.By("checking connectivity to its external VIP")
+
+		hostport := fmt.Sprintf("http://%s:%d/",
+			ingressIP,
+			svcTCPPort)
+		_, err = runCommand(false, "wget", "-O-", "-q", hostport)
```
s1061123, line 76: Same as bgp side. How about to add timeout option to make it shorter?

s1061123 (changes requested):

e2etest/bgp_test.go
markdgray: Good idea @s1061123 . I will update this in the next revision.

e2etest/layer2_test.go
markdgray: Yes

e2etest/bgp_test.go
```diff
@@ -0,0 +1,80 @@
...
+			DescribeSvc(f.Namespace.Name)
+		}
+	})
+
+	ginkgo.It("should work for type=Loadbalancer", func() {
+		namespace := f.Namespace.Name
+		serviceName := "external-local-lb"
+		jig := e2eservice.NewTestJig(cs, namespace, serviceName)
+
+		svc, err := jig.CreateLoadBalancerService(loadBalancerCreateTimeout,
```
s1061123, line 55: minor comment but should we make it one line(55+56)?

markdgray: I wasn't sure if there were any coding standard so I was keeping below 80 characters. I will join the lines and see if the linter complains,

russellb:
@s1061123 since you had started reviewing this, I'm going to give you a chance to give your +1 first before this merges

s1061123 (approved):
I've checked the diff again and it looks good to me, thank you for the changes!

s1061123:
LGTM!

.circleci/config.yml
```diff
@@ -12,6 +12,34 @@ jobs:
...
+      - checkout
+      - run: sudo apt-get update
+      - run: sudo apt-get install python3-pip
+      - run: sudo pip3 install invoke semver pyyaml
+      - run: GO111MODULE="on" go get sigs.k8s.io/kind@v0.10.0
+      - run: curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+      - run: sudo install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
+      - run: inv build
+      - run: inv dev-env -p bgp
+      - run: inv e2etest
```
russellb, line 28: When I tested this locally, I got one failure because my cluster wasn't ready when i ran e2etest.

```
  May 21 17:05:20.701: Unexpected error:                                                                                      
      <*errors.errorString | 0xc000460490>: {                                                                                 
          s: "there are currently no ready, schedulable nodes in the cluster",                                                
      }                                                                                                                       
      there are currently no ready, schedulable nodes in the cluster                                                          
  occurred                                                                                                                    
```

I wonder if this could turn into random failures in CI?  Maybe there should be an explicit check of some kind that the cluster is ready to proceed with running the e2etest suite?

russellb (changes requested):

.circleci/config.yml
markdgray: Hi @russellb, we could add that. There are some checks in the test that wait for objects (e.g. service/pod) to be created so I am not sure why this triggered. I have been trying to reproduce with no success. I tried with `inv dev-env -p layer2 && inv e2etest` and `inv dev-env -p layer2 e2etest`, I also tried with BGP protocol. How did you run this?

russellb: It was `inv dev-env -p bgp` and `inv e2etest` manually.  It passed the 2nd time I ran `e2etest`.  I haven't been able to reproduce it again though.  :(

russellb: Ok, I reproduced it by running `docker image prune -a` before `inv dev-env -p bgp && inv e2etest`.  I think pulling images is making it take a bit longer to come up and that's why it failed my first try and not later tries to reproduce.

.circleci/config.yml
```diff
@@ -12,6 +12,34 @@ jobs:
...
+      - checkout
+      - run: sudo apt-get update
+      - run: sudo apt-get install python3-pip
+      - run: sudo pip3 install invoke semver pyyaml
+      - run: GO111MODULE="on" go get sigs.k8s.io/kind@v0.10.0
+      - run: curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
+      - run: sudo install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
+      - run: inv build
+      - run: inv dev-env -p layer2
+      - run: inv e2etest
```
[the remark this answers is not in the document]
russellb: awesome, thank you for adding the circleci commands for dependency and kind installation

.circleci/config.yml
markdgray: I added a check in `tasks.py` when invoking `inv e2etest` that will wait for `condition=Ready` for all Pods in the `kube-system` and `metallb-system` namespaces (or timeout after 120s). It seems to work for me

russellb: Thanks.  I see in the logs that when running `e2etest`, it does try to wait, but it times out after 2 minutes.

I was able to reproduce most of the time with:

```
inv dev-env-cleanup && docker image prune -a && inv dev-env -p bgp && inv e2etest
```

but haven't been able to reproduce again with the extra 2 minute wait you added in tasks.py.

e2etest/layer2_test.go
```diff
@@ -0,0 +1,79 @@
...
+			framework.ExpectNoError(err)
+		}()
+
+		svcTCPPort := int(svc.Spec.Ports[0].Port)
+		ingressIP := e2eservice.GetIngressPoint(
+			&svc.Status.LoadBalancer.Ingress[0])
+
+		ginkgo.By("checking connectivity to its external VIP")
+
+		hostport := fmt.Sprintf("http://%s:%d/",
```
russellb, line 72: Can you update this to use JoinHostPort()?  That will make this IPv6 safe.  This is needed in bgp_test.go as well

https://golang.org/pkg/net/#JoinHostPort

russellb (changes requested):
Sorry, noticed one more thing ... them i'm good :)

.circleci/config.yml
markdgray: Ok, If it is timing out in the CI, I will increase it to 300 seconds

markdgray:
> Sorry, noticed one more thing ... them i'm good :)

Not at all. Thanks for taking the time to review and test!

e2etest/layer2_test.go
markdgray: Sure thing

russellb (approved):

markdgray:
@russellb Uggh .. I accidently introduced a bug but it was good that I did because it didnt break the CI even though the test failed. I will fix both.

russellb:
> @russellb Uggh .. I accidently introduced a bug but it was good that I did because it didnt break the CI even though the test failed. I will fix both.

ha, the one iteration i didn't test myself ... i just made sure CI was still green

PR #853, opened 2021-04-28
Fix image publish CI jobs

russellb:
The image publish jobs are currently failing due to the `python-pip`
package not being found.  I'm guessing this has to do with a newer
ubuntu version.  The patch updates these jobs to install `python3-pip`
instead, which is what the unit testing and linting jobs already use.


johananl (approved):
LGTM. Thanks!

PR #846, opened 2021-04-21
chart: clean up values schema of .Values.prometheus.*

gclawes:
.Values.prometheus.* underwent several re-writes during chart
development, the values.schema.json for this had fallen behind.

Changes:
* enforce labels are strings for prometheusAlert objects
* validate .Values.prometheus.podMonitor.*
* move PrometheusRule values from .Values.prometheus.* to
  .Values.prometheus.prometheusRule.*
* Clean up comment in values.yaml
* Fix type of .Values.prometheus.podMonitor.interval in values.yaml

Depends on #841

PR #854, opened 2021-04-29
chart: re-generate helm chart README.md with norwoodj/helm-docs

gclawes:
Re-generates helm chart README.md, as requested in #846.

#844 should be rebased after this is merged.


champtar:
Is it possible to add that to the CI ?

russellb:
> Is it possible to add that to the CI ?

Yes, please.  It seems like it'd be helpful to make sure this doesn't get missed.  Let's chat on slack if you want do discuss how we might do this.  It could be another validation to include in the existing `inv checkpatch` command that runs in CI already.

Issue #664, opened 2020-07-10
Unify logging


johananl:
We should take a look at the way we instantiate loggers and pass them around the codebase and make sure things make sense there.

For example, in the speaker code we instantiate a logger in `main.go`:

https://github.com/metallb/metallb/blob/80bfe1cf45bf4338e6eced11ab7a88d3d2356663/speaker/main.go#L65

We then use it directly within `main.go`...

https://github.com/metallb/metallb/blob/80bfe1cf45bf4338e6eced11ab7a88d3d2356663/speaker/main.go#L86

...and pass it around to other packages:

https://github.com/metallb/metallb/blob/80bfe1cf45bf4338e6eced11ab7a88d3d2356663/speaker/main.go#L138-L142

However, in some cases we seem to be doing something else:

https://github.com/metallb/metallb/blob/80bfe1cf45bf4338e6eced11ab7a88d3d2356663/speaker/main.go#L122-L123

We should try to use one "standard" way for logging in the project. If there are cases where we have to do something special, we should explain it in a comment.

johananl:
While we're at it, we may want to switch to the standard library `log` library unless there is a good reason to use `go-kit`. The `go-kit` logger seems to return an error for some reason, which causes a linting hassle since we aren't checking these errors.

champtar:
We have proper json logs, if we move to anything it's structured logs like k8s but why go back to free form log ?

johananl:
> We have proper json logs, if we move to anything it's structured logs like k8s but why go back to free form log ?

That's why I wrote "unless there is a good reason to use go-kit"...

My main concern is the fact that we currently have to either error-check every logging statement or deal with linting errors (see https://github.com/metallb/metallb/pull/855). I didn't say we should stop doing structured logging.

champtar:
The last exemple you are giving is normal, memberlist expect a stdlib logger so we wrap our go-kit logger allowing to have every memberlist logs tagged with "component: Memberlist".

champtar:
For me std log == free form, and you hinted at going to std log ;)

champtar:
I would also prefer to switch library that tag the whole codebase with linter silencer :)

champtar:
Also if we could have log level we could silence Memberlist by default

champtar:
> Maybe we could wrap this logger in a new metallb package and handle the error in that package.

We don't want to loose the call site (file/line) IMO

PR #855, opened 2021-04-29
Migrate linter from gometalinter to golangci-lint

markdgray:
In the process of adding a linter task to `tasks.py`, I discovered that `gometalinter` has been deprecated. This PR adds that task and migrates the linter.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -17,12 +17,11 @@ jobs:
       - image: cimg/go:1.16
     steps:
       - checkout
-      - run: curl -L https://git.io/vp6lP | sh # gometalinter
-      - run: PATH=./bin:$PATH gometalinter --deadline=5m --disable-all --enable=gofmt --enable=vet --vendor ./...
       - run: sudo apt-get update
       - run: sudo apt-get install python3-pip
       - run: sudo pip3 install invoke semver pyyaml
       - run: inv checkpatch
+      - run: inv lint -e host
   # FIXME: this is pinned to v3.0.0 because subsequent versions of ct pull in new versions of helm, which are subject
   #        to this bug: https://github.com/helm/helm/issues/8835, specifically for rbac object names with ':' in them
   helm-lint:
--- a/internal/k8s/k8s.go
+++ b/internal/k8s/k8s.go
@@ -245,9 +245,12 @@ func New(cfg *Config) (*Client, error) {
 	}
 
 	http.Handle("/metrics", promhttp.Handler())
-	go func() {
-		http.ListenAndServe(fmt.Sprintf("%s:%d", cfg.MetricsHost, cfg.MetricsPort), nil)
-	}()
+	go func(l log.Logger) {
+		err := http.ListenAndServe(fmt.Sprintf("%s:%d", cfg.MetricsHost, cfg.MetricsPort), nil)
+		if err != nil {
+			l.Log("op", "listenAndServe", "err", err, "msg", "cannot listen and serve", "host", cfg.MetricsHost, "port", cfg.MetricsPort)
+		}
+	}(c.logger)
 
 	return c, nil
 }
--- a/speaker/layer2_controller_test.go
+++ b/speaker/layer2_controller_test.go
@@ -21,9 +21,7 @@ func (sl *fakeSpeakerList) UsableSpeakers() map[string]bool {
 	return sl.speakers
 }
 
-func (sl *fakeSpeakerList) Rejoin() {
-	return
-}
+func (sl *fakeSpeakerList) Rejoin() {}
 
 func compareUseableNodesReturnedValue(a, b []string) bool {
 	if &a == &b {
--- a/tasks.py
+++ b/tasks.py
@@ -519,3 +519,25 @@ def checkpatch(ctx):
     except UnexpectedExit:
         # Will exit non-zero if no double-space-after-period lines are found.
         pass
+
+
+@task(help={
+    "env": "Specify in which environment to run the linter . Default 'container'. Supported: 'container','host'"
+})
+def lint(ctx, env="container"):
+    """Run linter.
+
+    By default, this will run a golangci-lint docker image against the code.
+    However, in some environments (such as the MetalLB CI), it may be more
+    convenient to install the golangci-lint binaries on the host. This can be
+    achieved by running `inv lint --env host`.
+    """
+    version = "1.39.0"
+
+    if env == "container":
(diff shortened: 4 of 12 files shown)


controller/main.go
```diff
@@ -28,7 +28,7 @@ import (
 	"go.universe.tf/metallb/internal/version"
 
 	"github.com/go-kit/kit/log"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
johananl, line 31: Your editor is doing this automatically most likely. Personally I don't mind merging this change but I wanted to verify this is intentional.

controller/main.go
```diff
@@ -46,8 +46,8 @@ type controller struct {
 }
 
 func (c *controller) SetBalancer(l log.Logger, name string, svcRo *v1.Service, _ *v1.Endpoints) k8s.SyncState {
-	l.Log("event", "startUpdate", "msg", "start of service update")
-	defer l.Log("event", "endUpdate", "msg", "end of service update")
+	l.Log("event", "startUpdate", "msg", "start of service update")   //nolint:errcheck
```
johananl, line 49: I'm not sure why a logger should return an error. Changing this is out of scope for this PR, however using `nolint` on every log statement from now on doesn't seem reasonable to me. OTOH I also don't want to disable this linter globally since it can save some nasty debugging in certain cases.

We have https://github.com/metallb/metallb/issues/664 for handling the logging part. I'm not sure whether to move forward with all these `nolint` statements or disable this linter globally as a workaround and add a TODO in the code until the logging task is done.

internal/k8s/k8s.go
```diff
@@ -245,9 +245,12 @@ func New(cfg *Config) (*Client, error) {
 	}
 
 	http.Handle("/metrics", promhttp.Handler())
-	go func() {
-		http.ListenAndServe(fmt.Sprintf("%s:%d", cfg.MetricsHost, cfg.MetricsPort), nil)
-	}()
+	go func(l log.Logger) {
+		err := http.ListenAndServe(fmt.Sprintf("%s:%d", cfg.MetricsHost, cfg.MetricsPort), nil)
+		if err != nil {
+			l.Log("Cannot listen and server on %s:%d", cfg.MetricsHost, cfg.MetricsPort) //nolint:errcheck
```
johananl, line 251: Could you use the existing logging style? We can discuss changing it but for now let's stay consistent.

Also - `s/server/serve/`.

speaker/layer2_controller_test.go
```diff
@@ -22,7 +22,6 @@ func (sl *fakeSpeakerList) UsableSpeakers() map[string]bool {
 }
 
 func (sl *fakeSpeakerList) Rejoin() {
-	return
```
johananl, line 25: I would collapse the braces if the function block is empty.

tasks.py
```diff
@@ -519,3 +519,11 @@ def checkpatch(ctx):
     except UnexpectedExit:
         # Will exit non-zero if no double-space-after-period lines are found.
         pass
+
+
+@task
+def linter(ctx):
```
johananl, line 525: Let's call this `lint` to be consistent with the rest of the tasks which are in the imperative tense.

johananl (changes requested):
Thanks a lot @markdgray!

I've left some comments.

controller/main.go
markdgray: There seems to be a way to exclude certain function calls from certain linters. However, I can't get the syntax to work in the short time I spent on it. I will look again later.

speaker/layer2_controller_test.go
markdgray: Done

internal/k8s/k8s.go
markdgray: Done. I hadn't notice the logging style. In fact, what I had was a bug because the arguments are supposed to be K/V pairs.

controller/main.go
markdgray: I figured it out by downloading errcheck. Going to push a new PR.

.circleci/config.yml
```diff
@@ -17,12 +17,13 @@ jobs:
...
       - checkout
-      - run: curl -L https://git.io/vp6lP | sh # gometalinter
-      - run: PATH=./bin:$PATH gometalinter --deadline=5m --disable-all --enable=gofmt --enable=vet --vendor ./...
       - run: sudo apt-get update
       - run: sudo apt-get install python3-pip
       - run: sudo pip3 install invoke semver pyyaml
       - run: inv checkpatch
+      # The version should match the version used in the linter task in tasks.py
+      - run: curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b $(go env GOPATH)/bin v1.39.0
+      - run: golangci-lint run --timeout 5m0s ./...
```
russellb, line 26: It would be great if this could just be `inv lint`.

Then in tasks.py, maybe it should have an option for whether it should run it in `docker` or whether it should run a pre-installed `golangci-lint`?

Similar to other spots in CI config, I just want a single place that ensures we run the same commands for both dev testing and CI testing.

russellb (changes requested):
other than this one nit, I'm good with this.  @johananl do you want to take another look?  I think your concerns were addressed.

.circleci/config.yml
markdgray: That was my plan originally. I had to try it again to remember why I could not do it like that. The reason is that we are running in a docker container. In order to pull a docker image we need to set up a remote docker environment. This remote docker environment does not have access to the files in the primary docker environment. This is documented at: https://circleci.com/docs/2.0/building-docker-images/#mounting-folders. So if you try `inv lint`, you get an error like this 

```
...
e07b05f2a22d: Pull complete
f298da50fa05: Pull complete
5f9f36afbb7f: Pull complete
Digest: sha256:06c0900d94953cfa35df107e36c6f8a3d191f4ce9736ce067eb1862da6e7e7a7
Status: Downloaded newer image for golangci/golangci-lint:v1.39.0
level=error msg="Running error: context loading failed: no go files to analyze"

Exited with code exit status 5
```

I had three options to resolve this:

1. Run the CI stage in a VM (I discounted because it would run slower)
2. Copy the files between containers (I discounted because it would run slower)
3. The current option.

I am happy to change to another option if you prefer as all options weren't great.

russellb: Yeah i was thinking something like ...

the default behavior:

```
$ inv lint --env container
```

or how we'd run it from the CI config:

```
$ inv lint --env host
```

markdgray: Updated.

russellb:
Thanks!

PR #856, opened 2021-05-05
dev-env: merge BGP dev-env doc into common doc

markdgray:
As there is now a layer 2 development environment, create a single
"README.md" that consolidates the dev-env documentation.


markdgray:
Thanks

PR #858, opened 2021-05-17
design: MetalLB release cadence proposal

fedepaol:
The document added here proposes a release cadence and a versioning
scheme that is tied to k8s release schedule. This ensures predictable
api compatibility.


fedepaol:
cc @russellb / @markdgray

johananl (commented):
Thanks @fedepaol.

While I do see value in tying the MetalLB releases to k8s versions, I'm reluctant to adopt this with the current allocation of maintainer time since we would very likely not keep up. Of course, we do our best to stay up to date with the latest and greatest from k8s, however I'm not sure we want to officially commit to a release cadence we know we can't deliver :grimacing: 

@rata @daxmc99 @russellb thoughts?

russellb:
I think the proposal makes a lot of sense.  It's one way to accomplish what was proposed in #813.

You're also right to be concerned about execution.  Maybe it's worth listing out exactly what this would take.

For each kube release:

1. Update CI in `main` to run against that kube version.
2. Once we're comfortable that metallb seems to be working fine with that release, create a new release branch
3. Make new release from release branch

TODO:

* Define criteria for step 2
* Decide how far back we'll backport bug fixes
* since we'll end up with more release branches than we would have with feature-driven releases based on the current development speed / maintainer capacity

My gut feeling is that this is worth it, and that "keeping up with kube" should really be our baseline minimum to keep up with for maintaining metallb.  I definitely wouldn't commit to bug fix backports though, perhaps at all.  I'd probably leave backports as a "considered on a case by case basis, depending on bug severity and maintainer capacity".

markdgray:
If the workload to keep up with the release cadence is too great, perhaps we could target every second release.

markdgray:
@fedepaol It would be good to state what (if any) policy there would be for stable releases. Would we support a stable release branch?

markdgray:
The other benefit to this proposal is that we can better align with the FRR release cadence.

russellb:
> If the workload to keep up with the release cadence is too great, perhaps we could target every second release.

I think I'm willing to sign up for covering this for now.  I think it's important enough.

fedepaol:
> I think the proposal makes a lot of sense. It's one way to accomplish what was proposed in #813.
> 
> You're also right to be concerned about execution. Maybe it's worth listing out exactly what this would take.
> 
> For each kube release:
> 
>     1. Update CI in `main` to run against that kube version.
> 
>     2. Once we're comfortable that metallb seems to be working fine with that release, create a new release branch
> 
>     3. Make new release from release branch
> 
> 
> TODO:
> 
>     * Define criteria for step 2
> 
>     * Decide how far back we'll backport bug fixes, since we'll end up with more release branches than we would have with feature-driven releases based on the current development speed / maintainer capacity
> 

Re: backporting bugs, I agree we can act on per - bug basis.
It would be cool to have multiple CI lanes on master to validate the compatibility with previous versions of k8s (ideally, up to n-3 which afaik is the supported window in k8s). Kind allows to do that by choosing the node image. 
That will give us elements to understand if a critical bug must be backported or not. 
If we are sure latest metallb works on all supported k8s version, we may just tell users to upgrade.

Does it make sense?

> 
> My gut feeling is that this is worth it, and that "keeping up with kube" should really be our baseline minimum to keep up with for maintaining metallb. I definitely wouldn't commit to bug fix backports though, perhaps at all. I'd probably leave backports as a "considered on a case by case basis, depending on bug severity and maintainer capacity".

fedepaol:
> @fedepaol It would be good to state what (if any) policy there would be for stable releases. Would we support a stable release branch?



> @fedepaol It would be good to state what (if any) policy there would be for stable releases. Would we support a stable release branch?

Not sure I understand the point here. 
Once we cut a release branch out of main, that is meant to become a stable branch were we only merge eventual urgent bugfixes (or at least, that's how I see it)

markdgray:
> > @fedepaol It would be good to state what (if any) policy there would be for stable releases. Would we support a stable release branch?
> 
> > @fedepaol It would be good to state what (if any) policy there would be for stable releases. Would we support a stable release branch?
> 
> Not sure I understand the point here.
> Once we cut a release branch out of main, that is meant to become a stable branch were we only merge eventual urgent bugfixes (or at least, that's how I see it)

My question was more about the policies. How often do we branch, how often do we maintain branch, when do we backport, what can we not backport, etc.

fedepaol:
> > > @fedepaol It would be good to state what (if any) policy there would be for stable releases. Would we support a stable release branch?
> > 
> > 
> > > @fedepaol It would be good to state what (if any) policy there would be for stable releases. Would we support a stable release branch?
> > 
> > 
> > Not sure I understand the point here.
> > Once we cut a release branch out of main, that is meant to become a stable branch were we only merge eventual urgent bugfixes (or at least, that's how I see it)
> 
> My question was more about the policies. How often do we branch, how often do we maintain branch, when do we backport, what can we not backport, etc.

Ok, let me try to add something to the doc so we can comment over there.

design/RELEASE_CADENCE.md
```diff
@@ -0,0 +1,67 @@
...
+## Summary
+
+The purpouse of this enhancement is to define a versioning and release schedule tied to
+kubernetes versions.
+
+## Motivation
+
+The reason for this proposal is to establish a dependency between a given MetalLB version and kubernetes versions.
+
+This will make MetalLB more reliable with regards with the set of kubernetes apis it can use or can't use. A good example for it is the introduction of the use of EndpointSlices, which are versioned as beta in k8s 1.17+ but moving to stable in 1.21.
```
rata, line 26: Sorry, but I don't follow. What would this proposal simplify, for example with endpoint slices?

design/RELEASE_CADENCE.md
```diff
@@ -0,0 +1,67 @@
...
+### Goals
+
+Define MetalLB release cadence and versioning as bounded to kubernetes versions.
+
+## Proposal
+
+Each MetalLB release branch correspond to a released k8s version.
+
+Every time a k8s version is released, the api used by metalLB are bumped to that release and the version used
+by kind in CI is bumped. This would prevent using deprecated apis.
```
rata, line 37: Hmm, that would prevent using _removed_ apis but not deprecated, right?

design/RELEASE_CADENCE.md
```diff
@@ -0,0 +1,67 @@
...
+
+This will make MetalLB more reliable with regards with the set of kubernetes apis it can use or can't use. A good example for it is the introduction of the use of EndpointSlices, which are versioned as beta in k8s 1.17+ but moving to stable in 1.21.
+
+### Goals
+
+Define MetalLB release cadence and versioning as bounded to kubernetes versions.
+
+## Proposal
+
+Each MetalLB release branch correspond to a released k8s version.
```
rata, line 34: Sorry for the silly question but from here I understand that we will have a 1-1 mapping between kubernetes minor releases and MetalLB. For example, k8s 1.20.x <---> MetalLB X. I guess if 1.21 is released, we will release MEtalLB X+1 from that part too.

However, later the proposal says:
> The most recent release of MetalLB must be able to support all the supported k8s versions (n-3 as of today). We can ensure that by having multiple CI lanes running kind
with different kubernetes versions.

Why would say "MetalLB X works with k8s 1.20" and have a 1-1 mapping, when it also works with n-3 releases?

design/RELEASE_CADENCE.md
rata, line 26: Is the motivation, maybe, a way to guarantee that the last MetalLB release doesn't use very old client-go versions or old versions of other dependencies?

If that is the case, though, the current motivation seems misleading. But I guess I'm missing something :)

rata (commented):
Sorry for the late review @fedepaol . I'll add this to the community meeting agenda today, just in case you can go, because I think I'm missing something and maybe talking in person is simpler

design/RELEASE_CADENCE.md
rata: Maybe the reason is: https://github.com/metallb/metallb/pull/858#discussion_r700287211 ? So, not "works with 1.20", but just having the last release "up to date"?

fedepaol:
Closing this, I think it's obsolete

PR #862, opened 2021-05-25
e2etest: Fix CI failure

markdgray:
This PR fixes a failure in the e2etest and also fixes a bug that caused the failure to silently pass the CI.

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -22,6 +22,7 @@ import (
 	"context"
 	"fmt"
 	"net"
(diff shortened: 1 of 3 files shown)


russellb:
Of course the final change that caused a bug is the one i didn't test manually :)

At least we identified that CI wasn't failing when it was supposed to!

markdgray:
> Of course the final change that caused a bug is the one i didn't test manually :)
> 
> At least we identified that CI wasn't failing when it was supposed to!

I know!

PR #863, opened 2021-05-26
Add kube-ovn to list

withlin:
I've been using [kube-ovn](https://github.com/kubeovn/kube-ovn) with MetalLB in BGP mode for about a year and no issues so far.


russellb:
thanks!

PR #864, opened 2021-05-28
Add an optional kind node image parameter to dev-env

fedepaol:
Being able to specify the image to use for the node allows us to create
test cluster pinned to specific kind versions.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -185,14 +185,17 @@ def validate_kind_version():
     "name": "name of the kind cluster to use.",
     "protocol": "Pre-configure MetalLB with the specified protocol. "
                 "Unconfigured by default. Supported: 'bgp','layer2'",
+    "node_img": "Optional node image to use for the kind cluster (e.g. kindest/node:v1.18.19)."
+                "The node image drives the kubernetes version used in kind."
 })
-def dev_env(ctx, architecture="amd64", name="kind", cni=None, protocol=None):
+def dev_env(ctx, architecture="amd64", name="kind", cni=None, protocol=None, node_img=None):
     """Build and run MetalLB in a local Kind cluster.
 
     If the cluster specified by --name (default "kind") doesn't exist,
     it is created. Then, build MetalLB docker images from the
     checkout, push them into kind, and deploy manifests/metallb.yaml
     to run those images.
+    The optional node_img parameter will be used to determine the version of the cluster.
     """
 
     validate_kind_version()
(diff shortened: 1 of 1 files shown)


fedepaol:
cc @markdgray

tasks.py
```diff
@@ -186,13 +186,14 @@ def validate_kind_version():
     "protocol": "Pre-configure MetalLB with the specified protocol. "
                 "Unconfigured by default. Supported: 'bgp','layer2'",
```
markdgray, line 187: Can you add a description of this option? I know it is in the general section below but I think convention would also put it in the `@task(help={` section

fedepaol: I missed that. Will do!

tasks.py
```diff
@@ -216,11 +217,14 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None, protocol=None):
...
             }
+        extra_options = ""
+        if node_img != None:
+            extra_options = "--image={}".format(node_img)
         config = yaml.dump(config).encode("utf-8")
         with tempfile.NamedTemporaryFile() as tmp:
             tmp.write(config)
             tmp.flush()
-            run("kind create cluster --name={} --config={}".format(name, tmp.name), pty=True, echo=True)
+            run("kind create cluster --name={} --config={} {}".format(name, tmp.name, extra_options), pty=True, echo=True)
```
markdgray, line 227: Why did you add "extra_options" instead of just "--image={}"?

fedepaol: because otherwise it wouldn't work if we don't specify the image.
I am open to suggestions, my python-fu is pretty bad :-)

tasks.py
markdgray: Thanks, an example image name would be useful as I am still trying to find an example to test.

fedepaol: You can find them here https://github.com/kubernetes-sigs/kind/releases
I tried and it seemed to work :P

tasks.py
markdgray: oh that makes sense. Im just as bad but what you have done makes sense to me

tasks.py
markdgray: you could add something like `(e.g. kindest/node:v1.18.19)`

fedepaol: yes will do

markdgray: Thanks, just tested it there and seems to work! cool.

fedepaol: Done

markdgray:
This looks good to me as long as it passes the tests.

russellb:
lgtm.  It would be good to add an example to dev-env/README.md, but don't need to block on it

PR #866, opened 2021-05-30
Update speaker.yaml

NorthFuture:
wrong value reference


russellb:
@gclawes can you take a look at this one?

gclawes:
A similar change is needed for the controller: https://github.com/metallb/metallb/blob/main/charts/metallb/templates/controller.yaml#L22

@NorthFuture could you submit that?  Otherwise I can open a separate PR.

russellb:
@gclawes I added a commit to this PR for the controller fix.  PR look good to you?

gclawes:
Looks good to me.

PR #872, opened 2021-06-05
Verify if interface master is up before skipping

### LINKED ISSUE #871: MetalLB fails to send ARPs if interface is attached to Open vSwitch
Thanks for an awesome piece of software.

When using Open vSwitch and attaching a port to it, the port ends up being assigned to the `ovs-system` port, which remains to be an interface that is down.

```
# ip link | grep ovs
18: ovs-system: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN mode DEFAULT group default qlen 1000
111: bond0.1002@bond0: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue master ovs-system state UP mode DEFAULT group default qlen 1000
```

This effectively means that `bond0.1002` is skipped because it has a `master` assigned to it, and `ovs-system` is skipped beacuse it is naturally `DOWN`, which means no ARP ever goes out.  I think it would make sense that we check first if the bridge the interface is attached to up before skipping it.

mnaser:
In some cases such as wen using Open vSwitch, it is possible that
the interface master will be in down state (ovs-system) but MetalLB
skips the port because it is already assigned a master which means
that no ARP response ever goes through it.

This patch instead makes sure that the master interface is actually
up before skipping it to make sure that we can send traffic across
in one of the interfaces.

Signed-off-by: Mohammed Naser <mnaser@vexxhost.com>

Fixed #871

### DIFF
--- a/internal/layer2/announcer.go
+++ b/internal/layer2/announcer.go
@@ -5,6 +5,7 @@ import (
 	"net"
 	"os"
 	"strconv"
+	"strings"
 	"sync"
 	"time"
 
@@ -73,7 +74,10 @@ func (a *Announce) updateInterfaces() {
 			continue
 		}
 		if _, err = os.Stat("/sys/class/net/" + ifi.Name + "/master"); !os.IsNotExist(err) {
-			continue
+			masterOperState, err := ioutil.ReadFile("/sys/class/net/" + ifi.Name + "/master/operstate")
+			if err == nil && strings.Contains(string(masterOperState), "up") {
+				continue
+			}
 		}
 		f, err := ioutil.ReadFile("/sys/class/net/" + ifi.Name + "/flags")
 		if err == nil {


internal/layer2/announcer.go
```diff
@@ -73,7 +74,10 @@ func (a *Announce) updateInterfaces() {
 			continue
 		}
 		if _, err = os.Stat("/sys/class/net/" + ifi.Name + "/master"); !os.IsNotExist(err) {
-			continue
+			masterOperState, err := ioutil.ReadFile("/sys/class/net/" + ifi.Name + "/master/operstate")
+			if err == nil && strings.Contains(string(masterOperState), "up") {
```
fedepaol, line 78: worth logging the error if not nil? Seems like it should never happen.

Not related to this PR, but netlinks are easier to handle in go with https://github.com/vishvananda/netlink (which is used by a lot of CNI plugins)

PR #876, opened 2021-06-08
release: Fix image reference regex

### LINKED ISSUE #874: v0.10.0 manifests point to main tag instead of release tag
I noticed that the v0.10.0 manifests' images are pointing at the `main` tag instead of the `v0.10.0` tag. Is this expected? Previously, manifests for a tagged release would always point at the tagged image corresponding to that release.

russellb:
The release process script has a regex for fixing image references in
manifests, but these were not updated to reflect the move to quay.io.
This change updates it so the image reference will properly get updated
on the next release.

Fixes #874


rata (approved):
LGTM, thanks!

PR #882, opened 2021-06-10
Allow running dev-env from any branch

### LINKED ISSUE #879: dev-env doesn't work against branches other than main
The MetalLB dev-env, `inv dev-env`, builds images from the local copy of MetalLB code and runs them in a kind based cluster.  I noticed that this doesn't work if you run the dev-env from a branch of MetalLB other than `main`.

This is because of the following line in `tasks.py`:

```
manifest = manifest.replace(":main", ":dev-{}".format(architecture))
```

This will not match anything in the manifests for a branch like "v0.10".  This should be made more flexible, perhaps with a regex based match on images.

msherif1234:
Fixed #879 
Unit-test
```
git checkout v0.10
inv dev-env
image: quay.io/metallb/speaker:dev-amd64
image: quay.io/metallb/controller:dev-amd64


git checkout main
inv dev-env
image: quay.io/metallb/speaker:dev-amd64
image: quay.io/metallb/controller:dev-amd64

```
Metallb generated manifests will change image's tag from v0.10.1 to dev-amd64

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -230,8 +230,8 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None, protocol=None, nod
 
     if mk_cluster and cni:
         run("kubectl apply -f e2etest/manifests/{}.yaml".format(cni), echo=True)
-
-    build(ctx, binaries=["controller", "speaker", "mirror-server"], architectures=[architecture])
+    binaries = ["controller", "speaker", "mirror-server"]
+    build(ctx, binaries, architectures=[architecture])
     run("kind load docker-image --name={} quay.io/metallb/controller:dev-{}".format(name, architecture), echo=True)
     run("kind load docker-image --name={} quay.io/metallb/speaker:dev-{}".format(name, architecture), echo=True)
     run("kind load docker-image --name={} quay.io/metallb/mirror-server:dev-{}".format(name, architecture), echo=True)
@@ -245,7 +245,9 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None, protocol=None, nod
 
         with open(manifests_dir + "/metallb.yaml") as f:
             manifest = f.read()
-        manifest = manifest.replace(":main", ":dev-{}".format(architecture))
+        for image in binaries:
+            manifest = re.sub("image: quay.io/metallb/{}:.*".format(image),
+                          "image: quay.io/metallb/{}:dev-{}".format(image, architecture), manifest)
         with open(tmpdir + "/metallb.yaml", "w") as f:
             f.write(manifest)
             f.flush()


tasks.py
```diff
@@ -231,7 +232,7 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None, protocol=None):
 
         with open(manifests_dir + "/metallb.yaml") as f:
             manifest = f.read()
-        manifest = manifest.replace(":main", ":dev-{}".format(architecture))
+        manifest = manifest.replace(":main", ":dev-{}-{}".format(branch, architecture))
```
russellb, line 235: This isn't quite what I had in mind when I filed the bug.  Take this scenario for example:

```
git checkout v0.10
... edit some metallb code ...
inv dev-env
```

At this point, I would expect my dev-env cluster to be running my modified code, but it won't.  It will be running whatever is in quay.io for `main`, because this search/replace won't match anything.  In the `v0.10` branch, the image isn't `:main`.

russellb (changes requested):

markdgray:
@msherif1234 Is that timeout in the CI a bug in the CI? Might be good to try and fix it or add an issue if you cannot.

msherif1234:
> @msherif1234 Is that timeout in the CI a bug in the CI? Might be good to try and fix it or add an issue if you cannot.

we probably need a new issue for it, I will get to it next

russellb:
> > @msherif1234 Is that timeout in the CI a bug in the CI? Might be good to try and fix it or add an issue if you cannot.
> 
> we probably need a new issue for it, I will get to it next

I'm not sure if it was a timeout.  The log is a little confusing, but I think `wget` in the test exited non-zero (4), which the wget manpage says is "Network Error".

I wonder if we should add some args to `wget` to make sure it retries a few times, just in case we're trying to access the service too fast?

tasks.py
```diff
@@ -239,13 +240,14 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None, protocol=None, nod
...
     manifests_dir = os.getcwd() + "/manifests"
+    curr_branch = subprocess.check_output("git rev-parse --abbrev-ref HEAD", shell=True).decode().strip("\n")
     with tempfile.TemporaryDirectory() as tmpdir:
         # Copy namespace manifest.
         shutil.copy(manifests_dir + "/namespace.yaml", tmpdir)
 
         with open(manifests_dir + "/metallb.yaml") as f:
             manifest = f.read()
-        manifest = manifest.replace(":main", ":dev-{}".format(architecture))
+        manifest = re.sub(":{}\.?[0-9]?".format(curr_branch), ":dev-{}".format(architecture), manifest)
```
russellb, line 250: Does this still work if what it needs to replace is `main` ?  Also, this looks a bit risky.  There are a *lot* of lines that include `:` in a yaml doc.  Can you make sure we're only matching the `image:` lines?

msherif1234: I

> Does this still work if what it needs to replace is `main` ? Also, this looks a bit risky. There are a _lot_ of lines that include `:` in a yaml doc. Can you make sure we're only matching the `image:` lines?

yes I tried both v0.10 and main as showing the UT description, this is matching for :branch_name so not sure where is the risk ?

russellb: sorry, i didn't read it correctly

russellb, line 250: I tested out the regex with some different scenarios and I think it will break if the current manifest has a two digit patch release, like `quay.io/metallb/controller:v0.10.10`.  It will leave the last `0` at the end.

russellb:
Thanks for the revisions!  lgtm now

russellb (approved):

PR #883, opened 2021-06-10
Ensure RBAC for Helm matches YAML manifests

### LINKED ISSUE #878: v0.10.0 Helm chart is missing some RBAC updates
While evaluating the v0.10.0 Helm chart for my usage, I noticed that the Helm chart is missing a few changes from the Kustomize version of the deployment.  The commit at https://github.com/metallb/metallb/commit/9ad25a9d removes the Service update permission from the controller ClusterRole in the Kustomize manifests as well as dead code that would have updated services.  The EndpointSlices feature commit here https://github.com/metallb/metallb/commit/312b03cd adds the ability for MetalLB to use EndpointSlices and changes the speaker ClusterRole in the Kustomize manifests so the speaker can access EndpointSlices in the cluster.

I also noticed a difference in the PodSecurityPolicy for the controller where the Kustomize version does not grant the ability to create a host port for the metrics port but the Helm chart does.  See these lines in the Helm chart PSP for the controller https://github.com/metallb/metallb/blob/main/charts/metallb/templates/psp.yaml#L15-L17 vs the PSP for the controller in the manifests which lacks that section.

treydock:
Fixes #878

Signed-off-by: Trey Dockendorf <tdockendorf@osc.edu>

### DIFF
--- a/charts/metallb/templates/rbac.yaml
+++ b/charts/metallb/templates/rbac.yaml
@@ -8,7 +8,7 @@ metadata:
 rules:
 - apiGroups: [""]
   resources: ["services"]
-  verbs: ["get", "list", "watch", "update"]
+  verbs: ["get", "list", "watch"]
 - apiGroups: [""]
   resources: ["services/status"]
   verbs: ["update"]
@@ -31,6 +31,9 @@ metadata:
 rules:
 - apiGroups: [""]
   resources: ["services", "endpoints", "nodes"]
+  verbs: ["get", "list", "watch"]
+- apiGroups: ["discovery.k8s.io"]
+  resources: ["endpointslices"]
   verbs: ["get", "list", "watch"]
 - apiGroups: [""]
   resources: ["events"]
@@ -78,7 +81,7 @@ rules:
   verbs: ["create"]
 - apiGroups: [""]
   resources: ["secrets"]
-  resourceNames: ["{{ include "metallb.secretName" . }}"]
+  resourceNames: [{{ include "metallb.secretName" . | quote }}]
   verbs: ["list"]
 - apiGroups: ["apps"]
   resources: ["deployments"]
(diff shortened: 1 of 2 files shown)


charts/metallb/templates/rbac.yaml
```diff
@@ -32,6 +32,9 @@ rules:
 - apiGroups: [""]
   resources: ["services", "endpoints", "nodes"]
   verbs: ["get", "list", "watch"]
+- apiGroups: ["discovery.k8s.io"]
+  resources: ["endpointslices"]
```
champtar, line 36: I know it match current manifests, but isn't it needed for the controller also ?

fedepaol: Nope, the same way endpoints are needed only for speakers.
The controller is started with ReadEndpoints = false which does not enable the listeners.
See https://github.com/metallb/metallb/blob/ebf611a825f04e62ad71f9eef1cea913ddab0611/controller/main.go#L154 vs 
https://github.com/metallb/metallb/blob/75e2dd302ec26ed979725f0f4bf7ab0718b7c0de/speaker/main.go#L134

champtar: controller doesn't need to look at the endpoints, only at the service, thanks for clearing that
LGTM then

treydock: Was there anything needed from me to move this pull request forward or does it look good as-is?

russellb: i think we're good

treydock:
Will this get released as like v0.10.2 ? I was really wanting to deploy with Helm when I did the v0.9.0 to v0.10.x update but figured need this change in there too.

russellb:
> Will this get released as like v0.10.2 ? I was really wanting to deploy with Helm when I did the v0.9.0 to v0.10.x update but figured need this change in there too.

Yeah.  I'm doing v0.10.2 right now.

treydock:
Excellent, thanks!

Issue #884, opened 2021-06-11
Support e2e test to run non kind clusters


s1061123:
Current e2e test is available only in kind cluster. e2e test is good not only for CI also good for testing its functionality from users' environment, such as not kind cluster (e.g. kubeadm, kubespray and so on).

russellb:
fixed by #885

PR #885, opened 2021-06-11
Change e2etest for non kind clusters

s1061123:
This change enables to run e2etest in non kind deployments
 - parameterize kubeconfig and system namespace
 - add option to change port of Pod because some distribution cannot open port tcp/80 in pod due to security considerations

This PR is for #884

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -55,10 +55,10 @@ var _ = ginkgo.Describe("BGP", func() {
 		jig := e2eservice.NewTestJig(cs, namespace, serviceName)
 
 		svc, err := jig.CreateLoadBalancerService(loadBalancerCreateTimeout,
-			nil)
+			tweakServicePort())
 		framework.ExpectNoError(err)
 
-		_, err = jig.Run(nil)
+		_, err = jig.Run(tweakRCPort())
 		framework.ExpectNoError(err)
 
 		defer func() {
@@ -70,11 +70,17 @@ var _ = ginkgo.Describe("BGP", func() {
 		ingressIP := e2eservice.GetIngressPoint(
 			&svc.Status.LoadBalancer.Ingress[0])
 
-		ginkgo.By("checking connectivity to its external VIP")
-
 		hostport := net.JoinHostPort(ingressIP, port)
 		address := fmt.Sprintf("http://%s/", hostport)
-		err = wgetRetry(true, address)
+
+		useDocker := true
+		if skipDockerCmd {
+			ginkgo.By("checking connectivity to its external VIP")
+			useDocker = false
+		} else {
+			ginkgo.By("checking connectivity to its external VIP with docker")
+		}
+		err = wgetRetry(useDocker, address)
 		framework.ExpectNoError(err)
 	})
 
--- a/e2etest/e2etest_suite_test.go
+++ b/e2etest/e2etest_suite_test.go
@@ -34,11 +34,17 @@ import (
 	e2econfig "k8s.io/kubernetes/test/e2e/framework/config"
 )
 
+// use ephemeral port for pod, instead of well-known port (tcp/80)
+var servicePodPort uint
+var skipDockerCmd bool
+
 // handleFlags sets up all flags and parses the command line.
 func handleFlags() {
 	e2econfig.CopyFlags(e2econfig.Flags, flag.CommandLine)
 	framework.RegisterCommonFlags(flag.CommandLine)
 	framework.RegisterClusterFlags(flag.CommandLine)
+	flag.UintVar(&servicePodPort, "service-pod-port", 80, "port number that pod opens, default: 80")
+	flag.BoolVar(&skipDockerCmd, "skip-docker", false, "et this to true if the BGP daemon is running on the host instead of in a container")
 	flag.Parse()
 }
 
--- a/tasks.py
+++ b/tasks.py
@@ -552,20 +552,33 @@ def lint(ctx, env="container"):
 
 
 @task(help={
-    "name": "name of the kind cluster to test.",
-    "export": "where to export kind logs"
+    "name": "name of the kind cluster to test (only kind uses).",
+    "export": "where to export kind logs.",
+    "kubeconfig": "kubeconfig location. By default, use the kubeconfig from kind.",
+    "system_namespaces": "comma separated list of Kubernetes system namespaces",
+    "service_pod_port": "port number that service pods open.",
+    "skip_docker": "don't use docker command in BGP testing."
 })
-def e2etest(ctx, name="kind", export=None):
+def e2etest(ctx, name="kind", export=None, kubeconfig=None, system_namespaces="kube-system,metallb-system", service_pod_port=80, skip_docker=False):
     """Run E2E tests against development cluster."""
-    validate_kind_version()
-    clusters = run("kind get clusters", hide=True).stdout.strip().splitlines()
-    if name in clusters:
-        kubeconfig = tempfile.NamedTemporaryFile()
-        run("kind export kubeconfig --name={} --kubeconfig={}".format(name, kubeconfig.name), pty=True, echo=True)
+    if skip_docker:
+        opt_skip_docker = "-skip-docker"
     else:
-        raise Exit(message="Unable to find cluster named: {}".format(name))
+        opt_skip_docker = ""
 
-    namespaces = ["kube-system", "metallb-system"]
+    if kubeconfig is None:
+        validate_kind_version()
+        clusters = run("kind get clusters", hide=True).stdout.strip().splitlines()
+        if name in clusters:
+            kubeconfig_file = tempfile.NamedTemporaryFile()
+            kubeconfig = kubeconfig_file.name
+            run("kind export kubeconfig --name={} --kubeconfig={}".format(name, kubeconfig), pty=True, echo=True)
+        else:
+            raise Exit(message="Unable to find cluster named: {}".format(name))
+    else:
+        os.environ['KUBECONFIG'] = kubeconfig
+
+    namespaces = system_namespaces.replace(' ', '').split(',')
     for ns in namespaces:
         run("kubectl -n {} wait --for=condition=Ready --all pods --timeout 300s".format(ns), hide=True)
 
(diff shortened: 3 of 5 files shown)


e2etest/bgp_test.go
```diff
@@ -74,7 +96,11 @@ var _ = ginkgo.Describe("BGP", func() {
 
 		hostport := net.JoinHostPort(ingressIP, port)
 		address := fmt.Sprintf("http://%s/", hostport)
-		_, err = runCommand(true, "wget", "-O-", "-q", address, "-T", "60")
+		useDocker := true
+		if skipDockerCmd {
+			useDocker = false
+		}
+		_, err = runCommand(useDocker, "wget", "-O-", "-q", address, "-T", "60")
```
markdgray, line 103: Is there a way to use the variable `skip-docker` directly to avoid the `if` clause above ? e.g.

`_, err = runCommand(!skipDocker, "wget", "-O-", "-q", address, "-T", "60")`

e2etest/e2etest_suite_test.go
```diff
@@ -50,6 +50,15 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
+// use ephemeral port (tcp/8080) for pod, instead of well-known port (tcp/80)
+var servicePodPort uint
+var skipDockerCmd bool
+
+func init() {
```
markdgray, line 57: Is this a Gingko convention? I don't see where it is called but I did test and it does all seem to be working correctly.

e2etest/e2etest_suite_test.go
```diff
@@ -50,6 +50,15 @@ func TestMain(m *testing.M) {
 	os.Exit(m.Run())
 }
 
+// use ephemeral port (tcp/8080) for pod, instead of well-known port (tcp/80)
```
markdgray, line 53: I think the comment is incorrect as the port is not 8080 (it's configurable). Also, it is not the ephemeral port.

tasks.py
```diff
@@ -552,20 +552,33 @@ def lint(ctx, env="container"):
 
 
 @task(help={
-    "name": "name of the kind cluster to test.",
-    "export": "where to export kind logs"
+    "name": "name of the kind cluster to test (only kind uses).",
+    "export": "where to export kind logs.",
+    "kubeconfig_path": "kubeconfig path. By default, load kubeconfig from kind.",
+    "system_namespace": "comma separated kubernetes system namespace list.",
```
markdgray, line 558: Suggests 
"system_namespaces": "comma separated list of Kubernetes system namespaces"

tasks.py
```diff
@@ -552,20 +552,33 @@ def lint(ctx, env="container"):
 
 
 @task(help={
-    "name": "name of the kind cluster to test.",
-    "export": "where to export kind logs"
+    "name": "name of the kind cluster to test (only kind uses).",
+    "export": "where to export kind logs.",
+    "kubeconfig_path": "kubeconfig path. By default, load kubeconfig from kind.",
```
markdgray, line 557: Suggest:
` "kubeconfig": "kubeconfig location. By default, use the kubeconfig from kind.",`

This will require some naming changes below but I think it looks better

tasks.py
```diff
@@ -552,20 +552,33 @@ def lint(ctx, env="container"):
...
     else:
-        raise Exit(message="Unable to find cluster named: {}".format(name))
+        opt_skip_docker = ""
+
+    if kubeconfig_path is None:
+        validate_kind_version()
+        clusters = run("kind get clusters", hide=True).stdout.strip().splitlines()
+        if name in clusters:
+            kubeconfig = tempfile.NamedTemporaryFile()
+            kubeconfig_path = kubeconfig.name
```
markdgray, line 574: Following on from my comment above about the names of the kubeconfig parameter for this task. This could be:

`kubeconfig = kubeconfig_file.name`

tasks.py
```diff
@@ -552,20 +552,33 @@ def lint(ctx, env="container"):
...
+            run("kind export kubeconfig --name={} --kubeconfig={}".format(name, kubeconfig.name), pty=True, echo=True)
+        else:
+            raise Exit(message="Unable to find cluster named: {}".format(name))
+    else:
+        os.environ['KUBECONFIG'] = kubeconfig_path
```
markdgray, line 579: It seems like you use the env variable to define the kubeconfig location here but the kubeconfig file above. Can you make it consistent between both paths?

e2etest/bgp_test.go
markdgray, line 103: Might be good to add some kind of log message so we can see what port is being used. (and maybe if docker is being used)

e2etest/bgp_test.go
```diff
@@ -54,11 +56,31 @@ var _ = ginkgo.Describe("BGP", func() {
 		serviceName := "external-local-lb"
 		jig := e2eservice.NewTestJig(cs, namespace, serviceName)
 
+		var svcTweak func(*v1.Service)
```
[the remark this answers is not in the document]
s1061123: changed.

e2etest/e2etest_suite_test.go
s1061123: It is golang's convention. see here https://golang.org/doc/effective_go#init

This is used not only here, also kubernetes https://github.com/kubernetes/kubernetes/blob/master/test/e2e/storage/testsuites/base.go#L37-L39

e2etest/e2etest_suite_test.go
s1061123: port 8080 could be ephemeral port based on RFC6056. https://datatracker.ietf.org/doc/html/rfc6056#section-2.1

I changed the comment.

tasks.py
s1061123: changed.

tasks.py
s1061123: changed.

tasks.py
s1061123: above kubeconfig , ` Run("kind export kubeconfig --name={} --kubeconfig={}".format(name, kubeconfig.name), pty=True, echo=True)` is to export (i.e. generate) kubeconfig from kind cluster, I suppose. `os.environ['KUBECONFIG'] = kubeconfig_path` is to set the specific kubeconfig files for following `kubectl` command. 

In Kind cluster, kubeconfig file is temporally file generated by `tempfile.NamedTemporaryFile()`, however, non kind cluster uses usual file (not temporally file), so we cannot make it consistent because kind's kubeconfig and non-kind's kubeconfig are totally different.

e2etest/bgp_test.go
s1061123: Add message to mention to use docker, but not added port because Service still uses port 80 because I only added `targetPort` to specify pod's port.

Please see Kubernetes document for the services. https://kubernetes.io/docs/concepts/services-networking/service/#defining-a-service

s1061123: Due to change the message, I need to keep `if skipDockerCmd`, so I can changed it no longer.

markdgray: ok, I didn't get that on the first read actually. I see now.

e2etest/e2etest_suite_test.go
markdgray: Ephemeral ports usually refer to the source port of a connection to a server (i.e. the client side port). Perhaps in this context (pod backing a service) this term is also used so its a bit confusing. How about:

"use configured port for pod instead of default port (tcp/80)"

Although, I am now also starting to doubt I understand the problem you are solving. Could you explain in the commit message why you need this PR?

tasks.py
markdgray: My main issue (nit) was in ```go test --provider=local -ginkgo.focus=BGP --kubeconfig={} -service-pod-port={} {}".format(kubeconfig, service_pod_port, opt_skip_docker))``` we pass the kubeconfig as a parameter and I thought we could just use the environment variable and remove that flag. However, I just tried and it doesn't work. I think what you have here is fine.

e2etest/e2etest_suite_test.go
markdgray: Ok, gotcha. This seems to be an established pattern. Why can't handleFlags() be used? I presume because we need to call `flag.UintVar` and `flag.BoolVar` early?

e2etest/bgp_test.go
```diff
@@ -54,11 +55,25 @@ var _ = ginkgo.Describe("BGP", func() {
...
+		if servicePodPort != 80 {
```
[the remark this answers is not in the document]
s1061123: okey, changed.

markdgray:
LGTM

e2etest/e2etest_suite_test.go
```diff
@@ -34,11 +34,17 @@ import (
...
+var servicePodPort uint
+var skipDockerCmd bool
+
 // handleFlags sets up all flags and parses the command line.
 func handleFlags() {
 	e2econfig.CopyFlags(e2econfig.Flags, flag.CommandLine)
 	framework.RegisterCommonFlags(flag.CommandLine)
 	framework.RegisterClusterFlags(flag.CommandLine)
+	flag.UintVar(&servicePodPort, "service-pod-port", 80, "port number that pod opens, default: 80")
+	flag.BoolVar(&skipDockerCmd, "skip-docker", false, "skip to use docker in BGP testing")
```
russellb, line 47: Maybe add something like: `Set this to true if the BGP daemon is running on the host instead of in a container.`.

e2etest/e2etest_suite_test.go
```diff
@@ -34,11 +34,17 @@ import (
...
+// use ephemeral port for pod, instead of well-known port (tcp/80)
+var servicePodPort uint
+var skipDockerCmd bool
+
 // handleFlags sets up all flags and parses the command line.
 func handleFlags() {
 	e2econfig.CopyFlags(e2econfig.Flags, flag.CommandLine)
 	framework.RegisterCommonFlags(flag.CommandLine)
 	framework.RegisterClusterFlags(flag.CommandLine)
+	flag.UintVar(&servicePodPort, "service-pod-port", 80, "port number that pod opens, default: 80")
```
russellb, line 46: Did you consider just changing the port numbers in all cases to something that would work for everyone instead of making it configurable?

s1061123: The security policy depends on each environments, platforms. Some platform might accept non well-known port and other platform might accept only more than tcp/3000. I don't know perfect port number which accepts everywhere, hence I make it configurable.

Please let me know if you know the perfect port number which not be blocked by any firewall (with any firewall configurations), cloud platform and other aspects. I'm happy to change it immediately.

s1061123: In addition, Kubernetes uses port 80, as default.

e2etest/e2etest_suite_test.go
s1061123: Changed.

e2etest/e2etest_suite_test.go
russellb: OK, not a big deal.  I know we really just have 2 systems we're dealing with at the moment, so I'd be fine with a simple change that worked for both CI systems.  What you have is fine too though.

PR #886, opened 2021-06-12
chart: new additionalLabels for podMonitors and PrometheusRules

ymettier:
Signed-off-by: Yves Mettier <ymettier@free.fr>

This PR allows to specify additionnal labels on podMonitors and PrometheusRules.

This is needed when you are using Prometheus-Operator and when you configured it to use specific podMonitors and specific PrometheusRules using a specific label.

Important note : I have not bumped the chart version in Chart.yaml (because I don't know when you may merge this PR).

### DIFF
--- a/charts/metallb/templates/podmonitor.yaml
+++ b/charts/metallb/templates/podmonitor.yaml
@@ -6,6 +6,9 @@ metadata:
   labels:
     {{- include "metallb.labels" . | nindent 4 }}
     app.kubernetes.io/component: controller
+    {{- if .Values.prometheus.podMonitor.additionalLabels }}
+{{ toYaml .Values.prometheus.podMonitor.additionalLabels | indent 4 }}
+    {{- end }}
 spec:
   jobLabel: {{ .Values.prometheus.podMonitor.jobLabel | quote }}
   selector:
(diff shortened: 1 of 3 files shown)


russellb:
@gclawes FYI

gclawes:
`Values.schema.json` should be updated whenever new fields are added to `values.yaml`.  @ymettier  let me know if you need help with this, or would like me to add a commit to this PR myself.

On the subject of `Chart.yaml` versions, this is incremented by the MetalLB release process, and not individual PRs for the charts.

ymettier:
I guess you will do it faster and better than I. @gclawes thanks in advance if you can do it : feel free to add a commit !

gclawes:
@russellb  I've pushed the relevant changes to `value.schema.json` and `Chart.yaml` to https://github.com/gclawes/metallb/tree/additionalLabels

russellb:
@gclawes i pushed your commits to this PR

charts/metallb/Chart.yaml
```diff
@@ -19,9 +19,11 @@ type: application
 # This is the chart version. This version number should be incremented each time you make changes
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
+# NOTE: this value is updated by the metallb release process
```
russellb, line 22: Thanks, @gclawes .  This is a nit for the future, but maybe we should change "0.9.6" to something else, like "0.0.0" ... I just thought of this, but now this version in main is behind what's in the v0.10 branch, since that automation just updates the release branch.

russellb:
Thanks, everyone. I'm going to wait for tests to all pass, but I'd also like to add something to the release notes.  I'll handle the release notes part.

charts/metallb/Chart.yaml
gclawes: Yeah, that makes sense.

PR #887, opened 2021-06-13
Fix for #801

### LINKED ISSUE #801: IP address not shared when multiple services use local external traffic policy
Thanks for filing an issue! A few things before we get started:

1. This is not the place to ask for troubleshooting. Unfortunately
   troubleshooting LB issues in k8s takes a lot of time and effort
   that we simply don't have. If your load-balancers aren't working,
   please don't file an issue unless you have identified a _specific_
   bug to fix. Otherwise, please ask for help in the `#metallb`
   channel on the Kubernetes Slack.
   > I have reached out on Slack for help on this issue, no resolution yet.
2. Please search the bug tracker for duplicates before filing your
   bug.
   > I searched and did not find anything.
3. If you're reporting a bug, please specify all of the following:
     - The bug itself, as detailed as you can.
     - Version of MetalLB
     > v0.9.5
     - Version of Kubernetes
     > v1.20.2
     - Name and version of network addon (e.g. Calico, Weave...)
     > Weave v2.8.0
     - Whether you've configured kube-proxy for iptables or ipvs mode
     > I am using the default iptables configuration.

Here is the problem. When I have two or more services with `externalTrafficPolicy: Local`, only one service has the shared IP assigned to it (see screenshot - I am using 127.0.0.1/32 as my shared IP for testing) and the others remain in pending status. When I remove that policy from my services, the shared IP is assigned to all the services. I don’t have port conflicts in my services.  

I believe that MetalLB should support sharing an IP across multiple services that require the local traffic policy. Is this supported?

**With Local Policy**
image

**Without Local Policy**
image

briantopping:
This PR fixes the problem with `ExternalTrafficPolicy: Local` services that also have a sharing key. 

Quite simply, the majority of the code here supports `BackendKey` making a far more accurate identity decision.  (https://github.com/kubevirt/cloud-provider-kubevirt/issues/18).

I appreciate that I didn't review these changes with the team before implementing it, I simply didn't have a good plan for this before starting. Apologies in advance for that. I'm happy to invest more in the PR and get the code restructured as necessary.

As far as gaining confidence in the PR, it seems like a good path would be to look at the state that is required for the new code to be reached. Since MetalLB doesn't currently handle `ExternalTrafficPolicy: Local` + sharing keys, it's believed that the changes only affect those who use this combination (and whatever overhead is added by new indexers). Changes:

* Indexers are introduced to cache information that was previously not searchable locally. Indexers add a lot of bulk to this commit but do not otherwise affect the logic. They exist for efficient object retrieval elsewhere in the code. 
* Node watcher used to exist only for speaker, but controller needs to know about the Pod CIDRs of nodes for this change to work. So the config interface is expanded to include the plugging into `NodeChanged` and moving that logic to the speaker as with the other `-Changed` interfaces. Since we generate events for all nodes now, we need to filter in the speaker, which only cares about the node that it's running on. 
* `BackendKey` was improved to more accurately return a unique key when the backends are indeed the same. The empty key is still returned in the case where `ExternalTrafficPolicy: Cluster`. To do so, it needed access to the clientset, so it had to become a method on the `controller` type. 
* There are `SetNode` and `SetEndpoint` interface methods in controller config. These should be mostly self-explanatory. `SetNode` maintains a subnet trie for each worker node. `SetEndpoint` queries against this to reverse resolve what node an endpoint is on. Endpoints have an `OwnerReference` that should point to the service, but it's not reliable enough to depend on, especially if the endpoints are maintained by code other than the k8s endpoint controller (which is legal). The goal here is to distill that all endpoints are on the same node.

Fixed #801 
Signed-off-by: Brian Topping <brian@coglative.com>


russellb:
@briantopping FYI on the test failures in case you didn't see them

briantopping:
@russellb i did see those, thanks. Given this draft, is this something the team could approve in concept and get back in the flow of a regular approval? I can work on cleaning it up from there.

briantopping:
@russellb @danderson I created a [thread on Slack in #gardener](https://kubernetes.slack.com/archives/CB57N0BFG/p1624235602223700) with some background of this PR. Maybe we'll be able to get some more opinions from that path. It's a pretty critical fix for Gardener users, the thread there is to make sure any solution is adequately (but not overly) scoped.

fedepaol:
Closing this as stale

PR #889, opened 2021-06-14
design: arp-load-sharing proposal

guymguym:
This PR proposes a design to add an option to MetalLB to support "ARP Load Sharing" mode.
It is a followup of a slack discussion with @russellb -
https://kubernetes.slack.com/archives/C8FSTGU56/p1623341570127200.

The details of the proposal are open to discussion, and I would be happy to hear if this is valuable for others in the community, so that I can fill up the more details, or what concerns this is raising.

TIA!
Guy

### DIFF
--- a/design/arp-load-sharing.md
+++ b/design/arp-load-sharing.md
@@ -1,0 +1,41 @@
+# ARP Load Sharing
+
+## Summary
+
+MetalLB in layer 2 mode does not support load balancing. See [here](https://metallb.universe.tf/concepts/layer2/).
+However, in some cases, L2 is the only option, yet we still require a load balancing solution which is internal to our infrastructure.
+
+A method that can achieve that is known as "ARP Load Sharing". See for example [here](https://docs.paloaltonetworks.com/pan-os/10-0/pan-os-admin/high-availability/ha-concepts/arp-load-sharing). This technique involves fiddling with how ARP replies are sent, which raises the concern that it might get blocked by the network switch or security policy since it seems like an [ARP Spoofing](https://en.wikipedia.org/wiki/ARP_spoofing) attack. See more information about "does arp spoofing work on all lans" [here](https://security.stackexchange.com/questions/133784/does-arp-spoofing-work-on-all-lans).
+
+This document proposes adding an option to MetalLB to support "ARP Load Sharing" mode.
+
+## Motivation
+
+When BGP is not available, in order to support load balancing and HA for a Kubernetes Service, we need to create as many Services as there are nodes, so that each node will be an owner of one VIP and that node will be able to get traffic directly from clients. The VIP is needed to support HA so that on node failure the VIP can move to another node. However, even with a service per node, we still need to add a higher level load balancer, which will balance those VIPs for the clients (for example using switches that can perform L3-routing, DNS round-robin, etc.).
+
+We want to simplify this setup, and have an option to "serve" a VIP from all nodes.
+
+### Goals
+
+- Add config option to enable/disable arp-load-sharing.
+- Assuming the network allows it, the clients should be balanced between the nodes using a single Service VIP.
+
+### Non-Goals
+
+- It is expected that the more sophisticated the network gateway is, that it might block this behavior, unless it is manually configured to allow it, but this will remain outside the scope of this proposal, as a manual integration to the network environment.
+
+## Proposal
+
+DRAFT - still some open questions:
+
+- Add config option to enable/disable arp-load-sharing
+  - where? can we tag a feature as alpha/experimental?
+- In this mode instead of single VIP owner, any node will be able to reply to ARP requests on the IP. We can think of two ways to answer:
+  - The simplest way is that all nodes will answer all the requests (with random jitter), and will effectively "race" on the ARP mapping for the clients, which will keep those mappings for some time until their ARP cache expires.
+  - More organized way could be that every node will enumerate the list of nodes where the service is available (for example based on externalTrafficPolicy or other restrictions) and then use the client IP hash to select just one node for each client.
+
+
+## Alternatives
+
+- Use BGP if available, although it is still quite rare in the enterprise.
+- Introduce LB router in L3+, which requires a VIP per node.


design/arp-load-sharing.md
```diff
@@ -0,0 +1,41 @@
+# ARP Load Sharing
+
+## Summary
+
+MetalLB in layer 2 mode does not support load balancing. See [here](https://metallb.universe.tf/concepts/layer2/).
```
russellb, line 5: Well, load balancing is done by the Service load balancer once it hits a Kubernetes node.  I'd say that the load balancing is limited by going through a single node before being load balanced across the cluster.

design/arp-load-sharing.md
```diff
@@ -0,0 +1,41 @@
...
+However, in some cases, L2 is the only option, yet we still require a load balancing solution which is internal to our infrastructure.
+
+A method that can achieve that is known as "ARP Load Sharing". See for example [here](https://docs.paloaltonetworks.com/pan-os/10-0/pan-os-admin/high-availability/ha-concepts/arp-load-sharing). This technique involves fiddling with how ARP replies are sent, which raises the concern that it might get blocked by the network switch or security policy since it seems like an [ARP Spoofing](https://en.wikipedia.org/wiki/ARP_spoofing) attack. See more information about "does arp spoofing work on all lans" [here](https://security.stackexchange.com/questions/133784/does-arp-spoofing-work-on-all-lans).
+
+This document proposes adding an option to MetalLB to support "ARP Load Sharing" mode.
```
russellb, line 10: Is this only expected to be useful when multiple clients are on the same layer2 domain as the cluster?  Otherwise, the client issuing the ARP request is just the router, right?

I see in the palo alto doc, it lists this limitation: "Use ARP load-sharing only when no Layer 3 device exists between the firewall and end hosts, that is, when end hosts use the firewall as their default gateway."

Does this match your use case?

design/arp-load-sharing.md
guymguym: Right, I mean network load-balancing, not the compute side of it.
I can clarify.

design/arp-load-sharing.md
guymguym: @russellb Thanks! that's a good point.

A router will appear as a **single** L2 client to the cluster, so this is indeed a limitation of this mode for clients behind a router. However for clients on the same LAN/VLAN this approach provides the best balancing L2 can buy.

My use-case is an object-store software, that runs on all the cluster nodes, so even the relatively simple clients balancing over nodes allows to scale to the full throughput of the cluster's network.

I'm pretty sure that this mode should be considered experimental until proven otherwise, it might get routers to block the cluster ports. But it will look like magic in the hands of a skilled admin.

russellb: Yeah, it's pretty clever for the environments where it would work.  I would definitely agree that it should be experimental.

So then we're left deciding whether to take this experimental mode and the added complexity it introduces, for a limited use case (only useful to network-local clients, and potentially problematic in some environments).  With my maintainer hat on, I think I'd want to see some more demand from people saying they would use this mode.

With my Red Hat on, I don't think I'd want to support this.  I'm pretty nervous of it resulting in expensive support engagements figuring out why someone's network is not behaving as we expect with this mode.

troppens:
I am working with @guymguym on use cases to process massive amounts of data which are born or accessed outside K8s, OpenShift, etc. One example is edge computing. External devices (e.g., sensors, cameras) generate data which should be processed on modern infrastructure. It is on prem, so we cannot use a load balancer provided by a public cloud. Solutions such as load balancing appliances and BGP have dependencies to infrastructure outside K8s, etc. K8s built-in capabilities which I know suffer from single-node bottlenecking. Having a solution such as proposed by Guy would simplify the configuration for respective data-intensive workloads.

design/arp-load-sharing.md
```diff
@@ -0,0 +1,41 @@
...
+### Non-Goals
+
+- It is expected that the more sophisticated the network gateway is, that it might block this behavior, unless it is manually configured to allow it, but this will remain outside the scope of this proposal, as a manual integration to the network environment.
+
+## Proposal
+
+DRAFT - still some open questions:
+
+- Add config option to enable/disable arp-load-sharing
+  - where? can we tag a feature as alpha/experimental?
```
rata, line 32: There is currently no way. I think adding "experimental" to the knob name is enough :)

design/arp-load-sharing.md
```diff
@@ -0,0 +1,41 @@
...
+- In this mode instead of single VIP owner, any node will be able to reply to ARP requests on the IP. We can think of two ways to answer:
+  - The simplest way is that all nodes will answer all the requests (with random jitter), and will effectively "race" on the ARP mapping for the clients, which will keep those mappings for some time until their ARP cache expires.
```
rata, line 34: Currently a failover (when a node stops responding ARPs and a new node starts answering them) implies downtime in most scenarios. Not because we are slow to detect this (we detect this in <1 sec for most clusters), but because of external network equipment routing to the new node.

I haven't read the linked docs yet, but wouldn't this cause that too? Or should this only be enabled on some routers where this behavior won't cause downtime?

design/arp-load-sharing.md
```diff
@@ -0,0 +1,41 @@
...
+
+This document proposes adding an option to MetalLB to support "ARP Load Sharing" mode.
+
+## Motivation
+
+When BGP is not available, in order to support load balancing and HA for a Kubernetes Service, we need to create as many Services as there are nodes, so that each node will be an owner of one VIP and that node will be able to get traffic directly from clients. The VIP is needed to support HA so that on node failure the VIP can move to another node. However, even with a service per node, we still need to add a higher level load balancer, which will balance those VIPs for the clients (for example using switches that can perform L3-routing, DNS round-robin, etc.).
```
rata, line 14: I don't follow this:
> we need to create as many Services as there are nodes, so that each node will be an owner of one VIP and that node will be able to get traffic directly from clients

creating as many services as nodes doesn't mean that each service owner will be announced on a different node (i.e. no two services will arrive in the same node), does it? Is there something about how the current hashing works that will cause that?

design/arp-load-sharing.md
```diff
@@ -0,0 +1,41 @@
...
+
+We want to simplify this setup, and have an option to "serve" a VIP from all nodes.
```
rata, line 16: What is the use case to do this? I mean, why is that serving from one node (and fast failover if that node fails) is problematic? Is it because your ingress is limited to the ingress of a single node? Is there some other reason?

Can we state what is the underlying problem we want to solve by adding arp load balancing?

design/arp-load-sharing.md
```diff
@@ -0,0 +1,41 @@
...
+
+### Goals
+
+- Add config option to enable/disable arp-load-sharing.
+- Assuming the network allows it, the clients should be balanced between the nodes using a single Service VIP.
+
+### Non-Goals
+
+- It is expected that the more sophisticated the network gateway is, that it might block this behavior, unless it is manually configured to allow it, but this will remain outside the scope of this proposal, as a manual integration to the network environment.
```
rata, line 25: I fully agree. Furthermore, I'd like to see that the code to implement this doesn't break the existing functionality. I'm okay with slight changes to current functionality, as far as they _really_ seem safe. This extension, if merged, should be as "surgical" as possible.

design/arp-load-sharing.md
```diff
@@ -0,0 +1,41 @@
+# ARP Load Sharing
+
+## Summary
+
+MetalLB in layer 2 mode does not support load balancing. See [here](https://metallb.universe.tf/concepts/layer2/).
+However, in some cases, L2 is the only option, yet we still require a load balancing solution which is internal to our infrastructure.
+
+A method that can achieve that is known as "ARP Load Sharing". See for example [here](https://docs.paloaltonetworks.com/pan-os/10-0/pan-os-admin/high-availability/ha-concepts/arp-load-sharing). This technique involves fiddling with how ARP replies are sent, which raises the concern that it might get blocked by the network switch or security policy since it seems like an [ARP Spoofing](https://en.wikipedia.org/wiki/ARP_spoofing) attack. See more information about "does arp spoofing work on all lans" [here](https://security.stackexchange.com/questions/133784/does-arp-spoofing-work-on-all-lans).
```
rata, line 8: The links there mention each node should have a unique mac for the interface in advertises, right now IIRC we don't set any MAC and advertise on all interfaces. How do you propose to select and use a unique mac when adding this to MetalLB? Also, does this need to know which interface we should send the GARP or not?

rata, line 8: Sorry to bother with so many questions, but in IPv6 we don't have GARP as the link mentions. Do you propose to use this in some way for IPv6 or is this an IPv4-only feature?

rata (commented):
@guymguym thanks for the proposal, seems quite interesting! Left some questions as I was not aware of ARP load balancing before :)

fedepaol:
Closing as outdated, and no replies were given

PR #891, opened 2021-06-15
Add fsGroup to controller deployment

### LINKED ISSUE #890: controller won't run on k8s > 1.20 without fsGroup
I identified an issue while upgrading our (outdated) 1.17 cluster to 1.21.
With the TokenRequest and TokenRequestProjection featuregates graduated to GA in 1.20 the SerivceAccount token is now mounted via the [projected volume API](https://kubernetes.io/docs/concepts/storage/volumes/#projected).
Since metallb runs as nobody user and the projected token is only readable by root(600) or root and fsGroup if set(640) we need to set the fsGroup in the helm chart, or at least, make securityContext configurable via values.yaml.

I'll happily provide a PR for the chart, I'm just curious if I'm doing something wrong and there's another way to run metallb on k8s > 1.20 or if just nobody run it on 1.20+ yet (what I couldn't imagine tbh).

TheKangaroo:
Fixes #890
we need to set fsGroup for the controller to read the k8s api token.

### DIFF
--- a/charts/metallb/templates/controller.yaml
+++ b/charts/metallb/templates/controller.yaml
@@ -33,9 +33,10 @@ spec:
       {{- end }}
       serviceAccountName: {{ template "metallb.controller.serviceAccountName" . }}
       terminationGracePeriodSeconds: 0
+{{- if .Values.controller.securityContext }}
       securityContext:
-        runAsNonRoot: true
-        runAsUser: 65534 # nobody
+{{ toYaml .Values.controller.securityContext | indent 8 }}
+{{- end }}
       containers:
       - name: controller
         image: {{ .Values.controller.image.repository }}:{{ .Values.controller.image.tag | default .Chart.AppVersion }}
--- a/charts/metallb/values.yaml
+++ b/charts/metallb/values.yaml
@@ -138,6 +138,11 @@ controller:
     # true, a name is generated using the fullname template
     name: ""
     annotations: {}
+  securityContext:
+    runAsNonRoot: true
+    # nobody
+    runAsUser: 65534
+    fsGroup: 65534
   resources: {}
     # limits:
       # cpu: 100m
--- a/manifests/metallb.yaml
+++ b/manifests/metallb.yaml
@@ -439,5 +439,6 @@ spec:
       securityContext:
         runAsNonRoot: true
         runAsUser: 65534
+        fsGroup: 65534
       serviceAccountName: controller
       terminationGracePeriodSeconds: 0
--- a/website/content/release-notes/_index.md
+++ b/website/content/release-notes/_index.md
@@ -16,6 +16,15 @@ Changes in behavior:
 
 Bug Fixes:
 
+## Version 0.10.3
+
+Bug Fixes:
+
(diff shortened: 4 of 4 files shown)


russellb:
@gclawes FYI on this PR.  I wonder if we should add a comment by the chart version to clarify that it gets updated as part of our release process and doesn't need to be updated in PRs?

gclawes:
The description says this PR is adding to the values, but I only see `controller.yaml` changing.

gclawes:
Edit: Ah, I see here: https://github.com/metallb/metallb/issues/890#issuecomment-862424413

@russellb given that this PR has been reduced in scope, want me to make the comment in another PR?

russellb:
> @russellb given that this PR has been reduced in scope, want me to make the comment in another PR?

Yes, that'd be great

russellb:
I'd like to add a release note about this change.  How about this, based on the bug report:

```
 -  Add `fsGroup` to the MetalLB controller deployment to address compatibility with Kubernetes 1.20 and later. With the TokenRequest and TokenRequestProjection featuregates graduated to GA in 1.20, the SerivceAccount token is now mounted via the projected volume API. The security context needed an update to ensure the controller is able to read the token.  (Issue #890)
```

gclawes:
LGTM on the actual helm chart diff and the release notes.  Should the manifests be updated too?

gclawes:
@russellb I will make the helm chart version note in #886

champtar:
s/SerivceAccount/ServiceAccount/

charts/metallb/templates/controller.yaml
```diff
@@ -36,6 +36,7 @@ spec:
       securityContext:
         runAsNonRoot: true
         runAsUser: 65534 # nobody
+        fsGroup: 65534
```
russellb, line 39: This will need to go in manifests/metallb.yaml too, right?

russellb:
> s/SerivceAccount/ServiceAccount/

thanks, fixed

TheKangaroo:
Sorry for the confusion to comment in https://github.com/metallb/metallb/issues/890 instead of the PR before.
I think it would be best to default fsGroup to 65534, but add the option to override it via values.yaml.
For the release note, I dug a bit deeper in the kubernetes Docs and think `TokenRequest` and `TokenRequestProjection` are just part of the problem. The root cause for this is the changed ServiceAccount admission controller behaviour (https://github.com/kubernetes/kubernetes/issues/70679) which graduated to beta in 1.21.

charts/metallb/templates/controller.yaml
TheKangaroo: thanks, I fixed that

russellb:
an updated release note ...

```
 -  Add `fsGroup` to the MetalLB controller deployment to address compatibility with Kubernetes 1.21 
    and later. See [Kubernetes issue #70679](https://github.com/kubernetes/kubernetes/issues/70679).
    This ensures the MetalLB controller can read the service account token volume.  (Issue #890)
```

Look OK to you, @TheKangaroo ?  or propose changes as you see fit

TheKangaroo:
Looks good to me, thank you :)

russellb:
also backported these changes to the v0.10 branch

PR #832, opened 2021-04-07
design: BGP FRR integration

markdgray:
This design document discusses integration of FRR as an
alternative BGP agent implementation for MetalLB.

Signed-off-by: Mark Gray <mark.d.gray@redhat.com>

PR #822, opened 2021-03-23
website: Fix a hugo error

russellb:
When I try to run `hugo server` right now, I get the following error:

  Error: Error building site: "/home/rbryant/go/src/github.com/metallb/metallb/website/content/_header.md:1:1": plain HTML documents not supported

It seems like this is an issue with a more recent version of hugo.
I'm using the following version from a Fedora package:

  Hugo Static Site Generator v0.80.0/extended linux/amd64 BuildDate: unknown

This change is based on a suggested workaround in the following issue:

  https://github.com/gohugoio/hugo/issues/7296

The workaround seems harmless, and may no longer be needed if the
following issue gets resolved at some point in the future:

  https://github.com/gohugoio/hugo/issues/6098

PR #892, opened 2021-06-15
Add wget retries to both layer2 and bgp e2e tests

msherif1234:

Add wget retries (4 attempts) to layer2 and bgp e2e tests, which was mentioned during the review of PR #882 
Unit-Test:

[unit-test-program](https://play.golang.org/p/TJTfphar8g4)

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -74,7 +74,7 @@ var _ = ginkgo.Describe("BGP", func() {
 
 		hostport := net.JoinHostPort(ingressIP, port)
 		address := fmt.Sprintf("http://%s/", hostport)
-		_, err = runCommand(true, "wget", "-O-", "-q", address, "-T", "60")
+		err = wgetRetry(true, address)
 		framework.ExpectNoError(err)
 	})
 
--- a/e2etest/utils.go
+++ b/e2etest/utils.go
@@ -7,6 +7,11 @@ import (
 	"k8s.io/kubernetes/test/e2e/framework"
 )
 
+const (
+	NetworkFailure = 4
+	retryLimit     = 4
+)
+
 // DescribeSvc logs the output of kubectl describe svc for the given namespace
 func DescribeSvc(ns string) {
 	framework.Logf("\nOutput of kubectl describe svc:\n")
@@ -24,4 +29,27 @@ func runCommand(bgp bool, name string, args ...string) (string, error) {
 	}
 	out, err := exec.Command(name, args...).CombinedOutput()
 	return string(out), err
+}
+
+func wgetRetry(bgp bool, address string) error {
+	retrycnt := 0
+	code := 0
+	var err error
+
+	// Retry loop to handle wget NetworkFailure errors
+	for {
+		_, err = runCommand(bgp, "wget", "-O-", "-q", address, "-T", "60")
+		if exitErr, ok := err.(*exec.ExitError); err != nil && ok {
+			code = exitErr.ExitCode()
+		} else {
+			break
+		}
+		if retrycnt < retryLimit && code == NetworkFailure {
+			framework.Logf(" wget failed with code %d, err %s retrycnt %d\n", code, err, retrycnt)
+			retrycnt++
+		} else {
+			break
+		}
+	}
+	return err
 }
--- a/e2etest/layer2_test.go
+++ b/e2etest/layer2_test.go
@@ -73,7 +73,7 @@ var _ = ginkgo.Describe("L2", func() {
 
 		hostport := net.JoinHostPort(ingressIP, port)
 		address := fmt.Sprintf("http://%s/", hostport)
-		_, err = runCommand(false, "wget", "-O-", "-q", address, "-T", "60")
+		err = wgetRetry(false, address)
 		framework.ExpectNoError(err)
 	})
 


e2etest/bgp_test.go
```diff
@@ -74,7 +74,7 @@ var _ = ginkgo.Describe("BGP", func() {
 
 		hostport := net.JoinHostPort(ingressIP, port)
 		address := fmt.Sprintf("http://%s/", hostport)
-		_, err = runCommand(true, "wget", "-O-", "-q", address, "-T", "60")
+		_, err = runCommand(true, "wget", "-O-", "-q", address, "-T", "60", "-t", "40")
```
russellb, line 77: I see the default for `-t` / `--tries` is 20.  I wonder if increasing this is what's needed, or if maybe adding `-w` / `--wait` ?  It's not super clear to me what the wait behavior is by default if a connection fails.  If it's really retrying 20 times and still not working, either we should wait longer between retries, or maybe it's a real failure.

russellb:
cc @markdgray who originally wrote this code

e2etest/bgp_test.go
msherif1234:
```
‘--waitretry=seconds’
If you don’t want Wget to wait between every retrieval, but only between retries of failed downloads, you can use this option. Wget will use linear backoff, waiting 1 second after the first failure on a given file, then waiting 2 seconds after the second failure on that file, up to the maximum number of seconds you specify.

By default, Wget will assume a value of 10 seconds.
```

10 seconds is more the enough

markdgray: Is there a link to the failed CI job? I went into circleci at https://app.circleci.com/pipelines/github/metallb/metallb?status=failed and I dont see it anymore (it failed in #882). Maybe they get purged after the PR is merged?

I think the `--waitretry` options seems more specific.

russellb: From what I remember on the failure in #822, it failed when wget exited non-zero (code 4, or "Network failure"), and didn't spot anything else obvious.  It also didn't collect all the logs from the cluster, but maybe that run was from before we merged the PR to collect logs ...

markdgray: It seems like `waitretry` should help and wont hurt. If it happens again, we can look through the logs again. It also looks like `connection refused` or `404` will not cause a retry . Maybe we could also add `--retry-connrefused`.

markdgray: @russellb Is this the same race condition you saw a few weeks ago?

russellb: I don't think it's what I saw when running tests a few weeks ago.

I just did a local test.  I ran `inv dev-env -p layer2` and then ran `wget http://172.18.0.6 -T 60`, which is an IP in the MetalLB config, but is not allocated to any Service.  The result is it fails with `failed: No route to host` and an exit code of 4.  It does *not* retry.  I'm guessing this is the scenario we saw happen in CI, where we tried to connect before the network was ready.

Adding `-t` does not make it retry, nor does `--retry-connrefused`.  I'm guessing wget treats this as a fatal error where retrying is pointless.  Maybe we need to add retrying into the test code if the failure code is 4 (Network failure) ?

markdgray: Ok. I think the bash loop might be the best way to do it. @msherif1234 Do you need help or will you do it?

russellb: bash loop or go loop?

markdgray: :D sorry. This morning I have been looking at Go, Python, Bash and C so I am all mixed up. Go would be better here, of course :D

msherif1234: I add retry loop for wget PTAL

markdgray:
Looking at this failure, Is the test framework raising an alarm because the test is going on for too long? Maybe we are never exiting the loop? The test is running for >13 minutes which seems like a lot.

msherif1234:
> Looking at this failure, Is the test framework raising an alarm because the test is going on for too long? Maybe we are never exiting the loop? The test is running for >13 minutes which seems like a lot.
Hmm I ran the new code on the playground even when I hardcode the error code 4 it exist after 4 attempts do u see any of the new log messages I couldn't find any ?

russellb:
For reference, here is another failure of the layer2 CI job.  https://app.circleci.com/pipelines/github/metallb/metallb/885/workflows/7196b339-c69b-48c5-a917-013ae427a3ff/jobs/2912

It's from #832, and the failure is the same as I've seen before.  `wget` exited with exit code 4 (Network Failure).  The retries here should fix this, assuming it's just that we tried too soon, and not that it's really broken.

e2etest/utils.go
```diff
@@ -25,3 +30,26 @@ func runCommand(bgp bool, name string, args ...string) (string, error) {
...
+
+func wgetRetry(bgp bool, address string) error {
+	retrycnt := 0
+	code := NetworkFailure
+	var err error
+
+	// Retry loop to handle wget NetworkFailure errors
+	for {
+		_, err = runCommand(bgp, "wget", "-O-", "-q", address, "-T", "60")
+		if err != nil && retrycnt < retryLimit && code == NetworkFailure {
```
russellb, line 42: I'm a little confused by `code == NetworkFailure` here.  This will *always* be true the first time through the loop.  We extract the exit code *after* the check, but then don't check it.  On the 2nd (or later) runs through, we're checking the exit code of the previous execution of wget.  I suspect this isn't what you intended?

msherif1234: yeah u are right logic wasn't exactly what I wanted I rearranged hopefully will be ok now

e2etest/utils.go
```diff
@@ -25,3 +30,26 @@ func runCommand(bgp bool, name string, args ...string) (string, error) {
...
+
+func wgetRetry(bgp bool, address string) error {
+	retrycnt := 0
+	code := 0
+	var err error
+
+	// Retry loop to handle wget NetworkFailure errors
+	for {
+		_, err = runCommand(bgp, "wget", "-O-", "-q", address, "-T", "60")
+		if exitErr, ok := err.(*exec.ExitError); err != nil && ok {
```
russellb, line 42: There's a check `err != nil`, but wouldn't this break if `err == nil` because of `err.(*exec.ExitError)` ?

but the tests are passing ... what am I missing :)

msherif1234: if there is no err "nil" we will break from for loop ? maybe I don't understand the question ? basically if we have err and we are able to extract the exit code then read the code and check for NetworkFailure

russellb: nevermind, i got it sorted out

PR #893, opened 2021-06-17
charts: Clear version in Chart.yaml in main branch

russellb:
These versions get updated automatically as part of the release process.
This will update release branches and tags, but never in main. To avoid
confusion, just leave the version as 0.0.0 in main, along side the
comment that says this is updated automatically.

### DIFF
--- a/charts/metallb/Chart.yaml
+++ b/charts/metallb/Chart.yaml
@@ -20,10 +20,10 @@ type: application
 # to the chart and its templates, including the app version.
 # Versions are expected to follow Semantic Versioning (https://semver.org/)
 # NOTE: this value is updated by the metallb release process
-version: 0.9.6
+version: 0.0.0
 
 # This is the version number of the application being deployed. This version number should be
 # incremented each time you make changes to the application. Versions are not expected to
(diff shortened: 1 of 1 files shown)


russellb:
@gclawes can you ack this?

gclawes:
I think this makes sense, the diff looks fine, though technically I don't think `0.0.0` is valid semver.

For all intents and purposes it doesn't matter, but if somebody tries to do something with these files in `main`, it may cause unexpected behaviors.

russellb:
> I think this makes sense, the diff looks fine, though technically I don't think `0.0.0` is valid semver.
> 
> For all intents and purposes it doesn't matter, but if somebody tries to do something with these files in `main`, it may cause unexpected behaviors.

any ideas?  We could comment out the line, and make the release script replace the whole line with an uncommented line maybe?

or we could just make it `REPLACED_BY_RELEASE_SCRIPT`  :-D

gclawes:
I think `0.0.0` is fine, I just wanted to note it in case anyone asks in the future.

russellb:
> I think `0.0.0` is fine, I just wanted to note it in case anyone asks in the future.

OK, I'll merge this ... we can always change again!

PR #898, opened 2021-06-27
layer2: stop spamLoop() ticker when not used

### LINKED ISSUE #745: Stop spamLoop() ticker when not used (golang 1.15+)
Now that #736 is merged, an improvement suggested by @russellb is to stop the ticker when not used.
https://github.com/metallb/metallb/blob/0c1ae8cc3b057b745ea4003ed65e81e404d424c9/internal/layer2/announcer.go#L144

Golang 1.15 introduces ticker.Reset(), which would make it easy to stop/start the ticker when needed

champtar:
Fixes #745

(tested on 1.19.10 cluster / CentOS 8)

### DIFF
--- a/internal/layer2/announcer.go
+++ b/internal/layer2/announcer.go
@@ -138,11 +138,16 @@ func (a *Announce) updateInterfaces() {
 func (a *Announce) spamLoop() {
 	// Map IP to spam stop time.
 	m := map[string]time.Time{}
-	// See https://github.com/metallb/metallb/issues/172 for the 1100 choice.
-	ticker := time.NewTicker(1100 * time.Millisecond)
+	// We can't create a stopped ticker, so create one with a big period to avoid ticking for nothing
+	ticker := time.NewTicker(time.Hour)
+	ticker.Stop()
 	for {
 		select {
 		case ip := <-a.spamCh:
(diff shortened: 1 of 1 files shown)


internal/layer2/announcer.go
```diff
@@ -138,11 +138,15 @@ func (a *Announce) updateInterfaces() {
 func (a *Announce) spamLoop() {
 	// Map IP to spam stop time.
 	m := map[string]time.Time{}
-	// See https://github.com/metallb/metallb/issues/172 for the 1100 choice.
-	ticker := time.NewTicker(1100 * time.Millisecond)
+	ticker := time.NewTicker(time.Hour)
```
russellb, line 141: I know this is a nit-pick, but I'd suggest changing this line back.  It made me wonder what the significance of an hour was, but I'm pretty sure you just picked a "big" number since it was going to be stopped right away.  I'd either add a comment clarifying that, or just change it back to the original value used elsewhere.

champtar: Added a comment, tell me if it's better

russellb: that works

russellb:
lgtm, thanks for following up on this!

PR #902, opened 2021-07-02
Make BGP source address configurable

### LINKED ISSUE #670: support selecting the source address when speaker sets up a bgp session

johananl:
This is a follow up for #672.

TODO:

- [x] Add docs
- [ ] ~Add tests~

>NOTE: There doesn't seem to be an easy way to test the source address functionality. We would likely need to do some significant refactoring to the tests and possibly to the way we deal with BGP sessions in the code itself.

Fixes #670.

### DIFF
--- a/internal/bgp/bgp.go
+++ b/internal/bgp/bgp.go
@@ -29,6 +29,7 @@ type Session struct {
 	routerID         net.IP // May be nil, meaning "derive from context"
 	myNode           string
 	addr             string
+	srcAddr          net.IP
 	peerASN          uint32
 	peerFBASNSupport bool
 	holdTime         time.Duration
@@ -170,7 +171,7 @@ func (s *Session) connect() error {
 	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
 	defer cancel()
 	deadline, _ := ctx.Deadline()
-	conn, err := dialMD5(ctx, s.addr, s.password)
+	conn, err := dialMD5(ctx, s.addr, s.srcAddr, s.password)
 	if err != nil {
 		return fmt.Errorf("dial %q: %s", s.addr, err)
 	}
@@ -357,9 +358,10 @@ func (s *Session) sendKeepalive() error {
 //
 // The session will immediately try to connect and synchronize its
 // local state with the peer.
-func New(l log.Logger, addr string, asn uint32, routerID net.IP, peerASN uint32, holdTime time.Duration, password string, myNode string) (*Session, error) {
+func New(l log.Logger, addr string, srcAddr net.IP, asn uint32, routerID net.IP, peerASN uint32, holdTime time.Duration, password string, myNode string) (*Session, error) {
 	ret := &Session{
 		addr:        addr,
+		srcAddr:     srcAddr,
 		asn:         asn,
 		routerID:    routerID.To4(),
 		myNode:      myNode,
@@ -525,8 +527,26 @@ type tcpmd5sig struct {
 // proper TCP MD5 options when the password is not empty. Works by manupulating
 // the low level FD's, skipping the net.Conn API as it has not hooks to set
 // the neccessary sockopts for TCP MD5.
-func dialMD5(ctx context.Context, addr, password string) (net.Conn, error) {
-	laddr, err := net.ResolveTCPAddr("tcp", "[::]:0")
+func dialMD5(ctx context.Context, addr string, srcAddr net.IP, password string) (net.Conn, error) {
+	// If srcAddr exists on any of the local network interfaces, use it as the
+	// source address of the TCP socket. Otherwise, use the IPv6 unspecified
+	// address ("::") to let the kernel figure out the source address.
+	// NOTE: On Linux, "::" also includes "0.0.0.0" (all IPv4 addresses).
+	a := "[::]"
+	if srcAddr != nil {
+		ifs, err := net.Interfaces()
+		if err != nil {
+			return nil, fmt.Errorf("Querying local interfaces: %w", err)
+		}
+
+		if !localAddressExists(ifs, srcAddr) {
+			return nil, fmt.Errorf("Address %q doesn't exist on this host", srcAddr)
+		}
+
+		a = fmt.Sprintf("[%s]", srcAddr.String())
+	}
+
+	laddr, err := net.ResolveTCPAddr("tcp", fmt.Sprintf("%s:0", a))
 	if err != nil {
 		return nil, fmt.Errorf("Error resolving local address: %s ", err)
 	}
@@ -670,4 +690,27 @@ func buildTCPMD5Sig(addr net.IP, key string) tcpmd5sig {
 	copy(t.key[0:], []byte(key))
 
 	return t
+}
+
+// localAddressExists returns true if the address addr exists on any of the
+// network interfaces in the ifs slice.
+func localAddressExists(ifs []net.Interface, addr net.IP) bool {
+	for _, i := range ifs {
+		addresses, err := i.Addrs()
+		if err != nil {
+			continue
+		}
+
+		for _, a := range addresses {
+			ip, ok := a.(*net.IPNet)
+			if !ok {
+				continue
+			}
+			if ip.IP.Equal(addr) {
+				return true
+			}
+		}
+	}
+
+	return false
 }
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -41,6 +41,7 @@ type peer struct {
 	MyASN         uint32         `yaml:"my-asn"`
 	ASN           uint32         `yaml:"peer-asn"`
 	Addr          string         `yaml:"peer-address"`
+	SrcAddr       string         `yaml:"source-address"`
 	Port          uint16         `yaml:"peer-port"`
 	HoldTime      string         `yaml:"hold-time"`
 	RouterID      string         `yaml:"router-id"`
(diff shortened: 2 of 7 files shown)


internal/bgp/bgp.go
```diff
@@ -525,8 +527,21 @@ type tcpmd5sig struct {
...
+	// If srcAddr exists on any of the local network interfaces, use it as the
+	// source address of the TCP socket. Otherwise, use the IPv6 unspecified
+	// address ("::") to let the kernel figure out the source address.
+	// NOTE: On Linux, "::" also includes "0.0.0.0" (all IPv4 addresses).
+	//
+	// When the source-address configuration option isn't set by the user,
+	// srcAddr becomes "<nil>", which makes localAddressExists() return false.
+	a := "[::]"
+	ifs, err := net.Interfaces()
+	if err == nil && localAddressExists(ifs, net.ParseIP(srcAddr)) {
```
rata, line 540: I'm not sure if the user sets this and the address doesn't exist, if we should fail instead. If we don't do that, then something that can easily happen is:
* I configure a src address A for a peer that is used only on one node
* Over time that peer applies to several nodes but all continues to work (src address A is ignored in all nodes, as not present)
* One of the nodes receives that IP address but doesn't connect to the peer (routes, whatever)
* By doing a very simple change (adding an ip to a node), you can break metallb on that node because of some legacy config that was left and metallb ignores without a warning nor anything.

Specifying a srcAddr is quite a clear indication to use it, I'd expect to use it or fail with an error, not to silently drop it.

Is there some reason to just not fail if the specified srcAddress can't be used? What do you think about raising an error instead?

internal/bgp/bgp.go
```diff
@@ -525,8 +527,21 @@ type tcpmd5sig struct {
 // proper TCP MD5 options when the password is not empty. Works by manupulating
 // the low level FD's, skipping the net.Conn API as it has not hooks to set
 // the neccessary sockopts for TCP MD5.
-func dialMD5(ctx context.Context, addr, password string) (net.Conn, error) {
-	laddr, err := net.ResolveTCPAddr("tcp", "[::]:0")
+func dialMD5(ctx context.Context, addr string, srcAddr string, password string) (net.Conn, error) {
```
rata, line 530: nit: two consecutive params of the same type (addr and srcAddr of type string) can be written to only write the type (string) once. Can you change it?

And we should probably ask a linter to check these things in the future.

internal/bgp/bgp.go
```diff
@@ -671,3 +686,28 @@ func buildTCPMD5Sig(addr net.IP, key string) tcpmd5sig {
...
 }
+
+// localAddressExists returns true if the address addr exists on any of the
+// network interfaces in the ifs slice. The function doesn't return errors and
+// instead returns false if an error has occurred while looking for the
+// address. This function is safe to call if addr is nil or malformed.
+func localAddressExists(ifs []net.Interface, addr net.IP) bool {
+	for _, i := range ifs {
+		addresses, err := i.Addrs()
+		if err != nil {
```
rata, line 697: Just curious: when can this happen? If the interface doesn't have any address, does this return an error?

I feel that can be a frequent case (like interfaces in a bond, or some unused vti, etc.). But that is guessing when that fails, maybe it doesn't fail in that case?

If it does, I think it will be fine to change the `return false` to just `continue`. In that way, we ignore that interface and if none matches we return false at the end of the function anyways.

internal/bgp/bgp.go
```diff
@@ -671,3 +686,28 @@ func buildTCPMD5Sig(addr net.IP, key string) tcpmd5sig {
...
+			return false
+		}
+
+		for _, a := range addresses {
+			ip, ok := a.(*net.IPNet)
+			if !ok {
+				return false
```
rata, line 704: a similar thing here: while this is more weird and probably never happens, it feels safer to just ignore ifaces with invalid IPs, and continue to look at the rest, instead of stop looking completely and return false.

I mean, there is no reason to not look at other iterfaces and if one of the others is well formed and all, use it, right?

rata (changes requested):
@johananl Thanks! I had a cursory look, overall seems fine. I left some concerns I had with edge cases

internal/bgp/bgp.go
johananl: >Over time that peer applies to several nodes

We clearly say in the docs that source IP should only be used with per-node peers, i.e. peers which apply to one node only.

>Is there some reason to just not fail if the specified srcAddress can't be used? What do you think about raising an error instead?

We could do that. I've opted for a less "intrusive" behavior because I assumed we may still be able to establish a BGP session using the kernel-selected source IP.

internal/bgp/bgp.go
johananl: Fixed.

Happy to add a linter for this. I couldn't find a linter which checks for this case though. Do you know one @rata?

internal/bgp/bgp.go
johananl: `Addrs()` returns an error so we want to do the error checking here. I'm not sure exactly when an error is returned.

Thanks, I actually changed `return false` to `continue` and apparently undid that change :shrug: Fixing.

johananl: Same goes for the 2nd `return false` - if one address is broken, maybe the next one is OK.

internal/bgp/bgp.go
flokli: I think there might even be valid use cases for using the same source ip address from multiple hosts - think about link-local ipv6 addresses.

johananl: > I think there might even be valid use cases for using the same source ip address from multiple hosts - think about link-local ipv6 addresses.

I agree, that's why I wrote "in most cases" [here](https://github.com/metallb/metallb/pull/902/files#diff-b3daf7076dbbe1b7fe39b03fe884a6a6fceef9859d489531069948319a6b38aaR266).

johananl: I've changed the behavior so that we log an error if the user provides a malformed IP or an address which doesn't exist locally.

internal/bgp/bgp.go
```diff
@@ -525,8 +527,26 @@ type tcpmd5sig struct {
...
+	// address ("::") to let the kernel figure out the source address.
+	// NOTE: On Linux, "::" also includes "0.0.0.0" (all IPv4 addresses).
+	a := "[::]"
+	if srcAddr != nil {
+		ifs, err := net.Interfaces()
+		if err != nil {
+			return nil, fmt.Errorf("Querying local interfaces: %w", err)
+		}
+
+		if !localAddressExists(ifs, srcAddr) {
```
rata, line 542: I don't know, but does this check pass with link-local addresses? I guess so, but wanted to double check just in case :)

internal/bgp/bgp.go
```diff
@@ -671,3 +691,28 @@ func buildTCPMD5Sig(addr net.IP, key string) tcpmd5sig {
 
 	return t
 }
+
+// localAddressExists returns true if the address addr exists on any of the
+// network interfaces in the ifs slice. The function doesn't return errors and
+// instead returns false if an error has occurred while looking for the
```
rata, line 697: This isn't true anymore, right? If an error happens, we ignore it and continue with the rest.

Maybe it is best to remove this sentence entirely? Since the return value is bool (not error) and is quite of a short function to read if needed

rata (approved):
LGTM, thanks! I've left two small comments (one is just a question), I don't think I need to review again even if you do one small change I asked in the comment (remove one sentence from docs).

I like the change to net.IP for the srcAddr param :)

internal/bgp/bgp.go
johananl: Ha. Indeed we had a bug with IPv6 addresses. Fixed now.

rata (approved):
LGTM, thanks!

flokli:
Can you make a release containing this?

PR #903, opened 2021-07-02
chart: fix validation of .Values.imagePullSecrets

### LINKED ISSUE #897: specifying imagePullSecrets in Helm Chart values does not work
I am specifying the imagePullSecrets as follows in my values-file:

imagePullSecrets:
  - name: test

and what I get is this error:

metallb:
- imagePullSecrets.0: Invalid type. Expected: string, given: object

I have tried many other ways too like using quotes, indendation etc but none of them work out. Can anyone also reproduce this issue?

 Here is helm version:
version.BuildInfo{Version:"v3.5.4", GitCommit:"1b5edb69df3d3a08df77c9902dc17af864ff05d1", GitTreeState:"clean", GoVersion:"go1.15.11"}

gclawes:
imagePullSecrets is an array of objects with a single property "name"

Fixed #897

### DIFF
--- a/charts/metallb/values.schema.json
+++ b/charts/metallb/values.schema.json
@@ -127,7 +127,14 @@
       "description": "Secrets used for pulling images",
       "type": "array",
       "items": {
-        "type": "string"
+        "type": "object",
+        "properties": {
+          "name": {
+            "type": "string"
+          }
+        },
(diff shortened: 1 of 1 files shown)


russellb:
also backported to v0.10

PR #904, opened 2021-07-04
adding option to set namespace of the prometheus instance

jr0dd:
This is a simple to change to just give an option to set the namespace where the rules and monitors will be installed


russellb:
@gclawes PTAL, thanks!

PR #911, opened 2021-07-12
fix: Specify type for fenced code blocks

### LINKED ISSUE #857: Add more contrast to code blocks with shell commands
Take this url for instance: https://metallb.universe.tf/installation/

image

It is just too hard to read that text. Could you please add some more contrast?

mikemckiernan:
Fixed #857

* Use `bash` instead of `shell`.
* Use `ini` on the Romana page. It's fishy, but the
  appearance is OK.
* Add `yaml` to a few and swap a few `yaml` to `bash`.

### DIFF
--- a/website/content/community/_index.md
+++ b/website/content/community/_index.md
@@ -146,9 +146,11 @@ are needed. Speakers need those and `-node-name`.
 
 For example:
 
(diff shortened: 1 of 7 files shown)


russellb (approved):
lgtm, thanks!

PR #913, opened 2021-07-13
doc: mostly load-balancer to load balancer

### LINKED ISSUE #912: doc: Revise use of "load balancer" and "load-balancer"
Documentation lingo:
* Use "load balancer" as a noun--as in "...you request a load balancer,...".
* Use "load-balancer" as an adjective--as in, "...load-balancer service...".

Those two catch my eye when they aren't applied quite correctly.  I plan to apply that change throughout.

I'd like to reserve the right to expand the scope on lingo--and maybe style--as time allows.  I'm time-boxing this to a few hours.

I can't imagine any resistance, but I'm opening this issue to feel that out before I open a PR.  Please let me know if there's a reason *not* to make the change.

mikemckiernan:
Fixes #912.

### DIFF
--- a/website/content/_index.md
+++ b/website/content/_index.md
@@ -16,23 +16,23 @@ explains what that implies.
 
 ## Why?
 
-Kubernetes does not offer an implementation of network load-balancers
+Kubernetes does not offer an implementation of network load balancers
 ([Services of type LoadBalancer](https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/))
-for bare metal clusters. The implementations of Network LB that
+for bare-metal clusters. The implementations of network load balancers that
 Kubernetes does ship with are all glue code that calls out to various
 IaaS platforms (GCP, AWS, Azure...). If you're not running on a
 supported IaaS platform (GCP, AWS, Azure...), LoadBalancers will
 remain in the "pending" state indefinitely when created.
 
-Bare metal cluster operators are left with two lesser tools to bring
+Bare-metal cluster operators are left with two lesser tools to bring
 user traffic into their clusters, "NodePort" and "externalIPs"
 services. Both of these options have significant downsides for
-production use, which makes bare metal clusters second class citizens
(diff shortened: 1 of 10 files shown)


website/content/_index.md
```diff
@@ -16,23 +16,23 @@ explains what that implies.
 
 ## Why?
 
-Kubernetes does not offer an implementation of network load-balancers
+Kubernetes does not offer an implementation of network load balancers
 ([Services of type LoadBalancer](https://kubernetes.io/docs/tasks/access-application-cluster/create-external-load-balancer/))
-for bare metal clusters. The implementations of Network LB that
+for bare-metal clusters. The implementations of network load balancer that
```
russellb, line 21: I think this should be plural?

```suggestion
for bare-metal clusters. The implementations of network load balancers that
```

website/content/configuration/_index.md
```diff
@@ -4,19 +4,19 @@ weight: 4
 ---
 
 MetalLB remains idle until configured. This is accomplished by
-creating and deploying a configmap into the same namespace
+creating and deploying a config map into the same namespace
```
russellb, line 7: This is referencing a type of Kubernetes resource.  Maybe it should be `ConfigMap` ?
```suggestion
creating and deploying a `ConfigMap` into the same namespace
```

website/content/_index.md
mikemckiernan: That's def an oops.  Thank you for noticing and pointing it out.

website/content/configuration/_index.md
mikemckiernan: I favor lower case text without styling whenever possible.  Overwhelmingly, references to API objects like config maps, deployments, pods, and so on are more readable as plain text.  I acknowledge that this is often subjective.

russellb: Thanks. I was mostly interested in your opinion and happy to go with what you suggest.

PR #916, opened 2021-07-15
Enforce comment style

johananl:
This PR adds linter rules for enforcing comment style.


champtar:
LGTM
It's important to keep the feedback loop short for contributors so anything that can be enforced by CI is a good thing IMO

PR #918, opened 2021-07-16
Enforce commit message style

johananl:
This PR enforces some rules around commit message style. I find it useful because I find myself commenting on commit style repeatedly in code reviews and it's easier to let a machine enforce at least some rules.

I've opted for GitHub Actions because it was very tricky to get `commitlint` to work in CircleCI and in Actions it just works :tm: 

A sample result of the new linting job can be seen in the status of this PR.

### DIFF
--- a/.github/workflows/commitlint.yaml
+++ b/.github/workflows/commitlint.yaml
@@ -1,0 +1,11 @@
+name: Lint Commit Messages
+on: [pull_request]
+
+jobs:
+  commitlint:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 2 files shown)


russellb:
lgtm.  I'm happy using github actions. Now that our jobs have picked up a bit, the limits on concurrency in CircleCI are starting to be noticeable.  Github actions currently allows 20 concurrent jobs, so starting to move things over there would speed up our CI, I think.

champtar:
body-full-stop lint rule fails on signed-off-by lines :(

PR #919, opened 2021-07-16
Do not enforce that body ends with a period

champtar:
This prevent the use of Signed-off-by

fixes #918


russellb:
FYI @johananl 

bummer that it doesn't understand additional headers.  We don't require Signed-off-by, but also don't want to prevent it.  We also need to allow for Co-authored-by (and probably more ...).

I'll merge this for now and we can turn this or something similar back on if we make it work with headers.

johananl:
@russellb I couldn't find a way to make the footer coexist with the period enforcement.

PR #921, opened 2021-07-19
Rename commitlint config file

johananl:


johananl:
Merging without a review since this is a trivial change and the CI is currently broken.

PR #922, opened 2021-07-20
Shared ip

### LINKED ISSUE #558: Multiple services with same IP announced multiple times

kvaps:
This PR implements the changes purposed by @liuyuan10 in https://github.com/metallb/metallb/pull/562 and adapt them for latest version `v0.10.2`
We're using this patch in production for a few month, it works perfectly.

```
When IP is shared by two services, it may be annonced from different
nodes causing conflicting arp responses.

This commit fixes by using service IP in the hash instead of service
name so that service sharing the same IP will have the same master.

For traffic cluster services, all nodes should be usable instead of
those running pods, so that services sharing IPs have the same set of
usable nodes.
```
Fixes #558

docker images prepared:

```
ghcr.io/kvaps/metallb-controller:v0.10.2-fix558
ghcr.io/kvaps/metallb-speaker:v0.10.2-fix558
```

### DIFF
--- a/speaker/bgp_controller.go
+++ b/speaker/bgp_controller.go
@@ -137,7 +137,7 @@ func hasHealthyEndpoint(eps k8s.EpsOrSlices, filterNode func(*string) bool) bool
 	return false
 }
 
-func (c *bgpController) ShouldAnnounce(l log.Logger, name string, svc *v1.Service, eps k8s.EpsOrSlices) string {
+func (c *bgpController) ShouldAnnounce(l log.Logger, name string, svc *v1.Service, eps k8s.EpsOrSlices, _ net.IP) string {
 	// Should we advertise?
 	// Yes, if externalTrafficPolicy is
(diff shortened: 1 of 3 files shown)


champtar:
This introduce 2 unneeded behavior change for all users:
- the node responding to arp might change for everyone
- when using traffic policy cluster we don't prefer node with endpoints anymore

Ideally changes in this PR should only impact shared IP services

PR #925, opened 2021-07-25
Fix typo on website

bootjp:
Github is more accurately referred to as GitHub.

### DIFF
--- a/website/config.toml
+++ b/website/config.toml
@@ -16,7 +16,7 @@ home = ["HTML", "RSS", "JSON"]
 
 [[menu.shortcuts]]
 pre = "<hr>"
-name = "<i class='fa fa-github'></i> Github repo"
+name = "<i class='fa fa-github'></i> GitHub repo"
 url = "https://github.com/metallb/metallb"
 weight = 10
 
--- a/website/content/community/_index.md
+++ b/website/content/community/_index.md
@@ -10,7 +10,7 @@ We would love to hear from you! Here are some places you can find us.
 Our mailing list is
 [metallb-users@googlegroups.com](https://groups.google.com/forum/#!forum/metallb-users). It's
 for discussions around MetalLB usage, community support, and developer
-discussion (although for the latter we mostly use Github directly).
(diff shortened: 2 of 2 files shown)


website/themes/docdock/exampleSite/content/content-organisation/_index.md
```diff
@@ -46,7 +46,7 @@ in the page frontmatter, add a `head` param to insert any HTML code before the m
 example to display a "Hello"
 
 	+++
-	title = "Github repo"
```
johananl, line 49: This file and the rest of the files under this path belong to a sample website which comes with the Hugo template we're using for MetalLB. I wouldn't touch them because they come from [upstream](https://github.com/vjeantet/hugo-theme-docdock/blob/master/exampleSite/content/content-organisation/_index.md) and in any case aren't used.

johananl (changes requested):
Thanks! We should drop the changes to the sample website. LGTM otherwise.

website/themes/docdock/exampleSite/content/content-organisation/_index.md
bootjp: Thankyou for review, i fixed it.

bootjp:
`website/themes/docdock/exampleSite/config.toml` was also from sample, I dropped this change as well.

johananl:
Thanks @bootjp.
If you could fix the linter error and squash the last two commits I can merge this.

bootjp:
@johananl 
I squashed commit.
Is there anything else I need to do?

johananl:
> @johananl
> I squashed commit.
> Is there anything else I need to do?

>If you could fix the linter error

Yes, this ^ :slightly_smiling_face:

bootjp:
Sorry, I overlooked that.

I corrected it a little, but if it is not enough, I would appreciate it if you could correct it.

bootjp:
Thank you.

johananl:
> Thank you.

Thank you!

PR #927, opened 2021-07-30
Design proposal: Expand BGP with BFD capabilities

fedepaol:
Extending BGP with BFD will make the sessions more reliable and it will make the detection of a neighbour failure faster, allowing users to provide a better service. Here, we submit a design proposal to support that case.


fedepaol:
cc @markdgray @russellb

design/bgp-bfd.md
```diff
@@ -0,0 +1,91 @@
+# BFD enablement in BGP
+
+## Summary
+
+The purpose of this enhancement is to use enable [BFD](https://en.wikipedia.org/wiki/Bidirectional_Forwarding_Detection) to be used
+in conjunction with BGP.
```
markdgray, line 6: I think we need to specify this is only for the FRR implementation and not the native implementation.

markdgray: Also we will probably need to follow this: https://github.com/metallb/metallb/blob/main/design/0001-frr.md#development-phases

design/bgp-bfd.md
```diff
@@ -0,0 +1,91 @@
...
+
+## Motivation
+
+BFD provides a quicker path failure detection than BGP, and using them together will allow users to provide a better
+service.
+
+BFD for BGP is supported by FRR [out of the box](http://docs.frrouting.org/en/latest/bfd.html#bgp-bfd-configuration).
+
+
```
markdgray, line 15: Superfluous line

design/bgp-bfd.md
```diff
@@ -0,0 +1,91 @@
...
+### Goals
+
+* Provide a way to establish BFD on BGP session.
+* Provide a way to configure the parameters of the BFD session associated with a BGP session.
+* Provide BFD on BGP only on top of the FRR implementation.
```
markdgray, line 20: Its specified in non-goals so we can probably drop it from here

design/bgp-bfd.md
fedepaol: > I think we need to specify this is only for the FRR implementation and not the native implementation even though you mention it below.

Ok, I'll add it into the summary as well.

design/bgp-bfd.md
```diff
@@ -0,0 +1,91 @@
...
+BFD provides a quicker path failure detection than BGP, and using them together will allow users to provide a better
+service.
+
+BFD for BGP is supported by FRR [out of the box](http://docs.frrouting.org/en/latest/bfd.html#bgp-bfd-configuration).
+
+
+### Goals
+
+* Provide a way to establish BFD on BGP session.
+* Provide a way to configure the parameters of the BFD session associated with a BGP session.
```
markdgray, line 19: I think there are a lot of tunables for this? Do you have any suggestion what would be the minimum we would need to expose? I think MetalLB has a reputation as simple to configure and "just works out of the box" so we should really try to minimize this as much as possible?

design/bgp-bfd.md
```diff
@@ -0,0 +1,91 @@
...
+When setting a BGP peer, an optional bfd-profile property will enable BFD:
+
+```yaml
+    peers:
+    - peer-address: 10.0.0.1
+      peer-asn: 64501
+      my-asn: 64500
+      bfd-profile: bfdprofile1
+```
+
```
markdgray, line 69: Would we need to communicate anything back to the cluster administrator. For example in the operator CRD status?

design/bgp-bfd.md
markdgray: Should this design include references to the operator and any work that needs to be done there?

fedepaol: > I think there are a lot of tunables for this? Do you have any suggestion what would be the minimum we would need to expose? I think MetalLB has a reputation as simple to configure and "just works out of the box" so we should really try to minimize this as much as possible?

It is in the design section below. There are 3 properties that we can set for each session. I guess we can define defaults or leverage and declare the frr defaults if we don't specify them.

fedepaol: > Should this design include references to the operator and any work that needs to be done there?

I will add a section, makes total sense

design/bgp-bfd.md
fedepaol: This is a good point, and it really depends on where the CRD is implemented. It would be cool to be able to report the status of the session (and this is probably something we can get also for a BGP session?), but if the CRD is implemented in the operator we can't have the operator retrieve that information out of the speakers / frr container.

design/bgp-bfd.md
fedepaol: > Also we will probably need to follow this: https://github.com/metallb/metallb/blob/main/design/0001-frr.md#development-phases

This feels way more straightforward. What we can do is declare the feature beta somehow, but apart from that the feature will get in all in once I think

design/bgp-bfd.md
```diff
@@ -0,0 +1,114 @@
...
+## Summary
+
+The purpose of this enhancement is to use enable [BFD](https://en.wikipedia.org/wiki/Bidirectional_Forwarding_Detection) to be used in conjunction with BGP.
+
+BFD will be available only in the upcoming [FRR based implementation](0001-frr.md).
+
+## Motivation
+
+BFD provides a quicker path failure detection than BGP, and using them together will allow users to provide a better
```
rata, line 11: Not really a review comment, but just a FYI in case you weren't aware, as it is not very documented IIRC. We have a know to configure the hold-time for BGP peers. The min is 3 seconds (IIRC that was imposed by BGP), and setting that to 3s really improves the failure detection.

You can see it here: https://github.com/metallb/metallb/blob/2671c3f22de729e5faad7b35732bba1314932d0d/internal/config/config.go#L46

Far from the ms failure detection we can get with BFD IIUC, though. But maybe the hold-time can help you to aliviate some issues meanwhile :)

design/bgp-bfd.md
```diff
@@ -0,0 +1,114 @@
...
+* Provide a way to establish BFD on BGP session.
+* Provide a way to configure the parameters of the BFD session associated with a BGP session.
+
+### Non-Goals
+
+* Implement BFD on the legacy BGP implementation.
+
+## Proposal
+
+### User Stories
```
rata, line 27: Maybe we want to expose some other prometheus metrics so users can monitor this aspect too?

I imagine a story saying something like: As a cluster administrator, I want to be able to read BFD status so I can create alerts using the BFD input too ?

rata (commented):
@fedepaol Sorry for the late review. It mostly LGTM, just left some questions to clarify some things

design/bgp-bfd.md
markdgray: Yeah this is a general problem that we will need to address.

design/bgp-bfd.md
fedepaol: @rata done

design/bgp-bfd.md
msherif1234: BGP peer can only select one BFD profile correct i.e 1:1 relationship ? if that is the case why don't we embed the BFD config to BGP-PEER object instead of creating new CR , we had similar to debate with @russellb about bgp-community and the decision was to avoid CR overhead , so just curious why this is any different ?

msherif1234: BFD profile can fall in the category where it needs to be defined only in one peer like other global configuration routerId, myASN, port ......etc

fedepaol: That is less likely to happen (at least compared to port, routerid, etc which are supposed to be unique).
So it's likely to be the same for multiple sessions. On top of that, that's the way FRR (which has more miles than us) handles that.
The only difference is that they leverage a default profile too iirc, but if you want to change the parameters you need to rely on a profile. Drifting from this will cause us a bit of a headache I am afraid.

design/bgp-bfd.md
```diff
@@ -0,0 +1,157 @@
...
+When declaring a BFD peer, all it takes to enable BFD is to add a `bfd` property on the neighbour:
+
+```yaml
+    neighbor <A.B.C.D|X:X::X:X|WORD> bfd profile BFDPROF
+```
+
+The proposal here is to define a bfd profile section in the config structure that looks like:
+
+```yaml
+    bfd-profiles:
```
cgoncalves, line 52: FRR has a few more profile configuration options available (e.g. `minimum-ttl`, `echo-mode`, `echo receive-interval`, `echo transmit-interval`). Should those also be supported in the BFD config structure for MetalLB?

The `minimum-ttl` may be potentially required for multi-hop sessions (e.g. BGP peering with DC gateway or border leaf, rather than with ToR switch) because the default value in FRR only allows one hop between the system and its peers).

design/bgp-bfd.md
```diff
@@ -0,0 +1,157 @@
+# BFD enablement in BGP
+
+## Summary
+
+The purpose of this enhancement is to use enable [BFD](https://en.wikipedia.org/wiki/Bidirectional_Forwarding_Detection) to be used in conjunction with BGP.
+
+BFD will be available only in the upcoming [FRR based implementation](0001-frr.md).
```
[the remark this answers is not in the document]
fedepaol: Not sure I understand the comment. Here we are saying BFD would be enabled in the frr based implementation only and not in the "legacy", in-tree one.

design/bgp-bfd.md
fedepaol: Uhm, I referred to the doc [here](https://docs.frrouting.org/en/latest/bfd.html#configuration)

```
 ! configure a peer with every option possible
 peer 192.168.0.4
  label peer-label
  detect-multiplier 50
  receive-interval 60000
  transmit-interval 3000
  shutdown
``` 

I guess it's yet another case where we should not trust the comments :-) 
Will extend the proposal with all the documented options, thanks!

fedepaol: Done

design/bgp-bfd.md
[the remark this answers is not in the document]
russellb: I guess @cgoncalves means that it could also be supported by some 3rd party speaker implementation?  That's true, but I don't think we need to make mention of it.  While using a 3rd party speaker with the metallb controller is possible, it's not explicitly supported.

design/bgp-bfd.md
russellb: A sentence could be added here to say:

`Prior to adding BFD support, you can configure the hold-time parameter to speed up failure detection, but the minimum or fastest value is 3 seconds.`

design/bgp-bfd.md
```diff
@@ -0,0 +1,164 @@
...
+service.
+
+BFD for BGP is supported by FRR [out of the box](http://docs.frrouting.org/en/latest/bfd.html#bgp-bfd-configuration).
+
+### Goals
+
+* Provide a way to establish BFD on BGP session.
+* Provide a way to configure the parameters of the BFD session associated with a BGP session.
+
+### Non-Goals
```
russellb, line 21: How about also:

`* Configure BFD independent of a BGP session`

I don't think we care to support configuring BFD idependent of using BGP, right?

design/bgp-bfd.md
```diff
@@ -0,0 +1,164 @@
...
+
+#### Story 1
+
+As a cluster administrator, I want to declare a BGP session to be backed up by BFD.
+
+#### Story 2
+
+As a cluster administrator, I want to be able to set the BFD parameters related to a BGP session.
+
+## Design Details
```
russellb, line 37: Can you add a comment somewhere about how MetalLB will behave if this gets configured with the legacy BGP stack in use?

design/bgp-bfd.md
russellb: I'm OK with keeping this one separate.  The BGP communities config was a bit of a different case since that was such a trivial piece of config info (simple aliases).  The BFD config is a bit more involved and the common case is probably just a single profile.

design/bgp-bfd.md
```diff
@@ -0,0 +1,164 @@
...
+    peers:
+    - peer-address: 10.0.0.1
+      peer-asn: 64501
+      my-asn: 64500
+      bfd-profile: bfdprofile1
+```
+
+#### Impacts on the operator
+
+Keeping the operator as the guinea pig for the CRD implementation, a new BFDProfile CRD will be introduced, with the form of:
```
russellb, line 78: You don't actually have to do this, but as a suggestion for the future, I'd mark names of CRDs as literals like this:

```suggestion
Keeping the operator as the guinea pig for the CRD implementation, a new `BFDProfile` CRD will be introduced, with the form of:
```

design/bgp-bfd.md
```diff
@@ -0,0 +1,164 @@
...
+**e2e tests**: E2E tests will be expanded to cover bfd, using external container(s)
+running FRR. The test will need to cover the cases where a node is dropped, verifying
+that the broken route is detected by BFD.
+**unit tests**: Unit tests will be added to any additional code in the MetalLB
+repository.
+
+## Drawbacks
+
+* This design would couple MetalLB with the FRR community, making MetalLB somewhat
+  dependent on the FRR community for bug fixes and further feature improvements.
```
russellb, line 150: I don't really see this as a drawback of this proposal specifically.  It's a drawback of the FRR proposal, and this just builds on top of it.

fedepaol: Lol I think this is a leftover from the other proposal I started from

design/bgp-bfd.md
```diff
@@ -0,0 +1,164 @@
...
+  namespace: metallb-system
+spec:
+  receive-interval: 150
+  transmit-interval: 150
+  detect-multiplier: 10
+```
+
+Similarly, the BGPPeer CRD that is getting introduced will be configured with a new optional `bfdProfile` field.
+
+Ideally, if/when the CRDs are moved back to MetalLB, it will be possible to enrich the BGPPeer CRD status with information containing the status of the BFD session.
```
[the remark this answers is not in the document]
fedepaol: What I meant is. ff the CRDs are handled by metallb itself, it will be possible to enrich the status with informations that metallb is aware of.
It is possible to publish that information (such as session state) as metrics, but it would be nice to do something like
`kubectl get peer foo` and see the status of the sessions at a glance. This is possible only if the CRDs are handled by metallb itself.

design/bgp-bfd.md
fedepaol: Will do

design/bgp-bfd.md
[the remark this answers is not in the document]
cgoncalves: Right. And a more general comment is we should strive to create implementation-agnostic APIs. I briefly checked some 3rd party SDN solutions with BFD support and the herein proposed API seems to accommodate them as well.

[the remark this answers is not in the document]
russellb: That's great.  A comment like "the proposal here has been spot checked against the configuration of other BFD implementations (including X Y Z), and this design seems to fit well with them, as well."

design/bgp-bfd.md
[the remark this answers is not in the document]
russellb: I mean the operator could consume the metrics and then reflect status via the CRD ... obviously not ideal, but doesn't seem to be completely blocked on living it metallb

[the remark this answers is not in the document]
russellb: lol, yes, you're right.  I wasn't arguing I had a good solution, I was just splitting hairs that it was technically possible!  Anyway, I'll resolve this thread.  :)

design/bgp-bfd.md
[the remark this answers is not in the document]
fedepaol: I am not totally comfortable with that, so I'd skip that part.

fedepaol:
@russellb @rata I _think_ I addressed all the comments here.

cgoncalves:
LGTM

PR #928, opened 2021-07-30
FRR: Reorganise directory structure

markdgray:
As per https://github.com/metallb/metallb/pull/832, this PR begins integration of FRR as an alternative BGP implementation in MetalLB. 

This PR does some initial work to reorganise the directory structure and add a flag to enable FRR.


speaker/main.go
```diff
@@ -72,6 +72,7 @@ func main() {
 		myNode     = flag.String("node-name", os.Getenv("METALLB_NODE_NAME"), "name of this Kubernetes node (spec.nodeName)")
 		port       = flag.Int("port", 7472, "HTTP listening port")
 		logLevel   = flag.String("log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", strings.Join(logging.Levels, ", ")))
+		bgpType    = flag.String("bgp-implementation-type", "native", "Specifies the BGP implementation type.")
```
russellb, line 75: Can we defer this to be part of a patch series that makes the new option do something?

russellb (changes requested):
The code shuffle seems harmless enough, but I'd prefer to defer the new option to a PR with a patch series to make the option do something.

speaker/main.go
markdgray: Sure. I was looking for early feedback on this but what I can do is split this PR in two and make the second one (i.e. the configuration option) a draft. I'll do this later in the week.

russellb: From our dev community call yesterday, i think we decided to change this to an env var instead of a config option, at least while it's still in development.  Then I'm fine putting it in early since it will be more hidden while it's still WIP.

markdgray: Ok, I will still split the PR so I will remove this commit.

internal/bgp/native/backoff.go
```diff
@@ -1,4 +1,4 @@
-package bgp
+package native
```
fedepaol, line 1: This will be used only as the exposed `session` interface from the controller, however I'd pick a more descriptive package name.
From the calling point of view it's gonna be `native.New(..)`, native.Session, etc which (I think) it's less descriptive than `nativebgp.New()` or something like that.

fedepaol:
One comment, looks good overall!

internal/bgp/native/backoff.go
champtar: I would go with bgpnative and bgpfrr

markdgray:
I have taken on your comments. However, I will close this PR and generate a new one with additional functionality.

PR #931, opened 2021-08-19
Create and delete metallb configurations when running e2etests

sabinaaledort:
1. Adding before suite to configure MetalLB for testing by applying
   the MetalLB ConfigMap including both bgp and layer2 protocols.

2. One part of the configuration is runnig a BGP router in a container
   for all of the speakers to peer with.

3. Another part of the configuration is getting a range of 10 unused ip
   in the kind network subnet to create the layer2 address-pool.

4. Adding after suite to delete the MetalLB ConfigMap and kill the BGP
   router container.

Wasn't tested yet with dev-scpripts or OCP, might need some adjustments.


russellb:
I'm not sure about running both `layer2` and `bgp` in the same configuration.  Once FRR is integrated, I expect we'll want to test `layer2` vs `bgp` using two different types of speaker deployments (with and without the FRR sidecar).

This also seems to introduce some duplication between the `e2etest` code and the `tasks.py` code for configuring MetalLB and FRR.  If there's good reason to be able to drive that from the test code, maybe it could be refactored so the dev-env setup could run just the configuration code?

sabinaaledort:
> I'm not sure about running both `layer2` and `bgp` in the same configuration. Once FRR is integrated, I expect we'll want to test `layer2` vs `bgp` using two different types of speaker deployments (with and without the FRR sidecar).
> 
> This also seems to introduce some duplication between the `e2etest` code and the `tasks.py` code for configuring MetalLB and FRR. If there's good reason to be able to drive that from the test code, maybe it could be refactored so the dev-env setup could run just the configuration code?

i will remove the MetalLB configuration from `tasks.py`

markdgray:
> I'm not sure about running both `layer2` and `bgp` in the same configuration. Once FRR is integrated, I expect we'll want to test `layer2` vs `bgp` using two different types of speaker deployments (with and without the FRR sidecar).
> 
> This also seems to introduce some duplication between the `e2etest` code and the `tasks.py` code for configuring MetalLB and FRR. If there's good reason to be able to drive that from the test code, maybe it could be refactored so the dev-env setup could run just the configuration code?

I think the `tasks.py` code is not very flexible to test all the test scenarios that we will need to run. For example, what if we want to test something that requires a configuration change to the FRR test instance and also MetalLB, we will need to redeploy both in a custom way. I guess, one option would be that we could make dev-env more configurable. However, I can imagine the list of command line options would start to get quite long. However, this would have the added advantage of being able to easily manually deploy the same configuration as what is being run in the test.

In this model, we would then run the e2etest go code which would exec.Command("inv dev-env --some-parameter --another-parameter") and then wait until the cluster was deployed before running the test for that configuration.

Alternatively, we could allow the e2etest go code, to run some docker commands in the FRR test instances in order to change the configuration. However, in this way, we will be unable to easily change the test topology (imagine a scenario in which we change the number of FRR test instances).

e2etest/config/frr/bgpd.conf
```diff
@@ -0,0 +1,43 @@
+! -*- bgp -*-
```
markdgray: There are a few typos in the commit message.

markdgray: Also, could you update it to include the reason why you need to do this as I think that is the most important thing to get across.

dev-env/README.md
```diff
@@ -10,19 +10,6 @@ of your git clone, which runs MetalLB in an unconfigured state:
 inv dev-env
 ```
```
markdgray, line 12: We will still need to be able to run the development environment in order to manually test and try things out.

e2etest/e2etest_suite_test.go
```diff
@@ -70,3 +83,218 @@ func TestE2E(t *testing.T) {
...
 }
+
+var _ = ginkgo.BeforeSuite(func() {
+	cs, err := framework.LoadClientset()
+	framework.ExpectNoError(err)
+	_, err = cs.CoreV1().ConfigMaps(testNameSpace).Get(context.TODO(), "config", metav1.GetOptions{})
+	if errors.IsNotFound(err) {
+		err = configureMetalLB(cs)
+		framework.ExpectNoError(err)
+	} else if err == nil {
```
markdgray, line 94: Maybe we should just fail if it has already been configured? That seems like an error condition.

e2etest/e2etest_suite_test.go
```diff
@@ -70,3 +83,218 @@ func TestE2E(t *testing.T) {
...
+	}
+})
+
+// Run a BGP router in a container for all of the speakers to peer with.
+func configureFRR(cs clientset.Interface) error {
+	// We need the ip of each node in the cluster to place them in the BGP router configuration file
+	// Each node will peer with this router.
+	nodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+
+	// TODO: The IP address handling will need updates to add support for IPv6.
```
markdgray, line 116: So this^ is required before merging?

e2etest/e2etest_suite_test.go
```diff
@@ -70,3 +83,218 @@ func TestE2E(t *testing.T) {
...
+	if err != nil {
+		return fmt.Errorf("Failed to create temp FRR config directory: %s", err)
+	}
+
+	err = CopyDir(consts.BGPTestFRRConfigDir, frrConfigTestDir)
+	if err != nil {
+		return fmt.Errorf("Failed to copy FRR config directory: %s", err)
+	}
+
+	bgpdConfigTestFile := fmt.Sprintf("%s/%s", frrConfigTestDir, consts.BGPDConfigFileName)
```
markdgray, line 141: Should this be created in `consts.BGPTestFRRConfigDir`?

e2etest/e2etest_suite_test.go
```diff
@@ -70,3 +83,218 @@ func TestE2E(t *testing.T) {
...
+	}
+
+	err = ioutil.WriteFile(bgpdConfigTestFile, bgpdConfig, 0666)
+	if err != nil {
+		return fmt.Errorf("Failed to update FRR bgpd config file: %s", err)
+	}
+
+	// Run the BGP router in a container.
+	volume := fmt.Sprintf("%s:/etc/frr", frrConfigTestDir)
+	out, err := runCommand(false, "docker", "run", "-d", "--privileged", "--network", "kind", "--rm", "--name", consts.FRRContainerName, "--volume", volume, "frrouting/frr:v7.5.1")
```
markdgray, line 158: We should add a const for the image name. e.g. frrouting/frr:v7.5.1. We will likely need to update this occasionally.

e2etest/e2etest_suite_test.go
```diff
@@ -70,3 +83,218 @@ func TestE2E(t *testing.T) {
...
+		}
+	}
+
+	return ipv4Range, ipv6Range, nil
+}
+
+// Configure MetalLB for testing.
+// Identify the unused network address range and use it in configmap.
+func configureMetalLB(cs clientset.Interface) error {
+	err := configureFRR(cs)
```
markdgray, line 227: Will we need to reconfigure FRR? If we do, we may want to move `configureFRR()` out of this function.

e2etest/e2etest_suite_test.go
```diff
@@ -70,3 +83,218 @@ func TestE2E(t *testing.T) {
...
+	if err != nil {
+		return fmt.Errorf("Failed to get layer2 address pool range: %s", err)
+	}
+
+	peerAddress, err := runCommand(false, "docker", "inspect", "-f", "{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}", consts.FRRContainerName)
+	if err != nil {
+		return fmt.Errorf("Failed to get FRR ip address: %s", err)
+	}
+
+	// Apply the MetalLB ConfigMap.
```
markdgray, line 243: It's probably not needed now, but we will need to configure MetalLB in different ways so I think the config map should be a parameter to this function.

e2etest/e2etest_suite_test.go
```diff
@@ -70,3 +83,218 @@ func TestE2E(t *testing.T) {
...
+		return fmt.Errorf("Failed to configure FRR: %s", err)
+	}
+
+	// TODO: should set IPv6 addresspool as well when adding support for IPv6.
+	ipv4Range, _, err := getL2AddressPoolRange()
+	if err != nil {
+		return fmt.Errorf("Failed to get layer2 address pool range: %s", err)
+	}
+
+	peerAddress, err := runCommand(false, "docker", "inspect", "-f", "{{range .NetworkSettings.Networks}}{{.IPAddress}}{{end}}", consts.FRRContainerName)
```
markdgray, line 238: Can you foresee a test case in which we have two different peers? i.e. two FRR Containers?

e2etest/e2etest_suite_test.go
markdgray: In general, I suspect that this function will need to be more configurable. It might be good to think about that now before we build on it.

e2etest/utils.go
```diff
@@ -76,3 +84,150 @@ func tweakRCPort() func(rc *v1.ReplicationController) {
...
+func sortIPs(ips []string) []string {
+	sort.Slice(ips, func(i, j int) bool {
+		return compareIPs(ips[i], ips[j]) < 0
+	})
+
+	return ips
+}
+
+// Copy all the files in src directory to the dest directory.
+func CopyDir(src string, dest string) error {
```
markdgray, line 191: Could we just do an os call to `cp` or is the a go library that already does this?

e2etest/utils.go
```diff
@@ -76,3 +84,150 @@ func tweakRCPort() func(rc *v1.ReplicationController) {
...
+	lastIP, err := incrementIPByValue(firstIP, network, rangeSize)
+	if err != nil {
+		return nil, fmt.Errorf("Failed to get ips range: %s", err)
+	}
+	ipsRange[1] = lastIP
+
+	return ipsRange, nil
+}
+
+// Compare two ip addresses.
```
markdgray, line 149: In what way, can you explain what comparison you are making?

e2etest/utils.go
```diff
@@ -76,3 +84,150 @@ func tweakRCPort() func(rc *v1.ReplicationController) {
 	}
 	return nil
 }
+
+// Increment ip and test against CIDR after to avoid overflows.
```
markdgray, line 88: I didn't review the code below in detail but it seems like there may be go library functions that could help here?

e2etest/e2etest_suite_test.go
sabinaaledort: the frrConfigTestDir is a temp dir getting a random name starting with the prefix "frr-conf":
`frrConfigTestDir, err = ioutil.TempDir(".", "frr-conf")`

e2etest/e2etest_suite_test.go
sabinaaledort: im not sure about IPv6 support status. this wasn't support in the dev-env configuration either.

e2etest/utils.go
sabinaaledort: i need to copy all the files from one directory to another so i use the read and write of "io/ioutil" go library. is there a better way?

e2etest/e2etest_suite_test.go
```diff
@@ -70,3 +83,218 @@ func TestE2E(t *testing.T) {
...
+			if err != nil {
+				return nil, nil, fmt.Errorf("Failed to get ipv6 range: %s", err)
+			}
+		}
+	}
+
+	return ipv4Range, ipv6Range, nil
+}
+
+// Configure MetalLB for testing.
```
fedepaol, line 224: This must be test-specific.
Different tests may need different configuration (and this is how we achieve coverage). The simplest example is having l2 tests / bgp tests, which may need different configurations, but also when we want to cover advanced configuraitons.

e2etest/utils.go
fedepaol: +1

fedepaol:
> I'm not sure about running both `layer2` and `bgp` in the same configuration. Once FRR is integrated, I expect we'll want to test `layer2` vs `bgp` using two different types of speaker deployments (with and without the FRR sidecar).

Agree on the configuration side. The configuration must be driven by the tests and (apart from willingly wanting to test the corner case of both l2 / bpg configured together) each test (or test class) must apply (and clean) its own configuration.

Re: the two types of deployments, I expect them to be driven from outside and be two different CI lanes, one with the current bgp implementation and the other with FRR. The test suite must be the same and not know about the implementation (unless we start introducing frr-only features)

e2etest/utils.go
```diff
@@ -1,11 +1,16 @@
 package e2e
 
 import (
+	"context"
 	"fmt"
 	"os/exec"
 
+	pkgerrors "github.com/pkg/errors"
```
fedepaol, line 8: Why you need to alias it? `errors` is generally intended to be this one, I'd alias anything else conflicts with this.

e2etest/utils.go
```diff
@@ -76,3 +136,31 @@ func tweakRCPort() func(rc *v1.ReplicationController) {
 	}
 	return nil
 }
+
+func updateConfigMapObjs(cs clientset.Interface, dataModifier func(m *configFile)) error {
```
fedepaol, line 140: I think the whole config parsing stuff deserves a separate package.
We should aim to keep the utils.go file slim / remove it, not adding new stuff to it.
https://dave.cheney.net/2019/01/08/avoid-package-names-like-base-util-or-common

e2etest/utils.go
```diff
@@ -76,3 +136,31 @@ func tweakRCPort() func(rc *v1.ReplicationController) {
...
+		cm.Data = make(map[string]string)
+	}
+
+	data := configFile{}
+	err = yaml.Unmarshal([]byte(cm.Data["config"]), &data)
+	framework.ExpectNoError(err)
+
+	dataModifier(&data)
+	resData, err := yaml.Marshal(data)
+	framework.ExpectNoError(err)
```
fedepaol, line 156: We are mixing framework.ExpectNoError with returning the error.
Just pick one.

e2etest/utils.go
```diff
@@ -76,3 +136,31 @@ func tweakRCPort() func(rc *v1.ReplicationController) {
...
 }
+
+func updateConfigMapObjs(cs clientset.Interface, dataModifier func(m *configFile)) error {
+	cm, err := cs.CoreV1().ConfigMaps(testNameSpace).Get(context.TODO(), "config", metav1.GetOptions{})
+	if err != nil {
+		return pkgerrors.Wrapf(err, "Failed to get MetalLB ConfigMap")
+	}
+
+	if len(cm.Data) == 0 {
+		cm.Data = make(map[string]string)
```
fedepaol, line 147: Do you need to do this? Here we are always overriding the whole config right?

fedepaol: Also, wouldn't everything be simpler if we always delete and create the configmap with the new config? Am I missing something?

e2etest/consts/const.go
```diff
@@ -0,0 +1,12 @@
+package consts
+
+const (
+	// BGPTestFRRConfigDir contains the name of the BGP configuration directory
+	BGPTestFRRConfigDir = "config/frr"
```
fedepaol, line 5: Please move the constants used in a single package to that package. We need a shared package only if multiple packages access the same constant.

e2etest/bgp_test.go
```diff
@@ -80,8 +131,117 @@ var _ = ginkgo.Describe("BGP", func() {
...
+
+		err = wgetRetry(useDocker, frrContainerTestName, address)
 		framework.ExpectNoError(err)
 	})
 
 })
+
+// Run a BGP router in a container.
+func startFRRContainer(containerName string, testDirName string) (string, error) {
+	srcFiles := fmt.Sprintf("%s/.", consts.BGPTestFRRConfigDir)
```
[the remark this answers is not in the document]
sabinaaledort: just tried that, apparently it removes dots. filepath.Join() is calling filepath.Clean() which eliminates each dot path name element (the current directory).

e2etest/bgp_test.go
```diff
@@ -21,38 +21,89 @@ package e2e
...
+		err := bgpCS.CoreV1().ConfigMaps(testNameSpace).Delete(context.TODO(), "config", metav1.DeleteOptions{})
+		framework.ExpectNoError(err)
+
 		if ginkgo.CurrentGinkgoTestDescription().Failed {
 			DescribeSvc(f.Namespace.Name)
 		}
 	})
 
 	ginkgo.It("should work for type=Loadbalancer", func() {
+		frrContainerTestName := "frr-test"
```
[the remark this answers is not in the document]
sabinaaledort: i think the name should be set in the test as we might have a test with 2 frr containers.

e2etest/bgp_test.go
```diff
@@ -80,8 +131,117 @@ var _ = ginkgo.Describe("BGP", func() {
...
+		return "", errors.Wrapf(err, "Failed to get FRR ip address")
+	}
+
+	containerIP = strings.Split(containerIP, "\n")[0]
+
+	return containerIP, nil
+}
+
+// Delete the BGP router configuration.
+func deleteFRRConfiguration(containerName string, testDirName string) error {
```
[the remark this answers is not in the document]
sabinaaledort: what config reloading mechanism do you mean?

tasks.py
```diff
@@ -385,6 +385,38 @@ def bgp_dev_env():
     # Apply the MetalLB ConfigMap
     run("kubectl apply -f %s/config.yaml" % dev_env_dir)
 
+def get_service_range(networkVersion=None):
```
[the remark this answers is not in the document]
sabinaaledort: do you mean use ipv4 and ipv6 instead of 4 and 6?

e2etest/bgp_test.go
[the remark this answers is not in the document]
fedepaol: If / when we have a test with two containers, we may have a set of tests depending on them.
Creating the external container (or containers) only once, will make the test execution quicker.

e2etest/config/frr/zebra.conf
```diff
@@ -0,0 +1,3 @@
+hostname Router
```
fedepaol, line 1: Do we need this at all?

e2etest/bgp_container.go
```diff
@@ -0,0 +1,189 @@
...
+	"strings"
+
+	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+const (
+	// BGPTestFRRConfigDir contains the name of the BGP configuration directory.
+	BGPTestFRRConfigDir = "config/frr"
```
fedepaol, line 36: If the consts are package local, no need to export them

e2etest/bgp_container.go
```diff
@@ -0,0 +1,189 @@
...
+		}
+	}
+
+	if len(nodesIPs) < 3 {
+		return "", fmt.Errorf("Expected 3 nodes, got %d", len(nodesIPs))
+	}
+
+	bgpPeersInfo := "router bgp 64512\n"
+	for _, ip := range nodesIPs {
+		bgpPeersInfo = fmt.Sprintf("%s neighbor %s remote-as 64512\n neighbor %s next-hop-self\n", bgpPeersInfo, ip, ip)
```
fedepaol, line 154: Is the `remote-as` equal the local router id?

e2etest/bgp_container.go
```diff
@@ -0,0 +1,189 @@
...
+	nodesIPs := make([]string, 0, len(nodes.Items))
+	for _, node := range nodes.Items {
+		for i := range node.Status.Addresses {
+			if node.Status.Addresses[i].Type == "InternalIP" {
+				nodesIPs = append(nodesIPs, node.Status.Addresses[i].Address)
+			}
+		}
+	}
+
+	if len(nodesIPs) < 3 {
```
fedepaol, line 148: I'd remove this constraint, at this point we don't really care about the number of nodes.

e2etest/bgp_container.go
```diff
@@ -0,0 +1,189 @@
...
+		return "", fmt.Errorf("Expected 3 nodes, got %d", len(nodesIPs))
+	}
+
+	bgpPeersInfo := "router bgp 64512\n"
```
fedepaol, line 152: How about using a template here too (with a local template, no need to have an external file)?

e2etest/bgp_test.go
```diff
@@ -44,12 +47,56 @@ var _ = ginkgo.Describe("BGP", func() {
 	})
 
 	ginkgo.AfterEach(func() {
+		err := cs.CoreV1().ConfigMaps(testNameSpace).Delete(context.TODO(), "config", metav1.DeleteOptions{})
+		framework.ExpectNoError(err)
+
+		Eventually(func() bool {
+			_, err := cs.CoreV1().ConfigMaps(testNameSpace).Get(context.TODO(), "config", metav1.GetOptions{})
+			return apierrr.IsNotFound(err)
+		}, 1*time.Minute, 5*time.Second).Should(BeTrue(), "Failed to delete MetalLB ConfigMap")
```
fedepaol, line 56: Let's hope it's quicker. Checking every minute will make this at least one minute slow.
Move to 1 sec

e2etest/bgp_container.go
```diff
@@ -0,0 +1,189 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+// https://github.com/kubernetes/kubernetes/blob/92aff21558831b829fbc8cbca3d52edc80c01aa3/test/e2e/network/loadbalancer.go#L878
+
+package e2e
```
fedepaol, line 19: I think this deserves to be in its own package.

I'd choose `frr` for the container operations, `frrconfig` or just `config` under frr for the config operations

e2etest/bgp_container.go
```diff
@@ -0,0 +1,189 @@
...
+
+import (
+	"context"
+	"fmt"
+	"html/template"
```
fedepaol, line 24: Use text/template

e2etest/bgp_container.go
```diff
@@ -0,0 +1,189 @@
...
+	if err != nil {
+		return errors.Wrapf(err, "Failed to change %s container volume permissions. %s", containerName, string(out))
+	}
+
+	return nil
+}
+
+// Delete the BGP router configuration.
+func stopFRRContainer(containerName string, testDirName string) error {
+	err := updateFRRContainerVolumePermissions(containerName)
```
fedepaol, line 111: Calling this right after the creation would make it possible to edit the file as part of the tests, right?

e2etest/bgp_container.go
```diff
@@ -0,0 +1,189 @@
...
+	BGPDConfigFileName = "bgpd.conf"
+	// FRRRoutingImage contains the name and tag of the FRR routing image.
+	FRRRoutingImage = "frrouting/frr:v7.5.1"
+	// FRRContainerMountDestination contains the mount destination path.
+	FRRContainerMountDestination = "/etc/frr"
+)
+
+// Run a BGP router in a container.
+// Returns the IPv4 and IPv6 addresses of the container.
+func startFRRContainer(cs clientset.Interface, containerName string, testDirName string, bgpConfig string) (string, string, error) {
```
fedepaol, line 47: I would let this just start the container with an empty config file, and implement another method `reloadFRRConfig` that basically leverages what @oribon used in https://github.com/metallb/metallb/pull/935 .
This will make handling external containers more flexible, and will also make it easier to use the same set of tests when we are running frr on localhost.

e2etest/bgp_test.go
sabinaaledort: it is checking every 5 seconds, for 1 minute

fedepaol: Ooops :-) Can we move it to 1 second?

e2etest/bgp_test.go
```diff
@@ -49,12 +54,71 @@ var _ = ginkgo.Describe("BGP", func() {
...
+				},
+			},
+		}
+		err = updateConfigMap(cs, configData)
+		framework.ExpectNoError(err)
+
+		bgpConfig, err := frrconfig.BGPPeersForAllNodes(cs)
+		framework.ExpectNoError(err)
+
+		err = frr.UpdateBGPConfigFile(frrContainer, frrTestConfigDir, bgpConfig)
```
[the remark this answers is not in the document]
sabinaaledort: i thought we might want to reload other files as well so i separated these two.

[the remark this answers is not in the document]
fedepaol: That won't prevent us to have the reload function called by the update one. I can't think about a case where we update the config but we don't want to reload it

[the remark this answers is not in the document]
fedepaol: so is there any reason why update can't also reload using the other function inside?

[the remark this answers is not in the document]
sabinaaledort: sounds good :+1:

fedepaol:
This will serve as a base for extending the coverage by changing configurations inside the tests.

LGTM
cc @russellb

PR #935, opened 2021-08-30
FRR: Add frr-reloader directory and basic manifests for MetalLB with FRR

oribon:
The frr-reloader will act as a proxy between the MetalLB speaker and the FRR sidecar container.
The speaker will render a valid FRR configuration file and place it as frr.conf in the volume it
shares with the reloader (/etc/frr_reloader). Then it would ask the reloader to load the given file
by sending it a SIGHUP. The speaker can know the reloader's PID by reading /etc/frr_reloader/reloader.pid.

### DIFF
--- a/frr-reloader/Dockerfile
+++ b/frr-reloader/Dockerfile
@@ -1,0 +1,6 @@
+FROM frrouting/frr:v7.5.1
+WORKDIR /
+RUN mkdir /var/log/frr
+COPY frr-reloader.sh .
+
+ENTRYPOINT ["/frr-reloader.sh"]
--- a/frr-reloader/frr-reloader.sh
+++ b/frr-reloader/frr-reloader.sh
@@ -1,0 +1,51 @@
+#!/bin/bash
+cleanup() {
+  echo "Caught an exit signal.."
+  rm -f "$PIDFILE"
+  rm -f "$LOCKFILE"
+  kill_sleep
+  exit
+}
+
+reload_frr() {
+  flock 200
+  echo "Caught SIGHUP and acquired lock! Reloading FRR.."
+  kill_sleep
+
+  echo "Checking the configuration file syntax"
+  if ! python3 /usr/lib/frr/frr-reload.py --test --stdout "$FILE_TO_RELOAD" ; then
+    echo "Syntax error spotted: aborting.."
+    return
+  fi
+
+  echo "Applying the configuration file"
+  if ! python3 /usr/lib/frr/frr-reload.py --reload --overwrite --stdout "$FILE_TO_RELOAD" ; then
+    echo "Failed to fully apply configuration file"
+    return
+  fi
+  
+  echo "FRR reloaded successfully!"
+} 200<"$LOCKFILE"
+
+kill_sleep() {
+  kill "$sleep_pid"
+}
+
+trap cleanup SIGTERM SIGQUIT SIGINT
+trap 'reload_frr &' HUP
+
+SHARED_VOLUME="${SHARED_VOLUME:-/etc/frr_reloader}"
+PIDFILE="$SHARED_VOLUME/reloader.pid"
+FILE_TO_RELOAD="$SHARED_VOLUME/frr.conf"
+LOCKFILE="$SHARED_VOLUME/lock"
+
+echo "PID is: $$, writing to $PIDFILE"
+printf "$$" > "$PIDFILE"
+touch "$LOCKFILE"
+
+while true
+do
+    sleep infinity &
+    sleep_pid=$!
+    wait $sleep_pid 2>/dev/null
+done
--- a/manifests/metallb-frr.yaml
+++ b/manifests/metallb-frr.yaml
@@ -1,0 +1,611 @@
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  labels:
+    app: metallb
+  name: controller
+  namespace: metallb-system
+spec:
+  allowPrivilegeEscalation: false
+  allowedCapabilities: []
+  allowedHostPaths: []
+  defaultAddCapabilities: []
+  defaultAllowPrivilegeEscalation: false
+  fsGroup:
+    ranges:
+    - max: 65535
+      min: 1
+    rule: MustRunAs
+  hostIPC: false
+  hostNetwork: false
+  hostPID: false
+  privileged: false
+  readOnlyRootFilesystem: true
+  requiredDropCapabilities:
+  - ALL
+  runAsUser:
+    ranges:
+    - max: 65535
+      min: 1
+    rule: MustRunAs
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    ranges:
+    - max: 65535
+      min: 1
+    rule: MustRunAs
+  volumes:
+  - configMap
+  - secret
+  - emptyDir
+---
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  labels:
+    app: metallb
+  name: speaker
+  namespace: metallb-system
+spec:
+  allowPrivilegeEscalation: false
+  allowedCapabilities:
+  - NET_RAW
+  allowedHostPaths: []
+  defaultAddCapabilities: []
+  defaultAllowPrivilegeEscalation: false
+  fsGroup:
+    rule: RunAsAny
+  hostIPC: false
+  hostNetwork: true
+  hostPID: false
+  hostPorts:
+  - max: 7472
+    min: 7472
+  - max: 7946
+    min: 7946
+  privileged: true
+  readOnlyRootFilesystem: true
+  requiredDropCapabilities:
+  - ALL
+  runAsUser:
+    rule: RunAsAny
+  seLinux:
+    rule: RunAsAny
+  supplementalGroups:
+    rule: RunAsAny
+  volumes:
+  - configMap
+  - secret
+  - emptyDir
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    app: metallb
+  name: controller
+  namespace: metallb-system
+---
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  labels:
+    app: metallb
+  name: speaker
+  namespace: metallb-system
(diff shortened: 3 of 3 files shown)


frr-reloader/Dockerfile
```diff
@@ -0,0 +1,6 @@
+FROM frrouting/frr:v7.5.1
+WORKDIR /
+RUN mkdir /var/log/frr
+COPY frr-reloader.sh .
+
+ENTRYPOINT ["/frr-reloader.sh"]
```
russellb, line 6: Could you make use of `frr-reload` from the frr container via a shared volume?  I'd rather not have to worry about keeping a custom image up to date with changes that happen upstream in frr.  It looks like the `frr-reload.py` script is only using Python standard libraries, so it doesn't seem too bad.

oribon: I believe it's possible, although I did choose that image because vtysh is included there. In a way sharing vtysh and the reload script through a volume would hurt the image's portability(?), since it'll rely on specific mounts and would be unusable on its own, but I'm fine with that if that's the direction you rather.

fedepaol: I believe what @russellb  is suggesting is to ship frr-reloader in the speaker image and use it directly as a daemon in the frr container, but it could be tricky because it can't be used as the entry point (since the other container must have the time to move the file to a shared container). I don't see a straightforward way to achieve that, if not by sharing a volume and do some polling to check if the speaker moved the file in the shared volume and then launch the process.
But maybe there's a more straightforward way to do that.

frr-reloader/frr-reloader.sh
```diff
@@ -0,0 +1,51 @@
...
+}
+
+trap cleanup SIGTERM SIGQUIT SIGINT
+trap 'reload_frr &' HUP
+
+SHARED_VOLUME="${SHARED_VOLUME:-/etc/frr_reloader}"
+PIDFILE="$SHARED_VOLUME/reloader.pid"
+FILE_TO_RELOAD="$SHARED_VOLUME/frr.conf"
+LOCKFILE="$SHARED_VOLUME/lock"
+
```
fedepaol, line 41: how about calling cleanup when we start? The trap should ensure that we never end up in a dirty state, but it would be safer

manifests/metallb-frr.yaml
```diff
@@ -0,0 +1,611 @@
+apiVersion: policy/v1beta1
```
fedepaol, line 1: Kustomize support is already in place (or at least, a kustomization.yaml file in the folder).
I wonder if we could have one default layer and one that overrides the speaker instead of two distinct manifests with duplicated resources.

frr-reloader/frr-reloader.sh
```diff
@@ -0,0 +1,51 @@
+#!/bin/bash
+cleanup() {
```
fedepaol, line 2: Neat! I am probably biased by my weak bash knowledge, any reason why we did not go with a go process?

frr-reloader/frr-reloader.sh
oribon: good idea, added clean_files func

manifests/metallb-frr.yaml
oribon: I think it's the right way to leverage Kustomize but would be more fitting in a later PR, since these manifests aren't actually usable now. Maybe once we actually get to a point where the FRR integration is functional and when we also change the Helm/dev-env manifests to include it. What do you think?

fedepaol: Sounds legit, but mention that on top of the manifest that this is temporary and we are gonna clean it later on before the FRR integration is finalized (or as part of that), maybe even as part of the file name so no one is tempted to use it.

frr-reloader/Dockerfile
fedepaol: Copying what we discussed offline: I'd keep a separate container in the pod for the loader, but based on the frr base image instead of the custom one.
In this way we don't mess up the entry point with multiple processes, and log wise the container logs only the reloader process, but we still don't have to maintain a separate image. WDTY @russellb ?

champtar: @fedepaol just using initContainers to move the script to a volume that you then mount on frr container would do the trick.

fedepaol: @champtar yes, moving the files is straightforward, but having two process in the same container is a bit less. I am a bit more inclined to have two contaiers using the same frr image, one for frr and the other with the reloader.

fedepaol: *straightforward after thinking about it I mean, I did not think about it in the first place :P

champtar: Having 2 containers using the same frr is cheap so go for it

oribon: done, removed the need for a custom image

manifests/metallb-frr.yaml
oribon: done

frr-reloader/Dockerfile
markdgray: Also to @russellb 's point (I think this is the point). Could we just share the FRR binaries, sockets, etc from the FRR container to the speaker container via a shared volume and then the speaker could call frr_reload.py directly?

fedepaol: > Also to @russellb 's point (I think this is the point). Could we just share the FRR binaries, sockets, etc from the FRR container to the speaker container via a shared volume and then the speaker could call frr_reload.py directly?

Doensn't it feel a bit too hacky? Having a proxy in the frr container (or in a copy of it) that performs the reload won't cause any dependency headaches, but we can't guarantee the opposite.

fedepaol: Yep, I am afraid that for this purpouse we need a better interface

msherif1234: the less coupling we have between the Metallb speaker and FRR is the better, also we are assuming single FRR instance/speaker is that long term accurate statement ? have we explore using something like ZMQ with configuration in memory and avoid using disk space ?

manifests/metallb-frr-WIP.yaml
```diff
@@ -0,0 +1,620 @@
...
+    # the normal daemon command is added to this at the end.
+  vtysh.conf: ""
+  zebra.conf: |+
+    ! -*- zebra -*-
+    !
+    ! zebra sample configuration file
+    !
+    hostname Router
+    password zebra
+    enable password zebra
```
cgoncalves, line 413: Is there a specific need/concern to secure vty with a password? If yes, it must not be hardcoded here.
The frr-reload.py won't be able to reset the vty password later on even if a new frr.conf templated out by the speaker sets one. See https://github.com/FRRouting/frr/blob/d2e41077549d65e08b14948597cacb3d34eeda52/tools/frr-reload.py#L1376-L1404).

manifests/metallb-frr-WIP.yaml
```diff
@@ -0,0 +1,620 @@
...
+              mountPath: /var/run/frr
+            - name: frr-conf
+              mountPath: /etc/frr
+            - name: reloader
+              mountPath: /etc/frr_reloader
+        - args:
+            - --port=7472
+            - --config=config
+            - --log-level=info
+          env:
```
cgoncalves, line 512: Need to set BGP type to frr, otherwise speaker will default to native.
```
- name: METALLB_BGP_TYPE
  value: frr
```

manifests/metallb-frr-WIP.yaml
```diff
@@ -0,0 +1,620 @@
...
+    #watchfrr_options=""
+
+    # for debugging purposes, you can specify a "wrap" command to start instead
+    # of starting the daemon directly, e.g. to use valgrind on ospfd:
+    #   ospfd_wrap="/usr/bin/valgrind"
+    # or you can use "all_wrap" for all daemons, e.g. to use perf record:
+    #   all_wrap="/usr/bin/perf record --call-graph -"
+    # the normal daemon command is added to this at the end.
+  vtysh.conf: ""
+  zebra.conf: |+
```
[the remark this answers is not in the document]
oribon: done, cherry-picked your commit

manifests/metallb-frr-WIP.yaml
oribon: will be as part of the actual integration PR.

fedepaol:
LGTM

cgoncalves:
LGTM

oribon:
@rata Thanks for the review! I'll try to address the batch updates issue in this comment.
Regarding the scenario here https://github.com/metallb/metallb/pull/935#issuecomment-942602695, did you remove the & from the trap when you ran it? Running it locally behaves as I expected showing 5 echo prints. If so, I'll try to explain why that happens from what I saw and read.

TLDR: & allows signals to trigger reload_frr immediately, flock keeps the order and creates a queue.

I think this part of your comment is true and a good starting point:
> The signals sent to a process are a set (a bitmask), there is no queue of signal, you just set a bit and that bit either set or not (technically RT signals are queued).

In other words (and assuming we are running without the &), the first signal triggers the trap (calling reload_frr to do stuff). Assuming the second signal arrives while the first reload_frr (trap) is still running and midst a blocking command the second trap is "queued". That's when that bit is set and further signals are ignored. That is the situation until in some point between the stuff that reload_frr runs it does not block (=the first trap is not blocking anymore) - and when that happens the second trap is triggered, both running reload_frr side-by-side and the bit is unset again (another signal at this point would be "queued" like before).

Going back to your scenario, the first signal triggered reload_frr, you probably saw "signal handler!" and the sleep started, sending multiple signals at this point only registers once (bit is set  and all the other are ignored), and when the first signal finished sleeping the second reload_frr started, that's when we see "signal handler!" again (from the second trap) and it sleeps again until both traps print "signal handler after sleep!" and return.

That's where the & in the trap comes in. In order to bypass that mechanism we need to "free up" (never set) that bit, by making the function that the trap calls to never block. If it never blocks, every signal that arrives immediately triggers the trap func (reload_frr & in our case) in the background and the bit never flips. This results in many trap calls running side-by-side.
That works but is a bit chaotic, and that's the reason for the flock - it keeps only one trap doing its thing while the other wait for their turn.

Issue #936, opened 2021-08-30
Config flag to disable usage of endpointslices


Christoph-Raab:
We're using metallb v0.10.2 in a 1.20.x cluster where endpointslices are technically available but currently disabled. 

To disable them in metallb as well we created a patch in our local metallb fork in form of a config flag in controller and speaker to make the usage configurable. 

With this issue I want to ask if that patch is of public interest and if we should contribute it to this repo. If so I will create a pull request.

champtar:
Can you detail why the auto detection is not working ? Having a tristate flag enable/auto/disable would be nice yes.

Christoph-Raab:
Sorry, maybe that was a bit unspecific. Endpointslices is disabled in kubeproxy via the feature gate. The endpointslices themself are not disabled and thus are created, but not used by kubeproxy and should not be used by metallb as well. 

The flag we created is a binary option where the usage of endpointslices is enabled by default and can be disabled. Our repo is private but I just forked this repo and applied the patch in my own github account, see this commit: https://github.com/Christoph-Raab/metallb/commit/396176c7da27a9f109531337d002da5de11a667e

champtar:
Ok so it's either auto or disable, default to auto, looks ok to me, I would say send the PR and we will see the maintainers opinion for the naming of the flag

rata:
PR was merged, closing this

PR #937, opened 2021-08-30
Add config flag to disable usage of endpointslices

Christoph-Raab:
As discussed in https://github.com/metallb/metallb/issues/936 there are certain scenarios where endpointslices are technical available but should not be used by metallb.

This PR adds a config flag to disable the usage of endpointslices completely. If the flag is not provided the behaviour is the same as before.

### DIFF
--- a/controller/main.go
+++ b/controller/main.go
@@ -122,13 +122,14 @@ func (c *controller) MarkSynced(l log.Logger) {
 
 func main() {
 	var (
-		port       = flag.Int("port", 7472, "HTTP listening port for Prometheus metrics")
-		config     = flag.String("config", "config", "Kubernetes ConfigMap containing MetalLB's configuration")
-		namespace  = flag.String("namespace", os.Getenv("METALLB_NAMESPACE"), "config / memberlist secret namespace")
-		kubeconfig = flag.String("kubeconfig", "", "absolute path to the kubeconfig file (only needed when running outside of k8s)")
-		mlSecret   = flag.String("ml-secret-name", os.Getenv("METALLB_ML_SECRET_NAME"), "name of the memberlist secret to create")
-		deployName = flag.String("deployment", os.Getenv("METALLB_DEPLOYMENT"), "name of the MetalLB controller Deployment")
-		logLevel   = flag.String("log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", strings.Join(logging.Levels, ", ")))
+		port            = flag.Int("port", 7472, "HTTP listening port for Prometheus metrics")
+		config          = flag.String("config", "config", "Kubernetes ConfigMap containing MetalLB's configuration")
+		namespace       = flag.String("namespace", os.Getenv("METALLB_NAMESPACE"), "config / memberlist secret namespace")
+		kubeconfig      = flag.String("kubeconfig", "", "absolute path to the kubeconfig file (only needed when running outside of k8s)")
+		mlSecret        = flag.String("ml-secret-name", os.Getenv("METALLB_ML_SECRET_NAME"), "name of the memberlist secret to create")
+		deployName      = flag.String("deployment", os.Getenv("METALLB_DEPLOYMENT"), "name of the MetalLB controller Deployment")
+		logLevel        = flag.String("log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", strings.Join(logging.Levels, ", ")))
+		disableEpSlices = flag.Bool("disable-epslices", false, "Disable the usage of EndpointSlices and default to Endpoints instead of relying on the autodiscovery mechanism")
 	)
 	flag.Parse()
 
@@ -154,12 +155,13 @@ func main() {
 	}
 
 	client, err := k8s.New(&k8s.Config{
-		ProcessName:   "metallb-controller",
-		ConfigMapName: *config,
-		ConfigMapNS:   *namespace,
-		MetricsPort:   *port,
-		Logger:        logger,
-		Kubeconfig:    *kubeconfig,
+		ProcessName:     "metallb-controller",
+		ConfigMapName:   *config,
+		ConfigMapNS:     *namespace,
+		MetricsPort:     *port,
+		Logger:          logger,
+		Kubeconfig:      *kubeconfig,
+		DisableEpSlices: *disableEpSlices,
 
 		ServiceChanged: c.SetBalancer,
 		ConfigChanged:  c.SetConfig,
--- a/internal/k8s/k8s.go
+++ b/internal/k8s/k8s.go
@@ -72,15 +72,16 @@ const (
 // Config specifies the configuration of the Kubernetes
 // client/watcher.
 type Config struct {
-	ProcessName   string
-	ConfigMapName string
-	ConfigMapNS   string
-	NodeName      string
-	MetricsHost   string
-	MetricsPort   int
-	ReadEndpoints bool
-	Logger        log.Logger
-	Kubeconfig    string
+	ProcessName     string
+	ConfigMapName   string
+	ConfigMapNS     string
+	NodeName        string
+	MetricsHost     string
+	MetricsPort     int
+	ReadEndpoints   bool
+	Logger          log.Logger
+	Kubeconfig      string
+	DisableEpSlices bool
 
 	ServiceChanged func(log.Logger, string, *v1.Service, EpsOrSlices) SyncState
 	ConfigChanged  func(log.Logger, *config.Config) SyncState
@@ -164,7 +165,8 @@ func New(cfg *Config) (*Client, error) {
 		c.syncFuncs = append(c.syncFuncs, c.svcInformer.HasSynced)
 
 		if cfg.ReadEndpoints {
-			if !UseEndpointSlices(c.client) {
+			// use DisableEpSlices to skip the autodiscovery mechanism. Useful if EndpointSlices are enabled in the cluster but disabled in kube-proxy
+			if cfg.DisableEpSlices || !UseEndpointSlices(c.client) {
 				epHandlers := cache.ResourceEventHandlerFuncs{
 					AddFunc: func(obj interface{}) {
 						key, err := cache.MetaNamespaceKeyFunc(obj)
--- a/speaker/main.go
+++ b/speaker/main.go
@@ -61,17 +61,18 @@ func main() {
 	prometheus.MustRegister(announcing)
 
 	var (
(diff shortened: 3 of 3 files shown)


russellb:
@fedepaol Can you take a look?

fedepaol:
> @fedepaol Can you take a look?

sure!

fedepaol:
@Christoph-Raab if I get this right, in the cluster you are mentioning the EndpointSliceProxying feature gate is disabled but EndpointSlice is enabled, so the CR is getting filled but it's not used.

Just to understand, any reason why you did not go with disabling EndpointSlice too instead of modifying metallb?

Christoph-Raab:
> Just to understand, any reason why you did not go with disabling EndpointSlice too instead of modifying metallb?

The main reason is that you can not disable the Endpointslices resource in the apiserver. So we would not only need to disable the endpointslices controller in the controller-manager, but also deny creating the resource and remove all existing ones. 

That's why we want a feature flag like core-dns has to disable the usage in the addon.

fedepaol:
> > if I get this right, in the cluster you are mentioning the EndpointSliceProxying feature gate is disabled but EndpointSlice is enabled, so the CR is getting filled but it's not used.
> 
> Exactly.
> 
> > Just to understand, any reason why you did not go with disabling EndpointSlice too instead of modifying metallb?
> 
> The main reason is that you can not disable the Endpointslices resource in the apiserver. So we would not only need to disable the endpointslices controller in the controller-manager, but also deny creating the resource and remove all existing ones.
> 
> That's why we want a feature flag like core-dns has to disable the usage in the addon.

But (and I may be wrong here), if you disable the controller in controller-manager via the EndpointSlice, ep slices won't be created (even though the resource is still available), and the automatic check should cover that:

https://github.com/metallb/metallb/blob/4de2c0c7fbd3e4eef624bdff40c1c02f98d252d8/internal/k8s/k8s.go#L611

Here we check if the kubernetes service doesn't have endpoint slices, and if so we assume that ep slices are disabled. Would that work or is that logic bugged somehow?

Christoph-Raab:
You are correct that if you disable the controller, EndpointSlices will not be automatically created by kubernetes. 

But the resource still exists in the apiserver. So any user or addon/operator could still create EndpointSlices. So you need to forbid that operation as well, remove all existing ones (because they are no longer manged by the controller/user/addon/operator) and hope, that no user/addon/operator depends on them or something breaks, if they are not there/can not be created.

It's much saver to disable them in metallb, where we are certain, that we don't want to use them and nothing breaks if we do so.

fedepaol:
> You are correct that if you disable the controller, EndpointSlices will not be automatically created by kubernetes.
> 
> But the resource still exists in the apiserver. So any user or addon/operator could still create EndpointSlices. So you need to forbid that operation as well, remove all existing ones (because they are no longer manged by the controller/user/addon/operator) and hope, that no user/addon/operator depends on them or something breaks, if they are not there/can not be created.
>

Well, endpoints slices are not meant to be created by anything else than k8s. You could say the same thing about endpoints, if somebody creates a new endpoint manually you break the service.
In particular, the check we are performing is against the "kubernetes" service which is always there.

I understand on the other hand the fact that having to delete them is clunky, and that we did not think about this corner case where a cluster could switch from having them enabled to disabled.
 
> It's much saver to disable them in metallb, where we are certain, that we don't want to use them and nothing breaks if we do so.

controller/main.go
```diff
@@ -122,13 +122,14 @@ func (c *controller) MarkSynced(l log.Logger) {
...
-		deployName = flag.String("deployment", os.Getenv("METALLB_DEPLOYMENT"), "name of the MetalLB controller Deployment")
-		logLevel   = flag.String("log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", strings.Join(logging.Levels, ", ")))
+		port          = flag.Int("port", 7472, "HTTP listening port for Prometheus metrics")
+		config        = flag.String("config", "config", "Kubernetes ConfigMap containing MetalLB's configuration")
+		namespace     = flag.String("namespace", os.Getenv("METALLB_NAMESPACE"), "config / memberlist secret namespace")
+		kubeconfig    = flag.String("kubeconfig", "", "absolute path to the kubeconfig file (only needed when running outside of k8s)")
+		mlSecret      = flag.String("ml-secret-name", os.Getenv("METALLB_ML_SECRET_NAME"), "name of the memberlist secret to create")
+		deployName    = flag.String("deployment", os.Getenv("METALLB_DEPLOYMENT"), "name of the MetalLB controller Deployment")
+		logLevel      = flag.String("log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", strings.Join(logging.Levels, ", ")))
+		disableSlices = flag.Bool("disable-slices", false, "Disable the usage of endpointslices")
```
fedepaol, line 132: I'd call this `disable-epslices` and mention the fact that it defaults to endpoints in the description without relying on the autodiscovery mechanism

fedepaol: Also, EndpointSlices or endpoint slices in the description

internal/k8s/k8s.go
```diff
@@ -603,7 +604,11 @@ func serviceNameForSlice(endpointSlice *discovery.EndpointSlice) (string, error)
 }
 
 // UseEndpointSlices detect if Endpoints Slices are enabled in the cluster.
-func UseEndpointSlices(kubeClient kubernetes.Interface) bool {
+func UseEndpointSlices(kubeClient kubernetes.Interface, cfg *Config) bool {
+	if cfg.DisableSlices {
```
fedepaol, line 608: Can you add a comment about the particular scenario (epslices enabled but disabled in the proxy) that we are covering with this parameter?

controller/main.go
Christoph-Raab: Done

internal/k8s/k8s.go
Christoph-Raab: Done

internal/k8s/k8s.go
```diff
@@ -164,7 +165,7 @@ func New(cfg *Config) (*Client, error) {
 		c.syncFuncs = append(c.syncFuncs, c.svcInformer.HasSynced)
 
 		if cfg.ReadEndpoints {
-			if !UseEndpointSlices(c.client) {
+			if !UseEndpointSlices(c.client, cfg) {
```
fedepaol, line 168: I'd change this to check the flag outside, seems more readable.

Something like `if cfg.DisableEPSlices || !UseEndpointSlices(c.client) {`

fedepaol:
@Christoph-Raab sorry for not reviewing this before

fedepaol:
> @Christoph-Raab sorry for not reviewing this before

Also, mind adding the rationale to the commit message? Same thing you added as a comment.

internal/k8s/k8s.go
Christoph-Raab: Having all logic to determine if EndpointSlices should be used in one method seams like the reasonalbe thing to do, so I put it there. But I can change that.

Should I add a comment like I did in `UseEndpointSlices`? Or is it fine without that?

fedepaol: Agree, but passing a flag just to return early won't add much value to the function itself I think. Might be different if we start using that in many different places, but that's not the current situation.

I'd move the comment too.

Christoph-Raab: Done

Christoph-Raab:
> > @Christoph-Raab sorry for not reviewing this before
> 
> Also, mind adding the rationale to the commit message? Same thing you added as a comment.

Sure, no problem

russellb:
just to be explicit, i'm looking for a "lgtm" from @fedepaol since @fedepaol did the endpointslices support

fedepaol:
lgtm , and thanks!

rata (approved):
LGTM, thanks @Christoph-Raab and @fedepaol ! Will merge now.

However, some questions for both regarding a possible follow-up PR:
1. Would it make sense in a future PR to try to detect this situation in "UseEndpointSlices()" without any need for a user-specified flag?

For example, we can look at the kubernetes service (to say an example) or some other service that should be present on any conformant clusters and see if endpoint slices are created for them? Or was that considered too hacky or there isn't such a service we can know to be present?


Or the issue is that endpoint slices are being created and updated, but we don't want to use them in metallb for some reason? What is the reason to not use them in metalLB if they are created and updated by k8s controllers?

I think I'm missing those details. Will merge this now, but please let me know what you think :)

rata:
Merging, LGTM to me, also LGTM from @fedepaol and IIUC also @russellb

fedepaol:
> For example, we can look at the kubernetes service (to say an example) or some other service that should be present on any conformant clusters and see if endpoint slices are created for them? Or was that considered too hacky or there isn't such a service we can know to be present?
> 
> Or the issue is that endpoint slices are being created and updated, but we don't want to use them in metallb for some reason? What is the reason to not use them in metalLB if they are created and updated by k8s controllers?
> 

The current code tries to do what you just described, against the k8s service, checking for the presence of epslices.

There is one corner case though, where you can have the epslices enabled but kubeproxy instructed to consume endpoints, that is enabled by a feature flag and there's no way (I know of) to find it programmatically.
I asked @Christoph-Raab to specify that in the commit message and as a comment in the code.

> I think I'm missing those details. Will merge this now, but please let me know what you think :)

rata:
@fedepaol oh, you are right, on my quick look I missed that the current code is doing just that. Thanks!

@Christoph-Raab just curious, in which scenario endpoint slices are created by the k8s controllers (I assume they have up to date information) but is problematic if MetalLB uses them? Do you hit a bug in MetalLB using endpoint slices? Or why MetalLB should not use them?

In any case, if kube-proxy is doing this, I guess we can too...

champtar:
(on my phone) Is the flag name the same as kube-proxy ? If not does it make sense to use the same flag name ?

Christoph-Raab:
@rata 

> @Christoph-Raab just curious, in which scenario endpoint slices are created by the k8s controllers (I assume they have up to date information) but is problematic if MetalLB uses them? Do you hit a bug in MetalLB using endpoint slices? Or why MetalLB should not use them?

We disabled endpointslices with kube-proxy but not in the whole cluster. So metallb would be using them but kube-proxy would still work with endpoints. We don't really trust that endpoints and endpointslices are in perfect sync so we rather want addons to use endpoints.

@champtar 

> Is the flag name the same as kube-proxy ? If not does it make sense to use the same flag name ?

Kube-proxy uses a flag in the config map:

```
    featureGates:
      EndpointSliceProxying: false
```

We chose `--disable-epslices` for metallb which describes quite good what it does.

Issue #408, opened 2019-02-26
Store BGP passwords in secrets


junkiebev:
I've deployed metallb in my cluster from version 0.8.4 and specified `configInline`.

If I am reading this correctly, it looks like if `configInline` is used, a config map is generated. I'm curious if it is possible to use secrets to specify `configInline.peers[0].password` values.

I'd presume that this would require some helm chart refactoring.

danderson:
It's not currently possible. It's a reasonable feature request, although full disclosure it likely won't happen any time soon, as my free time to work on MetalLB is extremely limited.

junkiebev:
Do you think it is just a chart modification or are you planning to handle it via environmental variables? Some other way? I might be able to take a stab at it if you can point me towards what you are thinking.

danderson:
It requires code changes to MetalLB to read the right secret(s), and to recycle BGP connections when the secrets change. It's not _especially_ hard, but it's also not a trivial manifest change.

Blorpy:
I realize this has been open for ages, but just throwing a +1 for this. 

It would be great if we could check the configmap into git, but doing so would currently require us to have the password in plaintext in there which is an issue since it's easily obtained. Since its also a very low bar to be able to source traffic from a k8s node IP, route hijacking and such security issues are a potential problem here. 

While we can apply prefix lists on the peer routers to limit what we accept/advertise to MetalLB, not having to store a secret in git in plaintext would be great!

junkiebev:
This was closed

PR #942, opened 2021-09-08
Add a pool configurability design proposal.

fedepaol:
As discussed in the last community meeting, this proposal tries to collect all the requests done in separate issues in a organic way, and come with a proposal that includes all of them to start a discussion.


fedepaol:
cc @johananl @rata @champtar @russellb

design/pool-configuration.md
```diff
@@ -0,0 +1,385 @@
+# Address Pool configuration options
```
markdgray, line 1: Just a general question for maintainers. Do we have a naming convention for design proposals?

design/pool-configuration.md
```diff
@@ -0,0 +1,386 @@
...
+
+## Proposal
+
+For each ip of an address pool, we need to address three main points:
+
+* Which nodes must be advertised as destinations of that ip
+* Where / to whom the ip address must be advertised to
+* How to allocate that ip to a given service
+
+The general idea of this proposal is that these degree of configuration are all related to the AddressPool object, and
```
markdgray, line 34: nit: degree -> degrees

design/pool-configuration.md
```diff
@@ -0,0 +1,386 @@
...
+
+### Node / Service IP mapping
+
+The most common request is being able to associate a given pool to a subset of nodes (which translates to selectively announce a subset of speakers to announce the ip from).
+
+This covers use cases where different set of nodes are exposed to different subnets, and the loadbalancer IP can be reached only via a subset of the nodes.
+
+[<img src="./images/ip_per_node.svg" width="400"/>](./images/ip_per_node.svg)
+
+This was raised in the following issues / PRs:
```
markdgray, line 45: @fedepaol Good work in collating all of these.

design/pool-configuration.md
```diff
@@ -0,0 +1,386 @@
...
+
+* [Issue: LoadBalancer service is not accessible using external IP, when endpoints (with a pod inside the cluster) is created manually](https://github.com/metallb/metallb/issues/717)
+* [Issue: Announce route for nodeless endpoints](https://github.com/metallb/metallb/issues/646)
+
+### IP Announcement Targets
+
+Currently, an IP assigned to a Service is advertised to all the possible destinations from all the nodes.
+The only exception is the BGP peer [node selector](https://metallb.universe.tf/configuration/#limiting-peers-to-certain-nodes).
+
+Here, we suggest to extend the configuration logic with a protocol dependent way to select the target(s) of the announcement.
```
markdgray, line 67: independent?

markdgray: Do you know why we can't use the BGP method with Layer2?

fedepaol: Not sure I understand the question here.
This part is about choosing the targets of the advertisement, which (I believe) is protocol dependent because of the different nature of the "targets" in the BGP and the L2 context.

The part above (node / addresspool mapping) is meant to be protocol independent, and as such the one below (pool / service association)

design/pool-configuration.md
```diff
@@ -0,0 +1,386 @@
...
+      - my-asn: example.com/local-asn
+        peer-asn: example.com/remote-asn
+        peer-address: example.com/address
+```
+
+will need to be extended with a name field (as much as the `peers` substructure) so it can be selected.
+
+### Conflicts with the FRR implementation
+
+All the BGP changes will happen from a "behavioural" point of view. Where needed, the changes will reflected in both the native and the new implementation.
```
markdgray, line 170: I think the FRR design proposal discussed not adding features to the native implementation when we started development on FRR. It might be that we need to do only a little to enable it but I wouldn't let the native implementation be a blocker.

design/pool-configuration.md
markdgray: Yeah and reading over this, I agree. I thought you were proposing doing something with L2 as well (selectively choosing interfaces) but you say below it wasn't included in this proposal. Do we think we will ever need or want to do this?

design/pool-configuration.md
```diff
@@ -0,0 +1,386 @@
...
+### Mixing nodes / target selection
+
+A more complete scenario would mix the two configuration degrees described above, advertising the same pool to different peers depending
+on the node the ip is being advertised from.
+
+[<img src="./images/pool_peer_node_selector.svg" width="400"/>](./images/pool_peer_node_selector.svg)
+
+No issues were raised but this combination is a more flexible variant of the [peer node selector](https://metallb.universe.tf/configuration/#limiting-peers-to-certain-nodes).
+
+With this combination, the user may choose a different set of targets depending on the node.
```
markdgray, line 103: So node allocation and peer selection are kind of orthogonal. The exceptions are due to a node being unable to reach a peer. So where would we check for those exceptions? The webhook that you mention below?

design/pool-configuration.md
fedepaol: Agree, I just have the feeling that all the implementation will happen above the bgp interface.

design/pool-configuration.md
```diff
@@ -0,0 +1,386 @@
...
+
+The BGP controller already contains a mapping between services and their BGP advertisement parameters taken from the address pool configuration.
+
+The structure is being filled [here](https://github.com/metallb/metallb/blob/5d236989e61683daf1eaad00777b81ce7dd98c5d/speaker/bgp_controller.go#L218).
+
+The idea here is to extend that with the list of peers that ip must be advertised to. In case the peer comes with a node selector, extra logic on the current node and the node selector will be in place to understand if to put the peer in the list of not.
+
+### Pool / Service - Namespace association
+
+The association of the pool with a particular service is orthogonal to the previously described feature.
```
markdgray, line 288: I am trying to convince myself of this statement :D. So, if I understand this correctly, if I assign an address pool to a particular service, we *can* reach that service from any node that has a backend for the service - this is regardless of what MetalLB does? .. which would seem to make them independent.

design/pool-configuration.md
fedepaol: I think that is clearer in the examples below

There are two types of node selection here:

* The pool node selector drives which nodes are allowed to advertise the ip
* The node selector in the bgp peer of a given pool, tells that the ip must be advertised to that particular peer from a given subset of nodes

A webhook would be able to spot configurations with no intersections between the two selectors resulting in a pool that won't be advertised anywhere.

design/pool-configuration.md
fedepaol: This section is only about the way we associate the pool to a given service. Today is just random. The first pool that has free ips is associated to a given service [here](https://github.com/metallb/metallb/blob/997a177a762c6c8215ca9438083545529b675376/internal/allocator/allocator.go#L274).

Once we have an ip associated to the service, MetalLB will advertise it from a node that:
- has a speaker running (L2 only)
- in case of local traffic policy, has a running backend
- in case the node selector is implemented, the node must match the selector

This phase comes after the pool allocation to the service. 
The only relation I can see is that if a user cares about pinning a particular pool to a service (or to a namespace), the reason is likely because they want to leverage some of the extra tweakings that we are proposing here.

design/pool-configuration.md
rata: AFAIK, no. But maybe we can move ones already implemented to a subdir, assume most will be implemented so this will still be manageable?

design/pool-configuration.md
```diff
@@ -0,0 +1,386 @@
...
+* Highlight the cases where the solution must be BGP or L2 specific
+
+### Non-Goals
+
+* Covering only a specific use case.
+* Providing a L2 only or BGP only solution where the problem might be expressed in a more generic way.
+
+## Proposal
+
+For each ip of an address pool, we need to address three main points:
```
johananl, line 28: This phrasing is ambiguous IMO.

Let IP A and IP B be IPs in address pool X. Which of the following is true?

1. IP A can be allocated/advertised separately from IP B. For example, the user can decide that IP A be advertised only from node A whereas IP B be advertised from all available nodes.
2. IP A and IP B are always treated the same way. IOW, the policy options which control IP allocation/advertisement apply to **all IPs** in a pool.

Personally, I think option 2 makes the most sense. If you want to treat an individual IP in a special way, you can always define a pool with just that IP.

johananl (commented):
Thanks for driving this @fedepaol!

I'm not done going through the proposal, however I wanted to provide feedback ASAP for the things I've already read. So, I'm submitting a couple of comments.
Will submit another review in the upcoming days after reviewing the rest :crossed_fingers:

design/pool-configuration.md
fedepaol: Yep, I agree and that's what I tried to express.
I'll rephrase to make it clear that any behavior we introduce will be consistent for all the IPs of a given pool.

design/pool-configuration.md
```diff
@@ -0,0 +1,386 @@
...
+
+* Which nodes must be advertised as destinations of that ip
```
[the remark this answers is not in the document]
fedepaol: Will change that.

design/pool-configuration.md
```diff
@@ -0,0 +1,386 @@
...
+* Where / to whom the ip address must be advertised to
```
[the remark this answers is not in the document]
fedepaol: Agree. I wasn't sure when I started writing this if we wanted to have some kind of filtering in ARP advertisement or not, ended up saying we don't in a following sectoin. Will change it.

design/pool-configuration.md
```diff
@@ -0,0 +1,386 @@
+# Address Pool configuration options
+
+## Summary
+
+Following @johananl's initiative in [this issue](https://github.com/metallb/metallb/issues/630)
+this enhancement tries to collect and classify various configuration degrees requested during time.
```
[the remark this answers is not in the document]
fedepaol: received over time? Pardon my English, not sure what to change here :P

design/pool-configuration.md
```diff
@@ -0,0 +1,390 @@
...
+          network: first
+          network: second
+```
+
+#### Implementation
+
+The [current implementation](https://github.com/metallb/metallb/blob/312b03cd3065687f25274486cd3ff5c79d6f6068/speaker/layer2_controller.go#L45) announces L2 pools only from those nodes
+with a running speaker and an active endpoint for the service.
+
+In order to be able to announce L2 from a subset of nodes, this logic must be changed, checking for active endpoints only if the TrafficPolicy is local (which is handled by the [BGP implementation](https://github.com/metallb/metallb/blob/5d236989e61683daf1eaad00777b81ce7dd98c5d/speaker/bgp_controller.go#L153)).
```
sabinaaledort, line 219: checking for active endpoints only if the TrafficPolicy is local was done:
https://github.com/metallb/metallb/blob/36166169e7c804bd727cdfe19df6133b069c4b1b/speaker/layer2_controller.go#L100

design/pool-configuration.md
```diff
@@ -0,0 +1,390 @@
...
+
+## Proposal
+
+For all the ips belonging to an address pool, we need to address three main points:
+
+* Which nodes should advertise the ip
+* To which BGP neighbors the IP address should be advertised
+* How to allocate that ip to a given service
+
+The general idea of this proposal is that these degrees of configuration are all related to the AddressPool object, and
```
johananl, line 35: I find the term "degree" confusing in this context. Also, I suggest we choose one term for the three items in the list above and use that same term throughout the document (right now we call these items "points" and also "configuration degrees"). Using a single term reduces ambiguity.

design/pool-configuration.md
```diff
@@ -0,0 +1,390 @@
...
+
+This third (currently not implemented) mode raised in [this discussion](https://github.com/metallb/metallb/pull/832#issuecomment-888897400) where
+there are scenarios in which only the ip pooling side of metallb is used, but the routes to those ips are set statically.
+
+On top of that, Calico suggests to deploy MetalLB without the speakers to leverage only the allocation feature exposed by the controller in
+[its official documentation](https://docs.projectcalico.org/networking/advertise-service-ips).
+
+### Mixing nodes / target selection
+
+A more complete scenario would mix the two configuration degrees described above, advertising the same pool to different peers depending
```
johananl, line 100:
```suggestion
A more complete scenario would mix the two configuration elements described above, advertising the same pool to different peers depending
```

See my previous comment about "degrees".

design/pool-configuration.md
```diff
@@ -0,0 +1,390 @@
...
+* Which nodes should advertise the ip
+* To which BGP neighbors the IP address should be advertised
+* How to allocate that ip to a given service
+
+The general idea of this proposal is that these degrees of configuration are all related to the AddressPool object, and
+that these configurations are independent between each other.
+
+### Node / Address Pool mapping
+
+The most common request is being able to associate a given pool to a subset of nodes (which translates to selectively announce a subset of speakers to announce the ip from).
```
johananl, line 40:
```suggestion
The most common request is being able to associate a given pool to a subset of nodes (which translates to selecting a subset of speakers so that these speakers - and only them - announce the IP).
```

I think there is a phrasing problem here.

design/pool-configuration.md
```diff
@@ -0,0 +1,390 @@
...
+
+This covers use cases where different set of nodes are exposed to different subnets, and the loadbalancer IP can be reached only via a subset of the nodes.
```
johananl, line 42:
```suggestion
This covers use cases where different sets of nodes have interfaces in different subnets, and the loadbalancer IP can be reached only via a subset of the nodes.
```

- I'm assuming you meant "different sets" and not "a different set" here.
- Let's make the term "exposed to" less ambiguous.

design/pool-configuration.md
```diff
@@ -0,0 +1,390 @@
...
+For all the ips belonging to an address pool, we need to address three main points:
+
+* Which nodes should advertise the ip
+* To which BGP neighbors the IP address should be advertised
+* How to allocate that ip to a given service
+
+The general idea of this proposal is that these degrees of configuration are all related to the AddressPool object, and
+that these configurations are independent between each other.
+
+### Node / Address Pool mapping
```
johananl, line 38: I'm a bit confused. Does this title refer to the 1st "point" in the previous list, namely "Which nodes should advertise the ip"? If this title and the following ones are an elaboration on the previous list, could we use the same text for the "points" and the titles of the respective sections? If this title and the following ones are *not* an elaboration of the list, could we explain what these are (e.g. are these use cases? Feature requests? Something else?). In the latter case, how do these sections relate to the 3 "points" above?

Also, what does the `/` denote here? Could the slash be substituted with the word "to"? If there is directionality to the mapping, could use use `->` or something similar? If not, maybe `<>` would be better.

design/pool-configuration.md
```diff
@@ -0,0 +1,390 @@
...
+* [PR: Address Pool with node selector filtering](https://github.com/metallb/metallb/pull/502)
+* [Issue: Layer 2: impossible to use MetalLB on a subset of Kubernetes nodes](https://github.com/metallb/metallb/issues/302)
+* [Issue: Layer2 Different subnets(label for LB, check for interface](https://github.com/metallb/metallb/issues/737)
+* [Issue: Look into a generic pool <> node affinity mechanism](https://github.com/metallb/metallb/issues/630)
+* [PR: MetalLB speakers working on a subset of nodes in layer2](https://github.com/metallb/metallb/pull/613)
+* [Issue: Annotate nodes with address pool/weighted address pools](https://github.com/metallb/metallb/issues/941)
+
+The following issues also fall in the same "speaker announcement logic" category:
+
+Service with external backends
```
johananl, line 58: I don't understand the meaning of this line. Is it necessary? Does this line classify the two issues which follow? Could we reduce ambiguity here?

design/pool-configuration.md
```diff
@@ -0,0 +1,390 @@
...
+* Covering only a specific use case.
+* Providing a L2 only or BGP only solution where the problem might be expressed in a more generic way.
+
+## Proposal
+
+For all the ips belonging to an address pool, we need to address three main points:
+
+* Which nodes should advertise the ip
+* To which BGP neighbors the IP address should be advertised
+* How to allocate that ip to a given service
```
johananl, line 33: Nit: Could we capitalize "IP" consistently throughout the document?

johananl (changes requested):
Thanks a lot for working on this @fedepaol. I really like your work around classifying the different issues we have around making address pools more flexible. This brings a lot of order and makes it easy to see the big picture :+1: 

IMO there is too much ambiguity in the "Proposal" section - see my inline comments. I haven't reviewed the sections which follow since I feel like I have to understand the "Proposal" section properly before I can evaluate the rest of the document. Specifically, I'm having a hard time understanding the **relation** between the various sections (i.e. "I understand the contents of this paragraph but I'm not sure why this paragraph is there").

If you could help me understand your intention by addressing my inline comments and ensure the various section titles make sense - that would be great. Please tag me again when done and I'll do another pass :pray:

design/pool-configuration.md
fedepaol: I'll try to re-organize always using the same title.

design/pool-configuration.md
fedepaol: It does, but those issues now closed and fixed by https://github.com/metallb/metallb/pull/976
Will remove it.

design/pool-configuration.md
fedepaol: Yep, I did it after redacting this proposal. Removing it.

fedepaol:
> Thanks a lot for working on this @fedepaol. I really like your work around classifying the different issues we have around making address pools more flexible. This brings a lot of order and makes it easy to see the big picture +1
> 
> IMO there is too much ambiguity in the "Proposal" section - see my inline comments. I haven't reviewed the sections which follow since I feel like I have to understand the "Proposal" section properly before I can evaluate the rest of the document. Specifically, I'm having a hard time understanding the **relation** between the various sections (i.e. "I understand the contents of this paragraph but I'm not sure why this paragraph is there").
> 
> If you could help me understand your intention by addressing my inline comments and ensure the various section titles make sense - that would be great. Please tag me again when done and I'll do another pass pray

I _think_ I addressed your comments (or maybe made it worse :P )
I used the same titles in both the proposal and the design, and it's way more clear.
This is ready for another pass, thanks a lot @johananl !

design/pool-configuration.md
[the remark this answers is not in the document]
johananl: Ah, right. Got it now. Yes, "received over time" is better.

design/pool-configuration.md
```diff
@@ -0,0 +1,359 @@
...
+* Covering only a specific use case.
+* Providing a L2 only or BGP only solution where the problem might be expressed in a more generic way.
+
+## Proposal
+
+For all the IPs belonging to an address pool, we need to address three main configuration degrees:
+
+* Announcing the IP from a subset of nodes
+* Announcing the IP to a subset of BGP neighbors
+* Using a given AddressPool only for a subset of services
```
johananl, line 33: Actually we seem to have **two** different concerns here:

1. Ensuring service X receives an IP from address pool Y.
2. Ensuring no other service besides X receives an IP from address pool Y.

The 1st concern is independent from the 2nd: I may want to ensure service X gets its IP from pool Y but I don't care if other services also get their IPs from pool Y.

design/pool-configuration.md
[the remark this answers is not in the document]
johananl: ARP is a broadcast-based layer 2 protocol. While it's technically possible e.g. to refuse to provide an ARP reply to a specific source MAC, doing so would be weird IMO since a security policy usually isn't enforced at that layer.

design/pool-configuration.md
fedepaol: Ugh I thought I accepted this

design/pool-configuration.md
```diff
@@ -0,0 +1,359 @@
...
+* Highlight the cases where the solution must be BGP or L2 specific
+
+### Non-Goals
+
+* Covering only a specific use case.
+* Providing a L2 only or BGP only solution where the problem might be expressed in a more generic way.
+
+## Proposal
+
+For all the IPs belonging to an address pool, we need to address three main configuration degrees:
```
johananl, line 29: I am not sure an IP is the "central" entity we should think about for all cases.

For example, in most commercial routers, controlling which IP prefixes are advertised to which BGP peers is done at the **peer** level, not the IP level. You typically attach a "route map" to a BGP peer. The route map is basically a policy document which defines which prefixes known to the local BGP process are allowed to be advertised to the peer to which the route map is attached.

AFAICT we have the following concerns in MetalLB:

1. Which speaker nodes should announce a prefix. This would be the equivalent of selecting which routers to apply a configuration to.
2. Which BGP peers to advertise a prefix to. This would be the equivalent of restricting BGP advertisements using a route map.
3. Which IP to allocate to which service. This would be the equivalent of an IPAM policy that's **completely unrelated** to advertisement.

I'd consider rephrasing as follows:

>MetalLB deals with the following concerns:

Then I would rephrase the list to be more generic as follows:

- Controlling which speaker nodes advertise a prefix.
- Controlling which peers a speaker node should advertise a prefix to.
- Controlling which IPs are allocated to which services.

Lastly, see my [note](https://github.com/metallb/metallb/pull/942/files#r773898754) below.

design/pool-configuration.md
```diff
@@ -0,0 +1,359 @@
...
+set custom properties only on one single service.
+
+That aside, this proposal covers the majority of the issues that were raised by the users (multiple times), and it won't prevent to add a finer level of configuration
+if ever comes the need.
+
+## Design Details
+
+### Announcing the IP from a subset of nodes
+
+Extending the address-pool object with an optional node selector field is straightforward:
```
johananl, line 162: Consider the following use case:

- **IP X** is allocated to some service.
- **Node A** is peered with **routers 1 and 2**. **Node B** exists as well.
- I want to advertise IP X **from node A** (and not node B) but **only to router 1**.

I'm not sure we can handle this use case in the proposed design, but even if we can, I'd say the following design would be more "natural" when thinking about how this kind of policy is handled by commercial routers:

We apply a policy **to node A**. This policy filters BGP advertisements to routers 1 and 2, allowing IP X to be advertised to router 1 but not to router 2.

"Translating" the above to MetalLB terms, I'd consider putting the "filtering" logic in the **peer**:

```yaml
    peers:
    - peer-address: 10.0.0.1
      peer-asn: 64501
      my-asn: 64500
      name: foo
      allowed-prefixes:
      - 172.16.0.1/32
      - 192.168.1.0/29
```

In case you want to have two different policies for node A and node B *for the same peer 10.0.0.1*, you would define **two peer sections**, each with a different node selector, with the same `peer-address` and different `allowed-prefixes` clauses.

WDYT?

johananl (commented):
Thanks for following up @fedepaol. It is much easier for me to reason about the design now.

I think we should reconsider where to put the prefix filtering logic. IMO the natural place for this logic would be the **peer**, not the **address pool**. See my inline comments.

I'm not reviewing the rest to save time since this is a crucial point in this proposal which may change the whole thing if we decide to touch it.

design/pool-configuration.md
fedepaol: Sounds clearer, will apply your suggestion. I'll reply on the bgp side on the other comment below.

design/pool-configuration.md
fedepaol: Makes sense, I think this will be covered by your rephrasing above. It doesn't add much value to split the item into two.

design/pool-configuration.md
fedepaol: The scenario you are describing is covered with the next section, "Announcing the IP to a subset of BGP neighbors".

The first thought that comes into my mind is that this won't work with L2, and a lot of users asked for this feature because in L2 only one node advertises and different nodes may be connected to different networks.

Said that, I am a bit torn about this. On one hand, it's more flexible and near to what it will look like in FRR.

On the other hand, it feels a bit too low level. Users will have to configure the addresspools, and then go manually and copy the cidrs to the peers. It feels more natural to say "I want this peer to advertise this pool only" (or as per my version, "I want this pool to be advertised only to these peers") because the pool is the higher level object the prefixes come from.

I saw value in having the addresspool the central entity of our configuration. Together with service mapping logic, it would allow metallb to be used in multi tenant scenarios that require different configurations for different tenants. 

With this, it would be doable, but more fragmented and error prone because we'd loose the natural association with the pool and with the namespace.

Would that work if `allowed-prefixes` accepts also strings and in that case the semantic is that it's the name of a pool? Or having an additional `allowed-pools` field together with that (maybe too many knobs)? We'd loose the addressPool being the center of the configuration, but we'd still be able to provide a higher level abstraction that will reduce configuration errors.

johananl: You've raised good points.

Following the discussion in the community meeting and since we're considering implementing this design using CRDs anyway, I think the best approach may be to separate the concerns of **allocating** IPs and **advertising** IPs. If we take this approach, the following happens:

- An address pool object becomes protocol-agnostic and specifies only allocation-related params
- Either we create a layer 2 equivalent of `peer` and control prefix filtering there (by referencing address pools), **or** we add a protocol-agnostic abstraction, say `kind: Advertisement`, which controls whether or not to advertise a prefix to a given router (and then we don't need a `peer` equivalent for layer 2 as long as there are no APR-specific knobs to tweak).

fedepaol: Yep, I'll rework this proposal to include a re-design of the data structures (I feel like it won't change to much whether we implement them as sections of the configmap or as separate crds), so we can iterate from there (next year :P ).

fedepaol:
@johananl @rata @gclawes I think this is ready for another pass.

I added the new data model changes here, as a result of our previous discussion. This is _one_ possible version, which I am pretty sure needs to converge.

I was torn whether to add a different design proposal, as this is becoming quite big, but I also have the feeling that it helps seeing it as a whole, to understand better where the new additions will end up. So far, the new changes are in a separate commit. When we agree (or start converging) on this, I'll squash.

design/pool-configuration.md
```diff
@@ -0,0 +1,533 @@
...
+  namespace: metallb-system
+spec:
+  peerAddress: 10.0.0.2
+  peerASN: 64501
+  myASN: 64500
+  routerID: 10.10.10.10
+  peerPort: 1
+  holdTime: "10s"
+  sourceAddress: 1.1.1.1
+  bfdProfile: bfdprofilefull
```
gclawes, line 323: What's the reason for a separate `BFDProfile` resource?  Is it possible to associate a `BFDProfile` with multiple BGP peers?

design/pool-configuration.md
```diff
@@ -0,0 +1,533 @@
...
+  name: bgpadvertisement
+  namespace: metallb-system
+spec:
+    pools:
+      - pool1
+      - pool2
+    peers:
+      - peer1
+      - peer2
+    - communities:
```
gclawes, line 241: This looks like a yaml typo, `- communities` is a list element mixed in with an object.

design/pool-configuration.md
```diff
@@ -0,0 +1,533 @@
...
```
gclawes, line 417: Ditto on the yaml typo, `- communities`

design/pool-configuration.md
fedepaol: Yes, or to put it the other way around, multiple bgp peers can use the same profile. It's also the same way frr handles it, which is not a good enough reason BUT with a 1:1 mapping between a peer and its bfd settings, we'd have to create multiple bfd profiles in the frr configuration (and this does seemed reasonable).

gclawes: Gotcha!  Yeah, I'm just trying to figure out if there's a way to prevent config sprawl.

design/pool-configuration.md
```diff
@@ -0,0 +1,533 @@
...
+        - 65535:65282
+    aggregationLength: 32
+    localPref: 100
+```
+
+Open question: I would like to preserve not having to specify all the ips / peers in order to match the current "ease of configuration" of metallb. One possible way is to assume that all the pools are advertised to all the peers if no selection is specified, the other one is to provide an example using the label selector.
```
rata, line 247: Big +1 to keep it easy!

design/pool-configuration.md
```diff
@@ -0,0 +1,533 @@
...
+        - namespaceB
+      serviceSelector:
+          tenant: foo
+```
+
+The controller will use the pool with highest priority in case of multiple matches. This is deterministic and will make the behaviour more clear to the users.
+
+Open question: what to do in case a selective pool finishes the IPs? As a user, if I choose a given pool from my services, I'd rather not have the IP assigned than getting one
+from another pool. One alternative is to keep this as default (i.e. if allocate-to is set and no IPs are available, do not go to the next pool). Another option is to have another
+flag that overrides this behaviour.
```
rata, line 475: I think if allocate-to is set and there are no IPs available AND there are no other pools referenced in the {BGP,L2}Advertisement, I'd go with no have the IP assigned. I think I prefer that over the flag. But no strong feelings if you think otherwise.

Another option might be to make the edit of an `AddressPool` fail if they remove an IP that is currently allocated? In that way, they can create a new service (or delete and re-create) even with a specific IP selected, then remove this service and edit the AddressPool (this time won't fail, as the IP to remove is not allocated).

Maybe that is better? What do you think?

rata (approved):
@fedepaol wow, impressive work. I like it quite a lot, thanks again! LGTM

I added some small comments, but overall big LGTM. MY personal preference is to keep things very simple for most users, and only users with a complicated setup pay the price of a complicated configuration. I see you agree on this, and it seems almost as simple as today (for example for L2 users, maybe 2 objects are needed for those use cases?).

Thanks again!

design/pool-configuration.md
```diff
@@ -0,0 +1,533 @@
...
+  minimumTtl: 10
+```  
+
+### New configuration degrees
+
+#### Announcing the IP from a subset of nodes
+
+The two Advertisement entities are extended with a selector of the nodes to announce from:
+
+A configuration would look like:
```
sabinaaledort, line 347: missing go structure (might be on purpose?)

design/pool-configuration.md
```diff
@@ -0,0 +1,533 @@
...
+  routerID: 10.10.10.10
+  peerPort: 1
+  holdTime: "10s"
+  sourceAddress: 1.1.1.1
+  bfdProfile: bfdprofilefull
+
+apiVersion: metallb.io/v1beta1
+kind: BFDProfile
+metadata:
+  name: bfdprofiledefault
```
sabinaaledort, line 328: the name of the profile here is inconsistent with the name in the BGPPeer example above.

design/pool-configuration.md
fedepaol: The problem I want to address here is, what if we have two valid AddressPools, one with the selector and the priority field set, the other one with just the selector (or the selector but a lower priority).
In my opinion we must just fail, because the behaviour is predictable and the user expressed his willing to use a specific address pool for a given service. Otherwise it would be harder for them to understand how MetalLB is behaving.

I'll go with the failure approach.

rata: Priority seems to be mandatory (it is an int, not an *int) so if the selector is present, a priority must be used (default 0?), right? I assume, then, it is the case with a lower priority you mentioned and not that other one.

Assuming that case, I'm not sure I'm missing something but just failing seems weird. LEt's say you have a namespace and 3 different address pools to use there. You _need_ to set a priority on all. Let's assume no two pool have the same priority (we would later need to address if having same prio is allowed or not), so you assign them all to one namespace.

The pool with highest prio runs out of IPs. Then, no new service will have an IP allocated. So, what is the point of having more address pools that are assigned to the namespace?

I guess it only makes sense to have several address pools assigned to one namespace if we also allow the prio to be the same. I don't think we should check any of this, just document whenever we write user-facing docs.

Or am I missing something @fedepaol ?

rata: If that makes sense and we allow several address pool to use the same prio, then I think that it can work.

It seems it can be tricky the algorithm to decide which AddressPool to use to allocate when a new service is created. But I trust you have thought that too :)

fedepaol: Priority has to be optional, because of the usual "keep it simple as long as we don't need complexity" logic.
What I am worried about here is a scenario where a user has some "special" pools for a namespace, and a bunch of other "regular" pools with no priority.

What I want to avoid is having the services in the special namespace pick an ip from the "regular" pools, that's why I am suggesting to fail.
But you are probably right, we can just document it and users will understand they had to put the right priorities and ranges in place. In that case, it will be a matter of changing the current logic to sort the pools by priority.

rata: @fedepaol I mean, from the golang POV if that field is not a pointer, it will be 0 when not present. 0 might be a reasonable default, though.

I'm kind of confused, though. I'm sure I'm missing something. But if we add a priority field and only allocate from the pool with highest prio, what is the point of having the other address pools matching the same selector with lower prio?

I mean, I don't follow why we have a prio field if basically setting it means that only the higher prio will be used and the rest ignored (seems like an easy way to ignore the rest of the pools).

What am I missing?

fedepaol: The point is, not having a selector currently means "select everything" (and we want to keep that).

With that in mind, you can have some "general pools" good for everything, and one namespace specific pool.
Also, the selector is a label selector, so it could be that we have a broader selector that also impacts (as a side effect) the "finer selected" pools.

An example:

Service A: label `important:true`, `notsoimportantbutimportant:true`
Service B: `notsoimportantbutimportant:true`
Service C

pool A -> selector: `important:true` , prio 100
pool B -> selector `notsoimportantbutimportant:true` (which includes some services collected by poolA), prio 50
pool C -> no selector

rata: Ohh, thanks, that is what I was missing!

Just curious, is that a real use case or just what happens with all the other stuff? I mean, it seems intricate if it is an use case, I'm just curious.

It seems reasonable. Also, if we want the alternative "to allocate from any of poolA or pool B", then pool B needs to have prio 100 (or pool A and pool B no prio).

I guess what we can't express is "take IPs from this pool, only when that pool is empty use this other pool". And that seems like a reasonable thing to leave out, AFAIK. Either services can have more labels (to play with selectors) and the user can request an specific IP.

Thanks for the explanations!

fedepaol: The real (and most common) use case is "I want to reserve some pools for some namespaces / services", and other pools to the rest.
The extra complication comes from the fact that we can have intersections in those selections, and if we allow them, it will be possible to incur in what I described.

In any case, documenting the priority and saying that we'll go on the next pool if the highest one is empty (instead of failing) is as much as reasonable, and I think it makes sense.

fedepaol: @rata changed.

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+
+Before addressing the new concerns, we will propose a new scheme for configuring MetalLB, which will be used as a starting point
+in the following parts of the proposal. For this reason, it makes sense to have it described in this proposal instead of a separate one.
+
+The new data model will be expressed in terms of kubernetes CRDs, deprecating the configuration via the metallb's ConfigMap.
+
+In the new data model, the following entities will be defined (more details on the implementation section):
+
+* `AddressPool`, responsible of the ip allocation to services
+* `L2Advertisement`, describing the intention to advertise an IP coming from an `AddressPool` via L2
```
johananl, line 42: >the intention to advertise an IP

An IP or a **prefix**? If we're really separating IP allocation from advertisement, I assume I should be able to aggregate multiple IPs in a `BGPAdvertisement` as a `/28` for example.

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+```go
+type addressPool struct {
+  Name              string
+  Addresses         []string
+  AvoidBuggyIPs     bool
+  AutoAssign        *bool
+}
+```
+
+Example CRD
```
johananl, line 194: Nit: This is a **CR**, not a CRD.

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+* [Issue: Dynamic BGP communities from services](https://github.com/metallb/metallb/issues/934)
+
+### Risks and Mitigations
+
+#### Regressions
+
+The e2e test suite is wide enough to ensure we have good coverage on metallb's main features. Given the new data model will be functionally equivalent
+to the old one, converting the test suite should give us enough confidence in that regard.
+
+#### Impacts on the users' configurations
```
johananl, line 126: Nit: "impact" is normally used as singular.

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+
+The operator already exposes CRDs, mapped directly to the current configmap entitites.
+Changing the CRDs layout will impact them, and in order to mitigate the behaviour, conversion webhooks will be implemented to support both the old and the new versions
+of the API, until the old version is deprecated.
+
+#### Bad configurations
+
+##### Service allocation
+
+When a user specify both a pool to be allocated to a specific service and one with no selector, the service may receive an IP from the broader one.
```
johananl, line 141:
```suggestion
When a user specifies both a pool to be allocated to a specific service and a pool with no selector, the service may receive an IP from the broader pool.
```

Fix grammar and reduce ambiguity.

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+
+##### Non overlapping filters
+
+Having an IP with no nodes advertising it is a risk. Some scenarios are:
+
+* Traffic Policy Local with pods not running on the same nodes selected by the selector
+* Node BGP peer selector with no intersection with the address selector
+
+It should be possible to detect non overlapping configurations, and raise an event when they happen. The metallb operator introduces a Validation WebHook
+that can be used to prevent those configurations to be created in first instance.
```
johananl, line 152:
```suggestion
that can be used to prevent those configurations from being created in first instance.
```

johananl: >in first instance

Did you mean "prevent from the start" or "prevent the 1st use case" (from the 2 cases mentioned above)?

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+
+### Conflicts with discovery
```
johananl, line 154: The main motivation for the peer autodiscovery feature was to provide a way for external entities to specify BGP configuration which MetalLB can pick up. With the introduction of CRDs, I'm wondering if it would make sense to let external entities (such as a Cloud Controller Manager) **create MetalLB CRs** instead of enriching `Node` objects.

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+
+The peer discovery feature currently being implemented in [this PR](https://github.com/metallb/metallb/pull/593) will introduce a way to
+fetch peers from the node's annotations.
+
+With this new structure, the PR will need to be adjusted to extend the `BGPPeer` entity in order to consume the node's annotations instead of hardcoded values.
+
+### Conflicts with the FRR implementation
+
+All the BGP changes will happen from a "behavioural" point of view. Where needed, the changes will reflected in both the native and the new implementation.
```
johananl, line 163:
```suggestion
All the BGP changes will happen from a "behavioural" point of view. Where needed, the changes will be reflected in both the native and the new implementation.
```

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+
+### Not covering all the angles
+
+The risk of not covering all possible cases / permutations is still there.
+
+That aside, this proposal covers the majority of the issues that were raised by the users (multiple times), and it won't prevent to add a finer level of configuration
```
johananl, line 169:
```suggestion
That aside, this proposal covers the majority of the issues that were raised by the users (multiple times), and it won't prevent adding a finer level of configuration
```

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+if ever comes the need. Moreover, providing a straightforward way to separate concerns will make it easier to accomodate new changes, and it will be more clear where
```
johananl, line 170:
```suggestion
if the need ever arises. Moreover, providing a straightforward way to separate concerns will make it easier to accomodate new changes, and it will be more clear where
```

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+The `AddressPool` is the entity responsible for the IP allocation part, regardless of how (and if) the ip is going to be advertised.
+
+Hence, we keep the same structure but remove the bgpAdvertisement section and the Protocol field.
+
+```go
+type addressPool struct {
+  Name              string
+  Addresses         []string
+  AvoidBuggyIPs     bool
+  AutoAssign        *bool
```
johananl, line 190: Why a bool *pointer*?

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+
+#### BGP Advertisement
+
+The `BGPAdvertisement` entity represents the intention to advertise a given IP via the BGP Protocol. It contains properties related to the type of advertisement and a reference to the address pools we want we advertise. This initial version matches the current behaviour, where an IP is advertised to all the `BGPPeer`s configured in MetalLB.
+
+```go
+type bgpAdvertisement struct {
+  Name              string
+  Pools             []string
+  PoolSelector      map[string]string
```
johananl, line 219: Why do we have both `Pools` and `PoolSelector`? What is the difference between the two?

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+  Name              string
+  MyASN         uint32
+  ASN           uint32
+  Addr          string
+  SrcAddr       string
+  Port          uint16
+  HoldTime      string
+  KeepaliveTime string
+  RouterID      string
+  Password      string
```
johananl, line 292: Following https://github.com/metallb/metallb/issues/408, I'm wondering if we should look into turning this field into a **secret ref** field from the start.

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+```
+
+A `L2Advertisement` with no pool selector is intended to select all the available `AddressPool`s.
+
+### BGP Peer
+
+The `BgpPeer` entity represents an external BGP peer we want to establish a session with. It contains the properties of the sessions, and it's a 1:1 mapping of the current bgp peer structure without the node selector. We also add the `bfdProfile` structure here, as it is part of the session definition.
+
+```go
+type bgpPeer struct {
```
johananl, line 282: What if I want a BGP peer to be configured only on node X, regardless of which prefixes are being advertised? In other words, how do I affect which nodes would initiate TCP connections on port 179 towards this peer? This concern is especially important if we support the `SrcAddr` field which is almost certainly node specific.

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+  MyASN         uint32
+  ASN           uint32
+  Addr          string
+  SrcAddr       string
+  Port          uint16
+  HoldTime      string
+  KeepaliveTime string
+  RouterID      string
+  Password      string
+  BFDProfile    string
```
johananl, line 293: I'm wondering if this should be a string, given that we're referencing a proper k8s API object here whose existence should likely be validated by a webhook upon creation of the BGP peer.

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+  namespace: metallb-system
+spec:
+  addresses:
+    - 2002:2:2::1-2002:2:2::100
+    - 172.20.0.100/24
+  autoAssign: false
+  avoidBuggyIPs: true
+  allocateTo:
+    priority: 50
+      namespaces:
```
johananl, line 485: This is misaligned: `namespaces` is a key at the same level as `priority` AFAICT.

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+        - namespaceA
```
johananl, line 486: Should we allow flexible selection using labels/annotations?

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+        - namespaceB
+      serviceSelector:
+          tenant: foo
+```
+
+The controller will use the pool with highest priority in case of multiple matches. This is deterministic and will make the behaviour more clear to the users.
```
johananl, line 492: "Higher is better" is counterintuitive to me. "Priority 1" is usually meant as "top priority". Also, lower is usually better in most routing-related matters (interface "metric", administrative distance, protocol-specific "cost" etc.). I'm aware there are exceptions but we probably want to go with the more common case.

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+The configmap will be deprecated.
+
+The CRDs are removed from the operator, which is in charge of deploying metallb only.
+
+#### CRD evolution
+
+During the new data model implementation (described in the "New Data Model" section), the e2e tests will be adjusted to use the new data model, and eventually a converter (like the one for configmap mode / crd mode)
+will be implemented to guarantee both the old and the new versions keep working.
+
+Conversion webhooks will be in place to guarantee backward compatibility for the current users of the MetalLB Operator.
```
johananl, line 535:
```suggestion
Conversion webhooks will be put in place to guarantee backward compatibility for the current users of the MetalLB Operator.
```

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+
+### New configuration degrees
+
+The remaining configuration degrees seem now to be independent:
+
+* BGPAdvertisement peers selector
+* LsAdvertisement node selector
```
johananl, line 542:
```suggestion
* L2Advertisement node selector
```

johananl (commented):
Very nice work @fedepaol.

I've done a full pass over the design and added some comments. My biggest concern is not being able to influence the formation of BGP sessions separately from prefix filtering. In addition I am not 100% "sold" on the idea of adding a node selector to `BGPAdvertisement` - see my inline comments.

design/pool-configuration.md
fedepaol: I'll add this consideration too, and if we move on with autodiscovery will take it into account.

design/pool-configuration.md
fedepaol: That's because the default value of AutoAssign is true (it's already like that in the current configuration), so we need a pointer to express "the user did not fill it" vs "the user asked for a false autoassign"

design/pool-configuration.md
fedepaol: My idea was to have both the simplicity of a list of names vs the flexibility of a label selector (which at this point can be expressed as `*metav1.LabelSelector`)

design/pool-configuration.md
fedepaol: I'd start with this version in order avoid adding even more load on this effort (as it's not a regression).
A new field with the secret name won't break the api.

design/pool-configuration.md
fedepaol: Then I'd create a BGP advertisement selecting all the pools, selecting only that particular BGP Peer AND the node selector I am proposing below.

design/pool-configuration.md
johananl: >A new field with the secret name won't break the api.

But changing the type of the `Password` field will... and I'm not sure we want to have two ways of specifying passwords in the end.

design/pool-configuration.md
johananl: Then why no pointer for `AvoidBuggyIPs`?

design/pool-configuration.md
johananl: Never mind, please disregard. I wasn't 100% clear on something here but now realize that all we need is reference a BFD profile name.

design/pool-configuration.md
```diff
@@ -0,0 +1,552 @@
...
+
+```go
+type bgpAdvertisement struct {
+  Name              string
+  Pools             []string
+  PoolSelector      map[string]string
+  AggregationLength *int
+  LocalPref         *uint32
+  Communities       []string
+  NodeSelector      *metav1.LabelSelector
```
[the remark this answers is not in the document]
fedepaol: My bad, the peer selector was already in my mind (see the CR example above):
```
```yaml
kind: BGPAdvertisement
metadata:
  name: bgpadvertisement
  namespace: metallb-system
spec:
    pools:
      - pool1
      - pool2
    peers:
      - peer1
      - peer2
 ```

With your version, we won't be able to express the case where we want to advertise only some ips from a subset of the nodes.
That's because the BGP Peer must be unique (per peer) while we can have multiple bgpAdvertisements (and I notice now the peer part is missing from this struct, need to amend that).

File: design/pool-configuration.md (hunk 46 above)
Comment (fedepaol) [reply design/pool-configuration.md:492]: Fine, but not priority will still mean the lowest possible priority.

File: design/pool-configuration.md (hunk 39 above)
Comment (fedepaol) [reply design/pool-configuration.md:190]: Because avoidbuggyips default is false, so it's fine both if the user puts false or not specifies it.

File: design/pool-configuration.md (hunk 44 above)
Comment (fedepaol) [reply design/pool-configuration.md:485]: good catch

File: design/pool-configuration.md (hunk 41 above)
Comment (fedepaol) [reply design/pool-configuration.md:291]: Yep, the effort should not be tremendous.
Changing it to `corev1.SecretRef`

File: design/pool-configuration.md (hunk 49 above)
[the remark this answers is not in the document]
Comment (fedepaol) [reply design/pool-configuration.md:358]: Also, the config is simpler if the focus is advertising to different peers from different nodes.
The most common use case I can envision is to have a pool advertised to a subset of peers from a subset of nodes, which will be simpler with the current proposed version, compared to having to add a nodeselector to each peer.

File: design/pool-configuration.md (hunk 46 above)
Comment (johananl) [reply design/pool-configuration.md]: Yes, I guess that makes sense.

Also, I'm not saying reversing the logic is *necessarily* the way to go, I just shared **my** understanding of this. Would you rather go with "higher is better"?

Comment (fedepaol) [reply design/pool-configuration.md]: I don't have a strong opinion about that. I guess I was influenced by localpref which I dealt with just recently. As long as we document it, works for me.

File: design/pool-configuration.md [hunk 50]
```diff
@@ -0,0 +1,553 @@
...
+
+#### Regressions
+
+The e2e test suite is wide enough to ensure we have good coverage on metallb's main features. Given the new data model will be functionally equivalent
+to the old one, converting the test suite should give us enough confidence in that regard.
+
+#### Impact on the users' configurations
+
+Deprecating the configmap will have impact on the users' current configurations. To mitigate this, we may consider having a controller that converts the configmap
+into the CRDs, or provide some offline tool in charge of that (the latter option is more desirable, as it will remove the need of reconciling changes into the configmap).
```
rata, line 129: I guess this implies for at least for 1 release it will be possible to use CRDs or a configmap to configure metalLB (but not both), right?

fedepaol: An automatic CLI converter could also help and possibly ease the transition, that's another option.

ChipWolf:
May I request priorities on the L2Advertisement nodeselector?

design/pool-configuration.md
```diff
@@ -0,0 +1,553 @@
...
+      - key: kubernetes.io/hostname
+        operator: In
+        values: [hostA, hostB]
+```
+
+```yaml
+apiVersion: metallb.io/v1beta1
+kind: L2Advertisement
+metadata:
+  name: lsadvertisement
```
liornoy, line 390: nit: l2advertisement

design/pool-configuration.md
```diff
@@ -0,0 +1,553 @@
...
+}
+```
+
+Example CR:
+
+```yaml
+apiVersion: metallb.io/v1beta1
+kind: L2Advertisement
+metadata:
+  name: lsadvertisement
```
liornoy, line 266: nit: l2advertisement

design/pool-configuration.md
rata: Thanks!

adampl:
What about the issue of interface selection in the L2 mode? I don't see any mention of it in this proposal.
The issue was discussed here: #277

design/pool-configuration.md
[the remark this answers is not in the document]
fedepaol: As a second thought, I think we need the node selector BOTH on the bgp advertisement and on the bgp peer, because they express two different concepts.

The BGPPeer object expresses the desire to establish a BGP connection, the BGPAdvertisement expresses the desire to advertise a prefix via BGP. We need both because we may either want to connect to peers from a subset of nodes, and orthogonally we may want to advertise prefixes only from a subset of nodes.

The first one is on par with what we have right now, and even if we could get away with not having the selector in the bgppeer, that would result in wasted resources of metallb attempting to connect to a peer from a node that won't be able to reach it.

fedepaol:
@gclawes @johananl I aligned the design doc to the latest changes (IPPool and secret as extra field alternative to password).

ChipWolf:
Out of interest, with the nodes effectively selecting pools based on the highest priority, what if the priority of each pool differs for each node?
For example; you have a pool of French IPs and German IPs, your node is distributed between France and Germany, your nodes have the ability to use the IPs interchangeably thanks to your vendor's network mesh, but you'd prefer them to select the IPs from the same geo as the node.

fedepaol:
> Out of interest, with the nodes effectively selecting pools based on the highest priority, what if the priority of each pool differs for each node? For example; you have a pool of French IPs and German IPs, your node is distributed between France and Germany, your nodes have the ability to use the IPs interchangeably thanks to your vendor's network mesh, but you'd prefer them to select the IPs from the same geo as the node (where possible).

Not sure what you mean. The node selector will be in the announcement.
So, the announcement (let's say L2) will be pinned both to a pool and to a set of nodes. If there's an announcement matching the ippool and the node, the speaker will announce. There is no priority.

fedepaol:
> > @gclawes @johananl I aligned the design doc to the latest changes (IPPool and secret as extra field alternative to password).
> 
> Just a note (as discussed on the community meeting), the kubectl short-name `ipool` will conflict with a similarly-named resource in calico.

@gclawes after thinking about it, the ux for calico users will be not ideal. I'll rename ippool to `ipaddresspool`. We will still have some time til the next release to let it converge if a better name comes to our mind.

fedepaol:
Merging this.
Thanks to all those who took part to the discussion, the proposal was shaped in the right direction as a collective effort, and the quality is way better than the initial version.

PR #943, opened 2021-09-10
Extend BGP test to check routes and session state

fedepaol:
Currently, BGP test only verifies that the remote peer can hit the service. We don't have guarantee that all the ip is advertised from all the nodes and that the BGP sessions are established.

Here we add helper packages to check both, providing a base for adding more testcases.

The frr parsing utilities are in a package of their own, and may be useful if parsing the FRR output is needed to provide status feedback or metrics.

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -26,11 +26,14 @@ import (
 	"time"
 
 	"github.com/onsi/ginkgo"
+	"go.universe.tf/metallb/e2etest/pkg/executor"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
 	e2eservice "k8s.io/kubernetes/test/e2e/framework/service"
 )
+
+const frrContainer = "frr"
 
 var _ = ginkgo.Describe("BGP", func() {
 	f := framework.NewDefaultFramework("bgp")
@@ -73,14 +76,14 @@ var _ = ginkgo.Describe("BGP", func() {
 		hostport := net.JoinHostPort(ingressIP, port)
 		address := fmt.Sprintf("http://%s/", hostport)
 
-		useDocker := true
+		exc := executor.ForContainer(frrContainer)
 		if skipDockerCmd {
 			ginkgo.By(fmt.Sprintf("checking connectivity to %s", address))
-			useDocker = false
+			exc = executor.Host
 		} else {
 			ginkgo.By(fmt.Sprintf("checking connectivity to %s with docker", address))
 		}
-		err = wgetRetry(useDocker, address)
+		err = wgetRetry(address, exc)
 		framework.ExpectNoError(err)
 	})
 
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -27,6 +27,8 @@ import (
 
 	"github.com/onsi/ginkgo"
 	"go.universe.tf/metallb/e2etest/pkg/executor"
+	"go.universe.tf/metallb/e2etest/pkg/frr"
+	"go.universe.tf/metallb/e2etest/pkg/routes"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
@@ -83,8 +85,29 @@ var _ = ginkgo.Describe("BGP", func() {
 		} else {
 			ginkgo.By(fmt.Sprintf("checking connectivity to %s with docker", address))
 		}
+
 		err = wgetRetry(address, exc)
 		framework.ExpectNoError(err)
-	})
 
+		allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+		framework.ExpectNoError(err)
+
+		advertised := routes.ForIP(ingressIP, exc)
+		err = routes.MatchNodes(allNodes.Items, advertised)
+		framework.ExpectNoError(err)
+
+		neighbors, err := frr.NeighborsInfo(exc)
+		framework.ExpectNoError(err)
+
+		err = frr.NeighborsMatchNodes(allNodes.Items, neighbors)
+		framework.ExpectNoError(err)
+
+		frrRoutes, _, err := frr.Routes(exc)
+		framework.ExpectNoError(err)
+
+		routes, ok := frrRoutes[ingressIP]
+		framework.ExpectEqual(ok, true, ingressIP, "not found in frr routes")
+		err = frr.RoutesMatchNodes(allNodes.Items, routes)
+		framework.ExpectNoError(err)
+	})
 })
--- a/e2etest/layer2_test.go
+++ b/e2etest/layer2_test.go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/onsi/ginkgo"
+	"go.universe.tf/metallb/e2etest/pkg/executor"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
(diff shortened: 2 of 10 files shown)


fedepaol:
cc @markdgray @oribon  @sabinaaledort , please take a look

fedepaol:
(hopefully CI doesn't fail :D )

e2etest/pkg/frr/bgp.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+// TODO: Leaving this package "test unaware" on purpouse, since we may find it
+// useful for fetching informations from FRR (such as metrics) and we may need to move it
+// to metallb.
+
+// NeighborForContainer returns informations for the given neighbor in the given
+// container. If containerName is "", it queries localhost.
+func NeighborForContainer(neighborName, containerName string) (*Neighbor, error) {
+	res, err := container.RunCommand(containerName, "vtysh", "-c", fmt.Sprintf("show bgp neighbor %s", neighborName))
```
oribon, line 17: why not use 'show bgp neighbor %s json' and parse that instead of the regexes?

e2etest/pkg/frr/bgp.go
```diff
@@ -0,0 +1,40 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to parse neighbour %s for %s", neighborName, containerName)
+	}
+	return neighbor, nil
+}
+
+// NeighborForContainer returns informations for the all the neighbors in the given
+// container. If containerName is "", it queries localhost.
+func NeighborsForContainer(containerName string) ([]*Neighbor, error) {
+	res, err := container.RunCommand(containerName, "vtysh", "-c", "show bgp neighbor")
```
oribon, line 31: ^^same question

e2etest/pkg/frr/bgp.go
fedepaol: because I was not aware of such option! Having a machine friendly format is definetely better, will change the parsing mechanism.

e2etest/pkg/frr/bgp.go
```diff
@@ -0,0 +1,40 @@
+package frr
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+	"go.universe.tf/metallb/e2etest/pkg/container"
+)
+
+// TODO: Leaving this package "test unaware" on purpouse, since we may find it
```
markdgray, line 10: s/purpouse/purpose

Also, maybe we could change the package name to make it less general (i.e. "frr")?

markdgray: If we want to make it general, we could connect to `bgpd` over the tcp/ip socket and pass the command. In this way, we wouldn't need to call `vtysh`.

e2etest/pkg/frr/bgp.go
markdgray: Also, if this may be general, it might be good to start creating public(?) functions in order to hide how we connect to frr. Something like: FrrShowBGPNeighbor()

Also, it would be OK to implement this at a later stage.

e2etest/pkg/routes/routes.go
```diff
@@ -0,0 +1,80 @@
+package routes
```
markdgray, line 1: I guess a lot of this will get reworked based on @oribon 's good comment about the JSON format. Therefore, I will not review this file.

e2etest/pkg/frr/bgp.go
fedepaol: The package name IS `frr`, not sure you are suggesting a less general name than frr or you mis-interpreted the file name as the package name. If the former, I think frr is a good way to collect all the frr interaction utilities.

fedepaol: Re: the socket, I'd tackle this separately, when the conversation around it is settled and when we decide to move this part of the code elsewhere. It could be that we find only the parsing part interesting, or that Ori throws everything away.

markdgray: i.e. "frr" meant the frr package. Yes, I think it should be less general.

fedepaol: otherwise I can split and have `bgpneighbour.ForContainer` and such

markdgray: I guess it depends if you are going to eventually pull this into the main code base. Maybe its ok as long as the package is a sub-package of the test framework?

fedepaol: If / when we move this to the main codebase we may need to rethink the package structure. I think we'll be able to recycle only the parsing part leaving the "execute in a container" side here.
So if I got it right, we leave frr for now, and we think about a more focused name when we move under metallb (I bet there'll be an frr package somewhere anyhow :-) )

markdgray: Yeah sounds like a plan

e2etest/bgp_test.go
```diff
@@ -73,15 +75,36 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			containerName = ""
 		} else {
 			ginkgo.By("checking connectivity to its external VIP with docker")
 		}
-		err = wgetRetry(useDocker, address)
+
+		err = wgetRetry(containerName, address)
+		framework.ExpectNoError(err)
+
+		allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
```
rata, line 89: I guess the e2e test is run with a timeout and aborts if this stalls? Or do we want to specify a timeout in the context here?

e2etest/layer2_test.go
```diff
@@ -73,7 +73,7 @@ var _ = ginkgo.Describe("L2", func() {
 
 		hostport := net.JoinHostPort(ingressIP, port)
 		address := fmt.Sprintf("http://%s/", hostport)
-		err = wgetRetry(false, address)
+		err = wgetRetry("", address)
```
rata, line 76: Can we split the signature change on a different commit? For sure the next time will help me to review, but is simpler also to catch bugs if commits are atomic and do only one "logical" thing.

e2etest/pkg/container/container.go
```diff
@@ -0,0 +1,16 @@
+package container
+
+import "os/exec"
+
+// RunCommand executes the given command / args in a container when the container name
+// is specified, on the current host otherwise.
+func RunCommand(containerName, cmd string, args ...string) (string, error) {
```
rata, line 7: Would it make sense to put the logic to run a command in the host in a different function? Like
```
func RunHostCommand(cmd string) (string, error) {
                // Didn't test this, but something like this to get the params
                cmd_split := strings.Split(cmd, " ")
                cmd = cmd_split[0]
                args = cmd_split[1:]

                // Or maybe CommandContext and use a timeout?
		out, err := exec.Command(cmd, args).CombinedOutput()
		return string(out), err
}

func RunCommand(containerName, cmd string) (string, error) {
     return RunHostCommand("docker exec" + " " + containerName + " " + "cmd")
}
```

Did it on github so whitespaces and such are horrible, but to show the idea

Having a different function for RunHostCommand and reusing the code seems less hacky than using containerName="" IMHO. Also, we can receive a string and the user doesn't need to split the commands perfectly so maybe it is slightly simpler to use?

What do you think?

rata (commented):
Partial review only, will continue in the coming days

e2etest/bgp_test.go
fedepaol: Yes, moreover it's kind local to the test runner. I'd avoid bloating the tests with timed contexts.

e2etest/layer2_test.go
fedepaol: Will do!

e2etest/pkg/container/container.go
fedepaol: This did not feel clean from the beginning, but the issue is that we are using these as base for executing commands in the container, wrapped by other calls (you may see it in the part you did not review yet). Moreover, we want to maintain the flexibility of running the same tests in a non dockerized environment.

I can't use an optional global container name either, because at some point we'll need to have multiple frr containers mimicking multiple bgp peers.
In any case, I'll try to rework this to see if I can make it slightly better.

fedepaol: Done via creating an `Executor` interface that represent what is gonna execute the actions we want to apply.
Feels less hacky than if-ing on the containerName while preserving the same degree of flexibility.

e2etest/layer2_test.go
fedepaol: Done

fedepaol:
> @fedepaol this one has a conflict now

Rebased, thanks!

PR #944, opened 2021-09-10
Extend the frr enhancement proposal with the list of tests

fedepaol:
Sending this as discussed with @markdgray . 
The enhancement proposal mentions extending the current e2e test suite.
Here, we provide a list of the coverage needed to ensure compatibility between the old implementation and the new implementation.

This is meant to be a starting point for defining the list of our tests. I think it's a good start, and meant to start a discussion to understand if new scenarios need to be added.

### DIFF
--- a/design/0001-frr.md
+++ b/design/0001-frr.md
@@ -310,3 +310,60 @@ releases is proposed:
    MetalLB is deprecated from this point onwards. FRR is no longer
    limitied by feature parity with the current native implementation of BGP and
    additional functionality may be developed.
+
+## Test Coverage
+
+The current e2e tests set up a bgp peer to all the nodes inside an FRR container, and try to hit the service.
+
+In order to guarantee parity between the two implementations, the coverage of the e2e tests must be
+extended.
+
+### Test strategy
+
+Every time we test an exposed service, we must verify that:
+
+* the service is reacheable from outside
+* the nodes advertising the service ip are the expected ones
+* all the nodes are paired with the expected bgp peers
+
+#### Changing configuration
+
+Every time the configuration is changed before testing the new scenario, we
+may need to wait an arbitrary time or to run the checks in an eventually loop,
+since there is no feedback of what configuration is being consumed.
+
+Another possible option is to remove the configuration completely and
+verify that hitting the service does not work anymore. This should ensure that
+it's possible to apply a new configuration.
+
+#### Coverage
+
+We must cover the following scenarios:
+
+* Cluster Traffic Policy
+* Local Traffic Policy
+* One Peer
+* Multiple Peers
+* Configuration Change (adding / removing new pools, adding / removing bgp peers)
+* Failover: killing one node and see that we are still able to reach the service
+* Different types of ranges (i.e. /24 CIDRs, /32 ranges)
+
+#### BGP Parameters
+
+We need to ensure that the [BGP parameters](https://github.com/metallb/metallb/blob/9126567db61017c15f732412bd04265a2c4a5031/manifests/example-config.yaml#L1) are correctly
+received by the peer.
+
+The following parameters must be covered:
+
+* node selector
+* source address
+* router-id
+* password (this may require another instance of frr / changing its configuration)
+* communities
+* local preferences
+* aggregation-length
+
+#### Metrics
+
+The current set of metrics must be covered by tests, in order to ensure that the
+new implementation does not regress from that perspective.


design/0001-frr.md
```diff
@@ -310,3 +310,58 @@ releases is proposed:
...
+#### Coverage
+
+We must cover the following scenarios:
+
+* Cluster Traffic Policy
+* Local Traffic Policy
+* One Peer
+* Multiple Peers
+* Configuration Change (adding / removing new pools, adding / removing bgp peers)
+* Failover: killing one node and see that we are still able to reach the service
```
markdgray, line 348: We should probably check that keepalives are received?

design/0001-frr.md
```diff
@@ -310,3 +310,58 @@ releases is proposed:
...
+
+#### Coverage
+
+We must cover the following scenarios:
+
+* Cluster Traffic Policy
+* Local Traffic Policy
+* One Peer
+* Multiple Peers
+* Configuration Change (adding / removing new pools, adding / removing bgp peers)
```
markdgray, line 347: Might be good to test publishing of /24s and /32s.

design/0001-frr.md
```diff
@@ -310,3 +310,58 @@ releases is proposed:
...
+* Failover: killing one node and see that we are still able to reach the service
+
```
markdgray, line 349: Should we check non-functional things like failover time or time for changes to propagate? I feel that this may be out of scope but it would be good to get your opinion?

design/0001-frr.md
```diff
@@ -310,3 +310,58 @@ releases is proposed:
...
+* source address
+* router-id
+* password (this may require another instance of frr / changing its configuration)
+* communities
+* local preferences
+* aggregation-length
+
+#### Metrics
+
+We need to verify the BGP and match the metrics available.
```
markdgray, line 367: You mean we need to match the metrics that are currently exposed by the native implementation?

design/0001-frr.md
```diff
@@ -310,3 +310,58 @@ releases is proposed:
...
+In order to guarantee parity between the two implementations, the coverage of the e2e tests must be
+extended.
+
+### Test strategy
+
+Every time we test an exposed service, we must verify that:
+
+* the service is reacheable from outside
+* the nodes advertising the service ip are the expected ones
+* all the nodes are paired with the expected bgp peers
```
markdgray, line 327: I guess there are some things we can read from the peer to ensure the connection is as expected. One thing for example is the connection state but maybe there are others? We currently don't do that but it could be useful.

design/0001-frr.md
fedepaol: Isn't checking the state of the peers / routes enough to verify that the new state is adjusted?

design/0001-frr.md
fedepaol: Yes, the connection state is already part of my PR https://github.com/metallb/metallb/pull/944 , plus there I check that routes are set as we expect both in FRR and inside the container.

design/0001-frr.md
fedepaol: Yes, I'll reword that to make it clearer.

design/0001-frr.md
fedepaol: I am a bit torn about this. 
I'd say, let's start with functional tests assuming that is fair to wait seconds for a config to be in place because we don't have feedback via API to understand if the configuration was applied or not.

design/0001-frr.md
fedepaol: Will add that

design/0001-frr.md
markdgray: Yes, I started reviewing your PR after this. Good idea!

design/0001-frr.md
markdgray: Yes, you are probably right.

rata (approved):
LGTM, ping me when you finish the changes you wanted to do (https://github.com/metallb/metallb/pull/944/files#r708108762).

IMO, it is enough for the tests be the authoritative source of what is being tested. Such a description can be in the PR description actually adding the tests. But if you prefer to add it to this doc before, I'm fine too :)

markdgray:
Thanks for doing this. Now we have some agreement on what we should be testing and as @rata stated, the tests will be the authoritative source.

fedepaol:
> IMO, it is enough for the tests be the authoritative source of what is being tested. Such a description can be in the PR description actually adding the tests. But if you prefer to add it to this doc before, I'm fine too :)

I agree, this was meant to be both a list of what we need to implement, and a discussion on what we may have missed.

fedepaol:
> LGTM, ping me when you finish the changes you wanted to do (https://github.com/metallb/metallb/pull/944/files#r708108762).
> 
> IMO, it is enough for the tests be the authoritative source of what is being tested. Such a description can be in the PR description actually adding the tests. But if you prefer to add it to this doc before, I'm fine too :)

Finished with the changes

design/0001-frr.md
```diff
@@ -310,3 +310,60 @@ releases is proposed:
...
+may need to wait an arbitrary time or to run the checks in an eventually loop,
+since there is no feedback of what configuration is being consumed.
+
+Another possible option is to remove the configuration completely and
+verify that hitting the service does not work anymore. This should ensure that
+it's possible to apply a new configuration.
+
+#### Coverage
+
+We must cover the following scenarios:
```
cgoncalves, line 341: Do we also need to add test coverage for IP address sharing?
https://metallb.universe.tf/usage/#ip-address-sharing

cgoncalves (commented):
A great addition to the design document! I particularly like the fact you're calling out the need to test cover many of the existing parameters like communities, localPref, etc.

design/0001-frr.md
fedepaol: I'd say no, simply because the allocation side of things is not impacted by switching bgp implementation.

Not saying it should not be tested, but here we are aiming at the set of tests that will limit (don't want to say avoid :P ) the regressions between the two implementations.

russellb (approved):
lgtm, only commented on one very minor typo that I'll fix myself before merging

PR #948, opened 2021-09-13
Add AddressPool validation check for invalid ranges where first address of the range is after the second

sabinaaledort:
An AddressPool where the first address of the range is after the second
should be an invalid range. In such case the LoadBalancer service is
unable to get an IP.

The service is not getting an ip with the following error:
`{"caller":"level.go:63","error":"no available IPs","level":"error","msg":"IP allocation failed","op":"allocateIP","service":"default/nginx","ts":"2021-09-13T08:40:02.487727369Z"}`

```
$ kubectl get service -A
NAMESPACE     NAME         TYPE           CLUSTER-IP      EXTERNAL-IP   PORT(S)                  AGE
default       nginx        LoadBalancer   10.96.240.166   <pending>     80:32319/TCP             53s
```

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -15,6 +15,7 @@
 package config // import "go.universe.tf/metallb/internal/config"
 
 import (
+	"bytes"
 	"errors"
 	"fmt"
 	"net"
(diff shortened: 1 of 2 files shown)


fedepaol:
Looks good to me

russellb:
This looks like a good candidate for a simple unit test.  Thanks!

rata:
@sabinaaledort thanks! It mostly LGTM, but can you fix the CI issue (from [here][link] it seems the issue is that "body must have leading blank line [body-leading-blank]")? A unit test like russell mentioned would be nice


[link]: https://github.com/metallb/metallb/pull/948/checks?check_run_id=3614544299

russellb:
Thank you for adding the test!

PR #951, opened 2021-09-14
Make single go.mod / go.sum files.

fedepaol:
Having a separate go module for e2e tests and for metallb makes it
harder to share the code between the two. Here we unify the go.mod file
and we pin k8s dependencies.


markdgray:
Looks good to me.

PR #957, opened 2021-09-17
Add ipv6 / dual stack lanes

fedepaol:
Here we add an option to specify the ipfamily of the kind cluster, we use it for CI and we adjust the tests / configuration to run against ipv6 only clusters.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -23,26 +23,22 @@ jobs:
       - setup_ci_image
       - run: inv test
       - run: cp manifests/metallb.yaml manifests/metallb.yaml.prev
-  e2etest-bgp:
+  e2etest:
+    parameters:
+      ip-family:
+        type: string
+      protocol:
+        type: string
     machine:
       image: ubuntu-2004:202104-01
     steps:
       - setup_ci_image
       - setup_kind
       - run: inv build
-      - run: inv dev-env -p bgp
-      - run: inv e2etest -e /tmp/kind_logs
-      - store_artifacts:
-          path: /tmp/kind_logs
-  e2etest-layer2:
-    machine:
-      image: ubuntu-2004:202104-01
-    steps:
-      - setup_ci_image
-      - setup_kind
-      - run: inv build
-      - run: inv dev-env -p layer2
-      - run: inv e2etest -e /tmp/kind_logs
+      - run: inv dev-env -p << parameters.protocol >> -i  << parameters.ip-family >>
+      - run: |
+          if [ "<< parameters.protocol >>" = "layer2" ]; then FOCUS=L2; else FOCUS=BGP; fi
+          inv e2etest -f $FOCUS -e /tmp/kind_logs
       - store_artifacts:
           path: /tmp/kind_logs
   lint-1-16:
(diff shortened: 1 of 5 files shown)


cgoncalves:
LGTM. Pretty neat the matrix option!

tasks.py
```diff
@@ -233,7 +233,7 @@ def dev_env(ctx, architecture="amd64", name="kind", cni=None, protocol=None, nod
         if cni:
             networking_config["disableDefaultCNI"] = True
         if ip_family != "ipv4":
-            networking_config["disableDefaultCNI"] = ip_family
+            networking_config["ipFamily"] = ip_family
```
markdgray, line 234: ah ok. This confused me in the previous commit. Not sure if you intend to squash commits but could you add this change to the first commit.

tasks.py
```diff
@@ -199,7 +199,7 @@ def validate_kind_version():
     "protocol": "Pre-configure MetalLB with the specified protocol. "
                 "Unconfigured by default. Supported: 'bgp','layer2'",
     "node_img": "Optional node image to use for the kind cluster (e.g. kindest/node:v1.18.19)."
-                "The node image drives the kubernetes version used in kind."
+                "The node image drives the kubernetes version used in kind.",
```
markdgray, line 202: This probably belongs in the previous commit?

tasks.py
```diff
@@ -600,20 +605,8 @@ def e2etest(ctx, name="kind", export=None, kubeconfig=None, system_namespaces="k
...
-    if metallb_config['address-pools'][0]['name'] == "dev-env-layer2":
-        run("cd `git rev-parse --show-toplevel`/e2etest &&"
-            "go test --provider=local -ginkgo.focus=L2 --kubeconfig={} -service-pod-port={}".format(kubeconfig, service_pod_port))
-    elif metallb_config['address-pools'][0]['name'] == "dev-env-bgp":
-        run("cd `git rev-parse --show-toplevel`/e2etest &&"
-            "go test --provider=local -ginkgo.focus=BGP --kubeconfig={} -service-pod-port={} {}".format(kubeconfig, service_pod_port, opt_skip_docker))
-    else:
-        print("dev-env environment not configured correctly. Try running `inv dev-env -p <layer2/bgp>`")
+    run("cd `git rev-parse --show-toplevel`/e2etest &&"
+            "go test {} --provider=local --kubeconfig={} --service-pod-port={} {}".format(ginkgo_focus, kubeconfig, service_pod_port, opt_skip_docker))
```
markdgray, line 609: I'm a bit on the fence with this change but I guess as we scale out the tests it will make it easier to test the various permutations.

.circleci/config.yml
```diff
@@ -23,26 +23,22 @@ jobs:
...
-      image: ubuntu-2004:202104-01
-    steps:
-      - setup_ci_image
-      - setup_kind
-      - run: inv build
-      - run: inv dev-env -p layer2
-      - run: inv e2etest -e /tmp/kind_logs
+      - run: inv dev-env -p << parameters.protocol >> -i  << parameters.ip-family >>
+      - run: |
+          if [ "<< parameters.protocol >>" = "layer2" ]; then FOCUS=L2; else FOCUS=BGP; fi
```
markdgray, line 40: to avoid this `if` could we just match the `FOCUS` names to the values in the matrix below?

fedepaol: That was the original implementation.
The way exclude works however, is that it requires an item for each permutation to exclude.
I can't say 
```
            exclude:
              - protocol: bgp
                ip-family: ipv6
```
I would also need to expand with different FOCUS-es which was as much as ugly. Also, those two parameters would always be coupled, so I think it's better to have only one saying which mode we are deploying / testing.

On top of that, this part is gonna disappear when https://github.com/metallb/metallb/pull/931 lands

tasks.py
fedepaol: What's that makes you not liking it?

tasks.py
fedepaol: Yes, my bad. I messed up when trying to split the commits.
Same for the comment below

russellb:
lgtm, and I believe past comments were addressed or we could address in follow-ups.  thanks!

PR #958, opened 2021-09-17
FRR: Implementation of BGP interface for FRR

markdgray:
This pull request is an initial attempt at implementing the MetalLB BGP interface for FRR.

It is not complete as:
* does not handle all BGP configurations (e.g. communities)
* does not integrate with FRR itself (only generated configuration) as this requires https://github.com/metallb/metallb/pull/935
* the translation between MetalLB configuration and FRR configuration may not be completely accurate. This will require further integration with FRR to confirm.


speaker/bgp_controller.go
```diff
@@ -295,9 +296,11 @@ func (c *bgpController) SetNode(l log.Logger, node *v1.Node) error {
 }
 
 var newBGP = func(logger log.Logger, addr string, srcAddr net.IP, myASN uint32, routerID net.IP, asn uint32, hold time.Duration, password string, myNode string, bgpType string) (session, error) {
+	level.Info(logger).Log("op", "startup", "msg", "Starting BGP session", "type", bgpType)
 	if bgpType == "native" {
-		level.Info(logger).Log("op", "startup", "msg", "Starting BGP session", "type", bgpType)
-		return native.New(logger, addr, srcAddr, myASN, routerID, asn, hold, password, myNode)
+		return bgpnative.New(logger, addr, srcAddr, myASN, routerID, asn, hold, password, myNode)
```
markdgray, line 301: @msherif1234 Had a comment about consistency of the naming convention for ASNs. I will address that in a later commit.

internal/bgp/bgpfrr/config.go
```diff
@@ -0,0 +1,127 @@
...
+{{range $n := .Neighbors -}}
+{{range .Advertisements }}
+  address-family {{.Version}} unicast
+    neighbor {{$n.Addr}} activate
+    network {{.Prefix}}
+  exit-address-family
+{{- end}}
+{{end}}
+{{end}}
+`
```
msherif1234, line 35: this good start but I would to see further research on how best to represent BGP configs in FRR terms, the other day I was checking there peer groups , route-map, bgp list and the list just goes on

internal/bgp/bgpfrr/config.go
```diff
@@ -0,0 +1,127 @@
...
+
+type advertisementConfig struct {
+	Version string
+	Prefix  string
+}
+
+// routerName() defines the format of the key of the "Routers" map in the
+// frrConfig struct.
+func routerName(srcAddr string, myAsn uint32) string {
+	return fmt.Sprintf("%d@%s", myAsn, srcAddr)
```
msherif1234, line 63: you won't have more than 1 myASN/speaker

internal/bgp/bgpfrr/frr.go
```diff
@@ -0,0 +1,150 @@
...
+	for _, session := range s.sessions {
+		var router *routerConfig
+		var neighbor *neighborConfig
+		var exist bool
+
+		routerName := routerName(session.srcAddr.String(), session.asn)
+		if router, exist = config.Routers[routerName]; !exist {
+			router = &routerConfig{
+				Asn:       session.asn,
+				RouterId:  session.srcAddr.String(),
```
msherif1234, line 123: RouterId and srcAddr are two different things

internal/bgp/bgpfrr/frr.go
```diff
@@ -0,0 +1,150 @@
...
+		Loglevel: "", // TODO.
+		Routers:  make(map[string]*routerConfig),
+	}
+
+	for _, session := range s.sessions {
+		var router *routerConfig
+		var neighbor *neighborConfig
+		var exist bool
+
+		routerName := routerName(session.srcAddr.String(), session.asn)
```
msherif1234, line 119: This configuration is one time config for BGP speaker it shouldn't change for all sessions and if it does this will be an error

internal/bgp/bgpfrr/frr.go
```diff
@@ -0,0 +1,150 @@
...
+				Advertisements: make(map[string]*advertisementConfig),
+			}
+			router.Neighbors[neighborName] = neighbor
+		}
+
+		for _, adv := range session.advertised {
+			if _, exist := neighbor.Advertisements[adv.Prefix.String()]; !exist {
+				neighbor.Advertisements[adv.Prefix.String()] = &advertisementConfig{
+					Version: "ipv4",
+					Prefix:  adv.Prefix.String(),
```
msherif1234, line 143: do u need to look at prefix to findout the AF instead of hardcoding IPv4 ?

internal/bgp/bgpfrr/config.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// neighborName() defines the format of key of the 'Neighbors' map in the
+// routerConfig struct.
+func neighborName(peerAddr string, peerAsn uint32) string {
+	return fmt.Sprintf("%d@%s", peerAsn, peerAddr)
+}
```
msherif1234, line 70: peerAddr should be unique enough I don't think its normal/possible to have same neighbor defined in the same speaker if its for different remote-as ?

internal/bgp/bgpfrr/config.go
```diff
@@ -0,0 +1,127 @@
...
+type frrConfig struct {
+	Loglevel string
+	Hostname string
+	Routers  map[string]*routerConfig
+}
+
+type routerConfig struct {
+	Asn       uint32
+	RouterId  string
+	Neighbors map[string]*neighborConfig
```
msherif1234, line 46: I am not sure if this make sense to be map I think slice would have been better here what do u think ?

internal/bgp/bgpfrr/config.go
```diff
@@ -0,0 +1,127 @@
...
+  exit-address-family
+{{- end}}
+{{end}}
+{{end}}
+`
+
+type frrConfig struct {
+	Loglevel string
+	Hostname string
+	Routers  map[string]*routerConfig
```
msherif1234, line 40: this just one router/speaker I don't think it needs to be a map

internal/bgp/bgpfrr/config.go
```diff
@@ -0,0 +1,127 @@
...
+type routerConfig struct {
+	Asn       uint32
+	RouterId  string
+	Neighbors map[string]*neighborConfig
+}
+
+type neighborConfig struct {
+	Asn            uint32
+	Addr           string
+	Advertisements map[string]*advertisementConfig
```
msherif1234, line 52: same here I think this is might be better off slice instead of map

internal/bgp/bgpfrr/frr.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+// New() creates a BGP session using the given session parameters.
+//
+// The session will immediately try to connect and synchronize its
+// local state with the peer.
+func New(l log.Logger, addr string, srcAddr net.IP, asn uint32, routerID net.IP, peerASN uint32, holdTime time.Duration, password string, myNode string) (*Session, error) {
+	session := &Session{
+		asn:        asn,
+		routerID:   routerID.To4(),
```
msherif1234, line 80: routerID and myNode aren't/session field

internal/bgp/bgpfrr/frr.go
fedepaol: +1

internal/bgp/bgpfrr/frr_test.go
```diff
@@ -0,0 +1,333 @@
...
+	"os/exec"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"go.universe.tf/metallb/internal/bgp"
+)
+
+const TESTDATA = "testdata/"
```
fedepaol, line 16: nit: const doesn't need to be uppercase

internal/bgp/bgpfrr/frr_test.go
```diff
@@ -0,0 +1,333 @@
...
+	testSetup(t)
+
+	l := log.NewNopLogger()
+	session, err := New(l, "10.2.2.254", net.IPv4(10, 1, 1, 254), 100, net.IPv4(10, 1, 1, 254), 200, time.Second, "password", "hostname")
+	if err != nil {
+		t.Fatalf("Could not create session: %s", err)
+	}
+
+	testCheckConfigFile(t)
+	session.Close()
```
fedepaol, line 71: Can you (here and in the next tests) `defer session.Close()`

internal/bgp/bgpfrr/frr_test.go
```diff
@@ -0,0 +1,333 @@
...
+}
+
+func TestSingleAdvertisementInvalidPrefix(t *testing.T) {
+	testSetup(t)
+
+	l := log.NewNopLogger()
+	session, err := New(l, "10.2.2.254", net.IPv4(10, 1, 1, 254), 100, net.IPv4(10, 1, 1, 254), 200, time.Second, "password", "hostname")
```
fedepaol, line 172: You can use `net.ParseIP`

internal/bgp/bgp.go
```diff
@@ -506,212 +33,16 @@ func (a *Advertisement) Equal(b *Advertisement) bool {
...
-	err = unix.Connect(fd, ra)
-
-	switch err {
-	case syscall.EINPROGRESS, syscall.EALREADY, syscall.EINTR:
-	case nil:
-		return net.FileConn(fi)
-	default:
-		return nil, os.NewSyscallError("connect", err)
+func Validate(adv *Advertisement) error {
+	if adv.Prefix.IP.To4() == nil {
```
fedepaol, line 37: Does it make sense to relax this constraint?

internal/bgp/bgpfrr/config.go
msherif1234: no each speak has one myASN and as far as I know we allow only one speaker/node atm

internal/bgp/bgpfrr/frr.go
msherif1234: I need to check I don't think its needs to be more than 32bits

internal/bgp/bgpfrr/config.go
fedepaol: But `myASN` is a property of the peer object, not of the speaker, so I _think_ each speaker advertises the same myAsn to the peer, but may have different `myAsn`s if multiple peers are configured. May be wrong here

internal/bgp/bgpfrr/frr.go
fedepaol: Yep what I meant is, if routerID is an ipv6 address To4() will return nil.

internal/bgp/bgpfrr/config.go
markdgray: This is a great conversation because it is a little confusing and its good to see I am not the only one confused!

From what I understand, a speaker can peer with multiple peers using different ASNs. i.e. what @fedepaol stated. What I could maybe do differently is use 'myASN' as a key in 'Routers'. I think it is ok to do that as we will not establish two sessions with the same 'myASN' but different 'routerID's? I *think* that is correct?

internal/bgp/bgpfrr/config.go
markdgray: The reason I used maps was that it was a convenient way to remove duplicates. 

For example, two calls to the API New() can create two sessions with the same router but different neighbors that should produce a valid FRR config file something like this: 

```
router bgp 100
  no bgp default ipv4-unicast
  bgp router-id 10.1.1.254

  neighbor 10.2.2.254 remote-as 200
  neighbor 10.4.4.254 remote-as 400
```
Alternatively, perhaps it is possible to structure the FRR config file like:

```
router bgp 100
  no bgp default ipv4-unicast
  bgp router-id 10.1.1.254

  neighbor 10.4.4.254 remote-as 400

router bgp 100
  no bgp default ipv4-unicast
  bgp router-id 10.1.1.254

  neighbor 10.2.2.254 remote-as 200

```
Or, alternatively, I could use a slice to iterate through the routers and look for a duplicate but I think the map will perform better. If you feel strongly about it, I can change it. Alternatively, we can change it in the future if it causes a problem?

internal/bgp/bgpfrr/config.go
markdgray: I'll resolve this comment as it is a duplicate of the discussion above.

internal/bgp/bgpfrr/config.go
markdgray: For this, I may be able to use a slice? Do you have a reason for the preference?

internal/bgp/bgpfrr/config.go
markdgray: I think it is possible (at least in FRR). See https://docs.frrouting.org/en/latest/bgp.html#clicmd-router-bgp-ASN-vrf-VRFNAME

internal/bgp/bgpfrr/config.go
```diff
@@ -0,0 +1,127 @@
...
+	var b bytes.Buffer
+	err = t.Execute(&b, data)
+
+	return b.String(), err
+}
+
+// buildConfigFile takes the FRR configuration structure 'f' and attempts to
+// build an equivalent FRR configuration file as a string.
+func buildConfigFile(f *frrConfig) (string, error) {
+	return templateConfig(f)
```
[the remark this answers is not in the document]
markdgray: Maybe no reason at all. I probably had other stuff in there in a previous iteration. I will look at refactoring it.

internal/bgp/bgpfrr/frr.go
```diff
@@ -0,0 +1,150 @@
...
+var state *frrState
+
+func init() {
+	state = &frrState{
+		sessions: map[string]*Session{},
+	}
+}
+
+type Session struct {
+	asn        uint32
```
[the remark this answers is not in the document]
markdgray: I have a general comment about this elsewhere so I will close this as a duplicate. I want to go through all the MetalLB speaker code because I *think* I saw different conventions in different places. I need to confirm. However, I think the `MyASN` `ASN` convention is the best.

internal/bgp/bgpfrr/frr.go
```diff
@@ -0,0 +1,150 @@
...
+	state.addSession(session)
+	state.regenerate()
+
+	err := configFRR()
+
+	return session, err
+}
+
+func (s *frrState) addSession(session *Session) {
+	s.sessions[sessionName(session.srcAddr.String(), session.asn, session.addr, session.peerASN)] = session
```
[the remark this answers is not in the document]
markdgray: I did have a mutex in a previous revision and removed it as I didn't think there was a possibility of any contention. However, I will double-check.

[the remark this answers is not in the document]
markdgray: Regarding the wrapper discussion, it's probably not needed anymore. I will look at removing.

internal/bgp/bgpfrr/frr.go
markdgray: Yeah that was a hack as we don't support IPv6 but I may as well fix it now. Thanks for catching it.

internal/bgp/bgpfrr/frr_test.go
markdgray: doesn't need to be lower case either :) I can change it if you want. Is there a convention?

internal/bgp/bgpfrr/config.go
markdgray: Thanks for the review @msherif1234. Yes, I assume there is a more optimal way to do this translation but I am not an expert on FRR so I don't know what it is. Even if this is non-optimal, it should server as a starting point and as we integrate into MetalLB and develop tests, we should be able to improve it.

I did consider netsted-templates while I was developing this but didn't think they were warranted at this stage. However, I will have another look and see if they will simplify things for us now.

internal/bgp/bgpfrr/frr.go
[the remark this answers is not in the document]
fedepaol: I think you are right, all the events eventually converge to the same queue and they are handled sequentially.
Before leaving the comment I checked the current implementation and they are mutexed. 
Anyhow. not sure if we want to be on the safe side here.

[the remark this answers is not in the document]
markdgray: @fedepaol https://github.com/metallb/metallb/blob/543fcf6fa4eb771d55e0da05e120a4221b623b99/internal/k8s/k8s.go#L417 seems to guarantee single-threaded operation. `key` is popped from the queue and each one is handled synchronously. Also the controller `bgp_controller.go` that handles these objects doesn't create any go routines. I think generally controllers are single threaded?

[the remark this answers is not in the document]
markdgray: I think I will keep the wrapper for now in these case as it add an additional layer of abstraction which may be useful and I think it makes the code a little clearer.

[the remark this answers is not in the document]
fedepaol: Can you then split the session name generation and the assignement?

internal/bgp/bgpfrr/frr_test.go
markdgray: That's good enough for me.

internal/bgp/bgpfrr/config.go
msherif1234: yeah under different VRF sure but we don't have VRF support at all , correct ?

internal/bgp/bgpfrr/config.go
msherif1234: myASN represent '1' this in FRR config
```
router bgp 1
 no bgp default ipv4-unicast
 bgp router-id 10.10.10.10
 neighbor 10.0.0.1 remote-as 64000
 neighbor 10::1 remote-as 20000
 neighbour 10.0.0.1 update-source 20.0.0.2

 address-family ipv4 unicast
  neighbor 10.0.0.1 route-map RMAP in
  neighbor 10.0.0.1 route-map RMAP out
  network 172.18.0.100/24
 exit-address-family

 address-family ipv6 unicast
  neighbor 10::1 activate
  network 1001:1::1/64
 exit-address-family
```

and what I am sayin we can't have today
router bgp 1
router bgp 2
router bgp 3
that means 3 BGP speakers running on one node, remote-as ofcourse you can have as many as you want of them/session it could be same like myASN iBGP session or different eBGP session, same router-id is one/VRF since we are just working with default VRF here then we just need one router-id

internal/bgp/bgpfrr/config.go
msherif1234: it will be great if we have main template and that can invoke per AF template and the AF template invoke peer template what do you think ?

internal/bgp/bgpfrr/config.go
fedepaol: If we are not accessing it by key but only iterating (and I think we are) I'd second @msherif1234 .
More than unnecessary hashing, it removes unnecessary complexity when reading the code.

internal/bgp/bgpfrr/frr.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func (s *frrState) deleteSession(name string) {
+	delete(s.sessions, name)
+}
+
+func (s *frrState) regenerate() {
```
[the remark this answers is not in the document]
markdgray: Maybe createConfig() as it will return an frrConfig object.

internal/bgp/bgpfrr/frr_test.go
markdgray: You mean instead of `net.IPv4`? Ok, I can do that. I think I just copied this pattern from somewhere else in MetalLB.

internal/bgp/bgpfrr/frr.go
```diff
@@ -0,0 +1,150 @@
...
+		addr:       addr,
+		srcAddr:    srcAddr,
+		peerASN:    peerASN,
+		holdTime:   holdTime,
+		logger:     log.With(l, "peer", addr, "localASN", asn, "peerASN", peerASN),
+		password:   password,
+		advertised: map[string]*bgp.Advertisement{},
+	}
+
+	state.addSession(session)
```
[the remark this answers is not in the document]
fedepaol: At this point we should focus on what's better in terms of readability / safety and not for speed / performance until we understand if that's really too slow / not acceptable.

Said that, what bothers me is a potential drift between the content of the map and what's handled outside of the package (i.e. we remove the element from the map but for some reason teh consumer is still handling the Session object). I think adding the check I am proposing in https://github.com/metallb/metallb/pull/958/files/7385507445eca643420e38cad71d598e3946dfcb#r711956461 should take care of that.

speaker/main.go
```diff
@@ -76,6 +76,14 @@ func main() {
 	)
 	flag.Parse()
 
+	/* Note: Changing the MetalLB BGP implementation type should be considered
+	 *       experimental.
+	 */
+	bgpType, present := os.LookupEnv("METALLB_BGP_TYPE")
```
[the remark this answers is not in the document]
markdgray: how about a panic in the switch statement

```
	switch bgpType {
	case bgpNativeImpl:
		return bgpnative.New(logger, addr, srcAddr, myASN, routerID, asn, hold, password, myNode)
	case bgpFRRImpl:
		return bgpfrr.New(logger, addr, srcAddr, myASN, routerID, asn, hold, password, myNode)
	default:
		panic(fmt.Sprintf("unsupported BGP implementation type: %s", bgpType))
	}
```

It's unrecoverable and will be caught at initialization so it should be ok?

internal/bgp/bgpfrr/frr_test.go
fedepaol: Seems more readable than thinking about net.IPV4 being a constructor

speaker/main.go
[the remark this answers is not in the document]
fedepaol: Sounds good

internal/bgp/bgpfrr/frr.go
markdgray: Yes. Good catch. I will update there and in the name generation for the maps.

internal/bgp/bgpfrr/frr.go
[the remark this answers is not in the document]
markdgray: Its not just that and I do appreciate your concern about early optimization. This is slightly different behaviour to the native implementation which (at least via that API) allows creation of a session without advertisement. Let me have a think about this as I want to make sure it won't cause some corner case behaviour.

internal/bgp/bgpfrr/config.go
markdgray: Yeah, I think that is fair for this one. For the others, I can remove the map if you think it would be better but it would mean iterating over lists to search for duplicates. wdyt?

internal/bgp/bgpfrr/config.go
markdgray: There are a few comments here which are basically discussing requirements. I am unsure what the requirements are and, IMHO, I think they should be defined by our BGP tests. Can I propose that I resolve these comments in this peer and I will raise them as comments in the PR defining BGP test cases? @msherif1234 @fedepaol what do you think?

internal/bgp/bgpfrr/config.go
markdgray: See previous discussion about this being a requirement discussion.

internal/bgp/bgpfrr/config.go
fedepaol: My opinion is that if we access the map by key (as you said, for finding duplicates), then it should remain a map

internal/bgp/bgpfrr/config.go
fedepaol: Fine by me. The api allows it. Let's add a test covering this scenario so we both validate our assumption and we match the current behaviour.

speaker/bgp_controller.go
```diff
@@ -33,6 +35,11 @@ import (
 	"github.com/go-kit/kit/log/level"
 )
 
+const (
+	bgpNativeImpl = "native"
```
fedepaol, line 39: Can we make this a type?

internal/bgp/bgpfrr/frr.go
```diff
@@ -0,0 +1,178 @@
...
+		return fmt.Errorf("unable to validate IP address")
+	}
+
+	if len(adv.Communities) > 63 {
+		return fmt.Errorf("max supported communities is 63, got %d", len(adv.Communities))
+	}
+	return nil
+}
+
+func (s *Session) Set(advs ...*bgp.Advertisement) error {
```
fedepaol, line 64: Another thing to note is, we are probably calling this (and New) multiple times whenever we receive a new configuration / a loadbalancer event (https://github.com/metallb/metallb/blob/717fd1dfae2852ce95f6e05c580d177d0d44c35a/speaker/bgp_controller.go#L258)
Would it make sense to extend the interface with a `commit` sort of mechanism?
I know this is falling in the land of early optimization, but the file will blink (at least) one time per peer.

internal/bgp/bgpfrr/frr.go
```diff
@@ -0,0 +1,178 @@
...
+
+func init() {
+	state = &frrState{
+		sessions: map[string]*Session{},
+	}
+}
+
+type Session struct {
+	myASN      uint32
+	routerID   net.IP // May be nil, meaning "derive from context"
```
cgoncalves, line 28: I ended up with an invalid router in my lab right off a fresh deployment:

```
/ # cat /etc/frr/frr.conf 

log stdout 
hostname 

router bgp 64512
  no bgp default ipv4-unicast
  bgp router-id <nil>

  neighbor 10.89.0.5:179 remote-as 64512
```

internal/bgp/bgpfrr/config.go
```diff
@@ -0,0 +1,121 @@
...
+hostname {{.Hostname}}
+
+{{range .Routers -}}
+router bgp {{.MyASN}}
+  no bgp default ipv4-unicast
+  bgp router-id {{.RouterId}}
+{{range .Neighbors }}
+  neighbor {{.Addr}} remote-as {{.ASN}}
+{{- end }}
+{{range $n := .Neighbors -}}
```
cgoncalves, line 26: There's an AF block per Kubernetes Service. I may be missing out on something trivial but it seems this could be optimized?

```
/ # cat /etc/frr/frr.conf 

log stdout 
hostname 

router bgp 64512
  no bgp default ipv4-unicast
  bgp router-id <nil>

  neighbor 10.89.0.5:179 remote-as 64512

  address-family ipv4 unicast
    neighbor 10.89.0.5:179 activate
    network 192.168.10.0/32
  exit-address-family
  address-family ipv4 unicast
    neighbor 10.89.0.5:179 activate
    network 192.168.10.1/32
  exit-address-family
```

```
$ kubectl get svc
NAME         TYPE           CLUSTER-IP      EXTERNAL-IP    PORT(S)                                  AGE
kubernetes   ClusterIP      10.96.0.1       <none>         443/TCP                                  31m
mirror       LoadBalancer   10.96.121.174   192.168.10.0   80:31853/TCP,81:32147/TCP,82:30054/TCP   2m50s
mirror-bkp   LoadBalancer   10.96.161.133   192.168.10.1   80:32409/TCP,81:31045/TCP,82:31755/TCP   2m50s
```

Maybe something like this below?

```
  address-family ipv4 unicast
    neighbor 10.89.0.5:179 activate
    network 192.168.10.0/32
    network 192.168.10.1/32
  exit-address-family
```

internal/bgp/bgpfrr/frr.go
cgoncalves: Just to highlight that there are two things here that need addressing: router-id set to `<nil>` and hostname not set

```
2021-09-24T13:01:44.954061686Z b'line 3: % Command incomplete: hostname \n\n'
```

internal/bgp/bgpfrr/config.go
```diff
@@ -0,0 +1,121 @@
...
+const configTemplate = `
+log stdout {{.Loglevel}}
+hostname {{.Hostname}}
+
+{{range .Routers -}}
+router bgp {{.MyASN}}
+  no bgp default ipv4-unicast
+  bgp router-id {{.RouterId}}
+{{range .Neighbors }}
+  neighbor {{.Addr}} remote-as {{.ASN}}
```
cgoncalves, line 24: This renders a line like this:

```
neighbor 10.89.0.5:179 remote-as 64512
```

which is malformed resulting in an invalid configuration:

```
2021-09-27T20:12:31.573021851Z line 13: Failure to communicate[13] to bgpd, line:  neighbor 10.89.0.5:179 remote-as 64512
```

The port number needs either to be stripped out (defaults to port 179) or a new line to override default port needs to be set. For example:

```
neighbor 10.89.0.5 remote-as 64512
neighbor 10.89.0.5 port 9179
neighbor 10.89.0.6 remote-as 64512
```

internal/bgp/bgpfrr/config.go
```diff
@@ -0,0 +1,121 @@
...
+{{- end }}
+{{range $n := .Neighbors -}}
+{{range .Advertisements }}
+  address-family {{.Version}} unicast
+    neighbor {{$n.Addr}} activate
```
cgoncalves, line 29: ditto (neighbor address port)

cgoncalves (changes requested):

speaker/bgp_controller.go
markdgray: ok .. good idea

internal/bgp/bgpfrr/config.go
```diff
@@ -0,0 +1,121 @@
...
+	"bytes"
+	"fmt"
+	"io/ioutil"
+	"os"
+	"text/template"
+)
+
+// TODO configFileName will need to be set to the filename of the FRR configuration
+// file in the container.
+var configFileName = "frr.conf"
```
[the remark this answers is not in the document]
markdgray: @cgoncalves: Thanks for trying to integrate this and flushing out these issues. I had not intended this PR to work in an integrated manner but merely design the rendering operation. However, this is very useful and I will probably make most of the updates.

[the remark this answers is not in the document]
markdgray: I will update this as suggested and remove the comment.

internal/bgp/bgpfrr/config.go
markdgray: Thanks

internal/bgp/bgpfrr/config.go
markdgray: Yeah, I absolutely think this is the case. I did consider this but it was just easier to do it in the less elegant way we are currently doing. As I mentioned in another comment, I don't think that the configuration that we are rendering in this PR is optimal. However, until we have a working e2e CI, I didn't want to focus on that too much as it will be much easier to refactor at that point. Feel free to modify it if you would like, if not. I can address in the future.

internal/bgp/bgpfrr/frr.go
markdgray: Yeah, I think it would but as you mention - early optimization. IIRC, the reloader will hide some of this as it only applies the differences but we would need to confirm and test.

I would suggest we address this in a follow-up PR. Let's get something working and passing a BGP CI first and we can make those updates later? What do you think?

internal/bgp/bgpfrr/frr.go
```diff
@@ -0,0 +1,150 @@
...
+}
+
+func (s *frrState) deleteSession(name string) {
+	delete(s.sessions, name)
+}
+
+func (s *frrState) regenerate() {
+	config := frrConfig{
+		Hostname: "", // TODO.
+		Loglevel: "", // TODO.
```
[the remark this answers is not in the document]
markdgray: Not sure where to add this knob though. Any suggestions? I guess the ConfigMap

internal/bgp/bgpfrr/config.go
markdgray: I started to look at this. There are a number of ways we could arrange the templates. However, at this stage, I don't think we know what the final configuration file should look like. Therefore, I suggest that we wait until we have merged this and integrated with FRR through a CI and we can refactor after this. I will add a note in the code outlining this suggestion. How does that sound?

internal/bgp/bgpfrr/config.go
markdgray: @msherif1234 ?

internal/bgp/bgpfrr/frr.go
fedepaol: Sounds good to me

markdgray: On second thoughts, it didn't seem so hard to do. I did an implementation of this. I added it as two separate commits in this PR. Maybe have a look at it and if you think it is worthwhile.

cgoncalves:
LGTM. This PR provides a good foundation. PR #935 will add the FRR reloader part and PR #969 will integrate the two by signaling the FRR reloader that an updated configuration has been generated and should be applied.

Thank you very much for considering my review comments and cherry-picking suggested commits into this PR!

internal/bgp/bgpfrr/frr.go
```diff
@@ -0,0 +1,242 @@
...
+		return fmt.Errorf("unable to validate IP address")
+	}
+
+	if len(adv.Communities) > 63 {
+		return fmt.Errorf("max supported communities is 63, got %d", len(adv.Communities))
+	}
+	return nil
+}
+
+func (s *Session) Commit() error {
```
fedepaol, line 70: I know I proposed the `Commit` name, but maybe `Flush` is a better name for it?
Commit reminds me of a db transaction, but here we are not really starting a transaction>

internal/bgp/bgpfrr/frr.go
fedepaol: Looks great, and I think it will save us from some headaches!

internal/bgp/bgpfrr/frr.go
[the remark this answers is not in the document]
fedepaol: Yep it could be _the_ configmap_, or a separate one used for debugging purpouse only.
I'd definetely address this in a separate PR, but I believe it's gonna make our life easier in the future.

fedepaol:
A super minor comment about the `Commint` name, but looks great!
LGTM

internal/bgp/bgpfrr/config.go
markdgray: In the process of prototyping using slices for advertisements, I seemed to come across a problem due to the non-deterministic ordering of an iteration over a map. That made me think that it may be better to uses slices in order to ensuring ordering. If the order changes, we will generate a new config file. I pushed this as a separate commit in the PR so let me know what you think.

fedepaol: I am a bit torn about this. I think that if using a map is a better fit, we should be using a map and sort before rendering, even if it costs us more. I don't think the performance penalty is big enough to sacrifice readability.

fedepaol: I think it's more readable in the sense the the sort would be isolated and for one specific purpouse. Again, I am fine with both solutions and just thinking out loudly, but I'd slightly prefer the map if we access it by key.

fedepaol:
Should we squash the commits?

markdgray:
@fede I will squash them up to the `commit()` one. However, I think we should revert the last one about lists/maps as we discussed previously?

fedepaol:
> @fede I will squash them up to the `commit()` one. However, I think we should revert the last one about lists/maps as we discussed previously?

Yep, makes sense to me

fede:
Hey, I think you are mentioning the wrong username :)

markdgray:
> Hey, I think you are mentioning the wrong username :)

You are right! @fedepaol

johananl:
FYI, I've started reviewing this.

johananl (commented):
Thanks a lot for this effort @markdgray.

I am not done reviewing, however I already have some important feedback so submitting the current state since there is a chance it's going to affect some major things in the implementation.

internal/bgp/bgpnative/native.go
```diff
@@ -0,0 +1,702 @@
+package bgpnative
```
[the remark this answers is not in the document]
cgoncalves: I had the same opinion here but decided to keep it for myself, but since Johannes expressed the same view I am now raising in support of just "native".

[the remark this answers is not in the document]
markdgray: I had `native` originally and someone else suggested `bgpnative`. Thats the only reason. Happy to change.

speaker/bgp_controller.go
```diff
@@ -266,6 +266,9 @@ func (c *bgpController) updateAds() error {
 		if err := peer.bgp.Set(allAds...); err != nil {
 			return err
 		}
+		if err := peer.bgp.Commit(); err != nil {
```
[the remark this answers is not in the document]
markdgray: Maybe I could add some documentation to explain the rationale. In the FRR case, calling `Set()` will cause the configuration file to be updated each time. So if we have multiple peers, the loop (`for _, peer := range c.peers`) we will call `Set()` multiple times. By using `Commit()` we can defer writing of the configuration file until we have `Set()` for all peers. For reference, here is some discussion on this: https://github.com/metallb/metallb/pull/958/commits/6873789a7c8b2268ed1428f81886398ebceb27f1#r715505502

markdgray:
> Thanks a lot for this effort @markdgray.
> 
> I am not done reviewing, however I already have some important feedback so submitting the current state since there is a chance it's going to affect some major things in the implementation.

Thanks for the review so far @johananl !

internal/bgp/bgpfrr/frr.go
```diff
@@ -0,0 +1,242 @@
...
+	sessions map[string]*Session
+}
+
+var state *frrState
+
+func init() {
+	state = &frrState{
+		sessions: map[string]*Session{},
+	}
+}
```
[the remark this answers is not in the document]
markdgray: Yeah, I did experiment with something like this initially. It introduced more churn than I wanted but as the design has progressed I think something like this would be useful.

speaker/bgp_controller.go
```diff
@@ -293,6 +298,12 @@ func (c *bgpController) SetNode(l log.Logger, node *v1.Node) error {
 	return c.syncPeers(l)
 }
 
-var newBGP = func(logger log.Logger, addr string, srcAddr net.IP, myASN uint32, routerID net.IP, asn uint32, hold time.Duration, password string, myNode string) (session, error) {
-	return native.New(logger, addr, srcAddr, myASN, routerID, asn, hold, password, myNode)
+var newBGP = func(logger log.Logger, addr string, srcAddr net.IP, myASN uint32, routerID net.IP, asn uint32, hold time.Duration, password string, myNode string, bgpType string) (session, error) {
```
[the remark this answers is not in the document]
markdgray: Yeah, i noticed you doing this a few days ago and it seemed useful. It sounds like you are making progress and this doesn't necessarily block this PR. Feel free to continue with it and we can merge with this when appropriate. Alternatively, I can continue with your work as part of this. Whatever makes sense.

internal/bgp/bgpnative/native.go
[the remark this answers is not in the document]
markdgray: I have changed backed to native. However, to resolve @fedepaol 's point. I did 

```	bgpfrr "go.universe.tf/metallb/internal/bgp/frr"
	bgpnative "go.universe.tf/metallb/internal/bgp/native"```

File: internal/bgp/bgpfrr/frr.go (hunk 20 above)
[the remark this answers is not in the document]
Comment (markdgray) [reply internal/bgp/bgpfrr/frr.go]: Ok, I implemented your check.

File: internal/bgp/bgpfrr/frr.go (hunk 33 above)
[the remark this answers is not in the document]
Comment (markdgray) [reply internal/bgp/bgpfrr/frr.go]: I continued my prototype and I think it looks better. As a result, I pushed a commit that implements a SessionManager. I think this addresses @johananl comments

File: internal/bgp/frr/frr.go [hunk 35]
```diff
@@ -0,0 +1,240 @@
...
+	"strconv"
+	"time"
+
+	"github.com/go-kit/kit/log"
+	"go.universe.tf/metallb/internal/bgp"
+)
+
+// As the MetalLB controller should handle messages synchronously, there should
+// no need to lock this data structure. TODO: confirm this.
+type FRRSessionManager struct {
```
fedepaol, line 18: This can be just SessionManager I think

internal/bgp/native/native.go
```diff
@@ -0,0 +1,708 @@
...
+				return true
+			}
+		}
+	}
+
+	return false
+}
+
+// The 'Native' implementation does not require a session manager .
+type NativeSessionManager struct {
```
fedepaol, line 703: I'd put this on top as it's the external interface

internal/bgp/native/native.go
```diff
@@ -0,0 +1,708 @@
...
+		return fmt.Errorf("sending keepalive to %q: %s", s.addr, err)
+	}
+	return nil
+}
+
+// NewSession() creates a BGP session using the given session parameters.
+//
+// The session will immediately try to connect and synchronize its
+// local state with the peer.
+func (sessionManager *NativeSessionManager) NewSession(l log.Logger, addr string, srcAddr net.IP, myASN uint32, routerID net.IP, asn uint32, holdTime time.Duration, password string, myNode string) (bgp.Session, error) {
```
fedepaol, line 364: I'd put this on top of the file as it's part of the exposed interface

internal/bgp/native/native.go
```diff
@@ -0,0 +1,708 @@
...
+	routerID         net.IP // May be nil, meaning "derive from context"
+	myNode           string
+	addr             string
+	srcAddr          net.IP
+	asn              uint32
+	peerFBASNSupport bool
+	holdTime         time.Duration
+	logger           log.Logger
+	password         string
+
```
fedepaol, line 41: nit: were this spaces already there?

internal/bgp/native/native.go
```diff
@@ -0,0 +1,708 @@
...
+	"github.com/go-kit/kit/log"
+	"github.com/go-kit/kit/log/level"
+	"go.universe.tf/metallb/internal/bgp"
+	"golang.org/x/sys/unix"
+)
+
+var errClosed = errors.New("session closed")
+
+// Session represents one BGP session to an external router.
+type Session struct {
```
fedepaol, line 30: This can be private now I think

fedepaol:
Looks great! Left a few comments, not really blocking.

internal/bgp/native/native.go
fedepaol: Also, this can just be SessionManager I think

markdgray:
Great, I'll update

internal/bgp/native/native.go
markdgray: Actually both `session` and `sessionManager` types can be private in both `native.go` and `frr.go`.

fedepaol:
LGTM

markdgray:
Thanks

PR #959, opened 2021-09-21
Fix namespace parameter name

cgoncalves:
Commit 51bfb61 consolidated several namespace options into one
('namespace'/'METALLB_NAMESPACE'). This commit fixes a leftover in the
website.


russellb:
Thanks!

PR #961, opened 2021-09-22
Add bgp metrics tests

fedepaol:
We extend the e2e tests with one test that validates the presence of the current metrics, create a service and checks the service announcement metrics.

This may require some rework when https://github.com/metallb/metallb/pull/931 lands.

### DIFF
--- a/dev-env/bgp/config.yaml.tmpl
+++ b/dev-env/bgp/config.yaml.tmpl
@@ -16,4 +16,4 @@ data:
       protocol: bgp
       addresses:
       - 192.168.10.0/24
-      - fc00:f853:0ccd:e799::/64
+      - fc00:f853:0ccd:e799::/124
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -23,12 +23,15 @@ import (
 	"fmt"
 	"net"
 	"strconv"
-	"time"
+
+	"go.universe.tf/metallb/e2etest/pkg/metrics"
 
 	"github.com/onsi/ginkgo"
+	dto "github.com/prometheus/client_model/go"
 	"go.universe.tf/metallb/e2etest/pkg/executor"
 	"go.universe.tf/metallb/e2etest/pkg/frr"
 	"go.universe.tf/metallb/e2etest/pkg/routes"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
@@ -39,13 +42,10 @@ const frrContainer = "frr"
 
 var _ = ginkgo.Describe("BGP", func() {
 	f := framework.NewDefaultFramework("bgp")
-	var loadBalancerCreateTimeout time.Duration
-
 	var cs clientset.Interface
 
 	ginkgo.BeforeEach(func() {
 		cs = f.ClientSet
-		loadBalancerCreateTimeout = e2eservice.GetServiceLoadBalancerCreationTimeout(cs)
 	})
 
 	ginkgo.AfterEach(func() {
@@ -55,19 +55,10 @@ var _ = ginkgo.Describe("BGP", func() {
 	})
 
 	ginkgo.It("should work for type=Loadbalancer", func() {
-		namespace := f.Namespace.Name
-		serviceName := "external-local-lb"
-		jig := e2eservice.NewTestJig(cs, namespace, serviceName)
-
-		svc, err := jig.CreateLoadBalancerService(loadBalancerCreateTimeout,
-			tweakServicePort())
-		framework.ExpectNoError(err)
-
-		_, err = jig.Run(tweakRCPort())
-		framework.ExpectNoError(err)
+		svc := createServiceWithBackend(cs, f.Namespace.Name)
 
 		defer func() {
-			err = cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
+			err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
 			framework.ExpectNoError(err)
 		}()
 
@@ -86,7 +77,7 @@ var _ = ginkgo.Describe("BGP", func() {
 			ginkgo.By(fmt.Sprintf("checking connectivity to %s with docker", address))
 		}
 
-		err = wgetRetry(address, exc)
+		err := wgetRetry(address, exc)
 		framework.ExpectNoError(err)
 
 		allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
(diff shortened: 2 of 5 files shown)


e2etest/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,79 @@
...
+}
+
+// GaugeForLabels retrieves the value of the Gauge matching the given set of labels.
+func GaugeForLabels(metricName string, labels map[string]string, metrics map[string]*dto.MetricFamily) (int, error) {
+	return metricForLabels(metricName, labels, metrics, func(m *dto.Metric) int {
+		return int(m.GetGauge().GetValue())
+	})
+}
+
+// CounterForLabels retrieves the value of the Gauge matching the given set of labels.
```
markdgray, line 39: Should this say "value of Counter"?

fedepaol: Oops yes, copy / paste :-)

fedepaol: Fixed

e2etest/pkg/metrics/metrics.go
```diff
@@ -0,0 +1,79 @@
...
+	corev1 "k8s.io/api/core/v1"
+
+	"github.com/pkg/errors"
+	"k8s.io/kubernetes/test/e2e/framework"
+)
+
+// MetricsForPod returns the parsed metrics for the given pod, scraping them
+// from the executor pod.
+func ForPod(executor, target *corev1.Pod) (map[string]*dto.MetricFamily, error) {
+	metricsUrl := path.Join(net.JoinHostPort(target.Status.PodIP, "7472"), "metrics")
```
markdgray, line 20: Is this a standard port? If not, is there any way to determine it?

fedepaol: It depends on the deployment (meaning, the deployment done via manifests currently uses that port, but it could change).
We could inspect the pod checking what pod is using, but I am not sure it's worth the effort. Another option is to have this overrideable from outside, but I'd wait to have the problem. If that helps, I can have the port as a constant.

e2etest/bgp_test.go
```diff
@@ -80,8 +71,94 @@ var _ = ginkgo.Describe("BGP", func() {
...
+				ginkgo.By(fmt.Sprintf("checking speaker %s", speaker.Name))
+
+				speakerMetrics, err := metrics.ForPod(controllerPod, speaker)
+				framework.ExpectNoError(err)
+				validateGaugeValue(1, "metallb_bgp_session_up", map[string]string{"peer": peerAddr}, speakerMetrics)
+				validateGaugeValue(0, "metallb_bgp_announced_prefixes_total", map[string]string{"peer": peerAddr}, speakerMetrics)
+			}
+
+			ginkgo.By("creating a service")
+			svc := createServiceWithBackend(cs, f.Namespace.Name) // Is a sleep required here?
```
markdgray, line 119: Check this todo? I don't think so as `svc, err := jig.CreateLoadBalancerService(timeout, tweakServicePort())` will timeout and fail if not created. You can confirm by looking at the funciton in jig https://github.com/kubernetes/kubernetes/blob/b0bc8adbc2178e15872f9ef040355c51c45d04bb/test/e2e/framework/service/jig.go#L245 and from what I see it waits with `return j.WaitForLoadBalancer(timeout)`

fedepaol: Yes, what I don't know is how long it will take for the metrics to appear.
After the service (and the endpoint) is created, it may take some time to metallb to start producing the metrics and we may scrape them too early (causing the test to flake).
I'd say let's keep an eye on this knowing that this could be the cause.

e2etest/bgp_test.go
```diff
@@ -80,8 +71,94 @@ var _ = ginkgo.Describe("BGP", func() {
...
+				speakerMetrics, err := metrics.ForPod(controllerPod, speaker)
+				framework.ExpectNoError(err)
+				validateGaugeValue(1, "metallb_bgp_session_up", map[string]string{"peer": peerAddr}, speakerMetrics)
+				validateGaugeValue(1, "metallb_bgp_announced_prefixes_total", map[string]string{"peer": peerAddr}, speakerMetrics)
+
+				updatesTotal, err := metrics.CounterForLabels("metallb_bgp_updates_total", map[string]string{"peer": peerAddr}, speakerMetrics)
+				framework.ExpectNoError(err)
+				framework.ExpectEqual(updatesTotal >= 1, true, "expecting ", updatesTotal, "greater than 1")
+
+				validateGaugeValue(1, "metallb_speaker_announced", map[string]string{"node": speaker.Spec.NodeName, "protocol": "bgp", "service": fmt.Sprintf("%s/%s", f.Namespace.Name, svc.Name)}, speakerMetrics)
```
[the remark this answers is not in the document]
fedepaol: I _think_ I covered all the bgp related metrics (at least, what I did was pausing after the service deployment and see what was there). In any case, this should provide enough infra to validate any metrics we want in the future with minimal effort.

russellb:
I rebased this branch on main to get it to run against the updated CI config which includes the IPv6 tests.  I wanted to make sure this test still passed OK

russellb:
@fedepaol test failed for some reason after rebasing on the ipv6 test jobs

fedepaol:
> @fedepaol test failed for some reason after rebasing on the ipv6 test jobs

Sep 27 19:52:50.025: Expected
        <int>: -9223372036854775808
    to equal
        <int>: 256

almost :-)

dev-env/bgp/config.yaml.tmpl
```diff
@@ -16,4 +16,4 @@ data:
       protocol: bgp
       addresses:
       - 192.168.10.0/24
-      - fc00:f853:0ccd:e799::/64
+      - fc00:f853:0ccd:e799::/124
```
russellb, line 19: but a `/64` is a valid configuration, right?  So is this just changing the test so that it doesn't hit a known bug?  :)

fedepaol: It's a valid configuration that translates into maxint64 as value for the metric (see https://github.com/metallb/metallb/blob/997a177a762c6c8215ca9438083545529b675376/internal/allocator/allocator.go#L326), and already covered by unit tests: https://github.com/metallb/metallb/blob/4de2c0c7fbd3e4eef624bdff40c1c02f98d252d8/internal/allocator/allocator_test.go#L1213

I found more reasonable to have an exact number for the initial test.

When https://github.com/metallb/metallb/pull/931 lands, we can have multiple configurations testing that corner case, or maybe not since it would be more expensive in terms of e2e test execution time and the parsing is already covered by unit tests.

russellb: I wasn't sure if the metric was broken for the maxint64 case based on the output:

> Sep 27 19:52:50.025: Expected
> : -9223372036854775808
> to equal
> : 256

Let me put it another way -- do you expect the metric to have a value that makes sense if someone configures a `/64` ?

fedepaol: If someone configures a /64 or higher, the exposed value is MaxInt64 (as is being set here https://github.com/metallb/metallb/blob/997a177a762c6c8215ca9438083545529b675376/internal/allocator/allocator.go#L326)

I assumed the error was because of:
- the new ipv6 range, resulting in MaxInt64
- a bad int translation in the e2e infrastructure (translating to int instead of int64).

I ended up changing the pool range, because I thought e2e would be more meaningful with a regular case than into the corner case.

Now, I should have double tested it before assuming it was a bug in the tests. I'll validate it manually and fix it / cover it with a test if it's a bug.

russellb: @fedepaol says he'll follow up with another fix if something is broken with the metric itself with a `/64`

fedepaol: Just to add that I dug into this a bit more. The metric exposed to the endpoint looks fine:

`metallb_allocator_addresses_total{pool="dev-env-bgp"} 9.223372036854776e+18`

What's not working, is the fact that the parsing library returns it as a float64 and as part of our tests we are trying to convert it to int64, which doesn't really work as expected https://github.com/golang/go/issues/29463

Metrics wise, everything is fine and there's no bug I think.

russellb: thank you for confirming!

PR #963, opened 2021-09-23
chart: add artifacthub-repo.yml

gclawes:
Contains artifacthub.io metadata according to requirements here:
* https://artifacthub.io/docs/topics/repositories/#helm-charts-repositories
* https://github.com/artifacthub/hub/blob/master/docs/metadata/artifacthub-repo.yml

I've included my own name/email on the owners list.  We should add other maintainers as well.

Fixed #875

PR #964, opened 2021-09-23
gh-pages: Add empty circleci configuration

russellb:
Add an empty circleci configuration to this branch to avoid circleci
reporting failures against this branch.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,0 +1,22 @@
+version: 2.1
+
+# CircleCI reports an error if there's no config in this branch,
+# so we include a noop config.
+
+jobs:
+  noop:
+    docker:
+      - image: cimg/go:1.16
+    steps:
+      - checkout
+      - run: echo "noop"
+
+workflows:
+  version: 2
+  noop-workflow:
+    jobs:
+      - noop:
+          filters:
+            branches:
+              only: branchdoesnotexist
+


fedepaol:
@russellb I think the right approach is to share the same circle-ci config and configure the filters to skip this branch:

https://github.com/metallb/metallb/blob/a33d307960a6288bb464f8da6b9959d6e83e4a60/.circleci/config.yml#L107
https://circleci.com/docs/2.0/configuration-reference/#branches

russellb:
OK finally ... that took way longer than it should have.  I blame lack of coffee.

Now circleci won't report failures against this branch due to lack of config.  The current config makes it run nothing.  I noticed this on #963.  cc @gclawes

russellb:
> @russellb I think the right approach is to share the same circle-ci config and configure the filters to skip this branch:
> 
> https://github.com/metallb/metallb/blob/a33d307960a6288bb464f8da6b9959d6e83e4a60/.circleci/config.yml#L107
> 
> https://circleci.com/docs/2.0/configuration-reference/#branches

yeah, could do that, but there's nothing to automatically keep them in sync, so i'm not sure there's much value in doing that.  It just puts a much more complicated (and eventually outdated) config in this branch with jobs that will never run.  I agree that would work though ...

fedepaol:
Makes sense.
LGTM

russellb:
thanks for the review!  Since this is just a trivial CI change to add a job that should never run, i'm just going to merge this :)

PR #966, opened 2021-09-28
TestNeighbours unit test: sort neighbours before checking

fedepaol:
The result is fetched as the iteration of a map, making the result unpredictable. Here we sort the result so the unit test is stable.

Also, removed unnecessary println.


russellb:
passed!  thanks

PR #967, opened 2021-09-28
Add BFD support

fedepaol:
This PR adds BFD support on top of BGP.

It will be possible to enable BFD for a given BGP peer, following the configuration described in the design proposal: https://github.com/metallb/metallb/blob/main/design/bgp-bfd.md

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -45,10 +45,11 @@ import (
 )
 
 const (
-	frrIBGP = "frr-iBGP"
-	frrEBGP = "frr-eBGP"
-	IBGPAsn = 64512
-	EBGPAsn = 64513
+	frrIBGP      = "frr-iBGP"
+	frrEBGP      = "frr-eBGP"
+	IBGPAsn      = 64512
+	EBGPAsn      = 64513
+	baseRouterID = "10.10.10.%d"
 )
 
 var (
@@ -74,9 +75,7 @@ var _ = ginkgo.Describe("BGP", func() {
 			rc: frrconfig.RouterConfig{
 				ASN:      IBGPAsn,
 				BGPPort:  179,
-				RouterID: "10.10.10.10",
 				Password: "ibgp-test",
-				HoldTime: "180s",
 			},
 		},
 		{
@@ -88,7 +87,6 @@ var _ = ginkgo.Describe("BGP", func() {
 			rc: frrconfig.RouterConfig{
 				ASN:      EBGPAsn,
 				BGPPort:  180,
-				RouterID: "11.11.11.11",
 				Password: "ebgp-test",
 			},
 		},
@@ -123,24 +121,24 @@ var _ = ginkgo.Describe("BGP", func() {
 
 	table.DescribeTable("A service of protocol load balancer should work with", func(ipFamily string, setProtocoltest string) {
 		var allNodes *corev1.NodeList
-		var peers []peer
-		for _, c := range frrContainers {
-			c.RouterConfig.IPFamily = ipFamily
-			c.NeighborConfig.IPFamily = ipFamily
-			address := c.Ipv4
-			if ipFamily == "ipv6" {
-				address = c.Ipv6
+		ginkgo.BeforeEach(func() {
+			var peers []peer
+			for _, c := range frrContainers {
+				c.RouterConfig.IPFamily = ipFamily
+				c.NeighborConfig.IPFamily = ipFamily
+				address := c.Ipv4
+				if ipFamily == "ipv6" {
+					address = c.Ipv6
+				}
+				peers = append(peers, peer{
+					Addr:     c.Ipv4,
+					ASN:      c.RouterConfig.ASN,
+					MyASN:    c.NeighborConfig.ASN,
+					Port:     c.RouterConfig.BGPPort,
+					Password: c.RouterConfig.Password,
+				})
 			}
-			peers = append(peers, peer{
-				Addr:     address,
-				ASN:      c.RouterConfig.ASN,
-				MyASN:    c.NeighborConfig.ASN,
-				Port:     c.RouterConfig.BGPPort,
-				RouterID: c.RouterConfig.RouterID,
-				Password: c.RouterConfig.Password,
-				HoldTime: c.RouterConfig.HoldTime,
-			})
-		}
+		})
 		configData := configFile{
 			Pools: []addressPool{
 				{
@@ -230,8 +228,7 @@ var _ = ginkgo.Describe("BGP", func() {
 			framework.ExpectNoError(err)
 			speakerPods = make([]*corev1.Pod, 0)
 			for _, item := range speakers.Items {
-				i := item
-				speakerPods = append(speakerPods, &i)
+				speakerPods = append(speakerPods, &item)
 			}
 		})
 
@@ -248,14 +245,18 @@ var _ = ginkgo.Describe("BGP", func() {
 				if ipFamily == "ipv6" {
 					address = c.Ipv6
 				}
+				holdTime := ""
+				if i > 0 {
+					holdTime = fmt.Sprintf("%ds", i*180)
+				}
 				peers = append(peers, peer{
 					Addr:     address,
 					ASN:      c.RouterConfig.ASN,
 					MyASN:    c.NeighborConfig.ASN,
 					Port:     c.RouterConfig.BGPPort,
-					RouterID: c.RouterConfig.RouterID,
+					RouterID: fmt.Sprintf(baseRouterID, i),
 					Password: c.RouterConfig.Password,
-					HoldTime: c.RouterConfig.HoldTime,
+					HoldTime: holdTime,
 				})
 				peerAddrs = append(peerAddrs, address+fmt.Sprintf(":%d", c.RouterConfig.BGPPort))
 			}
@@ -397,14 +398,18 @@ var _ = ginkgo.Describe("BGP", func() {
 				if ipFamily == "ipv6" {
 					address = c.Ipv6
 				}
+				holdTime := ""
+				if i > 0 {
+					holdTime = fmt.Sprintf("%ds", i*180)
+				}
 				peers = append(peers, peer{
 					Addr:     address,
 					ASN:      c.RouterConfig.ASN,
 					MyASN:    c.NeighborConfig.ASN,
 					Port:     c.RouterConfig.BGPPort,
-					RouterID: c.RouterConfig.RouterID,
+					RouterID: fmt.Sprintf(baseRouterID, i),
 					Password: c.RouterConfig.Password,
-					HoldTime: c.RouterConfig.HoldTime,
+					HoldTime: holdTime,
 				})
 			}
 			configData := configFile{
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -33,10 +33,12 @@ import (
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
+
 	dto "github.com/prometheus/client_model/go"
 	"go.universe.tf/metallb/e2etest/pkg/frr"
 	frrconfig "go.universe.tf/metallb/e2etest/pkg/frr/config"
 	frrcontainer "go.universe.tf/metallb/e2etest/pkg/frr/container"
+	bgpfrr "go.universe.tf/metallb/internal/bgp/frr"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
@@ -104,6 +106,7 @@ var _ = ginkgo.Describe("BGP", func() {
 		cs = f.ClientSet
 		frrContainers, err = createFRRContainers(containersConf)
 		framework.ExpectNoError(err)
+
 	})
 
 	ginkgo.AfterEach(func() {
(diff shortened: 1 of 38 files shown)


internal/bgp/frr/config.go
```diff
@@ -64,14 +67,41 @@ router bgp {{.MyASN}} view {{.MyASN}}
...
+    transmit-interval {{.TransmitInterval}}
+    {{end -}}
+    {{ if .DetectMultiplier -}}
+    detect-multiplier {{.DetectMultiplier}}
+    {{end -}}
+    {{ if .EchoMode -}}
+    echo-mode
+    {{end -}}
+    {{ if .EchoInterval -}}
+    echo-interval {{.EchoInterval}}
```
liornoy, line 89: I don't think the echo interval of the BFD in frr is configured this way.
In their docs it is splits to:
echo receive-interval <disabled|(10-60000)>
echo transmit-interval (10-60000)

fedepaol: That part changed between 7.5 (which we are pinning now) and latest.

e2etest/bgp_test.go
```diff
@@ -229,15 +225,19 @@ var _ = ginkgo.Describe("BGP", func() {
 
 			var peers []peer
 			var peerAddrs []string
-			for _, c := range frrContainers {
+			for i, c := range frrContainers {
```
oribon: General comment about the commit: `Filling ... in the external containers but using them as
configuration of speakers is misleading.` - that's misleading because it wasn't used here correctly :sweat_smile:
From the `containerConfig`'s (which represents the external FRR container) perspective, configuration for the speaker side should be under `NeighborConfig` (and not `RouterConfig`). That's because `RouterConfig` is how it should configure itself and `NeighborConfig` is the way the speaker should present itself to the container.

I think we should preserve that and not declare any parameters somewhere else.

e2etest/bgp_test.go
```diff
@@ -394,6 +355,101 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			}
+			err := updateConfigMap(cs, configData)
+			framework.ExpectNoError(err)
+
+			for _, c := range frrContainers {
+				pairExternalFRRWithNodes(cs, c, func(container *frrcontainer.FRR) {
+					container.NeighborConfig.BFDEnabled = true
+				})
+			}
+
```
oribon, line 383: No need to call `validateFRRPeeredWithNodes` for BFD?

e2etest/bgp_test.go
```diff
@@ -394,6 +355,101 @@ var _ = ginkgo.Describe("BGP", func() {
...
+					for _, peerConfig := range bfdPeers {
+						ginkgo.By(fmt.Sprintf("Checking bfd parameters on %s", peerConfig.Peer))
+						checkBFDConfigPropagated(toCompare, peerConfig)
+					}
+				}
+				return nil
+			}, 4*time.Minute, 1*time.Second).Should(BeNil())
+
+		},
+			table.FEntry("default",
```
oribon, line 417: FEntry?

fedepaol: ops

internal/bgp/frr/config.go
```diff
@@ -80,6 +110,17 @@ type routerConfig struct {
...
+type BFDProfile struct {
+	Name             string
+	ReceiveInterval  *uint32
+	TransmitInterval *uint32
+	DetectMultiplier *uint32
+	EchoInterval     *uint32
+	EchoMode         bool
+	PassiveMode      bool
+	MinimumTTL       *uint32
+}
```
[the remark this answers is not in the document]
fedepaol: To enforce domain and module separation.
One is the data structure we read from the configuration, the other one is the one we pass to the template. They serve different purpouses and I did not want the config data structures to leak into the frr package.

e2etest/bgp_test.go
fedepaol: I think I forgot about it

internal/bgp/frr/config.go
msherif1234: maybe I am missing something in the config there should be a matching peer section in the bfd config ? the above configure only the profile ?

fedepaol: You can associate a neighbor to a bfd profile, which is what happens in https://github.com/metallb/metallb/pull/967/files#diff-1d9a33c302812585b1b2d1897f84126492129d1f9415786a73dcfa26603686c9R51

fedepaol: Documented here https://docs.frrouting.org/en/latest/bfd.html#bgp-bfd-configuration

speaker/bgp_controller.go
```diff
@@ -224,6 +229,18 @@ func (c *bgpController) syncPeers(l log.Logger) error {
 	return nil
 }
 
+func (c *bgpController) syncBFDProfiles(profiles map[string]*config.BFDProfile) error {
+	if c.bgpType != bgpFrr && len(profiles) > 0 {
+		return errors.New("bfd Profiles not supported in native mode")
+	}
```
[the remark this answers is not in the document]
fedepaol: Good catch, it's a leftover of the previous check. Will remove it.

internal/config/config_test.go
```diff
@@ -631,6 +637,143 @@ address-pools:
...
+			desc: "BFD Profile with too low receive interval",
+			raw: `
+bfd-profiles:
+- name: default
+  receive-interval: 2
+`,
+		},
+		{
+			desc: "BFD Profile with too high range receive interval",
+			raw: `
```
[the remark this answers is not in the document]
fedepaol: Yes, need to change. I was torn about testing every single parameter, thought it was significant enough (and less boring) to validate the range function against only one of them, but I can add more tests if you think it's necessary.

e2etest/bgp_test.go
fedepaol: I think if we call the structure FRR container, a reader expects to find parameters affecting the container only.
We are currently leveraging some of those to set up the metallb config because they are symmetric (i.e. the password, the ASNs). Some other belong to metallb only and it does not make sense to have them here unless we think about a wider wrapper.

oribon:
lgtm

e2etest/config/frr/daemons
```diff
@@ -27,7 +27,7 @@ eigrpd=no
 babeld=no
 sharpd=no
 pbrd=no
-bfdd=no
+bfdd=yes
```
cgoncalves, line 30: Would it make sense to also enable the BFD daemon in dev-env/bgp/frr/daemons? Having it disabled means one would need to manually toggle this setting, restart the external FRR container and resume with the desired BFD configuration. This is only for testing deployments (dev-env) so impact is low to none.

fedepaol: It won't hurt I guess.

fedepaol: Done

e2etest/pkg/frr/bfd.go
```diff
@@ -0,0 +1,23 @@
...
+	res, err := exec.Exec("vtysh", "-c", "show bfd peers json")
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to query neighbours")
+	}
+	peers, err := bgpfrr.ParseBFDPeers(res)
+	if err != nil {
+		return nil, errors.Wrapf(err, "Failed to parse neighbours %s", res)
+	}
+	return peers, nil
+}
```
msherif1234, line 23: so the CI as it stands it make sure BFD sessions are created , how about the failure detection part how does that been verified/tested, do we know how long it takes for BFD enabled session to detect a failure ? any plans to cover this in CI or at least manual testing for now and doc it ?

fedepaol: Failover tests are in the backlog. We'd need to carefully think about them, because we are still subject to the time it takes ot execute the commands. Maybe setting a ludicrous hold time on the bgp side and to check that the failover happens in a lower time would work.
I'd keep this on the functional side though.

internal/config/config.go
```diff
@@ -370,6 +426,44 @@ func parseAddressPool(p addressPool, bgpCommunities map[string]uint32) (*Pool, e
...
 }
 
+func parseBFDProfile(p bfdProfile) (*BFDProfile, error) {
+	if p.Name == "" {
+		return nil, fmt.Errorf("missing bfd profile name")
+	}
+	res := &BFDProfile{}
+	res.Name = p.Name
+	var err error
+	res.DetectMultiplier, err = bfdIntFromConfig(p.DetectMultiplier, 2, 255)
```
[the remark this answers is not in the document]
fedepaol: Wouldn't frr apply the default settings itself if not specified?
I fix defaults here, metallb will have to work with a fixed frr version which has 1:1 mapping to those parameters. If we keep it like this, we won't add the parameters that are not specified, making the config rendering more compatible.

internal/bgp/frr/config.go
fedepaol: We are not configuring bfd peers explicitly because we are leveraging the bgp's bfd option described here https://docs.frrouting.org/en/latest/bfd.html#bgp-bfd-configuration

If you have a bfd profile defined, it's totally fine to add the option to the neighbour and that neighbor will be backed up by a bfd session. So the configuration is more compact without the need of adding the peer explicitly.
This is what we do here https://github.com/metallb/metallb/pull/967/commits/694ed3476a26297dfca1db98f858be801104f0ca#diff-1d9a33c302812585b1b2d1897f84126492129d1f9415786a73dcfa26603686c9R51

fedepaol: > do u know how this will work with view bgp do u have one bfd process for every bgp instance or its shared and in that case u pick peers from both ?

I don't know, I guess we'll figure it out if / when we re-enable views.

internal/bgp/frr/config.go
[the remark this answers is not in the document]
msherif1234: this is shuting down the bfd peer I am assume its used for diagnosticsnothing against metallb api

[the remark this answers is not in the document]
fedepaol: I am not sure I understood the question then. Why would we need to add Shutdown here?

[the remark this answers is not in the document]
msherif1234: its one of the BFD config option in the FRR

speaker/main.go
```diff
@@ -354,6 +355,14 @@ func (c *controller) SetConfig(l log.Logger, cfg *config.Config) k8s.SyncState {
 		return k8s.SyncStateErrorNoRetry
 	}
 
+	if c.bgpType == bgpNative {
```
msherif1234, line 358: bgpNative ?!? u meant bgpFrr ?

internal/bgp/frr/config.go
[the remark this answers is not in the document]
msherif1234: functionality wise probably not but was just curious why u drop it to begin with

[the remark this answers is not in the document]
fedepaol: I did not add it because we don't need it, as there's no need to add stuff we won't be using.

speaker/main.go
fedepaol: Nope. The check here is, if native check if something frr only is applied

e2etest/bgp_test.go
```diff
@@ -230,33 +218,21 @@ var _ = ginkgo.Describe("BGP", func() {
 			framework.ExpectNoError(err)
 			speakerPods = make([]*corev1.Pod, 0)
 			for _, item := range speakers.Items {
-				i := item
-				speakerPods = append(speakerPods, &i)
+				speakerPods = append(speakerPods, &item)
```
oribon, line 221: accidental change?

e2etest/bgp_test.go
```diff
@@ -152,8 +135,13 @@ var _ = ginkgo.Describe("BGP", func() {
...
-			Peers: peers,
+			Peers: peersForContainers(frrContainers, ipFamily),
+		}
+		for _, c := range frrContainers {
+			c.RouterConfig.IPFamily = ipFamily
+			c.NeighborConfig.IPFamily = ipFamily
 		}
+
 		for _, c := range frrContainers {
 			pairExternalFRRWithNodes(cs, c)
```
oribon, line 146: I think we should have this in the loop above like you did here: https://github.com/metallb/metallb/blob/d0622145349a3c6687409722178e66e8f2f0e822/e2etest/bgp_test.go#L452)

e2etest/bgp_test.go
```diff
@@ -465,6 +426,124 @@ var _ = ginkgo.Describe("BGP", func() {
 			),
 		)
 	})
+
+	ginkgo.Context("BFD", func() {
+		table.DescribeTable("should work with the given bfd profile", func(bfd bfdProfile, ipFamily string) {
+			configData := configFile{
+				Pools: []addressPool{
+					{
+						Name:     "bgp-test",
```
oribon, line 435: would be better to rename this something like `bfd-test` (maybe with -bfd.Name suffix)?

oribon (commented):
Other than these small comments, LGTM

e2etest/bgp_test.go
fedepaol: yep fixin git

e2etest/bgp_test.go
fedepaol: Yes did not notice while rebasing

e2etest/bgp_test.go
fedepaol: Done

oribon:
LGTM, the layer2 fail from the ci isn't related

PR #970, opened 2021-09-29
Fix race condition in L2/ARP announcer

### LINKED ISSUE #681: Race condition
https://github.com/metallb/metallb/blob/main/internal/layer2/announcer.go#L158
There is a race condition.

We have a big cluster with many pod creation/deletion operations, so interfaces list on a node changes very often.
In our cluster MetalLB sometimes doesn't announce IPs (L2/ARP) because, at the moment when `gratuitous` function executes, one of the interfaces in `a.arps` array becomes dead and this loop crashes with an error, so no more gratuitous announces won't send.

I think we should ignore these errors and don't crash a loop.

maxpain:
Fixed #681

### DIFF
--- a/internal/layer2/announcer.go
+++ b/internal/layer2/announcer.go
@@ -196,13 +196,13 @@ func (a *Announce) gratuitous(ip net.IP) error {
 	if ip.To4() != nil {
 		for _, client := range a.arps {
 			if err := client.Gratuitous(ip); err != nil {
-				return err
+				a.logger.Log("gratuitousAnnounce", "error", err, "ip", ip, "msg", "failed to make gratuitous IP announcement")
 			}
 		}
 	} else {
 		for _, client := range a.ndps {
 			if err := client.Gratuitous(ip); err != nil {
-				return err
+				a.logger.Log("gratuitousAnnounce", "error", err, "ip", ip, "msg", "failed to make gratuitous IP announcement")
 			}
 		}
 	}
--- a/internal/layer2/announcer.go
+++ b/internal/layer2/announcer.go
@@ -196,13 +196,13 @@ func (a *Announce) gratuitous(ip net.IP) error {
 	if ip.To4() != nil {
 		for _, client := range a.arps {
 			if err := client.Gratuitous(ip); err != nil {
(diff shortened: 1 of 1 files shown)


champtar:
It's not obvious you are fixing a race condition, can you explain a bit ?

maxpain:
@champtar 

We have a big cluster with many pod creation/deletion operations, so the interfaces list on a node often changes.
In our cluster, MetalLB sometimes doesn't announce IPs (L2/ARP) because, at the moment when a gratuitous function is executing, one of the interfaces in the `a.arps` array becomes dead, and this loop crashes with an error, so no more gratuitous announces on other interfaces will be sent.

Let's ignore these errors and don't crash the loop in case of an error.

fedepaol:
@maxpain mind squashing the commits and extending the commit message with the rationale you just described?

internal/layer2/announcer.go
```diff
@@ -196,13 +196,13 @@ func (a *Announce) gratuitous(ip net.IP) error {
 	if ip.To4() != nil {
 		for _, client := range a.arps {
 			if err := client.Gratuitous(ip); err != nil {
-				return err
+				level.Error(a.logger).Log("gratuitousAnnounce", "error", err, "ip", ip, "msg", "failed to make gratuitous IP announcement")
```
fedepaol, line 199: Would it make sense to return here?

internal/layer2/announcer.go
```diff
@@ -196,13 +196,13 @@ func (a *Announce) gratuitous(ip net.IP) error {
...
 			}
 		}
 	} else {
 		for _, client := range a.ndps {
 			if err := client.Gratuitous(ip); err != nil {
-				return err
+				level.Error(a.logger).Log("gratuitousAnnounce", "error", err, "ip", ip, "msg", "failed to make gratuitous IP announcement")
```
fedepaol, line 205: same

internal/layer2/announcer.go
fedepaol: I meant return nil

russellb: but then it would skip sending the GARP on all other interfaces?

russellb: we should probably pick up discussion on #974

PR #971, opened 2021-09-30
FRR: Prometheus metrics exporter

oribon:
* In order to keep feature-parity with the current
BGP implementation we need to collect and export
the same Prometheus metrics.

* With FRR, we can't use the same approach to
collect the metrics as we currently do by producing the metrics ourselves
(https://prometheus.io/docs/introduction/glossary/#direct-instrumentation).
Rather we need to "collect" them from FRR, which is best done
using a Prometheus exporter and writing a collector for the BGP daemon
(https://prometheus.io/docs/instrumenting/writing_exporters/).
Each time Prometheus scrapes the /metrics endpoint the metrics are
gathered by the collector from FRR using vtysh.
The collector needs to satisfy the Prometheus.Collector interface.

* Adding new collectors to the exporter per FRR daemon
is possible in the future.

* The exporter will be built and shipped with the
speaker, and would be deployed as a seperate sidecar
container that has access to the FRR sockets like the reloader.


internal/bgp/stats.go
```diff
@@ -2,42 +2,60 @@ package bgp
 
 import "github.com/prometheus/client_golang/prometheus"
 
+type stat struct {
```
russellb, line 5: Are the changes here just refactoring?  It'd be nice to split it into a separate commit to make it more clear what these changes were for.

russellb (commented):
A general question - did you do an evaluation of any existing FRR promethus exporters before starting a new one for metallb?

Could you explain the tradeoffs and why writing our own is best for our needs?

For example, I see https://github.com/tynany/frr_exporter, and maybe there's more ...

fedepaol:
> A general question - did you do an evaluation of any existing FRR promethus exporters before starting a new one for metallb?
> 
> Could you explain the tradeoffs and why writing our own is best for our needs?
> 
> For example, I see https://github.com/tynany/frr_exporter, and maybe there's more ...

We had a discussion about this with @oribon , with existing exportes we won't be able to mimic the same set of metrics that are currently available in metallb, which is the bare minimum we need to provide to the users in order to guarantee no regressions.

internal/bgp/bgpfrr/parse.go
```diff
@@ -65,19 +73,25 @@ func parseNeighbour(vtyshRes string) (*Neighbor, error) {
...
+			pfxSnt += s.SentPrefixCounter
+		}
 		return &Neighbor{
-			ip:        ip,
-			connected: connected,
-			localAS:   strconv.Itoa(n.LocalAs),
-			remoteAS:  strconv.Itoa(n.RemoteAs),
+			Ip:          ip,
+			Connected:   connected,
+			LocalAS:     strconv.Itoa(n.LocalAs),
```
markdgray, line 83: This is a different naming convention that is used compared to the rest of MetalLB (myasn, asn). However, I much prefer yours. Maybe we could consolidate somehow .. not sure.

internal/bgp/bgpfrr/parse.go
```diff
@@ -10,10 +10,12 @@ import (
 )
 
 type Neighbor struct {
-	ip        net.IP
-	connected bool
-	localAS   string
-	remoteAS  string
+	Ip          net.IP
```
markdgray, line 13: this is kind of like `neighborConfig`. Maybe we could combine?

internal/bgp/bgpfrr/parse.go
```diff
@@ -65,19 +73,25 @@ func parseNeighbour(vtyshRes string) (*Neighbor, error) {
 		if n.BgpState != bgpConnected {
 			connected = false
 		}
+		pfxSnt := 0
```
markdgray, line 76: nit: I much prefer longer names like `prefixSent`.

internal/bgp/bgpfrr/parse.go
```diff
@@ -24,10 +26,16 @@ type Route struct {
...
-	LocalAs    int    `json:"localAs"`
-	BgpVersion int    `json:"bgpVersion"`
-	BgpState   string `json:"bgpState"`
+	RemoteAs     int    `json:"remoteAs"`
+	LocalAs      int    `json:"localAs"`
+	BgpVersion   int    `json:"bgpVersion"`
+	BgpState     string `json:"bgpState"`
+	MessageStats struct {
+		UpdatesSent int `json:"updatesSent"`
+	} `json:"messageStats"`
```
markdgray, line 35: This change seems unrelated to the commit message?

internal/bgp/bgpfrr/parse.go
```diff
@@ -65,19 +73,25 @@ func parseNeighbour(vtyshRes string) (*Neighbor, error) {
 		if n.BgpState != bgpConnected {
 			connected = false
 		}
+		pfxSnt := 0
+		for _, s := range n.AddressFamilyInfo {
```
markdgray, line 77: This change seems unrelated to the commit message.

frr-metrics/exporter.go
```diff
@@ -0,0 +1,61 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+	"github.com/prometheus/exporter-toolkit/web"
+
+	"go.universe.tf/metallb/frr-metrics/collector"
+	"go.universe.tf/metallb/internal/logging"
+)
+
+var (
+	metricsPort   = flag.String("metrics-port", ":7473", "Address to listen on for web interface.")
```
markdgray, line 21: Does this mean that `metricsPort` needs to be specified as `:<port_number`? Maybe change this to `flag.Uint`?

internal/bgp/stats.go
oribon: done, moved to a separate commit

internal/bgp/bgpfrr/parse.go
oribon: moved to a separate commit and updated unit tests, thanks

frr-metrics/exporter.go
oribon: done

russellb:
will need a rebase now

oribon:
> will need a rebase now

Fixed, also changed the exporter to reflect the switch to bgp views and added a commit that integrates it in the manifests.
The last missing piece is enabling the metrics tests for FRR in the CI but they are flakey without https://github.com/metallb/metallb/pull/1012 (specifically we need the Gauge change) so we should probably wait for that (locally with some sleeps in the right place they pass).

oribon:
Reverted the views change since we're probably dropping it

oribon:
enabled metrics for FRR lane, fails on ipv6 are expected and should be fixed by https://github.com/metallb/metallb/pull/1040

fedepaol:
LGTM

Issue #968, opened 2021-09-29
Shared Layer 2 IPs should not be announced by multiple nodes


wzrdtales:
Reopen of #315 

# What is happening?

Multiple nodes try to claim the IP, which results in an unstable connection.

# Why is this happening?

The layer 2 ip was shared between two services in two different namespaces. One is pointing to a specific port, the other one is pointing to the ingress. To reproduce just create two namespaces with two different services and share the ip.

# What should happen?

The shared ip should be recognized at that by the speaker in reliable fashion and not being used duplicated.

fedepaol:
Hi @wzrdtales , just to understand better:
When you say that the ip is shared between the services, are you setting the same `loadBalancerIP` manually or are you using the `metallb.universe.tf/allow-shared-ip` annotation?

Would you mind sharing the services / pools definitions to avoid missing details in translation?
Thanks!

wzrdtales:
both, since both have to be defined

wzrdtales:
```yaml
spec:
  clusterIP: 10.43.25.93
  clusterIPs:
  - 10.43.25.93
  externalTrafficPolicy: Cluster
  loadBalancerIP: anonymized
  ports:
  - name: ssh
    nodePort: 30226
    port: 22
    protocol: TCP
    targetPort: 2222
  selector:
    app: gitlab-shell
    release: gitlab-prod
  sessionAffinity: ClientIP
  sessionAffinityConfig:
    clientIP:
      timeoutSeconds: 60
  type: LoadBalancer
```

```yaml
spec:
  clusterIP: 10.43.232.168
  clusterIPs:
  - 10.43.232.168
  externalTrafficPolicy: Cluster
  loadBalancerIP: anonymized
  ports:
  - name: http
    nodePort: 31453
    port: 80
    protocol: TCP
    targetPort: 80
  - name: https
    nodePort: 32283
    port: 443
    protocol: TCP
    targetPort: 443
  selector:
    app: ingress-nginx
  sessionAffinity: ClientIP
  sessionAffinityConfig:
    clientIP:
      timeoutSeconds: 60
  type: LoadBalancer
```


This is the config applied

wzrdtales:
plus the annotation     `metallb.universe.tf/allow-shared-ip: anonymized` on both services of course

fedepaol:
Thanks! Will try to reproduce this locally and submit a fix.

wer-anders:
[...]

PR #976, opened 2021-09-30
Testl2issue

fedepaol:
Fixes https://github.com/metallb/metallb/issues/968

This started with adding a test to reproduce the issue above, and after investigating it the fix ended up to be removing the need for locality for endpoints in case of traffic policy of type cluster, plus changing the way we determine if a speaker should advertise a service, consuming the external ip(s) instead of the service name + namespace to sort.

Note: I haven't aligned the unit tests (yet), will do if this changes make sense to the reviewers.


fedepaol:
uhm, tests failed :eyes:

russellb:
You’ve hit an area that has needed to be fixed for a long time.  We have many open issues and PRs that are related to this.  I had to spend some time going back through all the related issues to help decide what to do with this, so I’m going to share some of those notes here.

I’d love to sort all of this out in this PR and get a bunch of stuff closed out.

You’ve made two key changes here:

You changed the way we calculate the set of candidate nodes to be all nodes when the traffic policy is `cluster` instead of using only the nodes where a speaker is running along side an Endpoint backing that Service.
You changed the sort key of the candidate node list to be the node name + load balancer IP addresses instead of node name + service name.

On #1 -- related PRs include #781, #665, and #613.  Related open issues include #646 and #302.

One thing I liked about other changes compared to yours is to prefer the old algorithm if it produces at least one candidate node. That way we’ll still put the IP on a node with a local endpoint if possible. Otherwise, we can fall back to what you’ve proposed here if we’re allowed to (with the `cluster` traffic policy).  If you agree, updating to incorporate this behavior is my first request.


On #2, you called out issue #968.  There’s another class of issues related to change #2, which is to deal with shared IPs for services that use the Local traffic policy.  See issues #558, #801 and PRs #562, #887, and #922.

I do see that your change would resolve #968, but I don’t think we would fully resolve the issues related to sharing an IP between multiple IP addresses using the Local traffic policy.  Sharing an IP across multiple Local policy services is not going to always work, and we need to get better at detecting that.  Perhaps we can defer that to a future PR because I don’t think what you’re proposing here makes it any worse.

Making a change here will potentially cause many/most/all IPs to get reassigned to a new node.  It would be nice if we could sort out these local traffic policy issues at the same time so we impose the disruption in node selection just once. We should also document this in the release notes.

If you can, take a look at these other PRs and issues.  I’d like to try to minimize the number of times we release a change to node selection.  If you’d like, we could split out change #1 (with my suggested updates) into a separate PR to merge first.

russellb:
> > One thing I liked about other changes compared to yours is to prefer the old algorithm if it produces at least one candidate node. That way we’ll still put the IP on a node with a local endpoint if possible. Otherwise, we can fall back to what you’ve proposed here if we’re allowed to (with the `cluster` traffic policy). If you agree, updating to incorporate this behavior is my first request.
> 
> Oh, sounds cool. But please remind that if there are more than one backend pods, there's no guarantee at all that the traffic will be directed to the local pod (unless traffic policy is local).

Right.  I was thinking that at least some percentage of the traffic would stay local, which is a little better than none of it.

> > Making a change here will potentially cause many/most/all IPs to get reassigned to a new node. It would be nice if we could sort out these local traffic policy issues at the same time so we impose the disruption in node selection just once. We should also document this in the release notes.
> 
> Do you mean after an upgrade?

Yes

fedepaol:
> On 2, you called out issue #968. There’s another class of issues related to change #2, which is to deal with shared IPs for services that use the Local traffic policy. See issues #558, #801 and PRs #562, #887, and #922.

I gave a look at the others:
- https://github.com/metallb/metallb/issues/558 will be solved by this very same logic
- https://github.com/metallb/metallb/pull/562 and https://github.com/metallb/metallb/pull/922 are more or less a copy of both changes contained in this PR (and sorry, did not realize it was already tackled), I started investigating #968 

Said that (and having looked at the code again):

- having shared ip between services with local policy means that they have the same endpoints (https://github.com/metallb/metallb/blob/6126100fb706e9d30d3698b2363ad89d7d235517/internal/allocator/k8salloc/k8salloc.go#L27 / https://github.com/metallb/metallb/blob/997a177a762c6c8215ca9438083545529b675376/internal/allocator/allocator.go#L145
- this makes it impossible to have two services, with one of them being local, with a different set of endpoints (and the risk of announcing one from one node, the other from the other)
- This means that, assuming a stable ordering function of the nodes based on active speakers / endpoints in case of traffic policy locals, all the speakers will agree on which node should advertise a given service (provided that we use the ip as "service identifier".
- If such conflict exists, the second service won't get its ip and will remain in pending

Long story short, if we want to share the same ip to two services with a different backend selector, the second service won't get its ip. If the set of backends is the same, the winner node in each speaker selection will be the same assuming we don't use the service's name but the service's ip(s) / the sharing key as ordering criteria.


https://github.com/metallb/metallb/issues/801 (and it's related PR) handles a corner case.
MetalLB runs with the simplification that, if two services have local policy and want to share the same IP address, they MUST have the same set of pods as backend. 
My understanding is that https://github.com/metallb/metallb/pull/887 tries to check if the set of backend pods is the same, regardless of the fact that they are choosen using the same selector. This would not change the considerations above, would just make the logic a bit smarter.

I think the conclusion is, the current PR would address all the mentioned issues / prs you mentioned except 
https://github.com/metallb/metallb/issues/801 which would not affect the node selection behaviour.

russellb:
> > Thanks a lot for putting the time into looking over all of those issues and PRs. I'm OK with proceeding with this PR once you update the related tests. Can you also update the PR / commit message to reflect all of the things it will close?
> 
> How about mentioning the things we close in the PR only? I always thought as commit message explaining the rationale, but I can also mention issues if you think it's worth.

PR is fine

champtar:
You might want to share an IPv4 but have 2 different IPv6 between 2 services. The sort key should be the `metallb.universe.tf/address-pool` annotation if not empty else the svc name (like right now).

briantopping:
As the author of #801, I'd support what I did there being integrated against this PR. If there's anything I can contribute, I'd be happy to help.

fedepaol:
> You might want to share an IPv4 but have 2 different IPv6 between 2 services. The sort key should be the `metallb.universe.tf/allow-shared-ip` annotation if not empty else the svc name (like right now).

This is an interesting point (and currently covered behind the fact that we don't handle dual stack yet). I assumed that two services with the same ipv4 would also have had the same ipv6, but you are right, that's not always the case.

I'd rather try to change the interface and the logic to focus on the IP (because at the end of the day the shouldAnnounce logic is applied per load balancer ip https://github.com/metallb/metallb/blob/31f4de0b27986e185bb15db32d5d7156833edcb3/speaker/main.go#L272 ) so we avoid having crowded announcers in those cases where the sharing key is the same BUT the ips are different.
Does it make sense?

russellb:
> Thanks @fedepaol, that's fine for me if your proposed changes don't block a solution to #801. It sounds like you are saying that's not the case. Merging them at the same time would help ensure that doesn't happen. I've been waiting six months for some actionable feedback on that issue and I'm not sure why it never happened.

just a time issue ... the project has been incredibly constrained lately.  we're trying to start picking back up, and looking at increasing the size of the maintainers team to help.  apologies for stuff sitting so long

briantopping:
> just a time issue ... the project has been incredibly constrained lately. we're trying to start picking back up, and looking at increasing the size of the maintainers team to help. apologies for stuff sitting so long

I *totally* understand that. Just trying to "keep a place in line" and not add significant work on other equally important issues. 

If this issue doesn't block #801, I'm happy to merge against it. Just trying to determine what should be merged first.

russellb:
> > just a time issue ... the project has been incredibly constrained lately. we're trying to start picking back up, and looking at increasing the size of the maintainers team to help. apologies for stuff sitting so long
> 
> I _totally_ understand that. Just trying to "keep a place in line" and not add significant work on other equally important issues.
> 
> If this issue doesn't block #801, I'm happy to merge against it. Just trying to determine what should be merged first.

understood.  I jumped on this one because it addresses a bunch of open issues / PRs, including one opened back in 2018 (#302).

PR #983, opened 2021-10-06
website: Add Linux Foundation footer

rata:
This is needed as part of the CNCF on-boarding, as requested here:
	https://github.com/cncf/toc/issues/720


rata:
Attach there is a screenshot of how this is shown on the website with this change.



metallb_000

rata:
@russellb thanks! I added that to the homepage (as the link says) too.

I jumped into the CNCF call for new sandbox projects a few minutes ago and they confirmed the footer looks fine.

Here is how the homepage looks like with the "We are a CNCF..." text now:
metallb_001


Let me know what you think :)

rata:
Website updated now, this fixed it: https://github.com/metallb/metallb/pull/988

PR #982, opened 2021-10-06
website: Remove reference to removed file

rata:
The romana file was removed in PR:
	https://github.com/metallb/metallb/pull/950

However, this reference wasn't and makes hugo fail with:
	ERROR 2021/10/06 16:12:50 [en] REF_NOT_FOUND: Ref "configuration/romana.md" from page "release-notes/_index.md": page not found
	Error: Error building site: logged 1 error(s)

This patch just removes that reference to the file that is removed.

cc @fedepaol @russellb @v1k0d3n

Issue #984, opened 2021-10-06
Adjust linter check to allow commit msg with more than 72 chars


rata:
While it is nice to have well-formatted commit msgs, making it a hard check to always be less than 72 chars long makes it impossible to include links, which is just silly IMHO.

Links are useful to reference why something was needed, requested or to provide more info.

IMHO, we should wither remove this check and maybe just assert about commit title, not the body. What do others think?

PR #985, opened 2021-10-06
commitlint: Allow body lines longer than 72

rata:
This way, we allow using links in the commit msg.

We will warn if this happens, as a way to not completely remove this
check. If we see that this is usually ignored, we can remove this check
in a future commit.

cc @johananl 

I run into this when creating these PRs:
 https://github.com/metallb/metallb/pull/982#issuecomment-936531399
 https://github.com/metallb/metallb/pull/983

It is super painful not being able to paste links or verbatim errors (like I found on hugo)

### DIFF
--- a/commitlint.config.js
+++ b/commitlint.config.js
@@ -2,8 +2,9 @@ module.exports = {
     rules: {
         // Ensure we have a blank line between the header and the body.
        'body-leading-blank': [2, 'always'],
-       // Ensure we don't have lines longer than 72 chars.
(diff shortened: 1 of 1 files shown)


rata:
He is ok removing the check (forgot to link this issue: https://github.com/metallb/metallb/issues/984#issuecomment-939894988). Takikng that as a LGTM for this, please let me know if you prefer to remove the check entirely, I can open a new PR for that.

PR #990, opened 2021-10-13
Add E2E tests to validate different ip ranges modes (ranges vs cidrs)

sabinaaledort:
Test Different types of AddressPools IP ranges, set by CIDRs or ranges.


fedepaol:
@sabinaaledort can you adjust the PR description?

fedepaol:
Also, can we add the same to L2? I know we are focusing on bgp, but with low effort we can expand the l2 coverage too

PR #991, opened 2021-10-14
Add tests for ExternalTrafficPolicy=Local services

oribon:
Layer2 and BGP both respect ExternalTrafficPolicy=Local, adding e2e tests
that validate that they work as intended.

* L2: As usual only one node can announce the service's IP but it should only route to its local pods.
The test checks that the service is reachable, finds its advertised MAC from the local arp table,
finds who's the node with that MAC = who's the advertising node, hits the service again multiple times
using wget http://svc/hostname and validates that all the pods that responded are on that node.

* BGP: Only the nodes with local endpoints should announce the service's IP.
The test checks that the service is reachable, validates the the external FRR container receives
advertisemnts for that IP only from these nodes and validates that ip route show matches only them.

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -91,85 +91,133 @@ var _ = ginkgo.Describe("BGP", func() {
 		}
 	})
 
-	ginkgo.It("should work for type=Loadbalancer", func() {
-		configData := configFile{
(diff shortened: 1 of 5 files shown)


fedepaol:
LGTM

PR #969, opened 2021-09-29
FRR support for BGP mode deployments

cgoncalves:
A couple of things in this pull request:

- Cherry-picked https://github.com/metallb/metallb/pull/958 and https://github.com/metallb/metallb/pull/935 on top of latest main branch.
- Fixed linter errors 
- Fixed open known issues in the FRR-based BGP implementation
- Added support to FRR-based deployments to dev-env
- Expanded CI to run test automation against an FRR-based deployment

PR #992, opened 2021-10-14
WIP FRR: IPv6 unicast BGP Peers support

msherif1234:
those changes are in top of #969


msherif1234:
#997 will include all v6 changes including the testing env changes

PR #997, opened 2021-10-18
Added IPv6 aggregation length and enable IPv6 ci lane.

msherif1234:
Those diffs are based off some of the work done in PR metallb#590.
BGP FRR stack will be used for IPv6 unicast BGP peering instead of adding MP support  to existing Metallb stack
Metallb operator changes https://github.com/metallb/metallb-operator/pull/99


fedepaol:
The change makes sense, it's backwards compatible in the sense that there's no current ipv6 support in BGP.
The only doubt I have is letting this in before ipv6 is supported in BGP.

internal/config/config.go
```diff
@@ -370,31 +374,45 @@ func parseBGPAdvertisements(ads []bgpAdvertisement, cidrs []*net.IPNet, communit
...
 		}
 
 		if rawAd.AggregationLength != nil {
 			ad.AggregationLength = *rawAd.AggregationLength
 		}
 		if ad.AggregationLength > 32 {
-			return nil, fmt.Errorf("invalid aggregation length %q", ad.AggregationLength)
+			return nil, fmt.Errorf("invalid aggregation length %q for IPv4", ad.AggregationLength)
+		}
+		if rawAd.AggregationLengthV6 != nil {
```
fedepaol, line 400: if the parameter is frr only, let's fail if the user tries to add it in legacy mode.

fedepaol:
Also, can you change the commit message of the first commit to explain the rationale behind this change? This https://chris.beams.io/posts/git-commit/ is a post that was added to the contributing guidelines.

e2etest/bgp_test.go
```diff
@@ -94,69 +97,66 @@ var _ = ginkgo.Describe("BGP", func() {
...
 
-			allNodes, err = cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
-			framework.ExpectNoError(err)
-		})
+		allNodes, err = cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+		framework.ExpectNoError(err)
 
-		ginkgo.It("should work for ExternalTrafficPolicy=Cluster", func() {
-			svc, _ := createServiceWithBackend(cs, f.Namespace.Name, corev1.ServiceExternalTrafficPolicyTypeCluster)
+		ginkgo.By("should work for ExternalTrafficPolicy=Cluster")
```
fedepaol, line 129: Why this change? By doing this, before / after each won't run. I'd rather two tables / have the externalpolicy as another parameter of the table.

e2etest/bgp_test.go
```diff
@@ -94,69 +97,66 @@ var _ = ginkgo.Describe("BGP", func() {
...
 
-			validateService(cs, svc, allNodes.Items)
-		})
+		validateService(cs, svc, allNodes.Items)
 
-		ginkgo.It("should work for ExternalTrafficPolicy=Local", func() {
-			svc, jig := createServiceWithBackend(cs, f.Namespace.Name, corev1.ServiceExternalTrafficPolicyTypeLocal)
-			err := jig.Scale(2)
-			framework.ExpectNoError(err)
+		ginkgo.By("should work for ExternalTrafficPolicy=Local")
```
fedepaol, line 139: Same

e2etest/bgp_test.go
```diff
@@ -94,69 +97,66 @@ var _ = ginkgo.Describe("BGP", func() {
...
-			epNodes, err := jig.ListNodesWithEndpoint() // Only nodes with an endpoint should be advertising the IP
-			framework.ExpectNoError(err)
+		epNodes, err := jig.ListNodesWithEndpoint() // Only nodes with an endpoint should be advertising the IP
+		framework.ExpectNoError(err)
 
-			defer func() {
-				err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
-				framework.ExpectNoError(err)
-			}()
+		defer func() {
```
fedepaol, line 147: Here (and above) you are changing the scope of the defer to the outside context.

e2etest/bgp_test.go
```diff
@@ -94,69 +97,66 @@ var _ = ginkgo.Describe("BGP", func() {
 		}
 	})
 
-	ginkgo.Context("type=Loadbalancer", func() {
+	table.DescribeTable("set protocol loadbalancer", func(frrContainerIP *string, version string) {
```
fedepaol, line 100: I'd rename version to ipfamily.

e2etest/layer2_test.go
```diff
@@ -75,7 +75,7 @@ var _ = ginkgo.Describe("L2", func() {
 
 	ginkgo.Context("type=Loadbalancer", func() {
 		ginkgo.It("should work for ExternalTrafficPolicy=Cluster", func() {
-			svc, _ := createServiceWithBackend(cs, f.Namespace.Name, corev1.ServiceExternalTrafficPolicyTypeCluster)
+			svc, _ := createServiceWithBackend(cs, f.Namespace.Name, corev1.ServiceExternalTrafficPolicyTypeCluster, "layer2")
```
fedepaol, line 78: I'd keep this being the ip family

e2etest/utils.go
```diff
@@ -50,11 +50,17 @@ func wgetRetry(address string, exc executor.Executor) error {
...
 }
 
-func tweakServicePort(svc *v1.Service) {
+func tweakService(svc *v1.Service, version string) {
 	if servicePodPort != 80 {
 		// if servicePodPort is non default, then change service spec.
 		svc.Spec.Ports[0].TargetPort = intstr.FromInt(int(servicePodPort))
 	}
+
+	if version == "ipv6" {
```
fedepaol, line 59: are you sure this is needed? I would have said that creating a service in an ipv6 only cluster will create an ip6 service. Also, we have ipv6 tests in place in l2 and this was not required.

internal/config/config.go
```diff
@@ -370,31 +375,50 @@ func parseBGPAdvertisements(ads []bgpAdvertisement, cidrs []*net.IPNet, communit
...
 		}
 
 		if rawAd.AggregationLength != nil {
 			ad.AggregationLength = *rawAd.AggregationLength
 		}
 		if ad.AggregationLength > 32 {
-			return nil, fmt.Errorf("invalid aggregation length %q", ad.AggregationLength)
+			return nil, fmt.Errorf("invalid aggregation length %q for IPv4", ad.AggregationLength)
+		}
+		_, frrMode := os.LookupEnv("METALLB_BGP_TYPE")
```
fedepaol, line 401: Same as the other discussion, I'd just fill it and fail in the legacy config handler

e2etest/utils.go
msherif1234: it seems i don't need

speaker/bgp_controller.go
```diff
@@ -228,6 +228,17 @@ func (c *bgpController) SetBalancer(l log.Logger, name string, lbIP net.IP, pool
 	c.svcAds[name] = nil
 	for _, adCfg := range pool.BGPAdvertisements {
 		m := net.CIDRMask(adCfg.AggregationLength, 32)
+		if lbIP.To4() == nil {
+			if c.bgpType == bgpNative {
```
fedepaol, line 232: Here it might be too late. I'd do this check in https://github.com/metallb/metallb/blob/9cb1653a87673ec91971a60c873204e0752ff97b/speaker/main.go#L350 
where if we return an error the config gets staled and an alert is thrown.

e2etest/bgp_test.go
```diff
@@ -191,15 +198,21 @@ var _ = ginkgo.Describe("BGP", func() {
...
 				paired := frrIsPairedOnPods(cs, c)
 				framework.ExpectEqual(paired, true)
 			}
-		})
-	})
+		}
+	},
+		table.Entry("IPV4 - FRR ExternalTrafficPolicyCluster", "ipv4", "ExternalTrafficPolicyCluster"),
+		table.Entry("IPV4 - FRR ExternalTrafficPolicyLocal", "ipv4", "ExternalTrafficPolicyLocal"),
+		table.Entry("IPV4 - FRR Check speaker POD is running", "ipv4", "CheckSpeakerFRRPodRunning"),
```
fedepaol, line 212: Sorry, I was probably not clear in my previous comment. What I meant is to add FRR only in the name of the test that will work only in the FRR implementation (which is the one with `CheckSpeakerFRRPodRunning` where we check if FRR is running inside the pod).

All the others must not contain FRR or they'll be skipped even in the native lane, which is skipping everything with FRR in the test name.
Also, I'd rename `Check speaker POD is running` to `FRR running in the speaker POD`

msherif1234: `pairExternalFRRWithNodes` is invoked on all tests and will fail on native run

fedepaol: And that's why its entry (and only its entry) must contain FRR in the name. All the others should not otherwise the native will skip them (even if it makes sense to run it)

e2etest/bgp_test.go
```diff
@@ -82,11 +82,11 @@ var _ = ginkgo.Describe("BGP", func() {
 		{
 			name: frrEBGP,
 			nc: frrconfig.NeighborConfig{
-				ASN:      EBGPAsn,
+				ASN:      IBGPAsn,
```
fedepaol, line 85: Can you add a message to the commit explaining why we are removing the view support and what are the tradeoffs here?

.circleci/config.yml
```diff
@@ -42,7 +42,7 @@ jobs:
           if [ "<< parameters.bgp-type >>" = "native" ]; then SKIP="$SKIP\|FRR"; fi
           if [ "<< parameters.ip-family >>" = "ipv4" ]; then SKIP="$SKIP\|IPV6"; fi
           if [ "<< parameters.ip-family >>" = "dual" ]; then SKIP="$SKIP\|IPV6"; fi # TODO: until dual stack is supported, we can't run the ipv6 tests against the cluster
-          if [ "<< parameters.ip-family >>" = "ipv6" ]; then SKIP="$SKIP\|IPV4\|BGP"; fi # TODO: we are currently skipping BGP for ipv6, re enable when we enable the support
+          if [ "<< parameters.ip-family >>" = "ipv6" ] && [ "<< parameters.bgp-type >>" = "native" ]; then SKIP="$SKIP\|IPV4\|BGP"; fi # we are skipping BGP for ipv6 for native bgp stack
```
fedepaol, line 45: Can you split this commit in two, one for enabling CI and one for adding support to dev-env?

e2etest/bgp_test.go
```diff
@@ -95,11 +95,13 @@ var _ = ginkgo.Describe("BGP", func() {
 	}
 	ginkgo.BeforeEach(func() {
 		var err error
-		// TODO: When adding support for bgp + IPv6 need to validate hostIPv6 as well
 		if containersNetwork == "host" {
 			if net.ParseIP(hostIPv4) == nil {
 				framework.Fail("Invalid hostIPv4")
 			}
+			if net.ParseIP(hostIPv6) == nil {
```
fedepaol, line 102: Can you change the commit message saying that we enable v6 tests here?

e2etest/bgp_test.go
```diff
@@ -469,7 +502,9 @@ func validateFRRPeeredWithNodes(cs clientset.Interface, c *frrcontainer.FRR) {
 	}, 2*time.Minute, 1*time.Second).Should(BeNil())
 }
 
-func validateService(cs clientset.Interface, svc *corev1.Service, nodes []corev1.Node, c *frrcontainer.FRR) {
+func validateService(cs clientset.Interface, svc *corev1.Service, nodes []corev1.Node, c *frrcontainer.FRR, ipFamily string) {
+	var iProutes frr.Route
```
fedepaol, line 506: nit: ipRoutes

dev-env/testsvc_ipv6.yaml
```diff
@@ -0,0 +1,37 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
```
fedepaol, line 3: Can you change the commit message explaining why we need this change and the fact that is frr only?

speaker/main.go
```diff
@@ -352,6 +367,11 @@ func (c *controller) SetConfig(l log.Logger, cfg *config.Config) k8s.SyncState {
 		return k8s.SyncStateError
 	}
 
+	if !c.bgpImplementationSupportAggrLenV6Cfg(cfg.Pools) {
+		level.Error(l).Log("op", "setConfig", "error", "unsupported configuration", "msg", "configuration is missing, MetalLB will not function")
```
fedepaol, line 371: Can you extend the error with the reason? It will be easier for the users to self debug.
Also, I'd reword this to "configuration is ignored"

speaker/main.go
```diff
@@ -352,6 +367,11 @@ func (c *controller) SetConfig(l log.Logger, cfg *config.Config) k8s.SyncState {
 		return k8s.SyncStateError
 	}
 
+	if !c.bgpImplementationSupportAggrLenV6Cfg(cfg.Pools) {
+		level.Error(l).Log("op", "setConfig", "error", "unsupported configuration", "msg", "configuration is ignored, MetalLB native implementation is configure with AggregationLengthV6")
```
fedepaol, line 371: nit: is configured with

speaker/main.go
```diff
@@ -343,6 +345,19 @@ func poolFor(pools map[string]*config.Pool, ip net.IP) string {
 	return ""
 }
 
+func (c *controller) bgpImplementationSupportAggrLenV6Cfg(pools map[string]*config.Pool) bool {
+	if c.bgpType == bgpNative {
+		for _, p := range pools {
+			for _, adCfg := range p.BGPAdvertisements {
+				if adCfg.AggregationLengthV6 != 0 && adCfg.AggregationLengthV6 != 128 {
```
fedepaol, line 352: This would make 128 a valid value (even for the native implementation), which is not desirable. 
I'd rather apply the default in `SetConfig` and leave it empty when we read the config. 

Something like a "SetFRRDefaults" that we call over the configuration object if c.bgpType == bgpNative.

It's not symmetric with what we do with AggregationLength, but I think it's cleaner and we'll need it for other frr-only values.

msherif1234: this won't work becasue there is a little check in the config to make sure sure this length isn't less than cidr mask len, I will think about away out

fedepaol: This is as ugly. 
Let's not fail if the user sets a v6 aggregation lenght, let's fail if they set a v6 cidr in native mode.

msherif1234: there were many CI test cases when I just check for V6 I will defer this check for future PR instead of causing functional breakage or CI issues

fedepaol:
LGTM

PR #1008, opened 2021-10-28
Allow running FRR container on host network when running e2e tests 

sabinaaledort:
We want to be able to run the tests against a real cluster and not only against kind, and to use podman too to run the e2e tests.

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -48,6 +48,7 @@ const frrContainer = "frr"
 var (
 	frrContainerIPv4 string
 	frrContainerIPv6 string
+	frrTestConfigDir string
 )
 
 var _ = ginkgo.Describe("BGP", func() {
@@ -58,20 +59,23 @@ var _ = ginkgo.Describe("BGP", func() {
 		cs = f.ClientSet
 
 		var err error
-		// If the FRR container is not running, start it on the local host.
-		if skipDockerCmd {
-			if len(frrTestConfigDir) == 0 {
-				framework.Fail("Missing FRR config directory.")
-			}
-		} else {
-			frrTestConfigDir, err = ioutil.TempDir("", "frr-conf")
-			framework.ExpectNoError(err)
-			err = frr.StartContainer(frrContainer, frrTestConfigDir)
+		frrTestConfigDir, err = ioutil.TempDir("", "frr-conf")
+		framework.ExpectNoError(err)
+		err = frr.StartContainer(frrContainer, frrTestConfigDir, containerNetwork)
+		framework.ExpectNoError(err)
+
+		frrContainerIPv4 = hostIPv4
+		frrContainerIPv6 = hostIPv6
+
+		if containerNetwork != "host" {
+			frrContainerIPv4, frrContainerIPv6, err = frr.GetContainerIPs(frrContainer)
 			framework.ExpectNoError(err)
 		}
 
-		frrContainerIPv4, frrContainerIPv6, err = frr.GetContainerIPs(frrContainer)
-		framework.ExpectNoError(err)
+		// TODO: When adding support for bgp + IPv6 need to validate frrContainerIPv6 as well
+		if net.ParseIP(frrContainerIPv4) == nil {
+			framework.Fail("Invalid frrContainerIPv4")
+		}
 
 		err = frr.UpdateContainerVolumePermissions(frrContainer)
 		framework.ExpectNoError(err)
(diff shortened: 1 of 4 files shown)


e2etest/e2etest_suite_test.go
```diff
@@ -98,8 +103,20 @@ var _ = ginkgo.BeforeSuite(func() {
 		testNameSpace = ns
 	}
 
-	if dir := os.Getenv("FRR_CONFIG_DIR"); len(dir) != 0 {
-		frrTestConfigDir = dir
+	if net := os.Getenv("FRR_CONTAINER_NETWORK"); len(net) != 0 {
```
fedepaol, line 106: You can use os.LookupEnv here. Also, would it make sense to have it as a boolean flag instead of a string?

e2etest/e2etest_suite_test.go
```diff
@@ -98,8 +103,20 @@ var _ = ginkgo.BeforeSuite(func() {
...
+		containerNetwork = net
+		if containerNetwork == "host" {
+			if ip := os.Getenv("PROVISIONING_HOST_EXTERNAL_IPv4"); len(ip) != 0 {
```
fedepaol, line 109: just put it all uppercase

fedepaol: no need to be under containerNetwork == "host", no need to check the ip (at least here). I'd do the full validation here https://github.com/metallb/metallb/pull/1008/files#diff-a115369e9006a64f1497f61daada01a4136fa15a1feb79f86f11a73a0cdc5d36R68 and fail if some bits are missing

e2etest/e2etest_suite_test.go
```diff
@@ -40,6 +40,8 @@ import (
 )
 
 const defaultTestNameSpace = "metallb-system"
+const defaultContainerNetwork = "kind"
```
fedepaol, line 43: you can use const (

e2etest/pkg/executor/executor.go
```diff
@@ -17,16 +17,20 @@ func (hostExecutor) Exec(cmd string, args ...string) (string, error) {
 	return string(out), err
 }
 
-func ForContainer(containerName string) Executor {
-	return &containerExecutor{container: containerName}
+func ForContainer(containerName string, containerRuntime string) Executor {
```
fedepaol, line 20: I'd rather have a higher level object that returns Executors for a given runtime and having one global of these, or consuming the global containerRuntime internally here.
I don't think the runtime should be part of this since it will always be one on the same e2e test run.

e2etest/bgp_test.go
```diff
@@ -58,22 +59,20 @@ var _ = ginkgo.Describe("BGP", func() {
...
-		if skipDockerCmd {
-			if len(frrTestConfigDir) == 0 {
-				framework.Fail("Missing FRR config directory.")
-			}
+		frrTestConfigDir, err = ioutil.TempDir("", "frr-conf")
+		framework.ExpectNoError(err)
+		err = frr.StartContainer(frrContainer, frrTestConfigDir, containerNetwork, containerRuntime)
+		framework.ExpectNoError(err)
+
+		if containerNetwork == "host" {
```
fedepaol, line 67: Initialize them with the host and override them if network != host, it will avoid the else.

fedepaol:
Re: commit messages. Can you add the rationale behind applying the changes?
I think it will help adding context.
I.e. we want to run on host network so we can run the tests against a real cluster and not against kind
We want to be able to use podman too to run the tests

e2etest/pkg/executor/executor.go
```diff
@@ -26,7 +31,18 @@ type containerExecutor struct {
...
 func (e *containerExecutor) Exec(cmd string, args ...string) (string, error) {
+	containerRuntime := GetContainerRuntime()
+
 	newArgs := append([]string{"exec", e.container, cmd}, args...)
-	out, err := exec.Command("docker", newArgs...).CombinedOutput()
+	out, err := exec.Command(containerRuntime, newArgs...).CombinedOutput()
 	return string(out), err
 }
+
+func GetContainerRuntime() string {
```
fedepaol, line 41: you can move this to a 
```
func init() 
```
and avoid accessing the env variables each time.

e2etest/pkg/frr/container.go
```diff
@@ -23,16 +23,18 @@ const (
 	frrMountPath = "/etc/frr"
 )
 
+var containerRuntime = executor.GetContainerRuntime()
```
fedepaol, line 26: Let's have a `Runtime` global variable here, sounds better.
We init it in an init() function and we refer it in the executor.

sabinaaledort: maybe better keep it in the executor file? as we already import the executor to this file

fedepaol: Ah yes, in that case it won't work because go does not allow circular dependencies

e2etest/pkg/executor/executor.go
```diff
@@ -2,15 +2,30 @@
...
-var Host hostExecutor
+const defaultContainerRuntime = "docker"
+
+var (
+	Host             hostExecutor
+	ContainerRuntime string
+)
+
+func init() {
+	ContainerRuntime = defaultContainerRuntime
```
fedepaol, line 24: Just have `var ContainerRuntime = "docker"` outside

fedepaol:
LGTM
Let's wait for CI

PR #1011, opened 2021-10-29
fix invalid IP range check for exactly one IP

Pumba98:
Hi, the check introduced in 0.11.0 leads to a problem if you specifiy exactly one IP in your metallb IP pool

`{"caller":"level.go:63","configmap":"metallb-system/metallb","error":"parsing address pool #1: invalid CIDR \"192.168.1.210-192.168.1.210\" in pool \"default\": invalid IP range \"192.168.1.210-192.168.1.210\": start IP \"192.168.1.210\" is after the end IP \"192.168.1.210\"","event":"configStale","level":"error","msg":"config (re)load failed, config marked stale","ts":"2021-10-29T17:54:47.659792734Z"}`

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -459,7 +459,7 @@ func ParseCIDR(cidr string) ([]*net.IPNet, error) {
 		return nil, fmt.Errorf("invalid IP range %q: invalid end IP %q", cidr, fs[1])
 	}
 
-	if bytes.Compare(start, end) >= 0 {
+	if bytes.Compare(start, end) > 0 {
 		return nil, fmt.Errorf("invalid IP range %q: start IP %q is after the end IP %q", cidr, start, end)
 	}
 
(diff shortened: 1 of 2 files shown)


fedepaol:
Thanks for catching this!
Would you mind extending the unit tests with this case too so we avoid regressing again in the future? Should be an entry in the table under `config_test.go`

I'd also reword the commit message to "allow CIDR range to be a single IP" or something along that line, explaining why is needed (and what was broken) in the body.

russellb:
Thanks, and sorry for the regression!

russellb:
I'll do a v0.11.1 release with this fix soon.  In the meantime, if anyone hits this, a workaround would be to configure the single IPs in CIDR notation, like `192.168.1.210/32`.

Pumba98:
Yeah that's a working solution, noticed that too after opening the PR, I had a kinda weird/uncommon configuration on my side 😄

PR #1012, opened 2021-10-29
WIP: Tests debuggability improvements

fedepaol:
This pr adds a serie of small improvements in e2e tests
- using different names for different tests
- dumping more information in case of failure
- moving the gauge check under an eventually loop


fedepaol:
I want to see what happens with real failures, keeping it not mergeable in the meanwhile

PR #1014, opened 2021-10-30
Metallb FRR integration

fedepaol:
This PR connects the dots in FRR integration by collecting the remaining bits from https://github.com/metallb/metallb/pull/969 and:

- adding the reloader to the speaker image
- implementing the reload function
- adding a metallb_frr.yaml manifest to deploy metallb in frr experimental mode
- adding a CI lane that tests bgp in frr mode

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -27,18 +27,23 @@ jobs:
     parameters:
       ip-family:
         type: string
+      bgp-type:
+        type: string
     machine:
       image: ubuntu-2004:202104-01
     steps:
       - setup_ci_image
       - setup_kind
       - run: inv build
-      - run: inv dev-env -i  << parameters.ip-family >>
+      - run: inv dev-env -i  << parameters.ip-family >> -b  << parameters.bgp-type >>
       - run: |
-          if [ "<< parameters.ip-family >>" = "ipv4" ]; then SKIP="--skip IPV6"; fi
-          if [ "<< parameters.ip-family >>" = "dual" ]; then SKIP="--skip IPV6"; fi # TODO: until dual stack is supported, we can't run the ipv6 tests against the cluster
-          if [ "<< parameters.ip-family >>" = "ipv6" ]; then SKIP="--skip IPV4\|BGP"; fi # TODO: we are currently skipping BGP for ipv6, re enable when we enable the support
-          inv e2etest $SKIP -e /tmp/kind_logs
+          SKIP="none"
+          if [ "<< parameters.bgp-type >>" = "frr" ]; then SKIP="$SKIP\|metrics"; fi # TODO: enable metrics tests when implementing in FRR
+          if [ "<< parameters.ip-family >>" = "ipv4" ]; then SKIP="$SKIP\|IPV6"; fi
+          if [ "<< parameters.ip-family >>" = "dual" ]; then SKIP="$SKIP\|IPV6"; fi # TODO: until dual stack is supported, we can't run the ipv6 tests against the cluster
+          if [ "<< parameters.ip-family >>" = "ipv6" ]; then SKIP="$SKIP\|IPV4\|BGP"; fi # TODO: we are currently skipping BGP for ipv6, re enable when we enable the support
+          echo "Skipping $SKIP"
+          inv e2etest --skip $SKIP -e /tmp/kind_logs
       - store_artifacts:
           path: /tmp/kind_logs
   lint-1-16:
@@ -126,6 +131,7 @@ workflows:
           matrix:
             parameters:
               ip-family: [ipv4, ipv6, dual]
+              bgp-type: [native, frr]
       - publish-images:
           filters:
             branches:
--- a/internal/bgp/frr/config.go
+++ b/internal/bgp/frr/config.go
@@ -7,10 +7,13 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"strconv"
+	"syscall"
 	"text/template"
 )
 
-var configFileName = "/etc/frr/frr.conf"
+var configFileName = "/etc/frr_reloader/frr.conf"
+var reloaderPidFileName = "/etc/frr_reloader/reloader.pid"
 
 // TODO: We will probably need to update this template when we start to
 // integrate with FRR. The current template is a reasonable first pass
@@ -103,8 +106,28 @@ func writeConfig(config string, filename string) error {
 
 // reloadConfig requests that FRR reloads the configuration file. This is
 // called after updating the configuration.
-func reloadConfig() error {
-	// TODO
+var reloadConfig = func() error {
+	pidFile, found := os.LookupEnv("FRR_RELOADER_PID_FILE")
+	if found {
+		reloaderPidFileName = pidFile
+	}
+
+	pid, err := os.ReadFile(reloaderPidFileName)
+	if err != nil {
+		return err
+	}
+
+	pidInt, err := strconv.Atoi(string(pid))
+	if err != nil {
+		return err
+	}
+
+	// send HUP signal to FRR reloader
+	err = syscall.Kill(pidInt, syscall.SIGHUP)
+	if err != nil {
+		return err
+	}
+
 	return nil
 }
 
--- a/internal/bgp/frr/config.go
+++ b/internal/bgp/frr/config.go
@@ -38,7 +38,16 @@ router bgp {{.MyASN}}
   neighbor {{.Addr}} port {{.Port}}
 {{- end }}
 {{range $n := .Neighbors -}}
-{{range .Advertisements }}
+{{/* no bgp default ipv4-unicast prevents peering if no address families are defined. We declare an ipv4 one for the peer to make the pairing happen */}}
+{{- if eq (len .Advertisements) 0}}
+  address-family ipv4 unicast
+    neighbor {{$n.Addr}} activate
+  exit-address-family
+  address-family ipv6 unicast
+    neighbor {{$n.Addr}} activate
+  exit-address-family
+{{- end}}
+{{- range .Advertisements }}
   address-family {{.Version}} unicast
     neighbor {{$n.Addr}} activate
     network {{.Prefix}}
--- a/internal/bgp/frr/testdata/TestSingleAdvertisementInvalidPrefix.golden
+++ b/internal/bgp/frr/testdata/TestSingleAdvertisementInvalidPrefix.golden
@@ -11,5 +11,11 @@ router bgp 100
   neighbor 10.2.2.254 remote-as 200
   neighbor 10.2.2.254 port 179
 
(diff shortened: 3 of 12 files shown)


fedepaol:
Double checked the artifacts, at least the frr-ipv4 lane is really running in frr mode

internal/bgp/frr/config.go
```diff
@@ -114,35 +117,39 @@ func writeConfig(config string, filename string) error {
...
-
-	// send HUP signal to FRR reloader
-	err = syscall.Kill(pidInt, syscall.SIGHUP)
-	if err != nil {
-		return err
-	}
-
-	return nil
+var reloadConfig = func(sf *singleflight.Group) error {
+	_, err, _ := sf.Do("reload",
```
oribon, line 121: nice!

internal/bgp/frr/testdata/TestSingleAdvertisementInvalidPrefix.golden
```diff
@@ -11,5 +11,10 @@ router bgp 100
   neighbor 10.2.2.254 remote-as 200
   neighbor 10.2.2.254 port 179
 
+  address-family ipv4 unicast
+    neighbor 10.2.2.254 activate
+  address-family ipv6 unicast
+    neighbor 10.2.2.254 activate
```
msherif1234, line 17: this isn't right to active v4 neigh from v6 AF ?

internal/bgp/frr/config.go
```diff
@@ -35,7 +41,15 @@ router bgp {{.MyASN}}
...
 {{- end }}
 {{range $n := .Neighbors -}}
-{{range .Advertisements }}
+{{/* no bgp default ipv4-unicast prevents peering if no address families are defined. We declare an ipv4 one for the peer to make the pairing happen */}}
+{{- if eq (len .Advertisements) 0}}
+  address-family ipv4 unicast
+    neighbor {{$n.Addr}} activate
+  address-family ipv6 unicast
+    neighbor {{$n.Addr}} activate
+  exit-address-family
```
msherif1234, line 50: is this for testing ? we can't have bgp peer with having addresspool config if that happened it should have caught eariler or should have been caught if that isn't the case ?

internal/bgp/frr/config.go
```diff
@@ -103,15 +117,39 @@ func writeConfig(config string, filename string) error {
...
+			}
+
+			// send HUP signal to FRR reloader
+			err = syscall.Kill(pidInt, syscall.SIGHUP)
+			if err != nil {
+				return nil, err
+			}
+			return nil, nil
+		})
+	return err
```
msherif1234, line 146: this is interesting package so the trick here is the sleep 2s u wanted to prevent doing back2back reload for 2sec window ?

internal/bgp/frr/config.go
fedepaol: That's to mimic the same behaviour we have now.
Took me a while to find it. Without a service, frr won't establish a connection with the peer because of the `no bgp default ipv4-unicast`.

"By default, only the IPv4 unicast address family is announced to all neighbors. Using the ‘no bgp default ipv4-unicast’ configuration overrides this default so that all address families need to be enabled explicitly."
So without this, FRR will complain of no address families configured.

internal/bgp/frr/config.go
fedepaol: Single flight squash all the calls that happen while one is in progress.
Adding the sleep will make the call last longer and will squash all the configuration requests into a single one (with the drawback of 2 seconds timeout). As a second thought, I can make the timeout even shorter since we care about squashing the calls caused by the same config change.

internal/bgp/frr/config.go
msherif1234: that was my question I did noticed too if no service no bgp neighbors I couldn't answer if this is an ok thing or not why do we care of having bgp peering if we don't have any network to advertise , I don't know hence asked that queston

internal/bgp/frr/testdata/TestSingleAdvertisementInvalidPrefix.golden
fedepaol: I wasn't sure about this, but I think it's possible to advertise an ip6 family to a neighbour reachable via ipv4 (and viceversa) ?

internal/bgp/frr/config.go
msherif1234: does it squash or block till on flight call completes ?

internal/bgp/frr/testdata/TestSingleAdvertisementInvalidPrefix.golden
fedepaol: It would be the same as https://github.com/metallb/metallb/pull/1014/files/b88845f28a5aa8ac430c75678d8414dec856ea14#diff-1d9a33c302812585b1b2d1897f84126492129d1f9415786a73dcfa26603686c9R53 (but we need to validate it).

msherif1234: I understand the config just I doubt if it will work as is i have seen some doc about how to get this to work 
```
Enable
Configure terminal
Route-Map IPV6 permit 10
Set ipv6 next-hop FE80::F816:3EFF:FE40:2985
Exit
Router bgp 1
Neighbor 2.0.0.2 remote-as 2
Address-family ipv6 unicast
Neighbor 2.0.0.2 activate
Neighbor 2.0.0.2 route-map IPV6 out
Network 2002:2::2/128
```
might be way more involved than u think

fedepaol: This part is for enabling the session only (address-family without routes) to avoid the issue described here https://github.com/metallb/metallb/pull/1014#discussion_r740421183

I think we'll need to hammer our way through v6 / dual stack anyway, so I can either remove it or leave it as it is since it's not harming v4 e2e tests.

internal/bgp/frr/config.go
fedepaol: I am not sure it's right or not, but the current implementation does that and it did not seem bad to replicate that.
Also, that would be the default frr behaviour if the `no bgp default ipv4-unicast` parameter (which is needed to allow both ipv4 and ipv6 to co-exist) wasn't added.

internal/bgp/frr/config.go
fedepaol: The point about blocking makes total sense and I did not think about it. This will be likely to be called many subsequent times from the single goroutine, so the effect will be of delaying instead of really squashing them.
Let me put a goroutine in the middle so it will sort out the desired effect.

msherif1234: is there any possibility for this working async now that we could fire frr reload with hlaf baked config ? from what I have seen FRR will drop the entire config if one line is incorrect, just thinking loud here since we don't have any idea when the reload took effect debugging invalid config will be interesting

fedepaol: Configuration updates happen in discrete steps, and each step is supposed to provide a valid config (not meaning that we won't have bugs). What I am saying is that we won't produce a half baked config, we'll produce a config with less stuff that the caller expects:

What could happen here, is that the timeout falls in the middle (i.e. we are looping here https://github.com/metallb/metallb/blob/9cb1653a87673ec91971a60c873204e0752ff97b/speaker/bgp_controller.go#L172 and the threshold happens half way through https://github.com/metallb/metallb/blob/9cb1653a87673ec91971a60c873204e0752ff97b/speaker/bgp_controller.go#L269 )

With that, we would have less addresses announced, but the configuration (bugs aside) will still be valid.

In any case, I am torn about this because as it is now it's just introducing a delay and not working asynchronously and I don't want this to hold the integration to get in. Will remove the commit and rework it in a separate PR, using goroutines.

fedepaol: Removed the reloading optimization, will handle it in a separate PR

tasks.py
```diff
@@ -6,6 +6,7 @@
 import sys
 import yaml
 import tempfile
+
```
russellb, line 9: Can you please re-do this commit without all of the unrelated formatting changes?  @cgoncalves

tasks.py
```diff
@@ -251,9 +251,12 @@ def validate_kind_version():
     "node_img": "Optional node image to use for the kind cluster (e.g. kindest/node:v1.18.19)."
                 "The node image drives the kubernetes version used in kind.",
     "ip_family": "Optional ipfamily of the cluster."
-                 "Default: ipv4, supported families are 'ipv6' and 'dual'."
+                 "Default: ipv4, supported families are 'ipv6' and 'dual'.",
+    "bgp_type": "Type of BGP implementation to use."
+                "Supported: 'native' (default), 'frr'"
```
russellb, line 256: could have also been `-p bgp-native` and `-p bgp-frr` instead of a new option I guess, but this works too

fedepaol: We want to be able to differentiate, because CI now doesn't care about the protocol.
Using -p will add a pre-canned configuration which will interfere with the test execution.

russellb: good point, thanks

.circleci/config.yml
```diff
@@ -27,18 +27,23 @@ jobs:
...
       - run: inv build
-      - run: inv dev-env -i  << parameters.ip-family >>
+      - run: inv dev-env -i  << parameters.ip-family >> -b  << parameters.bgp-type >>
       - run: |
-          if [ "<< parameters.ip-family >>" = "ipv4" ]; then SKIP="--skip IPV6"; fi
-          if [ "<< parameters.ip-family >>" = "dual" ]; then SKIP="--skip IPV6"; fi # TODO: until dual stack is supported, we can't run the ipv6 tests against the cluster
-          if [ "<< parameters.ip-family >>" = "ipv6" ]; then SKIP="--skip IPV4\|BGP"; fi # TODO: we are currently skipping BGP for ipv6, re enable when we enable the support
-          inv e2etest $SKIP -e /tmp/kind_logs
+          SKIP="none"
+          if [ "<< parameters.bgp-type >>" = "frr" ]; then SKIP="$SKIP\|metrics"; fi # TODO: enable metrics tests when implementing in FRR
```
cgoncalves, line 41: I'm not convinced with this SKIP approach for two reasons:
1. hard to human-parse and likely not scaling that well
2. it will run CI jobs that all they do is setup the environment and skip all tests, while reporting as "succeed" giving wrong idea that they work and are supported.

Can't we add invalid/unsupported topologies to the exclude list?

cgoncalves:
Ugh, PR approved while reviewing it.

.circleci/config.yml
fedepaol: Those are temporary. My expectation is that we are gonna remove the skips in a matter of days, and for the metric thing it does not make any sense to add a new combination to skip.

PR #1016, opened 2021-10-30
Add Dockertest support

fedepaol:
In order to validate the correctness of the frr configuration files, we spin up a frr container and try to reload the file produced by the unit tests to ensure it's valid.


oribon:
lgtm

russellb:
How about adding a test case that feeds this check an intentionally invalid config file? That way we'll make sure the check fails properly, too.

fedepaol:
> How about adding a test case that feeds this check an intentionally invalid config file? That way we'll make sure the check fails properly, too.

Added a wip commit for validation, but now that I read the comment again you are asking for something else.
Will work on it

fedepaol:
@russellb flakes aside, I think I addressed your comment

russellb:
> @russellb flakes aside, I think I addressed your comment

thanks

PR #1017, opened 2021-10-30
Add keepalivetime configuration option to metallb

msherif1234:
Added BGP peer keepalivetime configuration option
the current metallb stack assumes its always 1/3 of the holdtime while this the recommended default setting by rfc4271 but it doesn't have to stay like that, this change export the config option so FRR can make use of it while not changing the current
keepalive alg used by Metallb native stack.

### DIFF
--- a/internal/bgp/frr/frr.go
+++ b/internal/bgp/frr/frr.go
@@ -27,6 +27,7 @@ type session struct {
 	srcAddr        net.IP
 	asn            uint32
 	holdTime       time.Duration
+	keepaliveTime  time.Duration
 	logger         log.Logger
 	password       string
 	advertised     []*bgp.Advertisement
@@ -108,7 +109,7 @@ func (s *session) Close() error {
 //
 // The session will immediately try to connect and synchronize its
 // local state with the peer.
-func (sm *sessionManager) NewSession(l log.Logger, addr string, srcAddr net.IP, myASN uint32, routerID net.IP, asn uint32, holdTime time.Duration, password string, myNode string) (bgp.Session, error) {
+func (sm *sessionManager) NewSession(l log.Logger, addr string, srcAddr net.IP, myASN uint32, routerID net.IP, asn uint32, holdTime time.Duration, keepaliveTime time.Duration, password string, myNode string) (bgp.Session, error) {
 	s := &session{
 		myASN:          myASN,
 		routerID:       routerID,
@@ -117,6 +118,7 @@ func (sm *sessionManager) NewSession(l log.Logger, addr string, srcAddr net.IP, 
 		srcAddr:        srcAddr,
 		asn:            asn,
 		holdTime:       holdTime,
+		keepaliveTime:  keepaliveTime,
 		logger:         log.With(l, "peer", addr, "localASN", myASN, "peerASN", asn),
 		password:       password,
 		advertised:     []*bgp.Advertisement{},
@@ -206,7 +208,7 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
 				Addr:           host,
 				Port:           uint16(portUint),
 				HoldTime:       uint64(s.holdTime / time.Second),
-				KeepaliveTime:  uint64(s.holdTime / (3 * time.Second)), // TODO use 1/3 of holdtime till we can configure it.
+				KeepaliveTime:  uint64(s.keepaliveTime / time.Second),
 				Password:       s.password,
 				Advertisements: make([]*advertisementConfig, 0),
 			}
--- a/internal/bgp/frr/frr_test.go
+++ b/internal/bgp/frr/frr_test.go
@@ -82,7 +82,7 @@ func TestSingleSession(t *testing.T) {
 
 	l := log.NewNopLogger()
 	sessionManager := NewSessionManager()
-	session, err := sessionManager.NewSession(l, "10.2.2.254:179", net.ParseIP("10.1.1.254"), 100, net.ParseIP("10.1.1.254"), 200, time.Second, "password", "hostname")
+	session, err := sessionManager.NewSession(l, "10.2.2.254:179", net.ParseIP("10.1.1.254"), 100, net.ParseIP("10.1.1.254"), 200, time.Second, time.Second, "password", "hostname")
 	if err != nil {
 		t.Fatalf("Could not create session: %s", err)
 	}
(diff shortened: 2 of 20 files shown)


internal/config/config.go
```diff
@@ -192,6 +195,18 @@ func parseHoldTime(ht string) (time.Duration, error) {
 	return rounded, nil
 }
 
+func parseKeepaliveTime(ka string) (time.Duration, error) {
+	if ka == "" {
```
fedepaol, line 199: Can the default logic be the same as the current stack (you mentioned it being 1/3 of holdtime)?

internal/bgp/native/native.go
```diff
@@ -64,19 +65,20 @@ func NewSessionManager() *sessionManager {
...
+		addr:          addr,
+		srcAddr:       srcAddr,
+		myASN:         myASN,
+		routerID:      routerID.To4(),
+		myNode:        myNode,
+		asn:           asn,
+		holdTime:      holdTime,
+		keepaliveTime: keepaliveTime,
+		logger:        log.With(l, "peer", addr, "localASN", myASN, "peerASN", asn),
+		newHoldTime:   make(chan bool, 1),
```
fedepaol, line 79: If the value is not supported with the native stack, we should fail in parsing the configuration.

internal/bgp/frr/frr.go
```diff
@@ -108,7 +109,7 @@ func (s *session) Close() error {
 //
 // The session will immediately try to connect and synchronize its
 // local state with the peer.
-func (sm *sessionManager) NewSession(l log.Logger, addr string, srcAddr net.IP, myASN uint32, routerID net.IP, asn uint32, holdTime time.Duration, password string, myNode string) (bgp.Session, error) {
+func (sm *sessionManager) NewSession(l log.Logger, addr string, srcAddr net.IP, myASN uint32, routerID net.IP, asn uint32, holdTime time.Duration, keepaliveTime time.Duration, password string, myNode string) (bgp.Session, error) {
```
fedepaol, line 112: Should we have also the configuration rendering?

internal/config/config.go
msherif1234: current stack doesn't configure it and the default is not configured is 30sec which is 1/3 of the default holdtime

internal/bgp/frr/frr.go
msherif1234: yeah I have that config with hold time in #1013

internal/bgp/native/native.go
msherif1234: ok

internal/config/config.go
fedepaol: > current stack doesn't configure it and the default is not configured is 30sec which is 1/3 of the default holdtime

so it's always 30 sec or it's 1/3 of the holdtime?

msherif1234: the default holdtime is 90s hence default ka is 30s if u changed the holdtime u need to change the ka value and at this point it does't have to be 1/3 it can be anything based on what you are peering with

fedepaol: Yes, but if the user specifies something different as HoldTime, the keep alive time is gonna always be 1/3 (https://github.com/metallb/metallb/blob/9cb1653a87673ec91971a60c873204e0752ff97b/internal/bgp/native/native.go#L363). 
So, to preserve the current behaviour, we should do the same. If a user specifies the HoldTime but not the keep alive time, we should use HT/3

msherif1234: at the description of this PR I said metallb it assumes its always 1/3 hence its not a config options this is not typical for other stacks they export it both timers to config with only one condition ka timer can't be less than ht

fedepaol: Not sure I am missing something. 
What I am saying here is, if a user specifies an HT but not a keep alive, let's use HT/3 as default instead of fixing it to 30 seconds, so one configuration that is working today will continue working when switching implementation.

internal/bgp/frr/frr.go
fedepaol: Fair enough, we are still in "stitching all together" mode.

internal/config/config.go
msherif1234: `neighbor PEER timers (0-65535) (0-65535)`
they are configured together so you either keep default for both or change both

fedepaol: They are configured together in FRR, not in MetalLB.
With the code of this PR, if a user sets heartbeat but not keepalive, keepalive is defaulted to 30 seconds.

We either force them to be configured both in MetalLB, or we use the same defaulting mechanism we had in place before, calculating keepalive to be heartbeat/3
I don't think it's fair to force it in metalLB since it would break the compatibility promise.

msherif1234: current stack completly ignore the keepalive settings this field in the config isn't user or referrenced anywhere in the metallb stack so the code will stll under the hood change ka to 1/3 of the new HT

internal/config/config.go
```diff
@@ -254,6 +274,7 @@ func Parse(bs []byte) (*Config, error) {
 }
 
 func parsePeer(p peer) (*Peer, error) {
+	_, frrMode := os.LookupEnv("METALLB_BGP_TYPE")
```
fedepaol, line 277: I'd avoid checking the env variable here, it's gonna be replaced by a flag (see https://github.com/metallb/metallb/pull/958#discussion_r723083723).

I think the best way would be to keep the config unaware of the implementation, and fail in the speakers `SetConfig` method: https://github.com/metallb/metallb/blob/9cb1653a87673ec91971a60c873204e0752ff97b/speaker/main.go#L347

specifically in the bgp handler: https://github.com/metallb/metallb/blob/9cb1653a87673ec91971a60c873204e0752ff97b/speaker/bgp_controller.go#L58

That would cause the parsing to fail, but the failure will happen in the frr-aware part (or un-aware) part. That would also require the default not to be set here but when we parse the config, in the bgpfrr side.

msherif1234: this is kiind of pretty early way b4 any parsing is done ? if we need this to be done in implementation independent way we may have to create interface for parsePeer() parseCommity()  parseAddresspool() only to handle the delta and keep the rest of the parsing common as is but that means the too stack will live together my understanding at somepoint we will just default to FRR so we need to be careful how much we need to invest on this atm we need only two additional knobs if that evolved we can look at more fancy approach not sure

fedepaol: I am just saying to move the check and the defaulting place to another function.
I'd say let's fill the config with all the parameters the user put in place, and defer to where we consume it to choose if it's good or not.
The parsing + consuming happens in the same stack of calls so it won't be a big deal from a UX point of view.

internal/bgp/frr/frr_test.go
```diff
@@ -82,7 +82,7 @@ func TestSingleSession(t *testing.T) {
 
 	l := log.NewNopLogger()
 	sessionManager := NewSessionManager()
-	session, err := sessionManager.NewSession(l, "10.2.2.254:179", net.ParseIP("10.1.1.254"), 100, net.ParseIP("10.1.1.254"), 200, time.Second, "password", "hostname")
+	session, err := sessionManager.NewSession(l, "10.2.2.254:179", net.ParseIP("10.1.1.254"), 100, net.ParseIP("10.1.1.254"), 200, time.Second, time.Second, "password", "hostname")
```
fedepaol, line 85: Can we make the holdtime and the keepalive time different so we know that we are rendering the right values?

fedepaol:
LGTM

fedepaol:
Filed https://github.com/metallb/metallb/issues/1047 , because for these (and for another set of parameters) we don't really know if they work e2e

PR #1018, opened 2021-10-31
Run tests with multiple configurable FRR containers

oribon:
Currently the tests run using one external BGP peer.
Running multiple external BGP peers allows us to cover
more scenarios in parallel such as internal/external BGP
and multiple MyASNs. In addition, doing it in a configurable
way will allow us to extend the tests to cover more scenarios
in the future by shuffling configuration for the speaker
and the external containers that serve as BGP peers.

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -21,7 +21,6 @@ package e2e
 import (
 	"context"
 	"fmt"
-	"io/ioutil"
 	"net"
 	"strconv"
 	"time"
@@ -29,13 +28,15 @@ import (
 	"go.universe.tf/metallb/e2etest/pkg/metrics"
 	"go.universe.tf/metallb/e2etest/pkg/routes"
 
+	"golang.org/x/sync/errgroup"
+
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/extensions/table"
 	. "github.com/onsi/gomega"
 	dto "github.com/prometheus/client_model/go"
-	"go.universe.tf/metallb/e2etest/pkg/executor"
 	"go.universe.tf/metallb/e2etest/pkg/frr"
 	frrconfig "go.universe.tf/metallb/e2etest/pkg/frr/config"
+	frrcontainer "go.universe.tf/metallb/e2etest/pkg/frr/container"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
@@ -43,41 +44,58 @@ import (
 	e2eservice "k8s.io/kubernetes/test/e2e/framework/service"
 )
 
-const frrContainer = "frr"
+const (
+	frrIBGP = "frr-iBGP"
+	frrEBGP = "frr-eBGP"
+	IBGPAsn = 64512
+	EBGPAsn = 64513
+)
 
 var (
-	frrContainerIPv4 string
-	frrContainerIPv6 string
-	frrTestConfigDir string
+	frrContainers []*frrcontainer.FRR
 )
+
+type containerConfig struct {
+	name string
+	nc   frrconfig.NeighborConfig
+	rc   frrconfig.RouterConfig
+}
 
 var _ = ginkgo.Describe("BGP", func() {
 	f := framework.NewDefaultFramework("bgp")
 	var cs clientset.Interface
-
+	containersConf := []containerConfig{
+		{
+			name: frrIBGP,
+			nc: frrconfig.NeighborConfig{
+				ASN: IBGPAsn,
+			},
+			rc: frrconfig.RouterConfig{
+				ASN:     IBGPAsn,
+				BGPPort: 179,
+			},
+		},
+		{
+			name: frrEBGP,
+			nc: frrconfig.NeighborConfig{
+				ASN: EBGPAsn,
+			},
+			rc: frrconfig.RouterConfig{
+				ASN:     IBGPAsn,
+				BGPPort: 180,
+			},
+		},
+	}
 	ginkgo.BeforeEach(func() {
-		cs = f.ClientSet
-
 		var err error
-		frrTestConfigDir, err = ioutil.TempDir("", "frr-conf")
-		framework.ExpectNoError(err)
-		err = frr.StartContainer(frrContainer, frrTestConfigDir, containerNetwork)
-		framework.ExpectNoError(err)
-
-		frrContainerIPv4 = hostIPv4
-		frrContainerIPv6 = hostIPv6
-
-		if containerNetwork != "host" {
-			frrContainerIPv4, frrContainerIPv6, err = frr.GetContainerIPs(frrContainer)
-			framework.ExpectNoError(err)
+		// TODO: When adding support for bgp + IPv6 need to validate hostIPv6 as well
+		if containersNetwork == "host" {
+			if net.ParseIP(hostIPv4) == nil {
+				framework.Fail("Invalid hostIPv4")
+			}
 		}
-
-		// TODO: When adding support for bgp + IPv6 need to validate frrContainerIPv6 as well
-		if net.ParseIP(frrContainerIPv4) == nil {
-			framework.Fail("Invalid frrContainerIPv4")
-		}
-
-		err = frr.UpdateContainerVolumePermissions(frrContainer)
+		cs = f.ClientSet
+		frrContainers, err = createFRRContainers(containersConf)
 		framework.ExpectNoError(err)
 	})
 
@@ -86,7 +104,7 @@ var _ = ginkgo.Describe("BGP", func() {
 		err := updateConfigMap(cs, configFile{})
 		framework.ExpectNoError(err)
 
-		err = frr.StopContainer(frrContainer, frrTestConfigDir)
+		err = stopFRRContainers(frrContainers)
 		framework.ExpectNoError(err)
 
 		if ginkgo.CurrentGinkgoTestDescription().Failed {
@@ -97,6 +115,15 @@ var _ = ginkgo.Describe("BGP", func() {
 	ginkgo.Context("type=Loadbalancer", func() {
 		var allNodes *corev1.NodeList
 		ginkgo.BeforeEach(func() {
+			var peers []peer
+			for _, c := range frrContainers {
+				peers = append(peers, peer{
+					Addr:  c.Ipv4,
+					ASN:   c.RouterConfig.ASN,
+					MyASN: c.NeighborConfig.ASN,
+					Port:  c.RouterConfig.BGPPort,
+				})
+			}
 			configData := configFile{
 				Pools: []addressPool{
 					{
@@ -108,18 +135,18 @@ var _ = ginkgo.Describe("BGP", func() {
 						},
 					},
 				},
-				Peers: []peer{
-					{
-						Addr:  frrContainerIPv4,
-						ASN:   64512,
-						MyASN: 64512,
-					},
-				},
+				Peers: peers,
 			}
+			for _, c := range frrContainers {
+				pairExternalFRRWithNodes(cs, c)
+			}
+
 			err := updateConfigMap(cs, configData)
 			framework.ExpectNoError(err)
 
-			pairExternalFRRWithNodes(cs)
+			for _, c := range frrContainers {
+				validateFRRPeeredWithNodes(cs, c)
+			}
 
 			allNodes, err = cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
 			framework.ExpectNoError(err)
@@ -133,7 +160,9 @@ var _ = ginkgo.Describe("BGP", func() {
 				framework.ExpectNoError(err)
 			}()
 
-			validateService(cs, svc, allNodes.Items)
+			for _, c := range frrContainers {
+				validateService(cs, svc, allNodes.Items, c)
+			}
 		})
 
 		ginkgo.It("should work for ExternalTrafficPolicy=Local", func() {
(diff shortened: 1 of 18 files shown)


e2etest/bgp_test.go
```diff
@@ -21,72 +21,107 @@ package e2e
...
-			}
-		} else {
-			frrTestConfigDir, err = ioutil.TempDir("", "frr-conf")
-			framework.ExpectNoError(err)
-			err = frr.StartContainer(frrContainer, frrTestConfigDir)
-			framework.ExpectNoError(err)
+		frrContainers = make([]*frrcontainer.Frr, 0)
+		g := new(errgroup.Group)
+		for _, conf := range containersConf {
+			conf := conf
```
sabinaaledort, line 94: seems odd

e2etest/pkg/frr/container/container.go
```diff
@@ -19,20 +20,78 @@ const (
...
+	configDir      string
+	NeighborConfig config.NeighborConfig
+	RouterConfig   config.RouterConfig
+	Ipv4           string
+	Ipv6           string
+}
+
+// New creates a new FRR container on the host and returns the corresponding *Frr.
+// A situation where a non-nil container and an error are returned is possible.
+func New(name string, nc config.NeighborConfig, rc config.RouterConfig, network string, hostipv4 string, hostipv6 string) (*Frr, error) {
```
sabinaaledort, line 41: maybe we can find a better name than new? when reading the test code it wasn't clear for me that this function is staring the container

e2etest/pkg/frr/container/container.go
```diff
@@ -42,50 +101,38 @@ func StartContainer(containerName string, testDirName string) error {
 }
 
 // Returns the IPv4 and IPv6 addresses of the container.
```
sabinaaledort, line 103: maybe we should update the comment as this function no longer returns the ips but setting them

e2etest/pkg/frr/container/container.go
```diff
@@ -111,9 +158,24 @@ func UpdateContainerVolumePermissions(containerName string) error {
...
+	out, err := exec.Command("docker", "exec", c.Name, "sh", "-c", cmd).CombinedOutput()
+	if err != nil {
+		return errors.Wrapf(err, "Failed to change %s container volume permissions. %s", c.Name, string(out))
+	}
+
+	return nil
+}
+
+// Updating the BGP config file.
+func (c *Frr) UpdateBGPConfigFile(bgpConfig string) error {
```
sabinaaledort, line 170: why did you move the function? might look confusing in the log

e2etest/bgp_test.go
oribon: https://golang.org/doc/faq#closures_and_goroutines

e2etest/pkg/frr/container/container.go
oribon: What do you suggest? For me it feels new container = start a container with additional stuff

e2etest/pkg/frr/container/container.go
oribon: done

e2etest/pkg/frr/container/container.go
oribon: done

e2etest/bgp_test.go
```diff
@@ -21,72 +21,107 @@ package e2e
...
+			},
+		},
+		{
+			name: frrEBGP,
+			nc: frrconfig.NeighborConfig{
+				ASN: EBGPAsn,
+			},
+			rc: frrconfig.RouterConfig{
+				ASN:     IBGPAsn,
+				BGPPort: 180,
```
msherif1234, line 85: hmm this BGPPort is different than the one /neighbor don't we need to keep this at 0 as we have today and just change the /neighbor port if configured ?
`When port number is 0, that means do not listen bgp port.`

e2etest/bgp_test.go
```diff
@@ -95,6 +130,15 @@ var _ = ginkgo.Describe("BGP", func() {
 	ginkgo.Context("type=Loadbalancer", func() {
 		var allNodes *corev1.NodeList
 		ginkgo.BeforeEach(func() {
+			var peers []peer
+			for _, c := range frrContainers {
+				peers = append(peers, peer{
+					Addr:  c.Ipv4,
+					ASN:   c.RouterConfig.ASN,
+					MyASN: c.NeighborConfig.ASN,
```
msherif1234, line 138: did u get the above swapped ?

e2etest/config/frr/daemons
```diff
@@ -38,7 +38,7 @@ vrrpd=no
 #
 vtysh_enable=yes
 zebra_options="  -A 127.0.0.1 -s 90000000"
-bgpd_options="   -A 127.0.0.1"
+bgpd_options="   -A 127.0.0.1 -p {{.BGPPort}}"
```
msherif1234, line 41: I don't think we need to set the above ?

e2etest/pkg/frr/config/config.go
```diff
@@ -18,30 +18,30 @@ import (
 // BGP router config.
 const bgpConfigTemplate = `
 router bgp {{.ASN}}
+  no bgp ebgp-requires-policy
```
msherif1234, line 21: curious what issues u faced to do the above I did bgp peering for both iBGP and eBGP w/o the need to this
I thought because we have `no bgp network import-check`, also I assume we are running as datacenter hence its off by default ? `This is enabled by default for the traditional configuration and turned off by default for datacenter configuration.`

e2etest/pkg/frr/config/config.go
```diff
@@ -105,3 +103,34 @@ func SetBGPConfig(testDirName string, config string) error {
...
+		BGPPort: rc.BGPPort,
+	}
+
+	err = tpl.Execute(f, info)
+	if err != nil {
+		return errors.Wrapf(err, "Failed to update %s", path)
+	}
+
+	return nil
+}
```
msherif1234, line 136: why we have this function here and its used only by test ?

e2etest/pkg/frr/container/container.go
```diff
@@ -19,20 +20,78 @@ const (
...
 	}
 
+	err = config.SetDaemonsConfig(testDirName, rc)
+	if err != nil {
+		return err
+	}
+
 	volume := fmt.Sprintf("%s:%s", testDirName, frrMountPath)
-	out, err := exec.Command("docker", "run", "-d", "--privileged", "--network", "kind", "--rm", "--name", containerName,
+	out, err := exec.Command("docker", "run", "-d", "--privileged", "--network", network, "--rm", "--name", containerName,
```
msherif1234, line 94: why do we need to run as privileged ?

msherif1234:
How did u test multiple MyASN w/oi using view feature in #1015 ?

e2etest/bgp_test.go
oribon: From the speaker's point of view he's the container's neighbor (MyASN should come from neighborconfig) and ASN,Addr corresponds to the RouterConfig and container's IP.

e2etest/bgp_test.go
oribon: This also relates to https://github.com/metallb/metallb/pull/1018#discussion_r740222489
that's the BGPPort that the external container will listen on so it is set here and reflected in the peer (c.RouterConfig.BGPPort) so the speaker would target the correct port.

e2etest/config/frr/daemons
oribon: ^^ this is to configure FRR to listen on a specified port, useful for checking the configmap's port option AND configuring 2 containers on the hostnetwork.

e2etest/pkg/frr/config/config.go
oribon: what do you mean?

e2etest/pkg/frr/container/container.go
oribon: that wasn't part of my commit :sweat_smile:

e2etest/bgp_test.go
msherif1234: what if we change the port later `neighbor 1.1.1.1 port 200` and what will happen if we have 
`neighbor 1.1.1.1 port 100`
`nieghbor 1.1.1.2 port 110`
which one will be used for BGPPort ?

e2etest/pkg/frr/config/config.go
msherif1234: just curious why you add it here not in the testfile itself

oribon:
> How did u test multiple MyASN w/oi using view feature in #1015 ?

It runs on the native bgp stack - it shows that it works with multiple MyASNs hence we'd want to have these tests pass when FRR with views land. The external containers here don't need to use views (they're using a single ASN).

e2etest/bgp_test.go
msherif1234: I understand ur explaintion IMO this will be source of continous confusion we should just consider them as two separate routers and use same syntax its known remote-as of neghibor has the match router instance

e2etest/pkg/frr/container/container.go
msherif1234: probably because the routes are added to kernel however with view usage we don't install routes in the kernel maybe we can remove that later i will check it out

oribon: The external containers that act as bgp peers are not related to how we configure FRR as a speaker (as a different BGP mode). Using views in the external containers would add up complexity because we'll need to find different ways to ping/wget/... since the routes won't be installed in the container.

e2etest/bgp_test.go
oribon: Could you explain why it's confusing? The containersConf describes which containers we want the tests to setup and with which configuration, for them the NeighborConfig is reflected in the bgp conf file (they need to know who are their neighbors), and the RouterConfig is how they should configure/present themselves (reflected in the daemons conf file).

e2etest/pkg/frr/container/container.go
msherif1234: well I was thinking we can use views in speaker configs and we can have two frr external containers ?

msherif1234:
> > How did u test multiple MyASN w/oi using view feature in #1015 ?
> 
> It runs on the native bgp stack - it shows that it works with multiple MyASNs hence we'd want to have these tests pass when FRR with views land. The external containers here don't need to use views (they're using a single ASN).

correct u don't need to use view in external frr routers

e2etest/bgp_test.go
msherif1234: never mind its clear now Thanks

e2etest/pkg/frr/config/config.go
msherif1234: it seems as of 4.7 the default that network has to exists, I think this need to be debugged further IMHO might be interaction between bgp and rip

e2etest/bgp_test.go
msherif1234: ok that is clear now thanks

e2etest/pkg/frr/config/config.go
msherif1234, line 21: can u add the following tp prepare for Ipv6
```
no bgp default ipv4-unicast
```
I don't know why use next-hop-self it was in the original template and I replaced it with just active

fedepaol: I think the two things are not related.
`no network import-check` is about being able to advertise networks without having them in the rib.

`no bgp ebgp-requires-policy` is about being able to accept routes when no filters are in place (so it's definetely something needed to add only to the external peer)

e2etest/bgp_test.go
```diff
@@ -21,72 +21,107 @@ package e2e
...
-			framework.ExpectNoError(err)
-			err = frr.StartContainer(frrContainer, frrTestConfigDir)
-			framework.ExpectNoError(err)
+		frrContainers = make([]*frrcontainer.Frr, 0)
+		g := new(errgroup.Group)
+		for _, conf := range containersConf {
+			conf := conf
+			g.Go(func() error {
+				c, err := frrcontainer.New(conf.name, conf.nc, conf.rc, containersNetwork, hostContainerIPv4, hostContainerIPv6)
+				if c != nil {
```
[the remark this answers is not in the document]
oribon: Because Start also calls updateVolumePermissions we could have a situation where a container is created (!=nil) but an error is returned. In that case we should add the container to the slice so it would be stopped.

e2etest/bgp_test.go
```diff
@@ -21,72 +21,107 @@ package e2e
...
-		if skipDockerCmd {
-			if len(frrTestConfigDir) == 0 {
-				framework.Fail("Missing FRR config directory.")
-			}
-		} else {
-			frrTestConfigDir, err = ioutil.TempDir("", "frr-conf")
-			framework.ExpectNoError(err)
-			err = frr.StartContainer(frrContainer, frrTestConfigDir)
-			framework.ExpectNoError(err)
+		frrContainers = make([]*frrcontainer.Frr, 0)
```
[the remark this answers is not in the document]
oribon: done

e2etest/pkg/frr/container/container.go
```diff
@@ -19,20 +20,78 @@ const (
 	frrConfigDir = "config/frr"
 	// FRR routing image.
 	frrImage = "frrouting/frr:v7.5.1"
+	// Host network name.
+	hostNetwork = "host"
 	// FRR container mount destination path.
 	frrMountPath = "/etc/frr"
 )
 
+type Frr struct {
```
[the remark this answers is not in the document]
oribon: done

e2etest/bgp_test.go
```diff
@@ -106,21 +150,18 @@ var _ = ginkgo.Describe("BGP", func() {
...
 			framework.ExpectNoError(err)
 
-			pairExternalFRRWithNodes(cs)
+			for _, c := range frrContainers {
+				pairExternalFRRWithNodes(cs, c)
+			}
 
 			allNodes, err = cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
 			framework.ExpectNoError(err)
+
```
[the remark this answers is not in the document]
oribon: done

e2etest/bgp_test.go
```diff
@@ -21,72 +21,107 @@ package e2e
...
 
 	ginkgo.AfterEach(func() {
 		// Clean previous configuration.
 		err := updateConfigMap(cs, configFile{})
 		framework.ExpectNoError(err)
 
-		if !skipDockerCmd {
-			err = frr.StopContainer(frrContainer, frrTestConfigDir)
-			framework.ExpectNoError(err)
+		g := new(errgroup.Group)
```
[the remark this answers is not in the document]
oribon: done

e2etest/pkg/frr/container/container.go
oribon: done

e2etest/pkg/frr/config/config.go
oribon: I think it'd be better to keep this PR focused on the current tests, but thanks for thinking ahead :smiley:

fedepaol:
CI is expected to fail, merging this

PR #1020, opened 2021-11-01
Update code to use k8s.io/api/discovery/v1 instead of v1beta1

### LINKED ISSUE #980: discovery.k8s.io/v1beta1 EndpointSlice is deprecated in v1.21+, unavailable in v1.25+; use discovery.k8s.io/v1 EndpointSlice
Once 1.21 is the oldest version supported we should migrate
```
W1005 08:22:22.301322       1 warnings.go:70] discovery.k8s.io/v1beta1 EndpointSlice is deprecated in v1.21+, unavailable in v1.25+; use discovery.k8s.io/v1 EndpointSlice
```

invidian:
Closes #980

### DIFF
--- a/speaker/layer2_controller.go
+++ b/speaker/layer2_controller.go
@@ -67,10 +67,10 @@ func usableNodes(eps epslices.EpsOrSlices, speakers map[string]bool) []string {
 				if !epslices.IsConditionReady(ep.Conditions) {
 					continue
 				}
-				nodeName := ep.Topology["kubernetes.io/hostname"]
-				if nodeName == "" {
+				if ep.NodeName == nil {
 					continue
 				}
+				nodeName := *ep.NodeName
 				if speakers != nil {
 					if hasSpeaker := speakers[nodeName]; !hasSpeaker {
 						continue
--- a/internal/k8s/controllers/endpoints.go
+++ b/internal/k8s/controllers/endpoints.go
@@ -7,7 +7,7 @@ import (
 
 	"go.universe.tf/metallb/internal/k8s/epslices"
 	v1 "k8s.io/api/core/v1"
-	discovery "k8s.io/api/discovery/v1beta1"
+	discovery "k8s.io/api/discovery/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/client"
--- a/internal/k8s/controllers/service_controller.go
+++ b/internal/k8s/controllers/service_controller.go
@@ -28,7 +28,7 @@ import (
 
 	v1 "k8s.io/api/core/v1"
 
-	discovery "k8s.io/api/discovery/v1beta1"
+	discovery "k8s.io/api/discovery/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
--- a/internal/k8s/epslices/endpoint_slices.go
+++ b/internal/k8s/epslices/endpoint_slices.go
@@ -6,7 +6,7 @@ import (
 	"fmt"
 
 	v1 "k8s.io/api/core/v1"
-	discovery "k8s.io/api/discovery/v1beta1"
+	discovery "k8s.io/api/discovery/v1"
 	"k8s.io/apimachinery/pkg/types"
 )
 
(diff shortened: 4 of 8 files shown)


invidian:
I probably rushed too much with this PR, as discovery/v1 is only available since K8s 1.21 and v1beta1 will be removed only in 1.25 release. And I'm not sure if it's worth it figuring out support for both in parallel. Even currently used version of `kind` in e2e tests deploys K8s 1.20 by default, hence the test failure I think.

BTW, I don't know why linter fails though. Running it locally with the same version does not report any issues to me and CI does not print why it fails :confused: 

See also https://kubernetes.io/docs/reference/using-api/deprecation-guide/#endpointslice-v125

fedepaol:
@invidian , I'd rather wait to have 1.20 out of the support loop, which should happen fairly soon.
At that point we can bump the version of kind and let this in.
There is also another change I think makes sense, which is starting to use the `nodeName` field as opposed to the `opology["kubernetes.io/hostname"]` one

fedepaol:
@invidian I think it's now time to revamp this since 1.20 is out of support lifecycle, are you willing to work on this again?

invidian:
I'll rebase and see how it goes then 👍

fedepaol:
> I'll rebase and see how it goes then +1

Awesome, thanks! I see kind already pins to 1.21, so there is no need to update it.

go.mod
```diff
@@ -18,54 +18,56 @@ require (
...
-	k8s.io/kube-proxy => k8s.io/kube-proxy v0.20.2
-	k8s.io/kube-scheduler => k8s.io/kube-scheduler v0.20.2
-	k8s.io/kubectl => k8s.io/kubectl v0.20.2
-	k8s.io/kubelet => k8s.io/kubelet v0.20.2
-	k8s.io/kubernetes => k8s.io/kubernetes v1.20.2
-	k8s.io/legacy-cloud-providers => k8s.io/legacy-cloud-providers v0.20.2
-	k8s.io/metrics => k8s.io/metrics v0.20.2
-	k8s.io/mount-utils => k8s.io/mount-utils v0.20.2
-	k8s.io/sample-apiserver => k8s.io/sample-apiserver v0.20.2
+	k8s.io/api => k8s.io/api v0.22.3
```
fedepaol, line 47: These should stick to v0.23.0, as much as the apis above

speaker/layer2_controller.go
```diff
@@ -66,10 +66,11 @@ func usableNodes(eps k8s.EpsOrSlices, speakers map[string]bool) []string {
 				if !k8s.IsConditionReady(ep.Conditions) {
 					continue
 				}
-				nodeName := ep.Topology["kubernetes.io/hostname"]
-				if nodeName == "" {
+				nodeNamePtr := ep.NodeName
```
fedepaol, line 69: This can be reduced to:
```
if ep.NodeName == nil {
   continue
}
```

and using `*ep.NodeName` below in place of nodeName

rata:
@fedepaol is 1.20 EOL yet? A new 1.20.x release was just created: https://groups.google.com/a/kubernetes.io/g/dev/c/Y1q7XUp5gjM/m/K_JZcn-5AgAJ

fedepaol:
> @fedepaol is 1.20 EOL yet? A new 1.20.x release was just created: https://groups.google.com/a/kubernetes.io/g/dev/c/Y1q7XUp5gjM/m/K_JZcn-5AgAJ

My understanding was that k8s maintain the n-3 latest release. According to https://endoflife.date/kubernetes , 1.20 active support ended in December, bug fixing will end next month.
Also note that in case ep slice is not available, we back off to endpoints (so it's not that metallb would stop working on older clusters).

fedepaol:
> > @fedepaol is 1.20 EOL yet? A new 1.20.x release was just created: https://groups.google.com/a/kubernetes.io/g/dev/c/Y1q7XUp5gjM/m/K_JZcn-5AgAJ
> 
> My understanding was that k8s maintain the n-3 latest release. According to https://endoflife.date/kubernetes , 1.20 active support ended in December, bug fixing will end next month. Also note that in case ep slice is not available, we back off to endpoints (so it's not that metallb would stop working on older clusters).

Also, https://kubernetes.io/releases/_print/

rata:
Ohh, great then! :)

speaker/layer2_controller.go
invidian: Right, done :+1: I was following old logic.

invidian:
I'm not sure why linter fails on CI (it runs quite old version BTW). This PR does not touch mentioned files...

And building was failing due to dependencies requiring Go 1.17 now, which wasn't quite explicitly said, so I had to downgrade the K8s dependencies to v0.22.x until Go is updated to 1.17.

fedepaol:
Apologies for following up so late, let's wait for CI and merge this. And thanks.

fedepaol:
np, I felt guilty enough not to merge it before :P

PR #1021, opened 2021-11-01
release-process: add notes on helm and slack

russellb:
Note that we must manually publish an update to the helm chart. Also
note that the version number should be updated in #metallb on slack.

### DIFF
--- a/website/content/community/release-process.md
+++ b/website/content/community/release-process.md
@@ -96,6 +96,13 @@ tags that don't exist until CircleCI makes them exist.
 Check on Quay for a `vX.Y.Z` tag on each image, or check on
 CircleCI that the deploy has completed.
 
+### Publish a helm chart update
+
+This is a manual step until some future time when we build an appropriate CI
+job that can do it for us. See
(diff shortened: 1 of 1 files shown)


cgoncalves:
Even though the release process is only actionable by the project maintainers, keeping it published on the website brings no harm but rather provides a rendered page and fosters greater community openness. On the flip side, referring to documentation in a docs/ directory is more intuitive. I wonder if the design/ directory should be moved into this new docs/ directory (and/or published on the website).

russellb:
> Even though the release process is only actionable by the project maintainers, keeping it published on the website brings no harm but rather provides a rendered page and fosters greater community openness. On the flip side, referring to documentation in a docs/ directory is more intuitive. I wonder if the design/ directory should be moved into this new docs/ directory (and/or published on the website).

Thanks. I like the idea of moving `design/` under `docs/`.  I should probably just split the move out since what I *really* wanted to get in was the content updates.

cgoncalves:
Sorry for being a PITA :sweat_smile:
LGTM.

fedepaol:
LGTM

PR #1024, opened 2021-11-03
Fix FRR manifest

fedepaol:
While fixing comments in https://github.com/metallb/metallb/pull/1014 , I cherry-picked a wrong version of the frr manifest, with the result that CI is running legacy mode even on the CI lanes (this is what we call volkswagen test https://github.com/auchenberg/volkswagen)

Here, we fix the manifest and we add a test (running only on the frr lanes) that validates the FRR container in the pods is actually working.

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -149,6 +149,11 @@ var _ = ginkgo.Describe("BGP", func() {
 
 			validateService(cs, svc, epNodes)
 		})
+
+		ginkgo.It("would check that FRR is running in the speaker", func() {
+			paired := frrIsPairedOnPods(cs, frrContainerIPv4)
+			framework.ExpectEqual(paired, true)
+		})
 	})
 
 	ginkgo.Context("metrics", func() {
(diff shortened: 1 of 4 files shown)


manifests/metallb-frr.yaml
```diff
@@ -538,6 +526,8 @@ spec:
           volumeMounts:
             - name: reloader
               mountPath: /etc/frr_reloader
+            - name: frr-conf
+              mountPath: /etc/frr
```
oribon, line 530: why is this needed?

fedepaol: probably a leftover of me hammering through making it work.
Will remove it.

fedepaol: Removed

russellb:
seems the "make sure frr is running" test failed?

> [Fail] BGP type=Loadbalancer [BeforeEach] would check that FRR is running in the speaker /home/circleci/project/e2etest/bgp_test.go:376

fedepaol:
> seems the "make sure frr is running" test failed?
> 
> > [Fail] BGP type=Loadbalancer [BeforeEach] would check that FRR is running in the speaker /home/circleci/project/e2etest/bgp_test.go:376

what's weird is, it failed on the part common to the loadbalancer tests https://github.com/metallb/metallb/blob/bbd0cf14a2c4eb7aaf96d4e1cacda5169fc2ce98/e2etest/bgp_test.go#L120

e2etest/bgp_test.go
```diff
@@ -149,6 +149,11 @@ var _ = ginkgo.Describe("BGP", func() {
 
 			validateService(cs, svc, epNodes)
 		})
+
+		ginkgo.It("would check that FRR is running in the speaker", func() {
+			paired := frrIsPairedOnPods(cs, frrContainerIPv4)
+			framework.ExpectEqual(paired, true)
```
russellb, line 155: could add the opposite check for non-frr test jobs (make sure the frr container isn't running when it's not supposed to be)

russellb (commented):
lgtm, one idea to enhance the check, but don't need to block on it

PR #1027, opened 2021-11-03
FRR: Don't listen bgp port

oribon:
The native bgp implementation does not listen on a bgp port,
meaning it is always the side initiating the BGP connection.
We should mimic this behavior in FRR mode as it preserves some
of the peer's configuration by using:
https://docs.frrouting.org/en/latest/bgp.html#cmdoption-bgpd-p.
For example peer-port: without this change there is no guarantee
that it is actually used for the connection - the other side might
be initiating it hence the peer's port will be random from the speaker's side (TCP).

### DIFF
--- a/manifests/metallb-frr.yaml
+++ b/manifests/metallb-frr.yaml
@@ -361,7 +361,7 @@ data:
     #
     vtysh_enable=yes
     zebra_options="  -A 127.0.0.1 -s 90000000"
-    bgpd_options="   -A 127.0.0.1"
+    bgpd_options="   -A 127.0.0.1 -p 0"
     ospfd_options="  -A 127.0.0.1"
     ospf6d_options=" -A ::1"
     ripd_options="   -A 127.0.0.1"


russellb:
> Overriding the default connect timer for neighbors (120s) with 10s should be fair enough.

This seems unrelated to the main point of the PR. Can you either split this out, or add more explanation for this change?

oribon:
> > Overriding the default connect timer for neighbors (120s) with 10s should be fair enough.
> 
> This seems unrelated to the main point of the PR. Can you either split this out, or add more explanation for this change?

Thanks updated the commit message, does it make sense?

fedepaol:
LGTM

msherif1234:
10 second looks very aggressive
TCP takes 30 seconds to realize tcp open failed and declares “open failed”
BGP will wait ConnectRetry timer (120 seconds) again to do another TCP OPEN ACTIVE
This situation will last till the TCP OPEN ACTIVE is successful
if we have  to change the default at least wait for TCP to declare it failed before try again

cgoncalves:
I agree with Mohamed. [RFC 4271](https://datatracker.ietf.org/doc/html/rfc4271#page-90) suggests 120 seconds for ConnectRetryTime and that is also the default value in FRR. It would be best not to override it but instead provide an option for users to override if needed. It can either be part of this PR or added to the backlog for later implementation.

oribon:
> 10 second looks very aggressive TCP takes 30 seconds to realize tcp open failed and declares “open failed” BGP will wait ConnectRetry timer (120 seconds) again to do another TCP OPEN ACTIVE This situation will last till the TCP OPEN ACTIVE is successful if we have to change the default at least wait for TCP to declare it failed before try again



> I agree with Mohamed. [RFC 4271](https://datatracker.ietf.org/doc/html/rfc4271#page-90) suggests 120 seconds for ConnectRetryTime and that is also the default value in FRR. It would be best not to override it but instead provide an option for users to override if needed. It can either be part of this PR or added to the backlog for later implementation.

Reverted that change, thanks for the info!

russellb:
thanks!  and nice job catching this

Issue #660, opened 2020-07-07
MetalLB reject new config when IP are already attributed


champtar:
A wrong ip was configured, then attributed by MetalLB

After fixing the configuration:
```
{"caller":"main.go:117","configmap":"metallb-system/config","error":"new config not compatible with assigned IPs: service \"ingress-nginx/ingress-nginx\" cannot own \"10.254.145.63\" under new config","msg":"applying new configuration failed","op":"setConfig","ts":"2020-07-07T11:26:49.584367712Z"}
```

Fixing the configuration requires restart of the controller.
This behavior is safe but really confusing, at least we need to be explicit in the log message.

MetalLB 0.9.3

johananl:
This is described in https://github.com/metallb/metallb/issues/462 AFAICT.

Issue #462, opened 2019-07-31
Remove the config reloading fail-static behavior


danderson:
Right now, MetalLB tries hard to fail static if you delete its configuration, or provide an invalid/unsafe configuration. It does so by ignoring the new configuration (or the deletion of the configuration), and continues serving based on the last known config.

This is theoretically good, because it's a good guard against user error. The problem is k8s gives us no way of VERY LOUDLY telling the cluster operator that something is wrong. The best we can do is flip a Prometheus metric that says "hey we're running on a stale config." In practice, ~nobody does the Prometheus setup to detect this, so it just leads to mass confusion when MetalLB seems to not be working correctly (not allocating IPs to new services, announcing some services but not others...)

I originally thought that using a k8s admission controller would let us eagerly reject invalid configs. Unfortunately that's subject to a race condition where we might accept a config, but then a service changes state in a way that makes the config unsafe before we process the actual reloading. So, we can't make it behave correctly in all cases.

Given that k8s doesn't give us the lever we require, the only sensible thing to do is to remove the fail-static behavior, and always do exactly what the user tells us: if you push a config that makes some service IPs no longer valid, we'll change the service IPs. If you delete the configmap, MetalLB will un-announce all services and just stop doing anything.

It sucks because it means things are a bit less safe, but otoh it'll make MetalLB behave exactly like all other k8s services, in that what you told MetalLB to do is exactly what you'll get.

CamelCaseNotation:
Is there a path forward for getting this done?

This current behavior causes problems when changing the addresses under an address pool in the configmap. Are you proposing the new behavior will cause any services which got an IP from the old pool to discard its current IP and get a new one?

danderson:
In general, MetalLB already checks existing services with an assigned IP, and verifies that the IP makes sense given the configuration (i.e. that this service is "allowed" to have this IP). If the check fails, it deletes the assigned IP and goes through the assignment algorithm as if the service was brand new.

so yes, the new behavior would be that if you delete a pool from the config, all services with IPs from that pool will immediately lose their IP. If another IP can be allocated, that will happen.

CamelCaseNotation:
> In general, MetalLB already checks existing services with an assigned IP, and verifies that the IP makes sense given the configuration (i.e. that this service is "allowed" to have this IP). If the check fails, it deletes the assigned IP and goes through the assignment algorithm as if the service was brand new.

This was not the behavior observed with the version of MetalLB we are running (v0.7.3).

[...]

Issue #308, opened 2018-09-06
Documenting the IP range change workflow


weisjohn:
**Is this a bug report or a feature request?**:

feature request

**What happened**:

@danderson very quickly and awesomely helped me understand how to migrate my metallb-managed services from one block of IPs to another.

**What you expected to happen**:

To spend hours rebuilding my cluster by hand after re-`kubeadm init`-ing and installing metallb.

**How to reproduce it (as minimally and precisely as possible)**:

Here's my first stab at documenting:

<hr />

## IP Address Migration

To migrate an IP address range for a set of services, you must remove the old ConfigMap, and replace it with a new one, then delete the Pods within the `metallb-system` namespace.

For example, using a Layer2 IP address range, beginning with the following ConfigMap:

```yml
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: metallb-system
  name: config
data:
  config: |
    address-pools:
    - name: default
      protocol: layer2
      addresses:
      - 10.0.0.10-10.0.0.19
```

And assuming a `new.yml` config file, with a new IP address range of `10.0.0.20-10.0.0.29`, to migrate your services to a new IP range, you should:

```bash
# note the old IPs allocated to the services
kubectl get svc

# delete the old configmap
kubectl -n metallb-system delete cm config

# apply the new configmap
kubectl apply -f new.yml

# delete the metallb pods
kubectl -n metallb-system delete pod --all

# watch the pods come back up
kubectl -n metallb-system get pods -w

# inspect new IPs of services
kubectl get svc
```

<hr />

**Anything else we need to know?**:

I need to know where to put the documentation.  Should it be in a page under usage https://metallb.universe.tf/usage/ ?  Should it be under configuration https://metallb.universe.tf/configuration/ ?

**Environment**:
- MetalLB version: v0.7.3
- Kubernetes version: v1.11.2
- BGP router type/version: Layer2
- OS (e.g. from /etc/os-release): n/a
- Kernel (e.g. `uname -a`): n/a

danderson:
This probably belongs in an "advanced" sub-page under "Usage". I'll get to that at some point, but if you want to send a PR, that's also fine by me.

sysnet4admin:
It could work as well like below : (FYI oooonly) 

**Change IP address**
```
[root@m-k8s ~]# cat metallb-l2.yaml | grep 192.168
      - 192.168.1.11-192.168.1.13
[...]

Issue #201, opened 2018-03-17
Better error messages when things go wrong with IP ranges in the config


danderson:
With #188, you can use explicit ranges of IPs instead of CIDR prefixes, but internally the config parser turns those ranges into a set of CIDR prefixes. This means that all subsequent validation of the config will spew errors about CIDR prefixes that don't exist verbatim in the configuration. This is obviously quite terrible.

ReticentFacade:
@fedepaol, can I work on this?

fedepaol:
@ReticentFacade sure thing!

fedepaol:
and sorry for the delay

ReticentFacade:
No problem! I just read the reply myself, I'll start working on it today.

ReticentFacade:
@fedepaol I'm a beginner in open source so if there's anything that you could recommend I read before working, that'd be helpful. I've read the contribution guidelines and started with the documentation as well.

fedepaol:
Closing this, we have a fail fast behavior which is pool specific in both the webhook and the config parsing.

The error is like `invalid CIDR %q in pool %q: %s", cidr, p.Name, err)` and then bubbled up here https://github.com/metallb/metallb/blob/7f87bf061165e50bd7b07d9018c0e8a2501f5058/internal/config/config.go#L312

PR #1028, opened 2021-11-03
Speaker: reprocess the configuration when the ip assigned to services

fedepaol:
When applying changes in bulk, a change in a service may be received / processed asynchronously from the configuration. We may want to be able to reprocess the configmap later, when the service is deleted.

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -621,12 +621,12 @@ func TestControllerConfig(t *testing.T) {
 	}
 
 	// Now that an IP is allocated, removing the IP pool is not allowed.
-	if c.SetConfig(l, &config.Config{}) != k8s.SyncStateError {
+	if c.SetConfig(l, &config.Config{}) != k8s.SyncStateErrorNoRetry {
 		t.Fatalf("SetConfig that deletes allocated IPs was accepted")
 	}
 
 	// Deleting the config also makes MetalLB sad.
-	if c.SetConfig(l, nil) != k8s.SyncStateError {
+	if c.SetConfig(l, nil) != k8s.SyncStateErrorNoRetry {
 		t.Fatalf("SetConfig that deletes the config was accepted")
 	}
 }
--- a/controller/main.go
+++ b/controller/main.go
@@ -104,12 +104,12 @@ func (c *controller) SetConfig(l log.Logger, cfg *config.Config) k8s.SyncState {
 
 	if cfg == nil {
 		level.Error(l).Log("op", "setConfig", "error", "no MetalLB configuration in cluster", "msg", "configuration is missing, MetalLB will not function")
-		return k8s.SyncStateError
+		return k8s.SyncStateErrorNoRetry
 	}
 
 	if err := c.ips.SetPools(cfg.Pools); err != nil {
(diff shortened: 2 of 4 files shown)


fedepaol:
Testing this PR on top of https://github.com/metallb/metallb/pull/1012

russellb:
Can you add a longer explanation of the scenario you're fixing? I don't really understand it.  I see that you've changed a couple of cases to not retry ... but what exactly was the test scenario that triggered a problem?  What was the problem?  Why does this fix it?

fedepaol:
I reworded the commit message, plus putting the references to the current code here

In a scenario where we erase the config and we remove the service at the same time, what could happen is:

- the config is handled first, here https://github.com/metallb/metallb/blob/9cb1653a87673ec91971a60c873204e0752ff97b/speaker/main.go#L358 , returning `k8s.SyncStateError` which is intended to be "this error is transitient" but in the specific case of configmap is intended as "this will never work", so it's transformed here https://github.com/metallb/metallb/blob/7fb3231966a0c2a6741122f4b81020abd3380011/internal/k8s/k8s.go#L537 in success so we'll never retry.
- Then we receive the service, we delete it but that does not trigger a reload of the configmap, because the only thing we reload is services as consequence of SyncStateReprocessAll: https://github.com/metallb/metallb/blob/7fb3231966a0c2a6741122f4b81020abd3380011/internal/k8s/k8s.go#L434 (forceSync only force services)

So the valid configuration change is lost forever. What we are doing here, is to add another SyncState result type that tells "never retry" that is returned for those scenarios where we thing we should really never retry, and we trust the result returned by the `configChanged` callback.

russellb:
Thank you.  I understand the changes now.

There are several open issues related to this, most notably #462, but also #660, #308, and #201.

I would say the behavior you've proposed here is a net improvement, in that a potentially valid configuration will be applied once IPs from the old config are no longer in use. That's different than what's proposed in #462, though maybe this is safest. This change would mean that we require users to explicitly remove all usage of IPs from the old configuration, while what was discussed in #462 was just force changing the IP allocations of all existing LoadBalancers to comply with the new configuration.  I'm not really sure what people would find least surprising.  This method seems a little "safer", but makes the process of changing IP ranges a lot more tedious.  Either way, we should document the behavior REALLY well on the web site.

I think this is something we really need to discuss with the whole maintainers group to make sure we're all really clear about the behavior we want. CC @rata @johananl

fedepaol:
#462 seems a bit more drastic, here we preserve the current behavior (you can't cancel a configmap if an address is in use) but tries to see if the error condition was temporary (i.e. because the configmap and the service are being changed at the same time). 

In any case the same race condition both this and #462 talk about is a bug. There is a sequence of service + configmap that is legal but makes the configmap invalid.
I hit this when trying to do something legit (actually, in one of the tests), which was deleting all the services and cleaning the configmap at the same time. 

An even better solution would be to reprocess the configmap whenever a service status change, but I am not sure it's worth the additional complexity. Here the "bad" configmap is being retried with exponential backoff.

rata (approved):
LGTM, thanks!

PR #1032, opened 2021-11-09
Add E2E configuration changes tests

sabinaaledort:
Create two address pools and request a specific address pool for each Loadbalancer service. Validate functionality.

Create two peers and validate each FRR container is paired with the cluster nodes.

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -171,7 +171,7 @@ var _ = ginkgo.Describe("BGP", func() {
 		framework.ExpectNoError(err)
 
 		if setProtocoltest == "ExternalTrafficPolicyCluster" {
-			svc, _ := createServiceWithBackend(cs, f.Namespace.Name, tweak)
+			svc, _ := createServiceWithBackend(cs, f.Namespace.Name, "external-local-lb", tweak)
 
 			defer func() {
 				err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
@@ -184,7 +184,7 @@ var _ = ginkgo.Describe("BGP", func() {
 		}
 
 		if setProtocoltest == "ExternalTrafficPolicyLocal" {
-			svc, jig := createServiceWithBackend(cs, f.Namespace.Name, tweak)
+			svc, jig := createServiceWithBackend(cs, f.Namespace.Name, "external-local-lb", tweak)
 			err = jig.Scale(2)
 			framework.ExpectNoError(err)
 
@@ -324,7 +324,7 @@ var _ = ginkgo.Describe("BGP", func() {
 			}
 
 			ginkgo.By("creating a service")
-			svc, _ := createServiceWithBackend(cs, f.Namespace.Name, testservice.TrafficPolicyCluster) // Is a sleep required here?
+			svc, _ := createServiceWithBackend(cs, f.Namespace.Name, "external-local-lb", testservice.TrafficPolicyCluster) // Is a sleep required here?
 			defer func() {
 				err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
 				framework.ExpectNoError(err)
@@ -407,7 +407,7 @@ var _ = ginkgo.Describe("BGP", func() {
 				validateFRRPeeredWithNodes(cs, c)
 			}
 
-			svc, _ := createServiceWithBackend(cs, f.Namespace.Name, tweak)
+			svc, _ := createServiceWithBackend(cs, f.Namespace.Name, "external-local-lb", tweak)
 
 			defer func() {
 				err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
@@ -511,7 +511,7 @@ var _ = ginkgo.Describe("BGP", func() {
 				})
 			}
 
-			svc, _ := createServiceWithBackend(cs, f.Namespace.Name, tweak)
+			svc, _ := createServiceWithBackend(cs, f.Namespace.Name, "external-local-lb", tweak)
 			defer func() {
 				err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
 				framework.ExpectNoError(err)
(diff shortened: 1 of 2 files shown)


e2etest/bgp_test.go
```diff
@@ -390,6 +390,167 @@ var _ = ginkgo.Describe("BGP", func() {
...
+
+	ginkgo.Context("validate configuration changes", func() {
+		ginkgo.AfterEach(func() {
+			// Clean previous configuration.
+			err := updateConfigMap(cs, configFile{})
+			framework.ExpectNoError(err)
+		})
+
+		ginkgo.It("request a specific address pool for each Loadbalancer service", func() {
+			ginkgo.By("configure first addresspool")
```
fedepaol, line 402: Can we do the two iterations inside a loop?
That'd remove some code duplication

e2etest/bgp_test.go
```diff
@@ -390,6 +390,167 @@ var _ = ginkgo.Describe("BGP", func() {
...
+
+			ginkgo.By("configure first service")
+			jig := e2eservice.NewTestJig(cs, namspace, "test-service1")
+			timeout := e2eservice.GetServiceLoadBalancerCreationTimeout(cs)
+			svc1, err := jig.CreateLoadBalancerService(timeout, func(svc *corev1.Service) {
+				tweakServicePort(svc)
+				svc.Spec.ExternalTrafficPolicy = corev1.ServiceExternalTrafficPolicyTypeCluster
+			})
+
+			framework.ExpectNoError(err)
```
fedepaol, line 448: Nit: move this the line above

e2etest/layer2_test.go
```diff
@@ -275,13 +275,154 @@ var _ = ginkgo.Describe("L2", func() {
...
+				firstAddressPool,
+			},
+		}
+
+		err := updateConfigMap(cs, configData)
+		framework.ExpectNoError(err)
+
+		namspace := f.Namespace.Name
+
+		ginkgo.By("configure first service")
```
fedepaol, line 303: same as above. Can we have both inside a loop?

e2etest/bgp_test.go
```diff
@@ -390,6 +390,130 @@ var _ = ginkgo.Describe("BGP", func() {
...
+					framework.ExpectEqual(ip, servicesIngressIP[j])
+
+					ginkgo.By(fmt.Sprintf("checking connectivity of service %d to its external VIP", j+1))
+					for _, c := range frrContainers {
+						validateService(cs, services[j], allNodes.Items, c)
+					}
+				}
+			}
+		})
+		ginkgo.It("configure peers one by one and validate FRR paired with nodes", func() {
```
fedepaol, line 490: nit: can you add a space?

e2etest/layer2_test.go
```diff
@@ -275,13 +275,102 @@ var _ = ginkgo.Describe("L2", func() {
...
+		}
+	},
+		table.Entry("IPV4", &ipv4ServiceRange),
+		table.Entry("IPV6", &ipv6ServiceRange))
 })
 
-func firstIPFromRange(ipRange string) string {
-	cidr, err := internalconfig.ParseCIDR(ipRange)
+func getIPFromRangeByIndex(ipRange string, index int) string {
+	cidrs, err := internalconfig.ParseCIDR(ipRange)
```
fedepaol, line 352: why is not enough to pick the first cidr anymore?
IIUC. by doing this we go over all the possible cidrs related to the range, which is different from picking the first and the nth item of the same cidr.

e2etest/bgp_test.go
```diff
@@ -390,6 +390,130 @@ var _ = ginkgo.Describe("BGP", func() {
...
 	})
+
+	ginkgo.Context("validate configuration changes", func() {
+		ginkgo.AfterEach(func() {
+			// Clean previous configuration.
+			err := updateConfigMap(cs, configFile{})
+			framework.ExpectNoError(err)
+		})
+
+		ginkgo.It("request a specific address pool for each Loadbalancer service", func() {
```
fedepaol, line 401: Can you change the description to something like: it should work after subsequent configuration updates (or something alogn that line)?

e2etest/bgp_test.go
```diff
@@ -390,6 +390,130 @@ var _ = ginkgo.Describe("BGP", func() {
...
+				}
+				err := updateConfigMap(cs, configData)
+				framework.ExpectNoError(err)
+
+				for _, c := range frrContainers {
+					pairExternalFRRWithNodes(cs, c)
+				}
+
+				ginkgo.By(fmt.Sprintf("configure service number %d", i+1))
+				jig := e2eservice.NewTestJig(cs, namspace, fmt.Sprintf("test-service%d", i+1))
```
fedepaol, line 450: Is there a reason why you can't reuse `createServiceWithBackend` ?

e2etest/layer2_test.go
sabinaaledort: for example, when passing the following range `172.18.0.6-172.18.0.16` to the ParseCIDR function we get the following cidrs:`[172.18.0.6/31 172.18.0.8/29 172.18.0.16/32]`
if we want to get the 5th ip in the range we can't use the first cidr as it has only two ips.

e2etest/bgp_test.go
sabinaaledort: i didn't want to add a name parameter to this function. do you think we should?

e2etest/layer2_test.go
fedepaol: so I think a better option is to pick the largest CIDR and use that as a range reference. Otherwise we (at extreme) risk picking the same ip because you jump from one CIDR to another

sabinaaledort: but the cidrs are not overlapping so i can't get the same ip, or they might me overlapping?

e2etest/bgp_test.go
sabinaaledort: it sounds more like the context description, i wanted to be more specific here describing what exactly the test does.

e2etest/bgp_test.go
fedepaol: That doesn't sound too bad. Another option is to add a slice of optional modifier functions so all the other calling sites won't be affected. Something like 
`createServiceWithBackend(cs clientset.Interface, namespace string, policy corev1.ServiceExternalTrafficPolicyType, modifiers ...func(*corev1.Service))

We can even refactor the traffic policy parameter to use that.
See https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis for more details.

Either ways are fine, right now it's a bit hard to follow.

e2etest/bgp_test.go
fedepaol: Yes but the point of this test is to try to change two different configurations I think. Otherwise, I think it's not totally clear.

e2etest/layer2_test.go
fedepaol: ugh, I completely mis-understood what ParseCIDR does.Ignore what I just said, thanks for explaining.

e2etest/layer2_test.go
```diff
@@ -275,13 +275,102 @@ var _ = ginkgo.Describe("L2", func() {
...
+		c = ipaddr.NewCursor([]ipaddr.Prefix{*ipaddr.NewPrefix(cidr)})
+		for j := i; j < index; j++ {
+			if c.Next() != nil {
+				err := c.Set(c.Next())
+				framework.ExpectNoError(err)
+				i++
+			}
+		}
+		if i == index {
+			break
```
fedepaol, line 366: We can just return here and return an error out of the loop

e2etest/bgp_test.go
```diff
@@ -464,18 +473,156 @@ var _ = ginkgo.Describe("BGP", func() {
 			),
 		)
 	})
+
+	ginkgo.Context("validate configuration changes", func() {
+		ginkgo.AfterEach(func() {
+			// Clean previous configuration.
+			err := updateConfigMap(cs, configFile{})
+			framework.ExpectNoError(err)
+		})
```
oribon, line 482: Why is this necessary? The one defined in the beginning of the Describe isn't enough?

sabinaaledort: That's an after each to we do a cleanup before the next test.

oribon: but doesn't the first aftereach already takes care of that? https://github.com/metallb/metallb/blob/9a08426c6d753d0e9323dac60152ba8f123cb507/e2etest/bgp_test.go#L113

fedepaol: It does

sabinaaledort: right, i will remove it

e2etest/bgp_test.go
```diff
@@ -465,18 +474,150 @@ var _ = ginkgo.Describe("BGP", func() {
...
+				}
+			}
+		},
+			table.Entry("IPV4", "ipv4"),
+			table.Entry("IPV6", "ipv6"))
+	})
 })
 
-func createServiceWithBackend(cs clientset.Interface, namespace string, policy corev1.ServiceExternalTrafficPolicyType) (*corev1.Service, *e2eservice.TestJig) {
+func createServiceWithBackend(cs clientset.Interface, namespace string, jigName string, config func(*corev1.Service)) (*corev1.Service, *e2eservice.TestJig) {
```
fedepaol, line 612: Sorry I wasn't clear, my suggestion about the callback(s) (as a variadic args) was to be able to override the service without adding an extra parameter.
I see here that for the name we need an extra param anyway, so there's no point in moving to the function.

sabinaaledort: i think it's more clean to have the function this way so we can set the service parameters. what do you think?

fedepaol: I think it will be clear if we'll need to add more parameters (which might be the case when https://github.com/metallb/metallb/pull/1065 lands. Adding this to change always the single parameter is a bit of an overcomplication imo.

In anycase, since we'll need to do that, I think the calling site would benefit from two global `config func(*corev1.Service)`, one as SetClusterTrafficPolicy and the other as SetLocalTrafficPolicy so you can use those as parameters for this function.

fedepaol:
@sabinaaledort looks good but needs rebase.
The other tests support dual stack too now, but I think it's not really necessary here since here we are testing how different config changes are handled.

fedepaol:
LGTM, thanks!

fedepaol:
Ugh, CI was green but this broke badly. We must force rebasing against main.

PR #1035, opened 2021-11-10
E2E test: make the check for sharing ip l2 test in eventually

fedepaol:
It may take some time to have the event we check available, here instead of checking once, we check inside an eventually loop to avoid flakes.

### DIFF
--- a/e2etest/layer2_test.go
+++ b/e2etest/layer2_test.go
@@ -20,6 +20,7 @@ package e2e
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"net"
 	"strconv"
@@ -28,6 +29,7 @@ import (
 	"github.com/mikioh/ipaddr"
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/extensions/table"
+	"github.com/onsi/gomega"
 	"go.universe.tf/metallb/e2etest/pkg/executor"
 	"go.universe.tf/metallb/e2etest/pkg/mac"
 	internalconfig "go.universe.tf/metallb/internal/config"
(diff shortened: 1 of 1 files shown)


e2etest/layer2_test.go
```diff
@@ -257,21 +259,31 @@ var _ = ginkgo.Describe("L2", func() {
 			})
 		framework.ExpectNoError(err)
 
-		events, err := cs.CoreV1().Events(namespace).List(context.Background(), metav1.ListOptions{FieldSelector: "reason=nodeAssigned"})
-		framework.ExpectNoError(err)
+		Eventually(func() error {
+			events, err := cs.CoreV1().Events(namespace).List(context.Background(), metav1.ListOptions{FieldSelector: "reason=nodeAssigned"})
+			framework.ExpectNoError(err)
```
oribon, line 264: I think you should return the err instead

e2etest/layer2_test.go
```diff
@@ -28,6 +29,7 @@ import (
 	"github.com/mikioh/ipaddr"
 	"github.com/onsi/ginkgo"
 	"github.com/onsi/ginkgo/extensions/table"
+	. "github.com/onsi/gomega"
```
oribon, line 32: I think you should not use dot here like the other file: https://github.com/metallb/metallb/blob/eb3c241b571f6838843002bedada5aee28ce14b5/e2etest/e2etest_suite_test.go#L32

oribon (commented):
other than that looks good

e2etest/layer2_test.go
```diff
@@ -257,21 +259,31 @@ var _ = ginkgo.Describe("L2", func() {
...
-				service2Announce = e.Message
+			if service1Announce == "" {
+				return errors.New("service1 not announced")
 			}
-		}
-		framework.ExpectNotEqual(service1Announce, "")
-		framework.ExpectNotEqual(service2Announce, "")
-		framework.ExpectEqual(service1Announce, service2Announce, "Services with the same IP announced from different nodes")
+			if service2Announce == "" {
+				return errors.New("service2 not announced")
```
oribon, line 279: why not fmt.Errorf?

oribon, line 279: why not fmt.Errorf?

e2etest/layer2_test.go
fedepaol: Being able to list nodes should not depend on the timing, that's why I left it that way.

e2etest/layer2_test.go
fedepaol: Will change

e2etest/layer2_test.go
oribon: IIUC there could be a scenario where listing the events fails due to some flake/connection error but in the next iteration of the Eventually it would have worked - if so do we want to fail the test because of that?

fedepaol: My idea was that it was less likely to happen (compared to the real reason why we have flakes), but changing it won't hurt.

fedepaol: Done

e2etest/layer2_test.go
fedepaol: Done

PR #1036, opened 2021-11-10
Set FRR version consistently across the manifests file

oribon:
We should use the same FRR version for all containers that
consume FRR as it is more effective (no need to pull multiple images/...).


oribon:
will be fixed by https://github.com/metallb/metallb/pull/997

PR #1037, opened 2021-11-10
L2 E2E: Set service port based on service-pod-port flag

sabinaaledort:
Currently the ports are hard coded in the test and therefore it is
failing in clusters where this port is not allowed.

Doing this will allow running this test based on the passed base port
by the user, which should be open.

### DIFF
--- a/e2etest/layer2_test.go
+++ b/e2etest/layer2_test.go
@@ -219,8 +219,8 @@ var _ = ginkgo.Describe("L2", func() {
 
 		ip := firstIPFromRange(*ipRange)
 		svc1, err := jig1.CreateLoadBalancerService(loadBalancerCreateTimeout, func(svc *corev1.Service) {
-			svc.Spec.Ports[0].TargetPort = intstr.FromInt(82)
-			svc.Spec.Ports[0].Port = 82
+			svc.Spec.Ports[0].TargetPort = intstr.FromInt(servicePodPort)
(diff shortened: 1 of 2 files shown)


e2etest/layer2_test.go
```diff
@@ -217,8 +217,8 @@ var _ = ginkgo.Describe("L2", func() {
 
 		ip := firstIPFromRange(*ipRange)
 		svc1, err := jig1.CreateLoadBalancerService(loadBalancerCreateTimeout, func(svc *corev1.Service) {
-			svc.Spec.Ports[0].TargetPort = intstr.FromInt(82)
-			svc.Spec.Ports[0].Port = 82
+			svc.Spec.Ports[0].TargetPort = intstr.FromInt(int(servicePodPort))
```
fedepaol, line 220: can we change the type of servicePodPort to int to avoid these extra casts? Would it introduce other extra casts?

sabinaaledort: we will still have to cast the Port which is an int32. and also i wanted to keep it as uint to avoid negative port.

fedepaol: Yes but that's a parameter the user is setting. The double cast is just anti-esthetic and I don't think there's much value in having that enforced.

sabinaaledort: ok so we can cast just the port

fedepaol:
LGTM

PR #1039, opened 2021-11-11
Buffer FRR reloading commands

fedepaol:
A change in config is reflected in several changes (and reloads) of the config map. Here, decouple the request for a reload to the reloading.
If a reload request comes in, we squash all the next reload requests and apply the final configuration only after 500 milliseconds.
If a new request comes after a reload, we start a new waiting period.

500 msec is short enough not to affect the user experience, and at the same time is long enough to collect all the variations related to a single config change.


internal/bgp/frr/config.go
```diff
@@ -158,26 +162,55 @@ var reloadConfig = func() error {
...
-		return err
+		level.Error(l).Log("op", "reload", "error", err, "cause", "writeConfig", "msg", "failed to reload config", "config", config)
+		return
 	}
 
 	err = reloadConfig()
 	if err != nil {
-		return err
+		level.Error(l).Log("op", "reload", "error", err, "cause", "reload", "msg", "failed to reload config", "config", config)
+		return
```
msherif1234, line 186: all the above has "failed to reload config"  probably better off with different error message to differentiate the different reason

internal/bgp/frr/frr.go
```diff
@@ -239,8 +237,17 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
 	return config, nil
 }
 
-func NewSessionManager() *sessionManager {
-	return &sessionManager{
-		sessions: map[string]*session{},
+var debounceTimeout = 500 * time.Millisecond
```
msherif1234, line 240: const ?

internal/bgp/frr/config.go
```diff
@@ -158,26 +162,55 @@ var reloadConfig = func() error {
...
+					return
+				}
+				config = newCfg
+				if !timerSet {
+					timeOut = time.After(reloadInterval)
+					timerSet = true
+				}
+			case <-timeOut:
+				body(config)
+				timerSet = false
```
msherif1234, line 212: do u need to check config is failed any chance ever to get time out and somehow we dropped reload signal ?

msherif1234:
just curious why did avoid using `singleflight` u could DoChan which I assume will achieve what debouncer do

internal/bgp/frr/frr.go
fedepaol: I left it as a var so we can put a lower variable in unit tests.
Otherwise unit tests execution would have take longer which is not desireable.

internal/bgp/frr/config.go
fedepaol: Not sure I undestand the comment.
We'd have the err field stating what's failing.

internal/bgp/frr/config.go
fedepaol: Since we decouple the calling site from the config execution, we can't really return any error.
This is not making things worse, because we already have the real reload decoupled by the go code (what we do is just call SIGUP). 
The only feedback we get is about what happens to the signal (which is unlikely to go wrong). Given the asyncrhonousness of the call, I preferred embedding the logger in the function and logging from there. That's why `body` does not return any error.

internal/bgp/frr/config.go
msherif1234: instead of fishing for single word difference between the different errors why not use 
```
for the 1st : "failed to setting the configuration template %s", config
for the 2nd: "failed to write config file %s",configFileName
for the 3d: "failed to reload FRR config err %s", err
```

fedepaol:
> just curious why did avoid using `singleflight` u could DoChan which I assume will achieve what debouncer do

Just realized I forgot to reply here. I totally mis-used single flight the last time.
The package is useful if you have the same requests coming from different goroutines, which is really different from what we want to do here. Here we have one single goroutine (our controller callback) that applies a set of subsequent changes that we want to squash. That's why we are decoupling putting a goroutine in the middle.

internal/bgp/frr/config.go
fedepaol: The log is structured, we have a cause and an err field to do that. I think we can even remove the msg because it won't add any new value.

internal/bgp/frr/config.go
oribon: What should the user do in the(unlikely) event where something did go wrong along the way?

fedepaol: I guess it depends on the nature of the error, and not dependant on this PR.
We already detach the request for a reload to the real reload, so there's already no way to understand why it failed.

fedepaol: Filed https://github.com/metallb/metallb/issues/1046 to keep track of this.

oribon: I was thinking about if something wrong happens before we reach the SIGHUP part (writeConfig fails for some reason/etc..)

fedepaol: Not sure I follow. We log for that in `generateAndReloadConfigFile`.
One other thing we could _ideally_ do is to retrigger the timeout in case of failure, not sure it's worth the effort. I suspect if something fails in that context, will continue to fail.

oribon: Thanks for explaining offline, I thought that in the current implementation we retry in case of an error but turns out we only log:
https://github.com/metallb/metallb/blob/f8e56adaf2d8a84f3d32e6a869fd8e31c5d06c3c/speaker/main.go#L367

russellb:
needs a rebase

fedepaol:
Known flake being addressed in https://github.com/metallb/metallb/pull/1057

PR #1041, opened 2021-11-11
E2E: write the config directly instead of using templates

fedepaol:
Here we render directly the bgpd conf and write it removing one extra jump through templates.


msherif1234:
LGTM

PR #1042, opened 2021-11-11
E2E: Dump the frr configuration in case of test failures

fedepaol:
In order to ease debugging, we dump the state of the external frr container in case of failure.

We also invert the order of beforeEach / afterEach and the framework creation so that our afterEach is registered before the framework's one and the namespace is not deleted when we bump.

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -25,6 +25,7 @@ import (
 	"strconv"
 	"time"
 
+	"go.universe.tf/metallb/e2etest/pkg/executor"
 	"go.universe.tf/metallb/e2etest/pkg/metrics"
 	"go.universe.tf/metallb/e2etest/pkg/routes"
 
@@ -680,7 +681,8 @@ func frrIsPairedOnPods(cs clientset.Interface, n *frrcontainer.FRR, ipFamily str
 		if ipFamily == "ipv6" {
 			address = n.Ipv6
 		}
-		toParse, err := framework.RunKubectl(testNameSpace, "exec", pods.Items[0].Name, "-c", "frr", "--", "vtysh", "-c", fmt.Sprintf("show bgp neighbor %s json", address))
+		podExecutor := executor.ForPod(testNameSpace, pods.Items[0].Name, "frr")
+		toParse, err := podExecutor.Exec("vtysh", "-c", fmt.Sprintf("show bgp neighbor %s json", address))
 		if err != nil {
 			return err
 		}
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -66,8 +66,8 @@ type containerConfig struct {
 }
 
 var _ = ginkgo.Describe("BGP", func() {
-	f := framework.NewDefaultFramework("bgp")
 	var cs clientset.Interface
+	var f *framework.Framework
 	containersConf := []containerConfig{
 		{
 			name: frrIBGP,
@@ -94,6 +94,36 @@ var _ = ginkgo.Describe("BGP", func() {
 			},
 		},
 	}
+
+	ginkgo.AfterEach(func() {
+		if ginkgo.CurrentGinkgoTestDescription().Failed {
+			for _, c := range frrContainers {
+				dump, err := frr.RawDump(c, "/etc/frr/bgpd.conf", "/tmp/frr.log")
+				framework.Logf("External frr dump for %s:\n%s %v", c.Name, dump, err)
+			}
+
+			speakerPods := getSpeakerPods(cs)
+			for _, pod := range speakerPods {
+				podExec := executor.ForPod(pod.Namespace, pod.Name, "frr")
+				dump, err := frr.RawDump(podExec, "/etc/frr/frr.conf", "/etc/frr/frr.log")
+				framework.Logf("External frr dump for pod %s\n%s %v", pod.Name, dump, err)
+			}
+			DescribeSvc(f.Namespace.Name)
+		}
+	})
+
+	ginkgo.AfterEach(func() {
+		ginkgo.By("Clearing the previous configuration")
+		// Clean previous configuration.
+		err := updateConfigMap(cs, configFile{})
+		framework.ExpectNoError(err)
+
+		err = stopFRRContainers(frrContainers)
+		framework.ExpectNoError(err)
+	})
+
+	f = framework.NewDefaultFramework("bgp")
+
 	ginkgo.BeforeEach(func() {
 		var err error
 		if containersNetwork == "host" {
@@ -107,20 +137,6 @@ var _ = ginkgo.Describe("BGP", func() {
 		cs = f.ClientSet
 		frrContainers, err = createFRRContainers(containersConf)
 		framework.ExpectNoError(err)
-
-	})
-
-	ginkgo.AfterEach(func() {
-		// Clean previous configuration.
-		err := updateConfigMap(cs, configFile{})
-		framework.ExpectNoError(err)
-
-		err = stopFRRContainers(frrContainers)
-		framework.ExpectNoError(err)
-
-		if ginkgo.CurrentGinkgoTestDescription().Failed {
-			DescribeSvc(f.Namespace.Name)
-		}
 	})
 
 	table.DescribeTable("A service of protocol load balancer should work with", func(ipFamily string, setProtocoltest string) {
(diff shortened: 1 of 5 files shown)


e2etest/pkg/frr/bgp.go
```diff
@@ -63,3 +63,26 @@ func Routes(exec executor.Executor) (map[string]Route, map[string]Route, error)
...
+	res = out
+	out, err = exec.Exec("vtysh", "-c", "show running-config")
+	if err != nil {
+		return "", errors.Wrapf(err, "Failed exec show bgp neighbor %s", res)
+	}
+	res = res + out
+	out, err = exec.Exec("cat", "/etc/frr/bgpd.conf")
+	if err != nil {
+		return "", errors.Wrapf(err, "Failed to cat bgpd.conf file %s", res)
+	}
```
msherif1234, line 84: while you are there can add this to the e2e config template
log timestamp precision 3

and cat the file too to see any bgpd errors

e2etest/pkg/frr/bgp.go
```diff
@@ -63,3 +63,26 @@ func Routes(exec executor.Executor) (map[string]Route, map[string]Route, error)
 	}
 	return v4Routes, v6Routes, nil
 }
+
+// RawDump dumps all the low leven info as a single string.
+// To be used for debugging in order to print the status of the frr instance.
+func RawDump(exec executor.Executor) (string, error) {
+	res := ""
+	out, err := exec.Exec("vtysh", "-c", "show bgp neighbor")
```
msherif1234, line 71: if u have issues with bgp that will error out and prevent us from getting all other debug info maybe move that to end as best effort so we can get a chance to see running conifg and other useful info ?

e2etest/bgp_test.go
```diff
@@ -93,32 +93,42 @@ var _ = ginkgo.Describe("BGP", func() {
...
 		err = stopFRRContainers(frrContainers)
 		framework.ExpectNoError(err)
+	})
 
-		if ginkgo.CurrentGinkgoTestDescription().Failed {
-			DescribeSvc(f.Namespace.Name)
+	f = framework.NewDefaultFramework("bgp")
+
+	ginkgo.BeforeEach(func() {
+		var err error
```
msherif1234, line 120: u don't need var u can initialize down when u use it ?

e2etest/bgp_test.go
```diff
@@ -93,32 +93,42 @@ var _ = ginkgo.Describe("BGP", func() {
...
 			}
+			DescribeSvc(f.Namespace.Name)
 		}
-		cs = f.ClientSet
-		frrContainers, err = createFRRContainers(containersConf)
-		framework.ExpectNoError(err)
 	})
 
 	ginkgo.AfterEach(func() {
+		ginkgo.By("Clearing the previous configuration")
```
msherif1234, line 108: I noticed if I ran the test multiple time we endup with lots of svc so we don't clean created service, can we add it here ?

e2etest/bgp_test.go
fedepaol: I can't because I need to assign the external `frrContainer` variable. If I do
```
		frrContainers, err = createFRRContainers(containersConf)
```
it will create an unused local version of `frrContainers`

e2etest/pkg/frr/bgp.go
fedepaol: I will

e2etest/pkg/frr/bgp.go
fedepaol: If there are issues with bgp we would still get an output, we get an error if we are really not able to execute the command.
I'll move anyhow, it shouldn't change much.

e2etest/bgp_test.go
fedepaol: That's if you get a failure / it gets interrupted in the middle, then framework doesn't clean the namespace.
I want to have a follow up to ensure external frr containers are wiped out when tests start, I'll try to incorporate that change there.

e2etest/pkg/frr/bgp.go
fedepaol: Done

e2etest/pkg/frr/bgp.go
fedepaol: I am logging under /tmp , /etc/frr is mounted so I think we can't write there

msherif1234:
what about collecting frr speaker side too ?

e2etest/bgp_test.go
```diff
@@ -93,32 +93,42 @@ var _ = ginkgo.Describe("BGP", func() {
...
-			if net.ParseIP(hostIPv4) == nil {
-				framework.Fail("Invalid hostIPv4")
-			}
-			if net.ParseIP(hostIPv6) == nil {
-				framework.Fail("Invalid hostIPv6")
+
+	ginkgo.AfterEach(func() {
+		if ginkgo.CurrentGinkgoTestDescription().Failed {
+			for _, c := range frrContainers {
+				dump, err := frr.RawDump(c)
```
oribon, line 100: Can you have this dumped to a file instead/in addition? I think it'll be easier to navigate a file (that is also uploaded to the ci's artifacts).

e2etest/bgp_test.go
```diff
@@ -93,32 +93,42 @@ var _ = ginkgo.Describe("BGP", func() {
...
-	ginkgo.BeforeEach(func() {
-		var err error
-		if containersNetwork == "host" {
-			if net.ParseIP(hostIPv4) == nil {
-				framework.Fail("Invalid hostIPv4")
-			}
-			if net.ParseIP(hostIPv6) == nil {
-				framework.Fail("Invalid hostIPv6")
+
+	ginkgo.AfterEach(func() {
```
oribon, line 97: Is there a guarantee that the first AfterEach is always called before the second one?

e2etest/pkg/frr/bgp.go
```diff
@@ -63,3 +63,34 @@ func Routes(exec executor.Executor) (map[string]Route, map[string]Route, error)
 	}
 	return v4Routes, v6Routes, nil
 }
+
+// RawDump dumps all the low leven info as a single string.
+// To be used for debugging in order to print the status of the frr instance.
+func RawDump(exec executor.Executor) (string, error) {
+	res := ""
+	out, err := exec.Exec("vtysh", "-c", "show running-config")
```
oribon, line 71: Will it add value having the command added before its output?
So the result would look something like:
```
show running-config
...
--------
cat /etc/frr/bgpd.conf
...
--------
...
```

e2etest/bgp_test.go
fedepaol: I am torn about this. It found dumping it as part of the log execution made quick debugging easier even without the need to download the files because you see the failed test and immediately go above and see the configuration.
Having it in a file means we need to download, cross check, etc. I'd keep it as it is for now.

e2etest/bgp_test.go
fedepaol: Yes. Ginkgo parses them in order.
That's also the reason why I moved the framework initialization, otherwise it would have cleared the namespace before we do the dump.

e2etest/pkg/frr/bgp.go
fedepaol: Yep, even with some kind of separator.

e2etest/bgp_test.go
```diff
@@ -680,7 +688,8 @@ func frrIsPairedOnPods(cs clientset.Interface, n *frrcontainer.FRR, ipFamily str
 		if ipFamily == "ipv6" {
 			address = n.Ipv6
 		}
-		toParse, err := framework.RunKubectl(testNameSpace, "exec", pods.Items[0].Name, "-c", "frr", "--", "vtysh", "-c", fmt.Sprintf("show bgp neighbor %s json", address))
+		podExecutor := executor.ForPod(testNameSpace, pods.Items[0].Name, "frr")
```
oribon, line 691: Can this be above the eventually?

e2etest/pkg/frr/bgp.go
```diff
@@ -75,3 +75,39 @@ func NeighborConnected(neighborJson string) (bool, error) {
 	}
 	return n.Connected, nil
 }
+
+// RawDump dumps all the low leven info as a single string.
```
oribon, line 79: typo? (leven)

e2etest/pkg/metrics/metrics.go
```diff
@@ -32,7 +32,8 @@ func ForPod(executor, target *corev1.Pod, namespace string) ([]map[string]*dto.M
 
 	for _, p := range ports {
 		metricsUrl := path.Join(net.JoinHostPort(target.Status.PodIP, strconv.Itoa(p)), "metrics")
-		metrics, err := framework.RunKubectl(namespace, "exec", executor.Name, "--", "wget", "-qO-", metricsUrl)
+		podExecutor := executor.ForPod(namespace, source.Name, source.Spec.Containers[0].Name)
```
oribon, line 35: Same, can this be above the range?

e2etest/pkg/metrics/metrics.go
```diff
@@ -11,15 +11,15 @@ import (
...
 	"github.com/prometheus/common/expfmt"
+	"go.universe.tf/metallb/e2etest/pkg/executor"
 	corev1 "k8s.io/api/core/v1"
 
 	"github.com/pkg/errors"
-	"k8s.io/kubernetes/test/e2e/framework"
 )
 
 // MetricsForPod returns the parsed metrics for the given pod, scraping them
 // from the executor pod.
```
oribon, line 21: nit: should be changed to source pod

oribon:
LGTM

PR #1043, opened 2021-11-12
Export MetallB FRR logging as envvar in FRR manifest

msherif1234:
user can overwrite the default logging level if needed via envvar settings

### DIFF
--- a/internal/bgp/frr/frr.go
+++ b/internal/bgp/frr/frr.go
@@ -168,8 +168,12 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
 
 	config := &frrConfig{
 		Hostname: hostname,
-		Loglevel: "informational", // TODO - make loglevel configurable via envvar.
+		Loglevel: "informational",
 		Routers:  make(map[string]*routerConfig),
+	}
+	frrLogLevel, found := os.LookupEnv("FRR_LOGGING_LEVEL")
+	if found {
+		config.Loglevel = frrLogLevel
 	}
 
 	for _, s := range sm.sessions {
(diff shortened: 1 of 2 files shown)


internal/bgp/frr/frr.go
```diff
@@ -168,9 +168,13 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
 
 	config := &frrConfig{
 		Hostname: hostname,
-		Loglevel: "informational", // TODO - make loglevel configurable via envvar.
+		Loglevel: "informational",
 		Routers:  make(map[string]*routerConfig),
 	}
+	frrLogLevel, found := os.LookupEnv("FRR_LOGGING_LEVEL")
```
russellb, line 174: How about we prefix all metallb env vars with `METALLB_` ?

msherif1234: currently all FRR env started with FRR_blah...u suggesting adding METALLB_FRR_balh....to new and existing envvar ?

russellb: That would be my preference, but I forgot that we had other vars as `FRR_` already, so don't worry about it

manifests/metallb-frr.yaml
```diff
@@ -497,6 +497,8 @@ spec:
               value: /etc/frr_reloader/reloader.pid
             - name: METALLB_BGP_TYPE
               value: frr
+            - name: FRR_LOGGING_LEVEL
```
russellb, line 500: We should document this somehow, even if it's a comment block in this manifest for now.  I'd expect it to list what the default value is, and what other valid choices are

PR #1045, opened 2021-11-14
Add Operator mode for e2e tests

oribon:
Currently how the e2e tests work is by changing MetalLB's ConfigMap directly.
Here we add a way for the tests to use the Operator's CRDs instead of
manipulating the ConfigMap. That'll enable us to have a true e2e scenario
for the Operator, meaning that from one side (the Operator's e2e) it
validates that it configures MetalLB's ConfigMap the way it thinks it should,
and with the new tests mode we could also know that that the operator is
configuring MetalLB properly from the other side (MetalLBs e2e) = the way
MetalLB expects it to be configured.

This may also serve as a base for when(if) the Operator is integrated with
MetalLB and not as a standalone.


oribon:
Marked as draft, waiting for BFDProfile CRD in Operator.

e2etest/pkg/config/update.go
```diff
@@ -0,0 +1,295 @@
...
+		}
+	}
+
+	/*
+		Since a peer doesn't have a name we need to clean them before creating.
+		Without this the webhook fails the request because we are trying to create
+		a BGPPeer with the same configuration under a different (generated) name.
+		TODO: is there a better way to handle this?
+	*/
+	err := o.DeleteAllOf(context.Background(), &operatorv1beta1.BGPPeer{}, client.InNamespace(o.namespace))
```
fedepaol, line 121: When does this happen? Does it mean that we are leaking peers between tests?
I assumed each test cleans the config afterwards.

e2etest/bgp_test.go
```diff
@@ -70,6 +71,7 @@ type containerConfig struct {
 }
 
 var _ = ginkgo.Describe("BGP", func() {
+	var uc config.UpdateCleaner
```
fedepaol, line 74: `configUpdater` ? uc is not really clear

e2etest/e2etest_suite_test.go
```diff
@@ -101,6 +112,9 @@ func TestE2E(t *testing.T) {
 }
 
 var _ = ginkgo.BeforeSuite(func() {
+	// Make sure the framework's kubeconfig is set.
+	framework.ExpectEqual(framework.TestContext.KubeConfig != "", true, fmt.Sprintf("%s env var not set", clientcmd.RecommendedConfigPathEnvVar))
```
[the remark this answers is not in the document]
oribon: done

e2etest/pkg/config/update.go
```diff
@@ -0,0 +1,295 @@
...
+	apierr "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+type Updater interface {
```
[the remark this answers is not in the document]
oribon: done, declaring a single interface now

e2etest/bgp_test.go
oribon: done

e2etest/pkg/config/update.go
oribon: It happens when running the `should work after subsequent configuration updates` tests.
The tests create configuration with the same peers but adding an addresspool each iteration. 
The current implementation doesn't care about this since it always writes over the current ConfigMap, 
but with CRs this causes a conflict because it also recreates the BGPPeers (with the same config) in each iteration.

e2etest/pkg/config/update.go
```diff
@@ -0,0 +1,287 @@
...
+	Update(f File) error
+	Clean() error
+}
+
+type cmUpdater struct {
```
fedepaol, line 27: configmapUpdater?

e2etest/pkg/config/update.go
```diff
@@ -0,0 +1,287 @@
...
+	clientset.Interface
+	name      string
+	namespace string
+}
+
+func NewConfigMap(c clientset.Interface, ns string) *cmUpdater {
```
fedepaol, line 33: UpdaterForConfigmap?

e2etest/pkg/config/update.go
```diff
@@ -0,0 +1,287 @@
...
+func (c cmUpdater) Clean() error {
+	return c.Update(File{})
+}
+
+type operatorUpdater struct {
+	client.Client
+	namespace string
+}
+
+func NewOperator(r *rest.Config, ns string) (*operatorUpdater, error) {
```
fedepaol, line 71: UpdaterForOperator

e2etest/pkg/config/update.go
oribon: done

e2etest/pkg/config/update.go
oribon: done

e2etest/pkg/config/update.go
oribon: done

fedepaol:
LGTM

Issue #1048, opened 2021-11-15
Helm: Include release namespace in all relevant resources


flyinggecko:
Hey,

I will try to explain this a bit:
We use gitlab-k8s-agents to bootstrap some applications in our clusters. To generate manifests, that get deployed, we use mostly helm like this for metallb:
```bash
> helm template --namespace metallb-system metallb metallb/metallb > manifest.yaml
```

This results in the following manifest: https://gist.github.com/flyinggecko/89f22815da8131d9e8fecb1fed6320b0

The problem is, that some resources now contain the namespace and some don't.
For example the roles and role-bindings contain the ns, but the deployment/daemonset and config don't.

This doesn't happen if you use helm the normal way e.g. `helm upgrade --install --namespace metallb-system metallb metallb/metallb`. 

Versions:
  - MetalLB: 0.11.0
  - Kubernetes: independent of version
  - Network: independent of used network
  - Kube-Proxy/Ipvs: independent of this

So, is this something you would except a merge request? I would then create one.

Best regard

fedepaol:
cc @gclawes

gclawes:
Good catch!  The `.metadata.namespace` field should be removed from all resources, helm should apply that on install to namespaced resources.

I also noticed that the default value of `.Values.prometheus.namespace` is `monitoring`, this was probably carried over from my test setup when I wrote the chart.  Properly, it should be un-set, and the chart should throw an error when `.Values.prometheus.podmonitor.enabled` is `true`.  I'll prepare a PR containing both of these fixes.

gclawes:
Ah, I see, `.Values.prometheus.namespace` was added here: https://github.com/metallb/metallb/commit/48cb1c3b8a97487a825f24860ecbf5c2b3919817.

@fedepaol, I think it makes sense for `.Values.prometheus.namespace` and `.Values.prometheus.serviceAccount` not to have default values, and instead have validation in `_helpers.tpl` functions that fail if they are not provided only when `.Values.prometheus.podMonitor.enabled` is `true`.  Thoughts?

fedepaol:
Ah dang, so I was responsible!
Yes, makes total sense

> Good catch! The `.metadata.namespace` field should be removed from all resources, helm should apply that on install to namespaced resources.
> 
[...]

PR #1049, opened 2021-11-16
Helm chart namespace fixes

gclawes:
Addresses namespace issues raised in #1048

1. Remove `.metadata.namespace` from resources, helm should generate thes
2. Remove default values of `.Values.prometheus.serviceAccount` and `.Values.prometheus.namespace`, we should not assume what users default values of these are.  Validate them with the `required` function instead.
```
$ helm lint .
==> Linting .

1 chart(s) linted, 0 chart(s) failed
$ helm lint . --set prometheus.podMonitor.enabled=true
engine.go:167: [INFO] Missing required value: .Values.prometheus.serviceAccount must be defined when .Values.prometheus.podMonitor.enabled == true
engine.go:167: [INFO] Missing required value: .Values.prometheus.namespace must be defined when .Values.prometheus.podMonitor.enabled == true
==> Linting .

1 chart(s) linted, 0 chart(s) failed
$ helm lint . --set prometheus.podMonitor.enabled=true --set prometheus.namespace=monitoring --set prometheus.serviceAccount=kube-prometheus
==> Linting .

1 chart(s) linted, 0 chart(s) failed
```

### DIFF
--- a/charts/metallb/templates/podmonitor.yaml
+++ b/charts/metallb/templates/podmonitor.yaml
@@ -71,7 +71,6 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
   name: {{ template "metallb.fullname" . }}-prometheus
-  namespace: {{ .Release.Namespace }}
 rules:
   - apiGroups:
       - ""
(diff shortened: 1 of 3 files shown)


fedepaol:
LGTM, thanks @gclawes

PR #1055, opened 2021-11-16
Use bgp-communities to update BGP community advertisements with FRR

msherif1234:

FRR template lacking ability to update bgp advertisment community, This PR extend the template 
to add the additional support and unit-test

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -56,6 +56,7 @@ const (
 	baseRouterID    = "10.10.10.%d"
 	v4PoolAddresses = "192.168.10.0/24"
 	v6PoolAddresses = "fc00:f853:0ccd:e799::/124"
+	CommunityNoAdv  = "65535:65282" // 0xFFFFFF02: NO_ADVERTISE
 )
 
 var (
@@ -693,6 +694,78 @@ var _ = ginkgo.Describe("BGP", func() {
 		},
 			table.Entry("IPV4", "ipv4"),
 			table.Entry("IPV6", "ipv6"))
+
+		table.DescribeTable("configure bgp community and verify it gets propagated",
+			func(addressPools []addressPool, ipFamily string) {
+				configData := configFile{
+					Peers: peersForContainers(frrContainers, ipFamily),
+					Pools: addressPools,
+				}
+				for _, c := range frrContainers {
+					pairExternalFRRWithNodes(cs, c, ipFamily)
+				}
+
+				err := updateConfigMap(cs, configData)
+				framework.ExpectNoError(err)
+
+				for _, c := range frrContainers {
+					validateFRRPeeredWithNodes(cs, c, ipFamily)
+				}
+
+				svc, _ := createServiceWithBackend(cs, f.Namespace.Name, "external-local-lb", testservice.TrafficPolicyCluster) // Is a sleep required here?
+
+				defer func() {
+					err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
+					framework.ExpectNoError(err)
+				}()
+
+				for _, i := range svc.Status.LoadBalancer.Ingress {
+					ginkgo.By("validate LoadBalancer IP is in the AddressPool range")
+					ingressIP := e2eservice.GetIngressPoint(&i)
+					err = validateIPInRange(addressPools, ingressIP)
+					framework.ExpectNoError(err)
+				}
+
+				allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+				framework.ExpectNoError(err)
+
+				for _, c := range frrContainers {
+					validateService(cs, svc, allNodes.Items, c)
+					Eventually(func() error {
+						return frr.ContainsCommunity(c, "no-advertise")
+					}, 4*time.Minute, 1*time.Second).Should(BeNil())
+				}
+			},
+			table.Entry("IPV4", []addressPool{
+				{
+					Name:     "bgp-test",
+					Protocol: BGP,
+					Addresses: []string{
+						"192.168.10.0/24",
+					},
+					BGPAdvertisements: []bgpAdvertisement{
+						{
+							Communities: []string{
+								CommunityNoAdv,
+							},
+						},
+					},
+				}}, "ipv4"),
+			table.Entry("IPV6", []addressPool{
+				{
+					Name:     "bgp-test",
+					Protocol: BGP,
+					Addresses: []string{
+						"fc00:f853:0ccd:e799::0-fc00:f853:0ccd:e799::18",
+					},
+					BGPAdvertisements: []bgpAdvertisement{
+						{
+							Communities: []string{
+								CommunityNoAdv,
+							},
+						},
+					},
+				}}, "ipv6"))
 	})
 })
 
--- a/e2etest/pkg/frr/bgp.go
+++ b/e2etest/pkg/frr/bgp.go
@@ -4,6 +4,7 @@ package frr
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 	"go.universe.tf/metallb/e2etest/pkg/executor"
@@ -110,4 +111,16 @@ func RawDump(exec executor.Executor, filesToDump ...string) (string, error) {
 	res = res + out
 
 	return res, nil
+}
+
+// ContainsCommunity check if the passed in community string exists in show bgp community.
+func ContainsCommunity(exec executor.Executor, community string) error {
+	res, err := exec.Exec("vtysh", "-c", "show bgp community-info")
+	if err != nil {
+		return err
+	}
+	if !strings.Contains(res, community) {
+		return errors.Wrapf(err, "show community %s doesn't include %s", res, community)
+	}
+	return nil
 }
--- a/internal/bgp/frr/config.go
+++ b/internal/bgp/frr/config.go
@@ -31,6 +31,19 @@ log file /etc/frr/frr.log {{.Loglevel}}
 log timestamp precision 3
 hostname {{.Hostname}}
 
+{{- range .Routers }}
+{{- range $n := .Neighbors }}
+{{- range $a := .Advertisements }}
+{{- if gt (len $a.Communities) 0}}
+route-map {{$n.Addr}}-map permit 10
+{{- range $c := $a.Communities }}
+  set community {{$c}} additive
+{{- end }}
+{{- end }}
+{{- end }}
+{{- end }}
+{{- end }}
+
 {{range .Routers -}}
 router bgp {{.MyASN}}
   no bgp ebgp-requires-policy
@@ -69,6 +82,9 @@ router bgp {{.MyASN}}
   address-family {{.Version}} unicast
     neighbor {{$n.Addr}} activate
     network {{.Prefix}}
+	{{- if gt (len .Communities) 0}}
+    neighbor {{$n.Addr}} route-map {{$n.Addr}}-map out
+    {{- end}}
   exit-address-family
 {{- end}}
 {{end -}}
@@ -138,8 +154,9 @@ type neighborConfig struct {
 }
 
 type advertisementConfig struct {
-	Version string
-	Prefix  string
+	Version     string
+	Prefix      string
+	Communities []string
 }
 
 // routerName() defines the format of the key of the "Routers" map in the
--- a/internal/bgp/frr/frr.go
+++ b/internal/bgp/frr/frr.go
@@ -12,7 +12,7 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"go.universe.tf/metallb/internal/bgp"
-	"go.universe.tf/metallb/internal/config"
+	metallbconfig "go.universe.tf/metallb/internal/config"
 )
 
 // As the MetalLB controller should handle messages synchronously, there should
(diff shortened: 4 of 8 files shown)


internal/bgp/frr/frr.go
```diff
@@ -252,17 +252,22 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
...
 			var version string
+			community := make([]string, 0)
 			if adv.Prefix.IP.To4() != nil {
 				version = "ipv4"
 			} else if adv.Prefix.IP.To16() != nil {
 				version = "ipv6"
 			}
-
+			for _, c := range adv.Communities {
+				a := c >> 16
```
sabinaaledort, line 262: maybe use better names than a and b? or add an explanation what's done here to create the community?

internal/bgp/frr/config.go
```diff
@@ -138,8 +154,9 @@ type neighborConfig struct {
 }
 
 type advertisementConfig struct {
-	Version string
-	Prefix  string
+	Version   string
+	Prefix    string
+	Community []string
```
sabinaaledort, line 159: should it be named Communities maybe?

internal/bgp/frr/frr.go
```diff
@@ -252,17 +252,22 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
 		   'neighbor.Advertisements' list. */
 		for _, adv := range s.advertised {
 			var version string
+			community := make([]string, 0)
```
sabinaaledort, line 255: same here, should it be named Communities maybe?

sabinaaledort:
maybe it makes sense to separate the config part and the tests into two commits?

sabinaaledort:
should it be documented in the website content?

internal/bgp/frr/frr.go
msherif1234: sure

msherif1234:
> should it be documented in the website content?
its already doc here https://metallb.universe.tf/configuration/#bgp-configuration

sabinaaledort:
> > should it be documented in the website content?
> > its already doc here https://metallb.universe.tf/configuration/#bgp-configuration

sorry missed it

e2etest/bgp_test.go
```diff
@@ -150,6 +152,14 @@ var _ = ginkgo.Describe("BGP", func() {
 					Name:      "bgp-test",
 					Protocol:  BGP,
 					Addresses: poolAddresses,
+					BGPAdvertisements: []bgpAdvertisement{
```
fedepaol, line 155: I'd rather have a single new tests with the communities (and in general, were we shuffle the parameters) than extending all the tests with communities so we make sure the basic use case (all default) works.

e2etest/bgp_test.go
```diff
@@ -150,6 +152,14 @@ var _ = ginkgo.Describe("BGP", func() {
...
+						{
+							Communities: []string{
```
fedepaol, line 157: Will these be available to parse on the receiver side?
If so the tests should verify that.

internal/bgp/frr/config.go
```diff
@@ -65,10 +78,13 @@ router bgp {{.MyASN}}
     neighbor {{$n.Addr}} activate
   exit-address-family
 {{- end}}
-{{- range .Advertisements }}
-  address-family {{.Version}} unicast
+{{- range $a := .Advertisements }}
```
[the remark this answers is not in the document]
msherif1234: not needed really

internal/bgp/frr/frr_test.go
```diff
@@ -184,6 +184,10 @@ func TestSingleAdvertisement(t *testing.T) {
 	adv := &bgp.Advertisement{
 		Prefix:  prefix,
 		NextHop: net.ParseIP("10.1.1.1"),
+		Communities: []uint32{
+			1111,
```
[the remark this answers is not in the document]
msherif1234: do u mean large number so it can be splitted to two values in : fmt instead of the higher 16bits being 0 ? Communities is slice of uint32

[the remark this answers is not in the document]
fedepaol: yes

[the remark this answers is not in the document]
fedepaol: and instead of having the number, having parseCommunity("1111:1111") to be more readable.

e2etest/bgp_test.go
msherif1234: added specific test case for community testing

e2etest/bgp_test.go
msherif1234: done w/o the need to snoop the packet on the receiver peer

e2etest/bgp_test.go
```diff
@@ -687,12 +689,76 @@ var _ = ginkgo.Describe("BGP", func() {
...
+				{
+					Name:     "bgp-test",
+					Protocol: BGP,
+					Addresses: []string{
+						"192.168.10.0/24",
+					},
+					BGPAdvertisements: []bgpAdvertisement{
+						{
+							Communities: []string{
+								CommunityNOPeer,
```
fedepaol, line 726: This should contain the name of a community defined in the communities section of the configuration, not the community itself.

e2etest/bgp_test.go
```diff
@@ -780,7 +846,7 @@ func validateFRRPeeredWithNodes(cs clientset.Interface, c *frrcontainer.FRR, ipF
 		framework.ExpectNoError(err)
 		err = frr.NeighborsMatchNodes(allNodes.Items, neighbors, ipFamily)
 		return err
-	}, 4*time.Minute, 1*time.Second).Should(BeNil())
+	}, 4*time.Minute, 1*time.Second).Should(chkFunc())
```
fedepaol, line 849: Checking that we are returning errors here may hide bugs, in general I'd rather be explicit in what we are looking for in a test and not checking for random failures.

As a side note, if we want to check a negative condition, we should use `Consistently` because here you may get a transitient error and exit immediately. When checking if an error persist, we must check it for x time.

e2etest/bgp_test.go
```diff
@@ -687,12 +689,76 @@ var _ = ginkgo.Describe("BGP", func() {
...
 				for j := 0; j <= i; j++ {
-					validateFRRPeeredWithNodes(cs, frrContainers[j], ipFamily)
+					validateFRRPeeredWithNodes(cs, frrContainers[j], ipFamily, BeNil)
 				}
 			}
 		},
 			table.Entry("IPV4", "ipv4"),
 			table.Entry("IPV6", "ipv6"))
+
+		table.DescribeTable("configure bgp community NOPEER attribute and verify FRR session can't get to established state",
```
fedepaol, line 699: I'd just add a test verifying that the community gets propagated to the other end.
I tried to tinker with FRR, one option is to run `show bgp community-info` and to check the output. In my case I got 
```
show bgp community-info 
Address Refcnt Community
[0x7f92505791d0] (1) no-advertise
```

So for example running it on the external containers before creating the service, check no communities are sent over, create the service and check there is a community.

e2etest/bgp_test.go
msherif1234: it can be the name or community itself both are supported

e2etest/bgp_test.go
```diff
@@ -687,12 +689,76 @@ var _ = ginkgo.Describe("BGP", func() {
...
+					Pools: addressPools,
+				}
+				for _, c := range frrContainers {
+					pairExternalFRRWithNodes(cs, c, ipFamily)
+				}
+
+				err := updateConfigMap(cs, configData)
+				framework.ExpectNoError(err)
+
+				for _, c := range frrContainers {
```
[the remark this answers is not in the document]
msherif1234: not for this specific community I used for testing here, I used specail value to prevent peering as I explained in the comment beside the code

e2etest/bgp_test.go
msherif1234: I used same peers with and w/o no peer community to proof the update message carry the new attribute

e2etest/bgp_test.go
msherif1234: no adv is open type as I stated above I used no peer

e2etest/bgp_test.go
fedepaol: ah, neat!

e2etest/bgp_test.go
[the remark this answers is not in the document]
fedepaol: Our code won't add the communities in the file if we don't advertise a service.

e2etest/bgp_test.go
fedepaol: you should gomega.Consistently instead of Eventually to check errors happening. Checking for random errors may cause false positives.

e2etest/bgp_test.go
fedepaol: yes I am suggesting using something that gets propagated so we have more confidence.

e2etest/bgp_test.go
```diff
@@ -955,3 +1030,19 @@ func uint32Ptr(n uint32) *uint32 {
 func boolPtr(b bool) *bool {
 	return &b
 }
+
+func frrIsCommunitySet(cs clientset.Interface, n *frrcontainer.FRR) {
+	pods := getSpeakerPods(cs)
+	podExecutor := executor.ForPod(testNameSpace, pods[0].Name, "frr")
```
fedepaol, line 1036: I would check this in the external containers rather than in the pods.

Also, I think the community-info part (by parsing the result) belongs here: https://github.com/metallb/metallb/blob/main/e2etest/pkg/frr/bgp.go

e2etest/bgp_test.go
```diff
@@ -955,3 +1030,19 @@ func uint32Ptr(n uint32) *uint32 {
...
+
+	Eventually(func() error {
+		res, err := podExecutor.Exec("vtysh", "-c", "show bgp  community-info")
```
fedepaol, line 1039: nit: extra space

e2etest/bgp_test.go
```diff
@@ -955,3 +1030,18 @@ func uint32Ptr(n uint32) *uint32 {
 func boolPtr(b bool) *bool {
 	return &b
 }
+
+func frrIsCommunitySet(cs clientset.Interface, n *frrcontainer.FRR) {
```
fedepaol, line 1034: I'd have the community fetching logic in this file: https://github.com/metallb/metallb/blob/main/e2etest/pkg/frr/bgp.go
Having a function returning a map community -> refcount.

e2etest/bgp_test.go
```diff
@@ -955,3 +1030,18 @@ func uint32Ptr(n uint32) *uint32 {
...
+	exec := executor.ForContainer("frr")
```
fedepaol, line 1035: there's no "frr" container on the host, you need to choose the eBgp or iBgp one. Also, you can use directly n as it implements the `Executor` interface.

e2etest/bgp_test.go
msherif1234: not sure why we need map and ref cnt here

e2etest/bgp_test.go
msherif1234: not needed anymore

e2etest/bgp_test.go
fedepaol: As discussed offline, we don't need having a map[community]bool / slice of available communities is enough. RefCnt doesn't matter for our purpouses.

e2etest/pkg/frr/bgp.go
```diff
@@ -111,3 +111,14 @@ func RawDump(exec executor.Executor, filesToDump ...string) (string, error) {
 
 	return res, nil
 }
+
+func CommunityInfo(exec executor.Executor, community string) error {
```
fedepaol, line 115: I'd rename this to "ContainsCommunity" since it does not return the info.
Also, can you add the doc text?

e2etest/bgp_test.go
```diff
@@ -693,6 +694,79 @@ var _ = ginkgo.Describe("BGP", func() {
...
+					pairExternalFRRWithNodes(cs, c, ipFamily)
+				}
+
+				err := updateConfigMap(cs, configData)
+				framework.ExpectNoError(err)
+
+				for _, c := range frrContainers {
+					validateFRRPeeredWithNodes(cs, c, ipFamily)
+				}
+				for _, c := range frrContainers {
```
fedepaol, line 714: same check added twice.

msherif1234: sorry :(

e2etest/bgp_test.go
```diff
@@ -955,3 +1029,9 @@ func uint32Ptr(n uint32) *uint32 {
 func boolPtr(b bool) *bool {
 	return &b
 }
+
+func frrIsCommunitySet(cs clientset.Interface, n *frrcontainer.FRR) {
+	Eventually(func() error {
```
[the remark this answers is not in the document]
msherif1234: it turns out if we pass function directly to Eventuallt it should be of 0 args, its expected function with no args and one or more return values I wendt back to the func() wrapper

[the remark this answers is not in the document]
fedepaol: Sorry, I just realized my comment wasn't clear. What I meant was to remove the `frrIsCommunitySet` function and add the evenutally loop directly in https://github.com/metallb/metallb/pull/1055/files#diff-a115369e9006a64f1497f61daada01a4136fa15a1feb79f86f11a73a0cdc5d36R734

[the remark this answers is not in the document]
msherif1234: its more cleaner this way and jive with the flow of the rest of functions where eventually were handled outside the test itself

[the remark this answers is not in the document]
fedepaol: Actually, there are a few places in the tests where eventually is explicit. This is compact enough not to hurt readability, and I am not sure there's value in having one more level of indirection.
We have wrappers when the code is duplicated and / or we need to do a bit more than eventually.

fedepaol:
LGTM, thanks!

Let's wait for CI

e2etest/bgp_test.go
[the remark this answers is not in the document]
msherif1234: ok moved the Evntually call to the test

fedepaol:
known flakes, merging.

PR #1056, opened 2021-11-16
Fail when trying to load frr only features in native mode

fedepaol:
Ideally, the checks of FRR only fields should happen in the config handler of the speaker, which is implementation aware.
This is not possible to achieve because of the defaulting mechanism in place.
Since it makes sense to have all the checks in the same place, and those checks must happen before the defaulting, we inject an implementation driven hook to check the raw config.

Current checks are meant to drop:
- ipv6 pools in case of native bgp mode
- ipv6 aggregation length
- BFD profiles


internal/config/validate_test.go
```diff
@@ -0,0 +1,110 @@
...
+			config: &configFile{
+				Peers: []peer{
+					{
+						Addr: "1.2.3.4",
+					},
+				},
+			},
+			mustFail: false,
+		},
+	}
```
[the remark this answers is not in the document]
russellb: nevermind, you're testing "DiscordFRROnly()" explicitly ...

russellb:
Can you take a look at the CI failures?

fedepaol:
> Can you take a look at the CI failures?

There's definetely an issue with tests now, as we were creating always pools with v6 / v4 addresses regardless of the family we are testing, and native now will reject any bgp pool containing v6 addresses.

fedepaol:
Should be fixed. There was another bfd flake I was not able to reproduce locally.
Need to push https://github.com/metallb/metallb/pull/1042 forward.

internal/config/validate.go
```diff
@@ -0,0 +1,54 @@
...
+			return fmt.Errorf("peer %s has keepalive set on native bgp mode", p.Addr)
+		}
+	}
+	if len(c.BFDProfiles) > 0 {
+		return errors.New("bfd profiles section set")
+	}
+	for _, p := range c.Pools {
+		for _, a := range p.BGPAdvertisements {
+			if a.AggregationLengthV6 != nil {
+				return fmt.Errorf("pool %s has aggregation lenghtv6 was set on native bgp mode", p.Name)
```
msherif1234, line 30: typo in the error msg

msherif1234:
very nice!! Thanks for consolidating config validation this validation function will keep growing for sure

internal/config/validate.go
fedepaol: Good catch, I also reworked the errors so it's clear what field they refers to.

russellb (approved):

PR #1058, opened 2021-11-17
FRR: Have FRR logs visible when running kubectl logs -c frr

oribon:
Currently the FRR logs aren't visible when running kubectl logs
(even when running in stdout logging mode) because the daemons run
under frr:frr and their stdout differs from watchfrr's (that runs
under a different user).
Bypassing this by tailing the log file seems good enough,
we could change this later if needed.

### DIFF
--- a/manifests/metallb-frr.yaml
+++ b/manifests/metallb-frr.yaml
@@ -404,10 +404,13 @@ data:
   vtysh.conf: |+
     service integrated-vtysh-config
   frr.conf: |+
+    ! This file gets overriden the first time the speaker renders a config.
+    ! So anything configured here is only temporary.
     frr version 7.5.1
     frr defaults traditional
     hostname Router
     line vty
+    log file /etc/frr/frr.log informational
 
 ---
 apiVersion: apps/v1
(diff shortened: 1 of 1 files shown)


manifests/metallb-frr.yaml
```diff
@@ -484,6 +487,11 @@ spec:
               mountPath: /var/run/frr
             - name: frr-conf
               mountPath: /etc/frr
+          # The command is FRR's default entrypoint & waiting for the log file to appear and tailing it.
+          # If the log file isn't created in 60 seconds the tail fails and the container is restarted.
+          # This workaround is needed to have the frr logs as part of kubectl logs -c frr < speaker_pod_name >.
+          command: ["/bin/sh"]
+          args: ["-c","/sbin/tini -- /usr/lib/frr/docker-start & attempts=0; until [[ -f /etc/frr/frr.log || $attempts -eq 60 ]]; do sleep 1 && attempts=$(( $attempts + 1 )); done && tail -f /etc/frr/frr.log"]
```
fedepaol, line 494: Just one ask. Can you make this multiline? I think it would be easier to read / understand.

oribon: changed, is it better now?

manifests/metallb-frr.yaml
```diff
@@ -404,10 +404,13 @@ data:
   vtysh.conf: |+
     service integrated-vtysh-config
   frr.conf: |+
+    ! This file gets overriden the first time the speaker renders a config.
+    ! So anything configured here is only temporary.
     frr version 7.5.1
     frr defaults traditional
     hostname Router
     line vty
+    log file /etc/frr/frr.log informational
```
fedepaol, line 413: This is gonna be overridden as soon as the speaker takes over right?

oribon: Yes, as soon as the speaker makes the first reload due to a config change.

fedepaol:
LGTM

fedepaol:
Also, checked the frr logs in the artifacts, they are present. Nice!

PR #1064, opened 2021-11-18
E2E: throw vtysh command errors in the eventually loop

fedepaol:
We start the external container and start sending commands.
There was one flake where bgpd was not ready yet, and in that case we fail on the error. Here we return the error and handle it in the eventually loop, so if it is transient the test will pass.


russellb:
lgtm assuming CI passes

PR #1065, opened 2021-11-19
Refactor Metallb svc controller and speaker to support dual-stack

msherif1234:
Refactor Metallb service controller to support dual stack, this PR is continuation of the effort started by 
PR #727 while trying to share most of the code between single stack and dual stack services, also it modifies to speaker code to return more than one address instead of one Loadbalancer IP.

### - Testing 

 - bring up kind dual-stack cluster 
 - create metallb configmap
 ```
  k -n metallb-system get configmap config -o yaml
 apiVersion: v1
 data:
   config: |
     peers:
     - peer-address: fc00:f853:ccd:e793::5
       peer-asn: 64512
       my-asn: 64512
     address-pools:
     - name: dev-env-bgp
       protocol: bgp
       addresses:
       - 192.168.10.0/24
       - fc00:f853:0ccd:e799::/124
 kind: ConfigMap
 ```
 - create dual-stack svc
 ```
  k create -f dev-env/testsvc_dualstack.yaml 
 ```
 - loadbalancer svc
 ```
  k get svc nginx 
 NAME    TYPE           CLUSTER-IP   EXTERNAL-IP                         PORT(S)        AGE
 nginx   LoadBalancer   10.96.6.42   192.168.10.0,fc00:f853:ccd:e799::   80:30253/TCP   18s
 ```

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -46,14 +46,14 @@ func ipnet(s string) *net.IPNet {
 	return n
 }
 
-func statusAssigned(ip string) v1.ServiceStatus {
+func statusAssigned(ips []string) v1.ServiceStatus {
+	lbIngressIPs := []v1.LoadBalancerIngress{}
+	for _, ip := range ips {
+		lbIngressIPs = append(lbIngressIPs, v1.LoadBalancerIngress{IP: ip})
+	}
 	return v1.ServiceStatus{
 		LoadBalancer: v1.LoadBalancerStatus{
-			Ingress: []v1.LoadBalancerIngress{
-				{
-					IP: ip,
-				},
-			},
+			Ingress: lbIngressIPs,
 		},
 	}
 }
@@ -133,6 +133,11 @@ func TestControllerMutation(t *testing.T) {
 				AutoAssign: false,
 				CIDR:       []*net.IPNet{ipnet("2000::1/128")},
 			},
+			"pool5": {
+				Protocol:   config.BGP,
+				AutoAssign: true,
+				CIDR:       []*net.IPNet{ipnet("1.2.3.0/31"), ipnet("1000::/127")},
+			},
 		},
 	}
 
@@ -163,8 +168,8 @@ func TestControllerMutation(t *testing.T) {
 			desc: "simple non-LoadBalancer",
 			in: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type:      "ClusterIP",
-					ClusterIP: "1.2.3.4",
+					Type:       "ClusterIP",
+					ClusterIPs: []string{"1.2.3.4"},
 				},
 			},
 		},
@@ -173,16 +178,16 @@ func TestControllerMutation(t *testing.T) {
 			desc: "simple LoadBalancer",
 			in: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type:      "LoadBalancer",
-					ClusterIP: "1.2.3.4",
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4"},
 				},
 			},
 			want: &v1.Service{
 				Spec: v1.ServiceSpec{
-					ClusterIP: "1.2.3.4",
-					Type:      "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4"},
+					Type:       "LoadBalancer",
 				},
-				Status: statusAssigned("1.2.3.0"),
+				Status: statusAssigned([]string{"1.2.3.0"}),
 			},
 		},
 
@@ -190,18 +195,18 @@ func TestControllerMutation(t *testing.T) {
 			desc: "request specific IP",
 			in: &v1.Service{
 				Spec: v1.ServiceSpec{
-					ClusterIP:      "1.2.3.4",
+					ClusterIPs:     []string{"1.2.3.4"},
 					Type:           "LoadBalancer",
 					LoadBalancerIP: "1.2.3.1",
 				},
 			},
 			want: &v1.Service{
 				Spec: v1.ServiceSpec{
-					ClusterIP:      "1.2.3.4",
+					ClusterIPs:     []string{"1.2.3.4"},
 					Type:           "LoadBalancer",
 					LoadBalancerIP: "1.2.3.1",
 				},
-				Status: statusAssigned("1.2.3.1"),
+				Status: statusAssigned([]string{"1.2.3.1"}),
 			},
 		},
 
@@ -211,7 +216,7 @@ func TestControllerMutation(t *testing.T) {
 				Spec: v1.ServiceSpec{
 					Type:           "LoadBalancer",
 					LoadBalancerIP: "please sir may I have an IP address thank you",
-					ClusterIP:      "1.2.3.4",
+					ClusterIPs:     []string{"1.2.3.4"},
 				},
 			},
 			wantErr: true,
@@ -223,15 +228,15 @@ func TestControllerMutation(t *testing.T) {
 				Spec: v1.ServiceSpec{
 					Type:           "LoadBalancer",
 					LoadBalancerIP: "1.2.3.4",
-					ClusterIP:      "1.2.3.4",
+					ClusterIPs:     []string{"1.2.3.4"},
 				},
-				Status: statusAssigned("1.2.3.1"),
+				Status: statusAssigned([]string{"1.2.3.1"}),
 			},
 			want: &v1.Service{
 				Spec: v1.ServiceSpec{
 					Type:           "LoadBalancer",
 					LoadBalancerIP: "1.2.3.4",
-					ClusterIP:      "1.2.3.4",
+					ClusterIPs:     []string{"1.2.3.4"},
 				},
 			},
 			wantErr: true,
@@ -246,8 +251,8 @@ func TestControllerMutation(t *testing.T) {
 					},
 				},
 				Spec: v1.ServiceSpec{
-					Type:      "LoadBalancer",
-					ClusterIP: "1.2.3.4",
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4"},
 				},
 			},
 			want: &v1.Service{
@@ -257,10 +262,10 @@ func TestControllerMutation(t *testing.T) {
 					},
 				},
 				Spec: v1.ServiceSpec{
-					Type:      "LoadBalancer",
-					ClusterIP: "1.2.3.4",
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4"},
 				},
-				Status: statusAssigned("1.2.3.0"),
+				Status: statusAssigned([]string{"1.2.3.0"}),
 			},
 		},
 
@@ -273,10 +278,10 @@ func TestControllerMutation(t *testing.T) {
 					},
 				},
 				Spec: v1.ServiceSpec{
-					ClusterIP: "1.2.3.4",
-					Type:      "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4"},
+					Type:       "LoadBalancer",
 				},
-				Status: statusAssigned("1.2.3.0"),
+				Status: statusAssigned([]string{"1.2.3.0"}),
 			},
 			want: &v1.Service{
 				ObjectMeta: metav1.ObjectMeta{
@@ -285,10 +290,10 @@ func TestControllerMutation(t *testing.T) {
 					},
 				},
 				Spec: v1.ServiceSpec{
-					ClusterIP: "1.2.3.4",
-					Type:      "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4"},
+					Type:       "LoadBalancer",
 				},
-				Status: statusAssigned("3.4.5.6"),
+				Status: statusAssigned([]string{"3.4.5.6"}),
 			},
 		},
 
@@ -301,8 +306,8 @@ func TestControllerMutation(t *testing.T) {
 					},
 				},
 				Spec: v1.ServiceSpec{
-					ClusterIP: "1.2.3.4",
-					Type:      "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4"},
+					Type:       "LoadBalancer",
 				},
 			},
 			wantErr: true,
@@ -312,17 +317,17 @@ func TestControllerMutation(t *testing.T) {
 			desc: "invalid IP assigned",
 			in: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type:      "LoadBalancer",
-					ClusterIP: "1.2.3.4",
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4"},
 				},
-				Status: statusAssigned("2.3.4.5"),
+				Status: statusAssigned([]string{"2.3.4.5"}),
 			},
 			want: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type:      "LoadBalancer",
-					ClusterIP: "1.2.3.4",
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4"},
 				},
-				Status: statusAssigned("1.2.3.0"),
+				Status: statusAssigned([]string{"1.2.3.0"}),
 			},
 		},
 
@@ -330,8 +335,8 @@ func TestControllerMutation(t *testing.T) {
 			desc: "invalid ingress state",
 			in: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type:      "LoadBalancer",
-					ClusterIP: "1.2.3.4",
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4"},
 				},
 				Status: v1.ServiceStatus{
 					LoadBalancer: v1.LoadBalancerStatus{
@@ -348,10 +353,10 @@ func TestControllerMutation(t *testing.T) {
 			},
 			want: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type:      "LoadBalancer",
-					ClusterIP: "1.2.3.4",
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4"},
 				},
-				Status: statusAssigned("1.2.3.0"),
+				Status: statusAssigned([]string{"1.2.3.0"}),
 			},
 		},
 
@@ -359,15 +364,15 @@ func TestControllerMutation(t *testing.T) {
 			desc: "former LoadBalancer, now NodePort",
 			in: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type:      "NodePort",
-					ClusterIP: "1.2.3.4",
+					Type:       "NodePort",
+					ClusterIPs: []string{"1.2.3.4"},
 				},
-				Status: statusAssigned("1.2.3.0"),
+				Status: statusAssigned([]string{"1.2.3.0"}),
 			},
 			want: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type:      "NodePort",
-					ClusterIP: "1.2.3.4",
+					Type:       "NodePort",
+					ClusterIPs: []string{"1.2.3.4"},
 				},
 			},
 		},
@@ -378,16 +383,16 @@ func TestControllerMutation(t *testing.T) {
 				Spec: v1.ServiceSpec{
 					Type:           "LoadBalancer",
 					LoadBalancerIP: "3.4.5.6",
-					ClusterIP:      "1.2.3.4",
+					ClusterIPs:     []string{"1.2.3.4"},
 				},
 			},
 			want: &v1.Service{
 				Spec: v1.ServiceSpec{
 					Type:           "LoadBalancer",
 					LoadBalancerIP: "3.4.5.6",
-					ClusterIP:      "1.2.3.4",
+					ClusterIPs:     []string{"1.2.3.4"},
 				},
-				Status: statusAssigned("3.4.5.6"),
+				Status: statusAssigned([]string{"3.4.5.6"}),
 			},
 		},
 
@@ -398,7 +403,7 @@ func TestControllerMutation(t *testing.T) {
 					Type:                  "LoadBalancer",
 					LoadBalancerIP:        "3.4.5.6",
 					ExternalTrafficPolicy: "Local",
-					ClusterIP:             "1.2.3.4",
+					ClusterIPs:            []string{"1.2.3.4"},
 				},
 			},
 			want: &v1.Service{
@@ -406,9 +411,9 @@ func TestControllerMutation(t *testing.T) {
 					Type:                  "LoadBalancer",
 					LoadBalancerIP:        "3.4.5.6",
 					ExternalTrafficPolicy: "Local",
-					ClusterIP:             "1.2.3.4",
+					ClusterIPs:            []string{"1.2.3.4"},
 				},
-				Status: statusAssigned("3.4.5.6"),
+				Status: statusAssigned([]string{"3.4.5.6"}),
 			},
 		},
 
@@ -428,7 +433,7 @@ func TestControllerMutation(t *testing.T) {
 				Spec: v1.ServiceSpec{
 					Type:           "LoadBalancer",
 					LoadBalancerIP: "1.2.3.1",
-					ClusterIP:      "3000::1",
+					ClusterIPs:     []string{"3000::1"},
 				},
 			},
 			wantErr: true,
@@ -440,7 +445,7 @@ func TestControllerMutation(t *testing.T) {
 				Spec: v1.ServiceSpec{
 					Type:           "LoadBalancer",
 					LoadBalancerIP: "1000::",
-					ClusterIP:      "1.2.3.4",
+					ClusterIPs:     []string{"1.2.3.4"},
 				},
 			},
 			wantErr: true,
@@ -450,17 +455,17 @@ func TestControllerMutation(t *testing.T) {
 			desc: "IP from wrong ip-family (ipv6) assigned",
 			in: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type:      "LoadBalancer",
-					ClusterIP: "1.2.3.4",
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4"},
 				},
-				Status: statusAssigned("1000::"),
+				Status: statusAssigned([]string{"1000::"}),
 			},
 			want: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type:      "LoadBalancer",
-					ClusterIP: "1.2.3.4",
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4"},
 				},
-				Status: statusAssigned("1.2.3.0"),
+				Status: statusAssigned([]string{"1.2.3.0"}),
 			},
 		},
 
@@ -468,17 +473,102 @@ func TestControllerMutation(t *testing.T) {
 			desc: "IP from wrong ip-family (ipv4) assigned",
 			in: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type:      "LoadBalancer",
-					ClusterIP: "3000::1",
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"3000::1"},
 				},
-				Status: statusAssigned("1.2.3.0"),
+				Status: statusAssigned([]string{"1.2.3.0"}),
 			},
 			want: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type:      "LoadBalancer",
-					ClusterIP: "3000::1",
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"3000::1"},
 				},
-				Status: statusAssigned("1000::"),
+				Status: statusAssigned([]string{"1000::"}),
+			},
+		},
+		// dual-stack test cases
+		{
+			desc: "deleted balancer",
+		},
+		{
+			desc: "simple dual-stack LoadBalancer",
+			in: &v1.Service{
+				Spec: v1.ServiceSpec{
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4", "3000::1"},
+				},
+			},
+			want: &v1.Service{
+				Spec: v1.ServiceSpec{
+					ClusterIPs: []string{"1.2.3.4", "3000::1"},
+					Type:       "LoadBalancer",
+				},
+				Status: statusAssigned([]string{"1.2.3.0", "1000::"}),
+			},
+		},
+		{
+			desc: "request IPs from specific pool",
+			in: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						"metallb.universe.tf/address-pool": "pool5",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4", "3000::1"},
+				},
+			},
+			want: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						"metallb.universe.tf/address-pool": "pool5",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4", "3000::1"},
+				},
+				Status: statusAssigned([]string{"1.2.3.0", "1000::"}),
+			},
+		},
+		{
+			desc: "request specific loadbalancer IP with dual-stack",
+			in: &v1.Service{
+				Spec: v1.ServiceSpec{
+					ClusterIPs:     []string{"1.2.3.4", "3000::1"},
+					Type:           "LoadBalancer",
+					LoadBalancerIP: "1.2.3.1",
+				},
+			},
+			wantErr: true,
+		},
+		{
+			desc: "request dual-stack loadbalancer with invalid ingress",
+			in: &v1.Service{
+				Spec: v1.ServiceSpec{
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"3000::1", "5.6.7.8"},
+				},
+				Status: v1.ServiceStatus{
+					LoadBalancer: v1.LoadBalancerStatus{
+						Ingress: []v1.LoadBalancerIngress{
+							{
+								Hostname: "foo.bar.local",
+							},
+							{
+								IP: "1000::",
+							},
+						},
+					},
+				},
+			},
+			want: &v1.Service{
+				Spec: v1.ServiceSpec{
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"3000::1", "5.6.7.8"},
+				},
+				Status: statusAssigned([]string{"1.2.3.0", "1000::"}),
 			},
 		},
 	}
@@ -509,13 +599,10 @@ func TestControllerMutation(t *testing.T) {
 				}
 			}
 
-			if test.want != nil && len(test.want.Status.LoadBalancer.Ingress) > 0 && test.want.Status.LoadBalancer.Ingress[0].IP != "" {
-				ip := net.ParseIP(test.want.Status.LoadBalancer.Ingress[0].IP)
-				if ip == nil {
+			if test.want != nil && len(test.want.Status.LoadBalancer.Ingress) > 0 {
+				ips := test.want.Status.LoadBalancer.Ingress
+				if len(ips) == 0 {
 					panic("bad wanted IP in loadbalancer status")
-				}
-				if !ip.Equal(c.ips.IP("test")) {
-					t.Errorf("%q: controller internal state does not match IP that controller claimed to allocate: want %q, got %q", test.desc, ip, c.ips.IP("test"))
 				}
 			}
 		}
@@ -547,8 +634,8 @@ func TestControllerConfig(t *testing.T) {
 	l := log.NewNopLogger()
 	svc := &v1.Service{
 		Spec: v1.ServiceSpec{
-			Type:      "LoadBalancer",
-			ClusterIP: "1.2.3.4",
+			Type:       "LoadBalancer",
+			ClusterIPs: []string{"1.2.3.4"},
 		},
 	}
 	if c.SetBalancer(l, "test", svc, k8s.EpsOrSlices{}) == k8s.SyncStateError {
@@ -615,7 +702,7 @@ func TestControllerConfig(t *testing.T) {
 	gotSvc = k.gotService(svc)
 	wantSvc := new(v1.Service)
 	*wantSvc = *svc
-	wantSvc.Status = statusAssigned("1.2.3.0")
+	wantSvc.Status = statusAssigned([]string{"1.2.3.0"})
 	if diff := diffService(wantSvc, gotSvc); diff != "" {
 		t.Errorf("SetBalancer produced unexpected mutation (-want +got)\n%s", diff)
 	}
@@ -654,8 +741,8 @@ func TestDeleteRecyclesIP(t *testing.T) {
 
 	svc1 := &v1.Service{
 		Spec: v1.ServiceSpec{
-			Type:      "LoadBalancer",
-			ClusterIP: "1.2.3.4",
+			Type:       "LoadBalancer",
+			ClusterIPs: []string{"1.2.3.4"},
 		},
 	}
 	if c.SetBalancer(l, "test", svc1, k8s.EpsOrSlices{}) == k8s.SyncStateError {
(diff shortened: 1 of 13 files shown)


e2etest/utils.go
```diff
@@ -50,11 +50,21 @@ func wgetRetry(address string, exc executor.Executor) error {
 	return err
 }
 
-func tweakServicePort(svc *v1.Service) {
+func tweakService(svc *v1.Service, ipFamily string) {
 	if servicePodPort != 80 {
 		// if servicePodPort is non default, then change service spec.
 		svc.Spec.Ports[0].TargetPort = intstr.FromInt(int(servicePodPort))
 	}
+	if ipFamily == "ipv6" {
```
fedepaol, line 58: I am not sure to always do this. I'd prefer having only specified in the dual stack case, and keeping the default in the v4 / v6 case.

e2etest/utils.go
```diff
@@ -50,11 +50,21 @@ func wgetRetry(address string, exc executor.Executor) error {
...
+		policy := v1.IPFamilyPolicyPreferDualStack
+		svc.Spec.IPFamilies = []v1.IPFamily{v1.IPv6Protocol, v1.IPv4Protocol}
```
fedepaol, line 60: Also, by doing this we'd have duplicate tests in the dual stack lane, because we are asking for both families in the v4 and in the v6 case.

internal/allocator/allocator.go
```diff
@@ -14,6 +14,16 @@ import (
 	"github.com/mikioh/ipaddr"
 )
 
+// IP family helps identifying single stack IPv4/IPv6 vs Dual-stack ["IPv4", "IPv6"] or ["IPv6", "Ipv4"].
+type IPFamily int
+
+const (
+	IPv4 IPFamily = iota
```
fedepaol, line 21: I'd use a string (still defined as a custom type) so it's easier to understand when we dump / log.

```
type Family string

const (
	IPV4      Family = "IPV4"
	IPV6      Family = "IPV6"
	DualStack Family = "DualStack"
)
```

internal/allocator/allocator.go
```diff
@@ -61,15 +71,39 @@ func New() *Allocator {
 	}
 }
 
+// GetSvcIPFamily returns the service address family based on spec ClusterIPs or ClusterIP.
+func (a *Allocator) GetSvcIPFamily(clusterIPs []string) (IPFamily, error) {
```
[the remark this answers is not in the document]
fedepaol: I think it makes sense to move this to its own package under internal.
Something like `ipfamily` as package, and `ForAddresses` for this function.

internal/allocator/allocator.go
```diff
@@ -61,15 +71,39 @@ func New() *Allocator {
...
+	case 1:
+		if ip.To4() != nil {
+			return IPv4, nil
+		} else {
+			return IPv6, nil
+		}
+	case 2:
+		ip2 := net.ParseIP(clusterIPs[1])
+		if ip2 == nil {
+			return Invalid, fmt.Errorf("Invalid address %q", clusterIPs[1])
```
fedepaol, line 87: Mention the name of the function in the error, it will make the debugging easier.

internal/allocator/allocator.go
```diff
@@ -61,15 +71,39 @@ func New() *Allocator {
...
+		}
+		if (ip.To4() == nil) == (ip2.To4() == nil) {
+			return Invalid, fmt.Errorf("Same address family")
```
fedepaol, line 90: Same, plus dump the slice.

internal/allocator/allocator.go
```diff
@@ -61,15 +71,39 @@ func New() *Allocator {
...
+		}
+		return DualStack, nil
+	default:
+		return Invalid, fmt.Errorf("Invaild clusterips string length %d", len(clusterIPs))
```
fedepaol, line 94: Same, name + clusterIPs

internal/allocator/allocator.go
```diff
@@ -109,61 +143,56 @@ func (a *Allocator) SetPools(pools map[string]*config.Pool) error {
...
-					otherSvcs = append(otherSvcs, otherSvc)
-				}
-			}
-			if len(otherSvcs) > 0 {
-				return fmt.Errorf("can't change sharing key for %q, address also in use by %s: %w", svc, strings.Join(otherSvcs, ","), err)
-			}
+	// Check the dual-stack constraints:
+	// - Two addresses
+	// - Different families, ipv4 and ipv6
+	if len(ips) > 1 {
```
fedepaol, line 181: Unroll the nested if.
Do len(ips) > 2 first and return.

internal/allocator/allocator.go
```diff
@@ -109,61 +143,56 @@ func (a *Allocator) SetPools(pools map[string]*config.Pool) error {
...
+		if len(ips) > 2 {
+			return fmt.Errorf("More than two addesses %q", ips)
 		}
+		if getIPFamily(ips[0]) == getIPFamily(ips[1]) {
```
fedepaol, line 185: This can then be if len == 2 && getIPFamily(ips[0]) == getIPFamily(ips[1])

internal/allocator/allocator.go
```diff
@@ -194,113 +222,116 @@ func (a *Allocator) Unassign(svc string) bool {
...
+func (a *Allocator) AllocateFromPool(svc string, ipFamily IPFamily, poolName string, ports []Port, sharingKey, backendKey string) ([]net.IP, error) {
 	if alloc := a.allocated[svc]; alloc != nil {
-		// Handle the case where the svc has already been assigned an IP but from the wrong family.
-		// This "should-not-happen" since the "ipFamily" is an immutable field in services.
-		if isIPv6 != ipIsIPv6(alloc.ip) {
-			return nil, fmt.Errorf("IP for wrong family assigned %s", alloc.ip.String())
+		if ipFamily != DualStack {
+			// Handle the case where the svc has already been assigned an IP but from the wrong family.
+			// This "should-not-happen" since the "ipFamily" is an immutable field in services.
+			if ipFamily != getIPFamily(alloc.ips[0]) {
```
fedepaol, line 255: I think this must check ipFamily of `alloc.ips` compared to the ipFamily parameter.

internal/allocator/allocator.go
```diff
@@ -194,113 +222,116 @@ func (a *Allocator) Unassign(svc string) bool {
...
-func cidrIsIPv6(cidr *net.IPNet) bool {
-	return cidr.IP.To4() == nil
-}
-func ipIsIPv6(ip net.IP) bool {
-	return ip.To4() == nil
-}
-
 // AllocateFromPool assigns an available IP from pool to service.
-func (a *Allocator) AllocateFromPool(svc string, isIPv6 bool, poolName string, ports []Port, sharingKey, backendKey string) (net.IP, error) {
+func (a *Allocator) AllocateFromPool(svc string, ipFamily IPFamily, poolName string, ports []Port, sharingKey, backendKey string) ([]net.IP, error) {
```
fedepaol, line 250: I'd refer ipFamily as serviceIPFamily or serviceFamily.

internal/allocator/allocator.go
```diff
@@ -194,113 +222,116 @@ func (a *Allocator) Unassign(svc string) bool {
...
 	for _, cidr := range pool.CIDR {
-		if cidrIsIPv6(cidr) != isIPv6 {
-			// Not the right ip-family
-			continue
-		}
-		c := ipaddr.NewCursor([]ipaddr.Prefix{*ipaddr.NewPrefix(cidr)})
-		for pos := c.First(); pos != nil; pos = c.Next() {
-			ip := pos.IP
-			if pool.AvoidBuggyIPs && ipConfusesBuggyFirmwares(ip) {
+		if len(ips) == 0 {
```
fedepaol, line 276: I think this is not working in the dual stack case.
What we need to do here is:
- find a v4 ip, a v6 ip (if single stack).
- find a v4 AND a v6 from the pool if dual stack.

Here we always check ips[0] and we never stop, so ips will grow more than two items if the pool has many cidrs.

You can do something like:

```
ips := []net.IP{}
	v4ToAdd := true
	v6ToAdd := true

	switch family {
	case ipfamily.IPV4:
		v6ToAdd = false
	case ipfamily.IPV6:
		v4ToAdd = false
	}

	for _, cidr := range pool.CIDR {
		if cidrIsIPv6(cidr) && !v6ToAdd {
			continue
		}
		if !cidrIsIPv6(cidr) && !v4ToAdd {
			continue
		}

		c := ipaddr.NewCursor([]ipaddr.Prefix{*ipaddr.NewPrefix(cidr)})
		for pos := c.First(); pos != nil; pos = c.Next() {
			ip := pos.IP
			if pool.AvoidBuggyIPs && ipConfusesBuggyFirmwares(ip) {
				continue
			}
			// Somewhat inefficiently brute-force by invoking the
			// IP-specific allocator.
			....
		}

	}
```

internal/allocator/allocator.go
```diff
@@ -194,113 +222,116 @@ func (a *Allocator) Unassign(svc string) bool {
...
+			ip := a.getIPFromCIDR(cidr, pool.AvoidBuggyIPs, svc, ports, sk)
+			if ip != nil {
+				ips = append(ips, ip)
 			}
 		}
 	}
-
-	// Woops, run out of IPs :( Fail.
-	return nil, fmt.Errorf("no available IPs in pool %q", poolName)
+	if len(ips) == 0 {
```
fedepaol, line 296: Here we must check also if we were asking for dual stack but we found only one good ip from the pool.

internal/allocator/allocator.go
```diff
@@ -383,3 +420,57 @@ func ipConfusesBuggyFirmwares(ip net.IP) bool {
 	}
 	return ip[3] == 0 || ip[3] == 255
 }
+
+func (a *Allocator) getIPFromCIDR(cidr *net.IPNet, avoidBuggyIPs bool, svc string, ports []Port, sk *key) net.IP {
```
fedepaol, line 424: I'd rename this to `ipForCIDR`

internal/allocator/allocator.go
```diff
@@ -383,3 +420,57 @@ func ipConfusesBuggyFirmwares(ip net.IP) bool {
...
+}
+
+func getcidrIPFamily(cidr *net.IPNet) IPFamily {
+	if cidr.IP.To4() == nil {
+		return IPv6
+	}
+	return IPv4
+}
+
+func getIPFamily(ip net.IP) IPFamily {
```
fedepaol, line 445: move this to ipfamily package

internal/allocator/allocator.go
```diff
@@ -383,3 +420,57 @@ func ipConfusesBuggyFirmwares(ip net.IP) bool {
...
+		}
+		if a.checkSharing(svc, pos.IP.String(), ports, sk) != nil {
+			continue
+		}
+		return pos.IP
+	}
+	return nil
+}
+
+func getcidrIPFamily(cidr *net.IPNet) IPFamily {
```
fedepaol, line 438: move this to ipfamily package

fedepaol (commented):
Did a quick round over the controller, haven't looked at the speaker or tests yet.

controller/service.go
```diff
@@ -47,76 +52,82 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
...
-	// User set or changed the desired LB IP, nuke the
-	// state. allocateIP will pay attention to LoadBalancerIP and try
-	// to meet the user's demands.
-	if svc.Spec.LoadBalancerIP != "" && svc.Spec.LoadBalancerIP != lbIP.String() {
-		level.Info(l).Log("event", "clearAssignment", "reason", "differentIPRequested", "msg", "user requested a different IP than the one currently assigned")
+		// User set or changed the desired LB IP, nuke the
+		// state. allocateIP will pay attention to LoadBalancerIP and try
+		// to meet the user's demands.
+		// For dual-stack it will match against the 1st allocated IP.
+		if svc.Spec.LoadBalancerIP != "" && svc.Spec.LoadBalancerIP != lbIPs[0].String() {
```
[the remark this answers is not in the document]
msherif1234: this check is already as part of IPFamilyForAddresses()

internal/allocator/allocator.go
msherif1234: I see that is much cleaner logic

controller/service.go
```diff
@@ -144,39 +166,46 @@ func (c *controller) clearServiceState(key string, svc *v1.Service) {
...
+	var err error
+	if svc.Spec.ClusterIPs != nil {
+		serviceIPFamily, err = ipfamily.IPFamilyForAddresses(svc.Spec.ClusterIPs)
+	} else {
 		// (we should never get here because the caller ensured that Spec.ClusterIP != nil)
 		return nil, fmt.Errorf("invalid ClusterIP [%s], can't determine family", svc.Spec.ClusterIP)
 	}
-	isIPv6 := clusterIP.To4() == nil
+
+	if err != nil {
```
fedepaol, line 179: We can avoid the extra nesting by checking for nill and returning (as the original code was doing).
if svc.Spec.ClusterIPs != nil {
	return nil, fmt.Errorf("invalid ClusterIP [%s], can't determine family", svc.Spec.ClusterIP)
}

internal/allocator/allocator.go
```diff
@@ -194,113 +187,118 @@ func (a *Allocator) Unassign(svc string) bool {
...
 // AllocateFromPool assigns an available IP from pool to service.
-func (a *Allocator) AllocateFromPool(svc string, isIPv6 bool, poolName string, ports []Port, sharingKey, backendKey string) (net.IP, error) {
+func (a *Allocator) AllocateFromPool(svc string, serviceIPFamily ipfamily.IPFamily, poolName string, ports []Port, sharingKey, backendKey string) ([]net.IP, error) {
 	if alloc := a.allocated[svc]; alloc != nil {
 		// Handle the case where the svc has already been assigned an IP but from the wrong family.
-		// This "should-not-happen" since the "ipFamily" is an immutable field in services.
-		if isIPv6 != ipIsIPv6(alloc.ip) {
-			return nil, fmt.Errorf("IP for wrong family assigned %s", alloc.ip.String())
+		// This "should-not-happen" since the "serviceIPFamily" is an immutable field in services.
+		if allocipsFamily, err := ipfamily.IPFamilyForAddressesIPs(alloc.ips); err != nil || allocipsFamily != serviceIPFamily {
```
fedepaol, line 219: nit: allocIPsFamily
I'd also call `IPFamilyForAddressesIPs` out of the if and check the error first, the family after in two separate ifs. Here we are swallowing the err and the if is a bit convoluted.

internal/allocator/allocator.go
```diff
@@ -194,113 +187,118 @@ func (a *Allocator) Unassign(svc string) bool {
...
 			continue
 		}
-		c := ipaddr.NewCursor([]ipaddr.Prefix{*ipaddr.NewPrefix(cidr)})
-		for pos := c.First(); pos != nil; pos = c.Next() {
-			ip := pos.IP
-			if pool.AvoidBuggyIPs && ipConfusesBuggyFirmwares(ip) {
-				continue
+		ip := a.getIPFromCIDR(cidr, pool.AvoidBuggyIPs, svc, ports, sk)
+		if ip != nil {
+			ips = append(ips, ip)
```
fedepaol, line 254: by doing this, we'll always append ips regardless of the family.
You need to skip if the ip is already added (or not requested):

```
                        if ipfamily.IPFamilyForCIDR(cidr) == ipfamily.IPv4 && !v4ToAdd {
				continue
			}
                        if ipfamily.IPFamilyForCIDR(cidr) == ipfamily.IPv6 && !v6ToAdd {
				continue
			}
                        ips = append(ips, ip)
```

As a second thought, maybe using a map[ipFamily]bool instead of two flags will make things more clean?

internal/ipfamily/ipfamily.go
```diff
@@ -0,0 +1,84 @@
...
+
+package ipfamily // import "go.universe.tf/metallb/internal/ipfamily"
+
+import (
+	"fmt"
+	"net"
+)
+
+// IP family helps identifying single stack IPv4/IPv6 vs Dual-stack ["IPv4", "IPv6"] or ["IPv6", "Ipv4"].
+type IPFamily string
```
fedepaol, line 11: Nit: just `Family`. From outside we'll call this as ipfamily.IPFamily which is a bit of a repetition.

internal/ipfamily/ipfamily.go
```diff
@@ -0,0 +1,84 @@
...
+
+const (
+	IPv4      IPFamily = "ipv4"
+	IPv6      IPFamily = "ipv6"
+	DualStack IPFamily = "dual"
+	Invalid   IPFamily = "invalid"
+)
+
+// IPFamilyForAddresses returns the address family given list of addresses strings.
+func IPFamilyForAddresses(ips []string) (IPFamily, error) {
```
fedepaol, line 21: Same: just "ForAddresses" (also for the functions below)

internal/ipfamily/ipfamily.go
```diff
@@ -0,0 +1,84 @@
...
+			return Invalid, fmt.Errorf("IPFamilyForAddresses: same address family %q", ips)
+		}
+		return DualStack, nil
+	default:
+		return Invalid, fmt.Errorf("IPFamilyForAddresses: invalid ips length %d %q", len(ips), ips)
+	}
+}
+
+// IPFamilyForAddressesIPs returns the address family from a given list of addresses IPs.
+func IPFamilyForAddressesIPs(ips []net.IP) (IPFamily, error) {
```
fedepaol, line 46: I'd convert ips to []string and use the one above, so the logic is in one place only.

speaker/layer2_controller.go
```diff
@@ -102,17 +102,18 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce
...
-
-		return bytes.Compare(hi[:], hj[:]) < 0
-	})
-
+	for _, toAnnounce := range lbIPs {
+		ipString := toAnnounce.String()
+		// Sort the slice by the hash of node + load balancer ips. This
+		// produces an ordering of ready nodes that is unique to all the services
+		// with the same ip.
+		sort.Slice(nodes, func(i, j int) bool {
```
fedepaol, line 110: This will sort nodes twice which is really not necessary.
I'd always take the first ip and that's it.
What's important is to have a well defined rule to understand which speaker should announce the ip, so taking the first ip will work in both single & dual stack.

speaker/main.go
```diff
@@ -327,24 +334,32 @@ func (c *controller) deleteBalancer(l log.Logger, name, reason string) k8s.SyncS
...
 
-	level.Info(l).Log("event", "serviceWithdrawn", "ip", c.svcIP[name], "reason", reason, "msg", "withdrawing service announcement")
+	level.Info(l).Log("event", "serviceWithdrawn", "ip", c.svcIPs[name], "reason", reason, "msg", "withdrawing service announcement")
 
 	return k8s.SyncStateSuccess
 }
 
-func poolFor(pools map[string]*config.Pool, ip net.IP) string {
+func poolFor(pools map[string]*config.Pool, ips []net.IP) string {
+	found := 0
```
fedepaol, line 354: You need to reset found in the beginning of the loop, otherwise you may risk to take two ips from two different pools.

internal/layer2/announcer.go
```diff
@@ -217,29 +217,31 @@ func (a *Announce) shouldAnnounce(ip net.IP) dropReason {
...
-	if _, ok := a.ips[name]; ok {
-		return
-	}
-	a.ips[name] = ip
+func (a *Announce) SetBalancer(name string, ips []net.IP) {
+	for _, ip := range ips {
+		// Call doSpam at the end of the function without holding the lock
+		defer a.doSpam(ip)
+		a.Lock()
+		defer a.Unlock()
```
fedepaol, line 225: Defer is executed when the function exits, so in case of two ips it will deadlock on the second a.Lock.
I think it will be simpler to keep this function as it is and call it twice from outside.

speaker/main.go
```diff
@@ -180,7 +181,7 @@ type controller struct {
 
 	protocols map[config.Proto]Protocol
 	announced map[string]config.Proto // service name -> protocol advertising it
-	svcIP     map[string]net.IP       // service name -> assigned IP
+	svcIPs    map[string][]net.IP     // service name -> assigned IP
```
fedepaol, line 184: nit: assigned IPs

fedepaol:
A general comment: we should also extend unit tests to cover dual stack.

controller/service.go
```diff
@@ -47,76 +53,88 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
...
+		lbIPs = append(lbIPs, net.ParseIP(ip))
 	}
-	if lbIP == nil {
+
+	if len(lbIPs) == 0 {
+		c.clearServiceState(key, svc)
+	}
+
+	lbIPsIPFamily, err := ipfamily.IPFamilyForAddresses(lbIPsString)
+	if err != nil {
```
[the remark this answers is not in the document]
msherif1234: in all error handling in this function we cleanup svc state I wil add return and log msg for boith

controller/service.go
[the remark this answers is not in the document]
msherif1234: sorry not following why is that ?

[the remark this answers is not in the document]
fedepaol: because here you only check ip lbIPs[0] matches, but that's not enough I think.
If you have a d/s service you still need to clear because we can't honor it.

controller/service.go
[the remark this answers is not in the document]
fedepaol: Not everywhere. In real error conditions we just return.
https://github.com/metallb/metallb/pull/1065/files#diff-e2b039cdee64b5555fd0499e1a75d9b5a42ac60c352a9680ccc9ebcac66cba42R127

Also, if we are not able to find the ipfamily of the service / lbips, we won't be able to apply any logic. For this reason, we must treat this error as non recoverable (and really unlikely to happen) and return.

controller/service.go
msherif1234: u meant == nil ok removed else

fedepaol: Yep sorry

internal/allocator/allocator.go
fedepaol: Sorry, did not notice https://github.com/metallb/metallb/pull/1065/files#diff-e11c83ce04539e8dc3dfc6cb1227f7057e4be177233b6d6765c4b3b1691e7411R247

The logic is good, I think using a map will make it slimmer.

controller/service.go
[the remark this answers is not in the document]
msherif1234: having an early return here fail the controller test I need to findout more

internal/layer2/announcer.go
msherif1234: ah sorry my bad I moved the loop to the caller that makes prefect sense

controller/service.go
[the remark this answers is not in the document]
msherif1234: we can't do early return because in case of no lb  been created at all and this is the 1st time

speaker/main.go
```diff
@@ -277,7 +282,7 @@ func (c *controller) SetBalancer(l log.Logger, name string, svc *v1.Service, eps
 		}
 	}
 
-	if svcIP, ok := c.svcIP[name]; ok && !lbIP.Equal(svcIP) {
+	if svcIPs, ok := c.svcIPs[name]; ok && !reflect.DeepEqual(lbIPs, svcIPs) {
```
fedepaol, line 285: `ip.Equal` does a bit more than bit comparison: https://cs.opensource.google/go/go/+/refs/tags/go1.17.3:src/net/ip.go;l=432
I don't have a strong opinino, but I think using it would put us on the safe side (considering the old implementation was doing the same). That would mean writing a custom comparison functoin for two []IP

speaker/main.go
```diff
@@ -327,24 +334,32 @@ func (c *controller) deleteBalancer(l log.Logger, name, reason string) k8s.SyncS
...
-func poolFor(pools map[string]*config.Pool, ip net.IP) string {
+func poolFor(pools map[string]*config.Pool, ips []net.IP) string {
 	for pname, p := range pools {
-		for _, cidr := range p.CIDR {
-			if cidr.Contains(ip) {
+		found := 0
+		for _, ip := range ips {
+			for _, cidr := range p.CIDR {
+				if cidr.Contains(ip) {
+					found++
```
fedepaol, line 359: same as above, break here.

internal/allocator/allocator.go
```diff
@@ -194,113 +187,120 @@ func (a *Allocator) Unassign(svc string) bool {
...
-				continue
-			}
-			// Somewhat inefficiently brute-force by invoking the
-			// IP-specific allocator.
-			if err := a.Assign(svc, ip, ports, sharingKey, backendKey); err == nil {
-				return ip, nil
+		ip := a.getIPFromCIDR(cidr, pool.AvoidBuggyIPs, svc, ports, sk)
+		if ip != nil {
+			ips = append(ips, ip)
+			if m[ipfamily.ForCIDR(cidr)] {
```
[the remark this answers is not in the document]
msherif1234: continue won't work if we have multiple cidr of the same family hence that is why I need to allocate it once for each address family and reset the flag

internal/allocator/allocator.go
```diff
@@ -358,13 +358,19 @@ func poolCount(p *config.Pool) int64 {
...
-		for _, cidr := range p.CIDR {
-			if cidr.Contains(ip) {
+		found := 0
+		for _, ip := range ips {
+			if p.AvoidBuggyIPs && ipConfusesBuggyFirmwares(ip) {
+				continue
+			}
+			for _, cidr := range p.CIDR {
+				if cidr.Contains(ip) {
+					found++
```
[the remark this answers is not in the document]
msherif1234: so keep it ?

speaker/layer2_controller.go
```diff
@@ -92,7 +92,7 @@ func usableNodes(eps k8s.EpsOrSlices, speakers map[string]bool) []string {
 	return ret
 }
 
-func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce net.IP, svc *v1.Service, eps k8s.EpsOrSlices) string {
+func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, lbIPs []net.IP, svc *v1.Service, eps k8s.EpsOrSlices) string {
```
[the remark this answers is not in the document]
msherif1234: this is multiple ips now not single one

internal/allocator/allocator.go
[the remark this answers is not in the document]
msherif1234: removed it

speaker/main.go
msherif1234: added customer compare

speaker/layer2_controller.go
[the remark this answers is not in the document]
fedepaol: Yep, it's multiple ips to announce

controller/service.go
```diff
@@ -47,76 +52,92 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
...
-	if (clusterIP.To4() == nil) != (lbIP.To4() == nil) {
+	lbIPsIPFamily := ipfamily.Unknown
+	if len(lbIPs) == 0 {
 		c.clearServiceState(key, svc)
-		lbIP = nil
+	} else {
+		lbIPsIPFamily, err = ipfamily.ForAddressesIPs(lbIPs)
+		if err != nil {
+			level.Info(l).Log("event", "clearAssignment", "reason", "nolbIPsIPFamily", "msg", "Failed to retrieve lbIPs family")
+			c.clearServiceState(key, svc)
```
fedepaol, line 72: no need to clear in case of error.

controller/service.go
```diff
@@ -47,76 +52,92 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
...
-	// (this should not happen since the "ipFamily" of a service is immutable)
-	if (clusterIP.To4() == nil) != (lbIP.To4() == nil) {
+	lbIPsIPFamily := ipfamily.Unknown
+	if len(lbIPs) == 0 {
 		c.clearServiceState(key, svc)
-		lbIP = nil
+	} else {
+		lbIPsIPFamily, err = ipfamily.ForAddressesIPs(lbIPs)
+		if err != nil {
+			level.Info(l).Log("event", "clearAssignment", "reason", "nolbIPsIPFamily", "msg", "Failed to retrieve lbIPs family")
```
fedepaol, line 71: This should be of level error, also, dump the error.

controller/service.go
```diff
@@ -47,76 +52,92 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
...
+			c.clearServiceState(key, svc)
+			return true
+		}
+		clusterIPsIPFamily, err := ipfamily.ForAddresses(svc.Spec.ClusterIPs)
+		if err != nil {
+			level.Info(l).Log("event", "clearAssignment", "reason", "noclusterIPsIPFamily", "msg", "Failed to retrieve clusterIPs family")
```
fedepaol, line 77: same

controller/service.go
```diff
@@ -144,39 +169,45 @@ func (c *controller) clearServiceState(key string, svc *v1.Service) {
...
+	}
 
 	// If the user asked for a specific IP, try that.
 	if svc.Spec.LoadBalancerIP != "" {
 		ip := net.ParseIP(svc.Spec.LoadBalancerIP)
 		if ip == nil {
 			return nil, fmt.Errorf("invalid spec.loadBalancerIP %q", svc.Spec.LoadBalancerIP)
 		}
-		if (ip.To4() == nil) != isIPv6 {
+		lbipFamily := ipfamily.ForAddress(ip)
```
fedepaol, line 191: nit: lbIPFamily

controller/service.go
```diff
@@ -144,39 +169,45 @@ func (c *controller) clearServiceState(key string, svc *v1.Service) {
...
+		if serviceIPFamily != lbipFamily {
 			return nil, fmt.Errorf("requested spec.loadBalancerIP %q does not match the ipFamily of the service", svc.Spec.LoadBalancerIP)
 		}
-		if err := c.ips.Assign(key, ip, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc)); err != nil {
+		lbips := []net.IP{ip}
```
fedepaol, line 195: nit: lbIPs

internal/allocator/allocator.go
```diff
@@ -358,14 +356,16 @@ func poolCount(p *config.Pool) int64 {
...
-		for _, cidr := range p.CIDR {
-			if cidr.Contains(ip) {
-				return pname
+		for _, ip := range ips {
+			if p.AvoidBuggyIPs && ipConfusesBuggyFirmwares(ip) {
+				continue
+			}
+			for _, cidr := range p.CIDR {
+				if cidr.Contains(ip) {
+					return pname
```
fedepaol, line 367: Sorry I wasn't clear in my previous comment. Did not mean to remove the check on the number of ips which made a lot of sense since we are looking for a pool that have both ips.
What I meant was to exit from the inner loop (the `p.CIDR` one) if we had one match, but continue on the outer (the `ips` one) so we avoid the case of different CIDRs matching the same ip.
What could happen in that case is that we match twice a v4 address, we return (because found==len(ips) but the v6 address is not in the pool.

internal/allocator/allocator.go
```diff
@@ -358,14 +356,16 @@ func poolCount(p *config.Pool) int64 {
 }
 
 // poolFor returns the pool that owns the requested IP, or "" if none.
-func poolFor(pools map[string]*config.Pool, ip net.IP) string {
+func poolFor(pools map[string]*config.Pool, ips []net.IP) string {
```
fedepaol, line 359: nit on comment: owns the requested IPs

speaker/main.go
```diff
@@ -327,31 +333,50 @@ func (c *controller) deleteBalancer(l log.Logger, name, reason string) k8s.SyncS
...
+					return pname
+				}
 			}
 		}
 	}
 	return ""
 }
 
+func compareIPs(ips1 []net.IP, ips2 []net.IP) bool {
+	for _, ip1 := range ips1 {
```
fedepaol, line 366: You should check also that len ips1 == len ips2

speaker/main.go
```diff
@@ -327,31 +333,50 @@ func (c *controller) deleteBalancer(l log.Logger, name, reason string) k8s.SyncS
...
+				if cidr.Contains(ip) {
+					return pname
+				}
 			}
 		}
 	}
 	return ""
 }
 
+func compareIPs(ips1 []net.IP, ips2 []net.IP) bool {
```
fedepaol, line 365: nit: ips1, ips2 []net.IP. No need to declare the type twice

speaker/main.go
```diff
@@ -327,31 +333,50 @@ func (c *controller) deleteBalancer(l log.Logger, name, reason string) k8s.SyncS
...
+	for _, ip1 := range ips1 {
+		found := false
+		for _, ip2 := range ips2 {
+			if ip1.Equal(ip2) {
+				found = true
```
fedepaol, line 370: you can break here

controller/service.go
msherif1234: hmm code above when it return it does also clear so I followed the same ?

fedepaol: oh, I looked at the first error occurrence here https://github.com/metallb/metallb/pull/1065/files#diff-e2b039cdee64b5555fd0499e1a75d9b5a42ac60c352a9680ccc9ebcac66cba42R134 and it wasn't, but there's another one later that clears.
It's fine this way I guess, no need to change.

controller/controller_test.go
```diff
@@ -450,35 +475,39 @@ func TestControllerMutation(t *testing.T) {
...
 			want: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type:      "LoadBalancer",
-					ClusterIP: "3000::1",
+					Type:       "LoadBalancer",
+					ClusterIP:  "3000::1",
+					ClusterIPs: []string{"3000::1"},
 				},
-				Status: statusAssigned("1000::"),
+				Status: statusAssigned([]string{"1000::"}),
```
fedepaol, line 510: can you extend these table tests to cover dual stack too?

internal/allocator/allocator_test.go
```diff
@@ -316,36 +318,50 @@ func TestAssignment(t *testing.T) {
...
+		// IP dual-stack test
+		{
+			desc: "s2 frees its IP",
+			svc:  "s2",
+			ips:  []string{},
+		},
+		{
+			desc: "ip dual-stack assign s1",
+			svc:  "s1",
+			ips:  []string{"1.2.3.4", "1000::4"},
```
fedepaol, line 332: can we add a couple of tests where we try to assign to ips, one matching one pool and the other matching the other to see if it fails?

internal/allocator/allocator_test.go
```diff
@@ -316,36 +318,50 @@ func TestAssignment(t *testing.T) {
...
 			continue
 		}
 
 		if err != nil {
-			t.Errorf("%q: Assign(%q, %q): %s", test.desc, test.svc, test.ip, err)
+			t.Errorf("%q: Assign(%q, %q): %s", test.desc, test.svc, test.ips, err)
 		}
-		if a := assigned(alloc, test.svc); a != test.ip {
-			t.Errorf("%q: ran Assign(%q, %q), but allocator has recorded allocation of %q", test.desc, test.svc, test.ip, a)
+		if a := assigned(alloc, test.svc); !reflect.DeepEqual(a, test.ips) {
```
fedepaol, line 363: I'd use the ips compare function you added

controller/controller_test.go
```diff
@@ -173,16 +174,18 @@ func TestControllerMutation(t *testing.T) {
 			desc: "simple LoadBalancer",
 			in: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type:      "LoadBalancer",
-					ClusterIP: "1.2.3.4",
+					Type:       "LoadBalancer",
+					ClusterIP:  "1.2.3.4",
```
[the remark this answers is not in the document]
msherif1234: we still using this early check with clusterIP https://github.com/metallb/metallb/blob/main/controller/service.go#L43 as safety net before going too deep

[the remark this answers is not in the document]
fedepaol: I'd say remove that reference to clusterIP (it was not shown in the diff). 
We are not using it anywhere apart from checking if it's empty.

controller/service.go
```diff
@@ -47,76 +52,92 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
...
 			level.Info(l).Log("event", "clearAssignment", "reason", "differentPoolRequested", "msg", "user requested a different pool than the one currently assigned")
 			c.clearServiceState(key, svc)
-			lbIP = nil
+			lbIPs = []net.IP{}
+		}
+		// User set or changed the desired LB IP, nuke the
+		// state. allocateIP will pay attention to LoadBalancerIP and try
+		// to meet the user's demands.
+		// For dual-stack it will match against the 1st allocated IP.
+		if svc.Spec.LoadBalancerIP != "" && (lbIPsIPFamily == ipfamily.DualStack || svc.Spec.LoadBalancerIP != lbIPs[0].String()) {
```
fedepaol, line 114: I'd split this in two:

svc.Spec.LoadBalancerIP != "" && lbIPsIPFamily == ipfamily.DualStack we fail.

if the ips are different we clear.

controller/service.go
```diff
@@ -144,39 +169,45 @@ func (c *controller) clearServiceState(key string, svc *v1.Service) {
...
+func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error) {
+	var serviceIPFamily ipfamily.Family
+	var err error
+	if svc.Spec.ClusterIPs == nil {
 		// (we should never get here because the caller ensured that Spec.ClusterIP != nil)
 		return nil, fmt.Errorf("invalid ClusterIP [%s], can't determine family", svc.Spec.ClusterIP)
 	}
-	isIPv6 := clusterIP.To4() == nil
+
+	serviceIPFamily, err = ipfamily.ForAddresses(svc.Spec.ClusterIPs)
```
fedepaol, line 180: nit: you can use := here and get rid of the variables above.

internal/allocator/allocator.go
```diff
@@ -194,113 +187,118 @@ func (a *Allocator) Unassign(svc string) bool {
...
-			ip := pos.IP
-			if pool.AvoidBuggyIPs && ipConfusesBuggyFirmwares(ip) {
-				continue
-			}
-			// Somewhat inefficiently brute-force by invoking the
-			// IP-specific allocator.
-			if err := a.Assign(svc, ip, ports, sharingKey, backendKey); err == nil {
-				return ip, nil
-			}
+		ip := a.getIPFromCIDR(cidr, pool.AvoidBuggyIPs, svc, ports, sk)
```
fedepaol, line 256: nit: I'd have `getIPFromCIDR` accept sharingkey and backendKey and construct the sk inside (or use sk also in allocate, but that's a bit more complex)

speaker/bgp_controller.go
```diff
@@ -237,33 +237,34 @@ func (c *bgpController) syncBFDProfiles(profiles map[string]*config.BFDProfile)
...
+			for comm := range adCfg.Communities {
+				ad.Communities = append(ad.Communities, comm)
+			}
+			sort.Slice(ad.Communities, func(i, j int) bool { return ad.Communities[i] < ad.Communities[j] })
+			c.svcAds[name] = append(c.svcAds[name], ad)
 		}
-		for comm := range adCfg.Communities {
-			ad.Communities = append(ad.Communities, comm)
+
+		if err := c.updateAds(); err != nil {
```
fedepaol, line 262: This should go outside of the loop I think.

speaker/layer2_controller.go
```diff
@@ -122,8 +123,10 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce
 	return "notOwner"
 }
 
-func (c *layer2Controller) SetBalancer(l log.Logger, name string, lbIP net.IP, pool *config.Pool) error {
-	c.announcer.SetBalancer(name, lbIP)
+func (c *layer2Controller) SetBalancer(l log.Logger, name string, lbIPs []net.IP, pool *config.Pool) error {
+	for _, lbIP := range lbIPs {
+		c.announcer.SetBalancer(name, lbIP)
```
fedepaol, line 128: I missed this before, but this won't work. The L2 controller assumes that a given service holds only one IP.
See when it does
```
	a.ips[name] = ip
```
I think it's fine deferring dual stack support for L2 outside of this PR, it's already hard to review.

fedepaol:
LGTM, let's rebase to import the tests and :crossed_fingers:

speaker/layer2_controller.go
msherif1234: i fixed this and make it a slice and added unit test case with svc and multiple ips

kub3let:
There is no documentation, testsvc_dualstack.yaml is not in the project.

What does the service deploy need to look like to set dualstack static ip's ? 

Like if I want to set the ip's 192.168.1.151 and 68d9:f187:b07b:15c2::151

msherif1234:
> There is no documentation, testsvc_dualstack.yaml is not in the project.
> 
> What does the service deploy need to look like to set dualstack static ip's ?
> 
> Like if I want to set the ip's 192.168.1.151 and 68d9:f187:b07b:15c2::151

- I was asked during review to not commit the dual stack test yaml file since we have dual-stack CI, I added the yaml file in this PR description for referrence.
- regarding svc static loadbalancer IP assignment, k8s loadbalancer svc API doesn't allow passing multiple loadbalancer IPs, so in dual stack case we will emit and error log when this happens

kub3let:
ok then I misunderstood, I will continue to use 2 services.

PR #1066, opened 2021-11-19
Update the security reporting recipients.

fedepaol:
The old section still contained Dave Anderson's email as the primary destination, it's sane to correct it.


russellb:
lgtm ... @johananl @rata your email addresses are included based on what you've used for recent commits.  I'm going to merge, so feel free to push a follow-up to change it if you want.

PR #1070, opened 2021-11-19
 E2E: Add an empty bgpd.conf file with zebra password

fedepaol:
Aims to fix a flake where BGP is not able to start, showing messages as
```
BGP: [EC 33554499] sendmsg_nexthop: zclient_send_message() failed
```

the current-config is as well empty, so it could be an issue related to the order in which bgp starts / the tests place a config under its hood. 
Here we start with a valid config that gets eventually overridden by the tests.

### DIFF
--- a/e2etest/config/frr/bgpd.conf
+++ b/e2etest/config/frr/bgpd.conf
@@ -1,0 +1,4 @@
+hostname bgpd
+password zebra
+log stdout debugging
+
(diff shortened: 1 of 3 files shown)


russellb:
lgtm once ci passes

fedepaol:
@russellb tried to run this 3 times, always passed.
I think it's good.

PR #1072, opened 2021-11-21
E2e tests code improvements

sabinaaledort:
1. Fix the error message produced when no controller pod found.
2. Add a validation to check at least one speaker pod found.
3. Use getSpeakerPods function to avoid code duplication.


fedepaol:
LGTM

Issue #881, opened 2021-06-10
Add helm based e2e CI job


russellb:
We have a couple of helm based CI jobs: `helm-lint` and `helm-conftest`.  Now that we have a couple of end-to-end CI jobs (`e2etest-layer2` and `e2etest-bgp`), we should add an e2etest job that uses the helm chart to deploy MetalLB.  We could just change one of the existing e2etest jobs to use helm to get additional test coverage without a new CI job required.

- [ ] Update the dev-env code in `tasks.py` to optionally use helm to deploy MetalLB instead of applying the manifest directly
- [ ] Update one of the e2etest CI jobs to use helm in the dev-env

fedepaol:
@sabinaaledort let's keep this for the missing CI to github actions

tylerauerbeck:
@fedepaol Is there anything specific that is still missing here? Would be happy to take a look.

fedepaol:
Nope, I think this is good to be close now. My comment in https://github.com/metallb/metallb/issues/881#issuecomment-993701573 is because in https://github.com/metallb/metallb/pull/1073 we added it to circle while we were moving towards gh actions.

PR #1073, opened 2021-11-21
Install frr through helm chart and add helm chart CI

sabinaaledort:
Update MetalLB helm chart to install FRR container, adding FRR and reloader containers to the speaker deployment. They also include the ConfigMap for the initial FRR files and an initContainer to set everything up.

Also adding a CI lane to deploy MetalLB with helm chart so we will be able to make sure the helm chart is update as well with every change done in manifests.


russellb:
This will close issue #881 

also cc @gclawes, since I think you'd be interested

charts/metallb/templates/controller.yaml
```diff
@@ -5,12 +5,12 @@ metadata:
   name: {{ template "metallb.fullname" . }}-controller
   labels:
     {{- include "metallb.labels" . | nindent 4 }}
-    app.kubernetes.io/component: controller
+    component: controller
```
gclawes, line 8: What's the reason for changing this?  `app.kubernetes.io/component` comes from the recommended labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

charts/metallb/templates/controller.yaml
```diff
@@ -57,13 +57,13 @@ spec:
           value: {{ template "metallb.fullname" . }}-controller
         {{- end }}
         ports:
-        - name: metrics
```
gclawes, line 60: Why change `metrics` -> `monitoring` for all the port names?

charts/metallb/values.yaml
```diff
@@ -225,3 +225,15 @@ speaker:
     periodSeconds: 10
     successThreshold: 1
     timeoutSeconds: 1
+
+# frr contains configuration specific to the MetalLB FRR container.
+frr:
```
gclawes, line 230: This should probably be moved from `.Values.frr` to `.Values.speaker.frr`, as it is a configuration specifically for the speaker.

charts/metallb/values.yaml
```diff
@@ -225,3 +225,15 @@ speaker:
     periodSeconds: 10
     successThreshold: 1
     timeoutSeconds: 1
+
```
gclawes, line 228: Any updates to `values.yaml` should have corresponding validations added in `values.schema.json` 

https://helm.sh/docs/topics/charts/#schema-files
https://www.arthurkoziel.com/validate-helm-chart-values-with-json-schemas/

charts/metallb/templates/speaker.yaml
```diff
@@ -118,6 +270,71 @@ spec:
...
+        {{- if .Values.frr.enabled }}
+        volumeMounts:
+          - name: reloader
+            mountPath: /etc/frr_reloader
+         {{- end }}
+      {{- if .Values.frr.enabled }}
+      - name: frr
+        securityContext:
+          capabilities:
+            add: ["NET_ADMIN", "NET_RAW", "SYS_ADMIN", "NET_BIND_SERVICE"]
```
gclawes, line 282: Samll formatting nitpick, this should probably be a mutli-line list to match the other `capabilities.add` fields elsewhere in the chart:
```
capabilities:
  add:
  - NET_ADMIN
  - NET_RAW
  - SYS_ADMIN
  - NET_BIND_SERVICE
```

Also, is `SYS_ADMIN` truly required for FRR to function properly?

charts/metallb/templates/speaker.yaml
```diff
@@ -1,31 +1,131 @@
+{{- if .Values.frr.enabled }}
+# FRR expects to have these files owned by frr:frr on startup.
+# Having them in a ConfigMap allows us to modify behaviors: for example enabling more daemons on startup.
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: frr-startup
+data:
```
gclawes, line 8: Do we want to allow the users to inject any custom configs in here?  Or should everything be hard-coded?

charts/metallb/templates/speaker.yaml
```diff
@@ -1,31 +1,131 @@
+{{- if .Values.frr.enabled }}
+# FRR expects to have these files owned by frr:frr on startup.
+# Having them in a ConfigMap allows us to modify behaviors: for example enabling more daemons on startup.
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: frr-startup
```
gclawes, line 7: ConfigMap `.metadata.name` should incorporate the release name using `{{ template "metallb.fullname" . }}-` like other resources do.  It should also include the standard set of labels, including `app.kubernetes.io/component`:

Example from speaker Daemonset:
```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: {{ template "metallb.fullname" . }}-speaker
  labels:
    {{- include "metallb.labels" . | nindent 4 }}
    app.kubernetes.io/component: speaker
...
```

gclawes (changes requested):

gclawes:
Added a review/change request, a few notes:
1. `.Values.frr.*` should probably be moved into `.Values.speaker.frr.*`, as it is only speaker-related
2. Any updates to `values.yaml` should be reflected in `values.schema.json`
3. We should retain the `app.kubernetes.io/component` label, as it is from the [Recommended Labels](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/)
4. I'd like to avoid changing port names from `metrics` -> `monitoring` if possible, it seems like an unnecessary non-functional change.  If there's a good reason to change it, though, I'm open to hearing it.
5. Any new resources introduces should not use hard-coded `.metadata` fields, they should be templated to include the release name and standard set of labels.

charts/metallb/templates/controller.yaml
gclawes: Ah, I see the reason now:
```
Helm chart: Update the metrics port name for E2E tests
Prometheus by default looks for a port named monitoring and so does
the E2E tests.
```

charts/metallb/templates/controller.yaml
gclawes: If the E2E tests are looking for `component`, they should probably be changed to reflect the standard expected by the Kubernetes documentation.

fedepaol: Agree. I _think_ e2e tests are crafted against the deployment via the manifest, which if I remember correctly use the `component: controller` .
It's probably worth fixing the manifests.

gclawes: Does it make sense to transition to using the helm chart to generate the manifests, via `helm template`.  I believe projects like calico do this, as well.

fedepaol: Yep, let me file an issue for that. It would be nice to have CI that ensure they do not drift.

gclawes: We can provide different manifest flavors by templating with different `values.yaml` files.  The `charts/metallb/values.yaml` file is the one that gets officially "packaged" in the chart, but we can put others in the same directory as reference/manifest sources.

fedepaol: Neat, I hoped that could have been done as we are gonna need at least a couple of flavours (with / without FRR).
Thanks @gclawes !

charts/metallb/values.yaml
sabinaaledort: you are right. i will fix it. thanks

charts/metallb/templates/controller.yaml
sabinaaledort: ok i will update it the label in the manifests and E2E tests.

charts/metallb/templates/speaker.yaml
sabinaaledort: i will fix the list.
@fedepaol or @oribon do you know about `SYS_ADMIN` for FRR?

charts/metallb/values.yaml
sabinaaledort: thanks

charts/metallb/templates/speaker.yaml
sabinaaledort: i think we don't want custom configs here.
@fedepaol or @oribon what do you think?

charts/metallb/templates/speaker.yaml
sabinaaledort: thanks

charts/metallb/templates/speaker.yaml
oribon: I got this error when I initially started working on adding the frr container:

```
privs_init: initial cap_set_proc failed: Operation not permitted
Wanted caps: cap_net_admin,cap_net_raw,cap_sys_admin=p
Have   caps: cap_chown,cap_dac_override,cap_fowner,cap_fsetid,cap_kill,cap_setgid,cap_setuid,cap_setpcap,cap_net_bind_service,cap_net_raw,cap_sys_chroot,cap_mknod,cap_audit_write,cap_setfcap=ip
starting staticd since zebra is running2021/08/04 09:51:56 STATIC: [EC 100663303] Could not lock pid_file /var/run/frr/staticd.pid (Resource temporarily unavailable), exiting.  Please ensure that the daemon is not already running
privs_init: initial cap_set_proc failed: Operation not permitted
Wanted caps: cap_net_bind_service,cap_net_admin,cap_net_raw,cap_sys_admin=p
Have   caps: cap_chown,cap_dac_override,cap_fowner,cap_fsetid,cap_kill,cap_setgid,cap_setuid,cap_setpcap,cap_net_bind_service,cap_net_raw,cap_sys_chroot,cap_mknod,cap_audit_write,cap_setfcap=ip
```

charts/metallb/templates/speaker.yaml
oribon: I think everything under this ConfigMap should be hard-coded, don't think we'd want to support changing stuff here (yet?)

sabinaaledort:
CI is failing due to:
https://github.com/metallb/metallb/pull/1111

charts/metallb/templates/speaker.yaml
gclawes: @oribon  @sabinaaledort ok, we can keep this static for now.

gclawes:
@sabinaaledort everything looks good to me!

fedepaol:
@sabinaaledort mind fixing the conflict here?

fedepaol:
@sabinaaledort we may need to duplicate the new helm lane also to github actions, but I don't want it to hold this PR.
I'll file an issue and assign it to you.

fedepaol:
LGTM

PR #1074, opened 2021-11-22
CI: change the permissions on kind logs when tests fails

fedepaol:
The change permission command is skipped if the tests fail, which makes the export of some logs fail. We want the command to always be executed, in particular when tests fail.


oribon:
LGTM, did you validate that it works when a test fails?

PR #1075, opened 2021-11-22
website: Update installation on Openshift instructions

liornoy:
Add another requirement for a change needed to be done to run
MetalLB on OpenShift. 
By also deleting the `spec.template.spec.securityContext.fsGroup` in the metallb manifest 
It's because currently, in the manifests, the fsGroup value is 65534 which is out of range 
for Openshift (5000 - 5999).


fedepaol:
LGTM, thanks!

PR #1077, opened 2021-11-23
E2E Test: provide support for dual stack

fedepaol:
Instead of driving the ipfamily from outside, we use the family of the loadbalancer ip to derive what family it is, and we use both in case of dual stack services.
We add more entries to the test tables forcing the service to be dual stack in case of dual stack entries, and adding extra entries to check single v6 in case of dual stack (for the v4, we defer to the fact that a service gets created as single stack v4 in case of dual stack).

Additionally, the old ipFamily parameter is now used to choose the ipfamily of the pairing between the external FRRs and the speakers, making it possible to test scenarios where we advertise v6 addresses on a v4 connection and viceversa.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,10 +38,10 @@ jobs:
       - run: inv dev-env -i  << parameters.ip-family >> -b  << parameters.bgp-type >> -l all
       - run: |
           SKIP="none"
-          if [ "<< parameters.bgp-type >>" = "native" ]; then SKIP="$SKIP\|FRR\|BFD"; fi
-          if [ "<< parameters.ip-family >>" = "ipv4" ]; then SKIP="$SKIP\|IPV6"; fi
-          if [ "<< parameters.ip-family >>" = "dual" ]; then SKIP="$SKIP\|IPV6"; fi # TODO: until dual stack is supported, we can't run the ipv6 tests against the cluster
-          if [ "<< parameters.ip-family >>" = "ipv6" ]; then SKIP="$SKIP\|IPV4"; fi
+          if [ "<< parameters.bgp-type >>" = "native" ]; then SKIP="$SKIP\|FRR\|BFD\|DUALSTACK"; fi
+          if [ "<< parameters.ip-family >>" = "ipv4" ]; then SKIP="$SKIP\|IPV6\|DUALSTACK"; fi
+          if [ "<< parameters.ip-family >>" = "dual" ]; then SKIP="$SKIP\|IPV6\|DUALSTACK"; fi # TODO: until DUALSTACK stack is supported, we can't run the ipv6 / dual stack tests against the cluster
+          if [ "<< parameters.ip-family >>" = "ipv6" ]; then SKIP="$SKIP\|IPV4\|DUALSTACK"; fi
           if [ "<< parameters.ip-family >>" = "ipv6" ] && [ "<< parameters.bgp-type >>" = "native" ]; then SKIP="$SKIP\|BGP"; fi # we are skipping BGP for ipv6 for native bgp stack
           echo "Skipping $SKIP"
           # `-E env "PATH=$PATH"` bypasses commands not found under sudo, `sudo` needed because arping requires CAP_NET_RAW
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -39,6 +39,7 @@ import (
 	"go.universe.tf/metallb/e2etest/pkg/frr"
 	frrconfig "go.universe.tf/metallb/e2etest/pkg/frr/config"
 	frrcontainer "go.universe.tf/metallb/e2etest/pkg/frr/container"
+	testservice "go.universe.tf/metallb/e2etest/pkg/service"
 	bgpfrr "go.universe.tf/metallb/internal/bgp/frr"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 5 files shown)


.circleci/config.yml
```diff
@@ -38,10 +38,10 @@ jobs:
       - run: inv dev-env -i  << parameters.ip-family >> -b  << parameters.bgp-type >> -l all
       - run: |
           SKIP="none"
-          if [ "<< parameters.bgp-type >>" = "native" ]; then SKIP="$SKIP\|FRR\|BFD"; fi
-          if [ "<< parameters.ip-family >>" = "ipv4" ]; then SKIP="$SKIP\|IPV6"; fi
-          if [ "<< parameters.ip-family >>" = "dual" ]; then SKIP="$SKIP\|IPV6"; fi # TODO: until dual stack is supported, we can't run the ipv6 tests against the cluster
-          if [ "<< parameters.ip-family >>" = "ipv6" ]; then SKIP="$SKIP\|IPV4"; fi
+          if [ "<< parameters.bgp-type >>" = "native" ]; then SKIP="$SKIP\|FRR\|BFD\DUALSTACK"; fi
+          if [ "<< parameters.ip-family >>" = "ipv4" ]; then SKIP="$SKIP\|IPV6\|STACK"; fi
```
msherif1234, line 42: typo here and below STACK==DUALSTACK ?

e2etest/bgp_test.go
```diff
@@ -442,38 +462,56 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			),
+			table.Entry("DUALSTACK - test AddressPool defined by address range", []addressPool{
+				{
+					Name:     "bgp-test",
+					Protocol: BGP,
+					Addresses: []string{
+						"192.168.10.0-192.168.10.18",
+						"fc00:f853:0ccd:e799::0-fc00:f853:0ccd:e799::18",
+					},
+				}}, "ipv4", testservice.TrafficPolicyCluster,
```
msherif1234, line 475: why we passing ipv4 here and below for dual-stack

e2etest/pkg/routes/routes.go
```diff
@@ -25,23 +25,22 @@ func init() {
...
 
 	routes := make([]net.IP, 0)
 
-	dst := net.ParseIP(target)
-	framework.ExpectNotEqual(dst, nil, "Failed to convert", target, "to ip")
-
-	re := Ipv4Re
-	if dst.To4() == nil { // assuming it's an ipv6 address
-		re = Ipv6Re
-	}
```
msherif1234, line 44: why we dropped v6 support here

e2etest/pkg/service/tweak.go
```diff
@@ -0,0 +1,26 @@
...
+
+func ForceV6(svc *corev1.Service) {
+	f := corev1.IPFamilyPolicySingleStack
+	svc.Spec.IPFamilyPolicy = &f
+	svc.Spec.IPFamilies = []corev1.IPFamily{corev1.IPv6Protocol}
+}
+
+func DualStack(svc *corev1.Service) {
+	f := corev1.IPFamilyPolicyPreferDualStack
+	svc.Spec.IPFamilyPolicy = &f
```
msherif1234, line 21: we need to set IPFamilies too ?

e2etest/bgp_test.go
fedepaol: not necessarily. The ipfamily now drives only how the bgp session is established, which is orthogonal to the family of the service.

.circleci/config.yml
fedepaol: yep, I changed the name on the way and this is a leftover, will fix.

e2etest/pkg/routes/routes.go
fedepaol: We did not. I moved the logic above, based on dst being v4 or v6.

fedepaol: See https://github.com/metallb/metallb/pull/1077/files/4d8c2b0f70c844d2836307686a34e9c7cbd87f07#diff-463ff4e14d1a72292475b266c5584dda7ed79f381f0759e8b3d14932dfd617e9R29

e2etest/pkg/service/tweak.go
fedepaol: IPFamilies, wasn't sure. PreferDualStack should work but we won't know if it silently fails, so I think require is a better choice. Will change.

.circleci/config.yml
```diff
@@ -38,10 +38,10 @@ jobs:
       - run: inv dev-env -i  << parameters.ip-family >> -b  << parameters.bgp-type >> -l all
       - run: |
           SKIP="none"
-          if [ "<< parameters.bgp-type >>" = "native" ]; then SKIP="$SKIP\|FRR\|BFD"; fi
-          if [ "<< parameters.ip-family >>" = "ipv4" ]; then SKIP="$SKIP\|IPV6"; fi
-          if [ "<< parameters.ip-family >>" = "dual" ]; then SKIP="$SKIP\|IPV6"; fi # TODO: until dual stack is supported, we can't run the ipv6 tests against the cluster
-          if [ "<< parameters.ip-family >>" = "ipv6" ]; then SKIP="$SKIP\|IPV4"; fi
+          if [ "<< parameters.bgp-type >>" = "native" ]; then SKIP="$SKIP\|FRR\|BFD\|DUALSTACK"; fi
+          if [ "<< parameters.ip-family >>" = "ipv4" ]; then SKIP="$SKIP\|IPV6\|DUALSTACK"; fi
+          if [ "<< parameters.ip-family >>" = "dual" ]; then SKIP="$SKIP\|IPV6\|DUALSTACK"; fi # TODO: until DUALSTACK stack is supported, we can't run the ipv6 / dual stack tests against the cluster
```
msherif1234, line 43: do we still need this TODO ? given i will follow with my PR once this merged ?

fedepaol: Yep, in your PR you should unblock DUALSTACK so CI will run dual stack tests.
If I don't do this, CI will fail until your pr gets in.

PR #1084, opened 2021-11-26
Bump kind to v0.11.1

cgoncalves:
MetalLB CI is currently pinned to install Kind v0.10.0, a release based on Kubernetes v1.20.2. Bumping to v0.11.1 (latest released as of today) based on Kubernetes v1.21.1.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ commands:
   setup_kind:
     description: "Install kind"
     steps:
-      - run: GO111MODULE="on" go get sigs.k8s.io/kind@v0.10.0
+      - run: GO111MODULE="on" go get sigs.k8s.io/kind@v0.11.1
       - run: curl -LO "https://dl.k8s.io/release/$(curl -L -s https://dl.k8s.io/release/stable.txt)/bin/linux/amd64/kubectl"
       - run: sudo install -o root -g root -m 0755 kubectl /usr/local/bin/kubectl
 jobs:


.circleci/config.yml
```diff
@@ -12,7 +12,7 @@ commands:
   setup_kind:
     description: "Install kind"
     steps:
-      - run: GO111MODULE="on" go get sigs.k8s.io/kind@v0.10.0
+      - run: GO111MODULE="on" go get sigs.k8s.io/kind@latest
```
fedepaol, line 15: It's ok to bump, but I'd rather pin to a specific kind version to have reliable CI and be deliberate on when we bump.

cgoncalves: Have there been occasions in the past where a released version negatively impacted MetalLB CI?

Following the latest kind version would allow us to more easily flag deprecated/removed Kubernetes APIs (among other possible breakages). We can still accomplish the same with pinned versions, but requires us to bump it every now and then. I'm fine with such a compromise.

fedepaol: There was on MetalLB CI in the past :P, so we can't tell. 
In my opinion the problem with latest is that (and with no pinning to specific versions in general) is that, in the unfortunate case something goes wrong with kind (and not necessarily with the compatibility between us and k8s), our CI will start breaking (or flaking) without us understanding why and needing to jump in firefighting mode.

fedepaol:
LGTM

PR #1086, opened 2021-11-29
Skip Nodes with LabelNodeExcludeBalancers label

zuzzas:
Have an option to skip peering with Nodes that are labeled by `node.kubernetes.io/exclude-from-external-load-balancers`.

[Service controller](https://github.com/kubernetes/kubernetes/blob/72da0b1bb06607f3f3e067f1bb5ce329ec861e1b/staging/src/k8s.io/cloud-provider/controllers/service/controller.go#L670-L672) that is used by each Cloud Provider Interface implementer use this logic, I'd like to have it integrated into MetalLB as well.

The dependency injection part of the submitter code looks ugly. If you've got ideas on how to improve, I'll do them in a heartbeat.

### DIFF
--- a/internal/k8s/k8s.go
+++ b/internal/k8s/k8s.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"net"
 	"net/http"
+	"time"
 
 	"go.universe.tf/metallb/internal/config"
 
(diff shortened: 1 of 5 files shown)


fedepaol:
Hi @zuzzas , thanks for the PR. 

We have something similar in the works, which is described in the currently opened design proposal here https://github.com/metallb/metallb/blob/512b276feead332c045243c3979c6b80a452c8cc/design/pool-configuration.md#node--address-pool-mapping (full design proposal https://github.com/metallb/metallb/pull/942).

I'd rather leverage the extended flexibility of choosing the nodes from the addresspool object, and have the selection work for both BGP and L2. The behaviour is still not "officially approved" so I'd wait before starting implementing it.

If you need a short term solution, using that label `node.kubernetes.io/exclude-from-external-load-balancers` as node selector for the speaker daemonset would make the trick, because only the nodes with a running speaker will announce.

fedepaol:
Closing this, as it's superseeded by https://github.com/metallb/metallb/pull/1302

PR #1087, opened 2021-11-29
Update FRR image URL to quay.io

sabinaaledort:
There is a rate limit in dockerhub. To avoid it we should pull the FRR image from quay.io.


fedepaol:
LGTM

PR #1082, opened 2021-11-25
Document the CNCF steps to add new maintainers

rata:
I've opened a PR to add fede to the CNCF/foundation repo: https://github.com/cncf/foundation/pull/258/files

And asked him to do the other steps I document in this PR.

Closes: #994

PR #1089, opened 2021-11-29
website: Documents new steps when adding maintainers

rata:
Try number two to do this: https://github.com/metallb/metallb/pull/1082

I consider the previous LGTM still valid for this PR so I'll merge this ASAP, let me know otherwise and I can change this in the future and fix what you think in a follow-up PR.

It is a simple step documentation, that hopefully now renders fine with the hugo version used by the server.


rata:
This new PR indeed fixed the rendering issue, this is live now!

I cherry-picked this to branch v0.11 and live-website

PR #1092, opened 2021-11-30
Unit tests: make the time for waiting the file longer

fedepaol:
We experienced some CI flakes where the rendered file was not available yet. By making this timeout longer, the tests should be more stable.


fedepaol:
cc @msherif1234

msherif1234:
is 100msec is enough will be nice if we measure the time when running UT in CI

fedepaol:
> is 100msec is enough will be nice if we measure the time when running UT in CI

The issue of this timeout is, we must give enough time to the debouncer to update the file. 10 msec was already 10 times the debounce timeout which we are forcing to 1msec in tests, but I suspect we are hitting some scheduler limit with times so low. 
Another option is to have some sort of channel where we publish the updates, but I feel it's a bit too overengineered for the sake of the tests.

internal/bgp/frr/frr_test.go
```diff
@@ -59,7 +59,7 @@ func testSetup(t *testing.T) {
 }
 
 func testCheckConfigFile(t *testing.T) {
-	time.Sleep(10 * time.Millisecond) // give time to the debouncer to write the file
+	time.Sleep(100 * time.Millisecond) // give time to the debouncer to write the file
```
russellb, line 62: sleeps make me sad.  What about making sure the timestamp on the file has been updated?

msherif1234:
/LGTM

> > is 100msec is enough will be nice if we measure the time when running UT in CI
> 
> The issue of this timeout is, we must give enough time to the debouncer to update the file. 10 msec was already 10 times the debounce timeout which we are forcing to 1msec in tests, but I suspect we are hitting some scheduler limit with times so low. Another option is to have some sort of channel where we publish the updates, but I feel it's a bit too overengineered for the sake of the tests.

sure since ut is a requirement now lets get this in and monitor it

internal/bgp/frr/frr_test.go
fedepaol: :-) There should be no files, so we can even check if the file appears after a longer timeout.

internal/bgp/frr/frr_test.go
```diff
@@ -51,18 +53,26 @@ func testGenerateFileNames(t *testing.T) (string, string) {
...
+			return false, nil
+		}
+		if err != nil {
+			return false, err
+		}
+		return true, nil
+	})
 
+	if err != nil {
+		t.Fatalf("Failed to wait the file %s", configFile)
```
msherif1234, line 74: maybe "Failed to wait for configfile %s err %s" ?

internal/bgp/frr/frr_test.go
```diff
@@ -51,18 +53,26 @@ func testGenerateFileNames(t *testing.T) (string, string) {
...
 	osHostname = testOsHostname
 }
 
 func testCheckConfigFile(t *testing.T) {
-	time.Sleep(10 * time.Millisecond) // give time to the debouncer to write the file
 	configFile, goldenFile := testGenerateFileNames(t)
+	err := wait.PollImmediate(10*time.Millisecond, 2*time.Second, func() (bool, error) {
+		_, err := os.Stat(configFile)
+		if errors.Is(err, os.ErrNotExist) {
+			return false, nil
```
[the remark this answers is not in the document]
fedepaol: It does not stop:
```
// PollImmediate tries a condition func until it returns true, an error, or the timeout
// is reached.
```
so if we return false, nil the polling continue.

internal/bgp/frr/frr_test.go
fedepaol: yes, good point. Changing it.

internal/bgp/frr/frr_test.go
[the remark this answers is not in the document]
msherif1234: so we don't need this special check then and just err != nil check after is sufficient ?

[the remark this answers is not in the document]
fedepaol: My idea was, continue if the error is the file is not there, but raise immediately if you get another error.

PR #1093, opened 2021-11-30
Let speakers settle in L2 ExternalTrafficPolicy=Local tests 

oribon:
Since we're creating multiple pods, the announcing
speaker can change until all the pods are ready and
the situation is stabilized.
Adding a 5 seconds sleep should be enough for the
speakers to settle on who announces the service,
and should be enough for it to complete sending the
gratuitous ARPs as that's the time set in spamLoop().

This should solve the flake we're experiencing recently in the CI.

### DIFF
--- a/e2etest/layer2_test.go
+++ b/e2etest/layer2_test.go
@@ -128,26 +128,39 @@ var _ = ginkgo.Describe("L2", func() {
 			port := strconv.Itoa(int(svc.Spec.Ports[0].Port))
 			ingressIP := e2eservice.GetIngressPoint(
 				&svc.Status.LoadBalancer.Ingress[0])
-
-			ginkgo.By("checking connectivity to its external VIP")
-
 			hostport := net.JoinHostPort(ingressIP, port)
 			address := fmt.Sprintf("http://%s/", hostport)
+
+			ginkgo.By(fmt.Sprintf("checking connectivity to its external VIP %s", hostport))
 			err = wgetRetry(address, executor.Host)
 			framework.ExpectNoError(err)
 
-			advNode, err := advertisingNodeFromMAC(epNodes, ingressIP, executor.Host)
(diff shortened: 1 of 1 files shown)


e2etest/layer2_test.go
```diff
@@ -120,6 +120,9 @@ var _ = ginkgo.Describe("L2", func() {
 			epNodes, err := jig.ListNodesWithEndpoint() // Only nodes with an endpoint could be advertising the IP
 			framework.ExpectNoError(err)
 
+			// Give the speakers enough time to settle and for the announcer to complete its gratuitous.
+			time.Sleep(5 * time.Second)
```
fedepaol, line 124: Can we leverage eventually / consistently instead of the sleep?

oribon: Not sure, what would we put there?

fedepaol: Like putting the block
```
                       advNode, err := advertisingNodeFromMAC(epNodes, ingressIP, executor.Host)
			framework.ExpectNoError(err)

			for i := 0; i < 5; i++ {
				name, err := getEndpointHostName(hostport, executor.Host)
				framework.ExpectNoError(err)

				ginkgo.By(fmt.Sprintf("checking that pod %s is on node %s", name, advNode.Name))
				pod, err := cs.CoreV1().Pods(f.Namespace.Name).Get(context.TODO(), name, metav1.GetOptions{})
				framework.ExpectNoError(err)
				framework.ExpectEqual(pod.Spec.NodeName, advNode.Name, fmt.Sprintf("traffic arrived to a pod on node %s which is not the announcing node %s", pod.Spec.NodeName, advNode.Name))
			}
```

In an eventually, with an interval of 0.5 seconds and max timeout of 5 seconds.
That would make the test execution slower in fast cases.

fedepaol:
LGTM

PR #1094, opened 2021-12-01
Specify the addresspool in each L2 test

oribon:
Currently the first BeforeEach creates a generic addresspool
that is sometimes overridden by the test. We should have
all tests declare the addresspools for themselves,
with a clear name that ties it to them.


fedepaol:
Failures not related.
LGTM

PR #1095, opened 2021-12-01
Run one single external container in case of hostnetwork

fedepaol:
When running hostnetworked, frr on metallb side squashes the ip and
configures one single neighbor. Here we add one single container in case
tests are running in hostnetwork mode.

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -72,30 +72,28 @@ type containerConfig struct {
 var _ = ginkgo.Describe("BGP", func() {
 	var cs clientset.Interface
 	var f *framework.Framework
-	containersConf := []containerConfig{
-		{
-			name: frrIBGP,
-			nc: frrconfig.NeighborConfig{
-				ASN:      IBGPAsn,
-				Password: "ibgp-test",
-			},
-			rc: frrconfig.RouterConfig{
-				ASN:      IBGPAsn,
-				BGPPort:  179,
-				Password: "ibgp-test",
-			},
+	ibgpContainerConfig := containerConfig{
+		name: frrIBGP,
+		nc: frrconfig.NeighborConfig{
+			ASN:      IBGPAsn,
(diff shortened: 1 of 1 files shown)


e2etest/bgp_test.go
```diff
@@ -72,30 +72,28 @@ type containerConfig struct {
...
+	ibgpContainerConfig := containerConfig{
+		name: frrEBGP,
+		nc: frrconfig.NeighborConfig{
+			ASN:      IBGPAsn,
+			Password: "ebgp-test",
+		},
+		rc: frrconfig.RouterConfig{
+			ASN:      EBGPAsn,
+			BGPPort:  180,
+			Password: "ebgp-test",
```
msherif1234, line 96: if its ibgp container why its called frrEBGP and ASN should be the same for router and nbr ?

fedepaol: Because I mismatched the names :-)

e2etest/bgp_test.go
```diff
@@ -72,30 +72,28 @@ type containerConfig struct {
...
-			rc: frrconfig.RouterConfig{
-				ASN:      IBGPAsn,
-				BGPPort:  179,
-				Password: "ibgp-test",
-			},
+	ebgpContainerConfig := containerConfig{
+		name: frrIBGP,
+		nc: frrconfig.NeighborConfig{
+			ASN:      IBGPAsn,
+			Password: "ibgp-test",
```
msherif1234, line 79: name frrIBGP ? and ASN need to different maybe we should just call them ASN1 and ASN2 ?

fedepaol: This is the very same configuration we had in place before, I just cut the slice in two elements. Apart from naming the elements, I think everything else does not belong to this PR (or at least to this commit).

msherif1234:
LGTM

oribon:
LGTM

PR #1096, opened 2021-12-01
Add support to generate bgpd corefile for CI

msherif1234:
Modify circle config.yaml to allow core file generation at /tmp folder and modify rawDump to check if there is any core is created 
Unit-test
```
bash-5.1# pkill -SIGSEGV bgpd
bash-5.1# ls -l /tmp/
total 5832
-rw-------    1 root     frr      107610112 Dec  7 01:31 core-bgpd.74.f874d0073cc1.1638840711
 gdb -q -n -ex bt /usr/lib/frr/bgpd /tmp/core-bgpd.74.f874d0073cc1.1638840711
[New LWP 74]
[New LWP 77]
[New LWP 75]
[New LWP 76]
Missing separate debuginfo for the main executable file
Try: dnf --enablerepo='*debug*' install /usr/lib/debug/.build-id/c2/5fc6b15e6584ebfa378da6575ef7ea76a495d3
Core was generated by `/usr/lib/frr/bgpd -d -F traditional -A 127.0.0.1'.
Program terminated with signal SIGABRT, Aborted.
#0  0x00007f6f6422ef98 in ?? ()
[Current thread is 1 (LWP 74)]
#0  0x00007f6f6422ef98 in ?? ()
#1  0x00007ffd41cb1520 in ?? ()
#2  0x0000000000000020 in ?? ()
#3  0x0000000000000000 in ?? ()
(gdb) 
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -49,6 +49,7 @@ jobs:
           if [ "<< parameters.deployment-method >>" = "helm" ]; then FLAGS="--helm-install"; fi
           inv dev-env -i  << parameters.ip-family >> -b  << parameters.bgp-type >> -l all $FLAGS
       - run: |
+          echo '/etc/frr/core-%e.%p.%h.%t' | sudo tee /proc/sys/kernel/core_pattern
           SKIP="none"
           FLAGS=""
           if [ "<< parameters.bgp-type >>" = "native" ]; then SKIP="$SKIP\|FRR\|BFD\|DUALSTACK"; fi
--- a/e2etest/pkg/frr/container/container.go
+++ b/e2etest/pkg/frr/container/container.go
@@ -113,7 +113,7 @@ func startContainer(cfg Config, testDirName string) error {
 	}
 
 	volume := fmt.Sprintf("%s:%s", testDirName, frrMountPath)
-	args := []string{"run", "-d", "--privileged", "--network", cfg.Network, "--rm", "--name", cfg.Name, "--volume", volume, frrImage}
+	args := []string{"run", "-d", "--privileged", "--network", cfg.Network, "--rm", "--ulimit", "core=-1", "--name", cfg.Name, "--volume", volume, frrImage}
 	if cfg.IPv4Address != "" {
 		args = append(args, "--ip", cfg.IPv4Address)
 	}
--- a/tasks.py
+++ b/tasks.py
@@ -427,7 +427,7 @@ def bgp_dev_env(ip_family):
     run('for frr in $(docker ps -a -f name=frr --format {{.Names}}) ; do '
         '    docker rm -f $frr ; '
         'done', echo=True)
-    run("docker run -d --privileged --network kind --rm --name frr --volume %s:/etc/frr "
+    run("docker run -d --privileged --network kind --rm --ulimit core=-1 --name frr --volume %s:/etc/frr "
             "quay.io/frrouting/frr:stable_7.5" % frr_volume_dir, echo=True)
 
     if ip_family == "ipv4":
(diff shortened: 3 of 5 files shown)


manifests/metallb-frr.yaml
```diff
@@ -446,6 +446,9 @@ spec:
           emptyDir: {}
         - name: metrics
           emptyDir: {}
+        - name: frr-corefile
```
fedepaol, line 449: These changes will be translated in the operator's bindata and it doesn't feel right.
Can we adjust the path under `/etc/frr` so we don't need to touch the manifests?

Same for ulimit below.

msherif1234: the volume can be changed but we have to make sure we set RLIMIT for core to unlimiited not sure that can be done anywhere else outside the container init script here

fedepaol: You can kubectl exec into the pod for example, still from CI and before running the tests.
Also, the issues we had on CI were related to the external frrs, the container inside the pods never showed the issue so far.

msherif1234: I was adding this capability to production FRR pod so if it crash in production we need to have core file

fedepaol: so it's not CI only? It feels weird that we need to set something explicit in the container, I wonder how the other components / platform handle this (i.e. ovn, to mention one).
We are not the only one having this problem, so it's probably worth investigating if it's only because we are running kind.

msherif1234: i checked on real OCP cluster and the core pattern and ulimit has been set so yeah no need to change manifests so I limited this change just for CI components

e2etest/pkg/frr/container/container.go
```diff
@@ -107,13 +107,18 @@ func startContainer(cfg Config, testDirName string) error {
 		return errors.Wrapf(err, "Failed to copy FRR config directory. %s", string(res))
 	}
 
+	coreFilePattern := frrMountPath + "/core-%e.%p.%h.%t"
+	res, err = exec.Command("echo", coreFilePattern, "|", "sudo", "tee", "/proc/sys/kernel/core_pattern").CombinedOutput()
```
fedepaol, line 111: Can we do this in CI? Not sure what will happen if we run it on a host where we are not root (i.e. ocp CI)

fedepaol: Also, we would be doing this one time for container, whereas we can do it only once.

tasks.py
```diff
@@ -423,11 +423,12 @@ def bgp_dev_env(ip_family):
     with open("%s/bgpd.conf" % frr_volume_dir, 'w') as f:
         f.write(bgpd_config)
 
+    run("echo '/etc/frr/core-%e.%p.%h.%t' | sudo tee /proc/sys/kernel/core_pattern")
```
fedepaol, line 426: Same, we don't know where this will be running and we don't always have root permissions.

.circleci/config.yml
```diff
@@ -49,6 +49,7 @@ jobs:
           if [ "<< parameters.deployment-method >>" = "helm" ]; then FLAGS="--helm-install"; fi
           inv dev-env -i  << parameters.ip-family >> -b  << parameters.bgp-type >> -l all $FLAGS
       - run: |
+          echo '/etc/frr/core-%e.%p.%h.%t' | sudo tee /proc/sys/kernel/core_pattern
```
fedepaol, line 52: Can you do this also on gh actions? On long term, we will be dropping circleci

fedepaol:
LGTM, thanks!

PR #1097, opened 2021-12-01
Set RouterID in the external containers

fedepaol:
This is probably the cause of the flakes we are experiencing in CI, as discussed with @ton31337 on slack.

Reference FRR issue https://github.com/FRRouting/frr/issues/10142

### DIFF
--- a/e2etest/pkg/frr/container/container.go
+++ b/e2etest/pkg/frr/container/container.go
@@ -69,6 +69,9 @@ func Start(name string, nc config.NeighborConfig, rc config.RouterConfig, networ
 		}
 	}
 
+	// setting routerid after calculating ips
+	frr.RouterConfig.RouterID = frr.Ipv4
+
 	err = frr.updateVolumePermissions()
 	if err != nil {
 		return frr, err
--- a/e2etest/pkg/frr/config/config.go
+++ b/e2etest/pkg/frr/config/config.go
@@ -31,6 +31,7 @@ log timestamp precision 3
 route-map RMAP permit 10
 set ipv6 next-hop prefer-global
 router bgp {{.ASN}}
+  bgp router-id {{.RouterID}}
   no bgp ebgp-requires-policy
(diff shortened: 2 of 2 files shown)


russellb:
lgtm

cgoncalves:
LGTM

e2etest/pkg/frr/container/container.go
```diff
@@ -69,6 +69,9 @@ func Start(name string, nc config.NeighborConfig, rc config.RouterConfig, networ
 		}
 	}
 
+	// setting routerid after calculating ips
+	frr.RouterConfig.RouterID = frr.Ipv4
```
msherif1234, line 73: routerID is constant has nothing to do with interface IPs i.e u don't have to wait to intf IPs to set it can be any 32bits value in ipv4 dotted fmt

russellb: isn't that a typical way to generate / set the ID though?

cgoncalves: True although it is commonly set to an IPv4 address. Makes it easier to identify routers and prevent duplicates.

msherif1234: typical way in router land was to use loopback ip which some random ip 10.10.10.10 something like that again its just uint32 used in the bgp open message

msherif1234: @fedepaol I checked slack and I saw the crash with error msg not sure what led to that but I found hard to believe this will crash bgpd we ran into bad bgp id alot when we were using views but that never crashed bgpd the session just never get to established

fedepaol: I had a private chat with @ton31337 where he was saying that's the cause of the error, looking at the extended logs I put in the frr issue.

fedepaol: I'll merge this, it won't hurt anyhow and it's test only.

PR #1099, opened 2021-12-02
Allow specification of priorityClassName for speaker and controller

### LINKED ISSUE #995: helm option for priorityClassName
Hi,
It would be great if the helm chart allowed priorityClassName to be set for the controller and speakers.

e.g.
```yaml
spec:
  template:
    spec:
      priorityClassName: system-node-critical
```

Here's an example from another chart:
https://github.com/jetstack/cert-manager/blob/master/deploy/charts/cert-manager/templates/webhook-deployment.yaml#L44

ianroberts:
When installing using the plain manifests you can add a `priorityClassName` to the speaker and controller pods via a kustomization, but there's currently no way to do the same when installing via helm.  This PR adds the necessary snippets to the helm templates to inject these settings via e.g. `--set speaker.priorityClassName=system-node-critical`, and the equivalent for `controller`.

Fixed #995


ianroberts:
Not sure why the tests are failing, I haven't touched anything that would affect the tests as installing with the stock `values.yaml` will not change anything from the previous defaults.

fedepaol:
> Not sure why the tests are failing, I haven't touched anything that would affect the tests as installing with the stock `values.yaml` will not change anything from the previous defaults.

yes we have a couple of flakes we are chasing. The do not seem related.

fedepaol:
cc @gclawes

PR #1100, opened 2021-12-03
FRR: block incoming routes

fedepaol:
BGP routes must only be published. We should not accept incoming routes, as FRR is host networked and that would result in unwanted routes at host level of the cluster.

We use prefix-list to block the routes.

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -946,6 +946,66 @@ var _ = ginkgo.Describe("BGP", func() {
 					},
 				}}, "ipv6"))
 	})
+
+	table.DescribeTable("MetalLB FRR rejects any routes advertised by any neighbor", func(addressesRange, pairingIPFamily, toInject string) {
+		configData := config.File{
+			Pools: []config.AddressPool{
+				{
+					Name:     "bgp-test",
+					Protocol: config.BGP,
+					Addresses: []string{
+						addressesRange,
+					},
+				},
+			},
+			Peers: peersForContainers(frrContainers, pairingIPFamily),
+		}
+		neighborAnnounce := func(frr *frrcontainer.FRR) {
+			frr.NeighborConfig.ToAdvertise = toInject
+		}
+
+		for _, c := range frrContainers {
+			pairExternalFRRWithNodes(cs, c, pairingIPFamily, neighborAnnounce)
+		}
+
+		err := configUpdater.Update(configData)
+		framework.ExpectNoError(err)
+
+		for _, c := range frrContainers {
+			validateFRRPeeredWithNodes(cs, c, pairingIPFamily)
+		}
+		svc, _ := createServiceWithBackend(cs, f.Namespace.Name, "external-local-lb")
+
+		defer func() {
+			err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
+			framework.ExpectNoError(err)
+		}()
+
+		speakerPods := getSpeakerPods(cs)
+		Consistently(func() error {
+			for _, pod := range speakerPods {
+				podExec := executor.ForPod(pod.Namespace, pod.Name, "frr")
+				routes, frrRoutesV6, err := frr.Routes(podExec)
+				framework.ExpectNoError(err)
+
+				if pairingIPFamily == "ipv6" {
+					routes = frrRoutesV6
+				}
+
+				for _, route := range routes {
+					if route.Destination.String() == toInject {
+						return fmt.Errorf("Found %s in %s routes", toInject, pod.Name)
+					}
+				}
+			}
+			return nil
+		}, 30*time.Second, 1*time.Second).ShouldNot(HaveOccurred())
+
+	},
+		table.Entry("IPV4", "192.168.10.0/24", "ipv4", "172.16.1.10/32"),
+		table.Entry("IPV6", "fc00:f853:0ccd:e799::/116", "ipv6", "fc00:f853:ccd:e800::1/128"),
+	)
+
 })
 
 func createServiceWithBackend(cs clientset.Interface, namespace string, jigName string, tweak ...func(svc *corev1.Service)) (*corev1.Service, *e2eservice.TestJig) {
--- a/internal/bgp/frr/config.go
+++ b/internal/bgp/frr/config.go
@@ -35,10 +35,11 @@ hostname {{.Hostname}}
 {{- range $n := .Neighbors }}
 {{- range $a := .Advertisements }}
 {{- if gt (len $a.Communities) 0}}
-route-map {{$n.Addr}}-map permit 10
+route-map {{$n.Addr}}-out permit 10
 {{- range $c := $a.Communities }}
   set community {{$c}} additive
 {{- end }}
+route-map {{$n.Addr}}-in deny 20
 {{- end }}
 {{- end }}
 {{- end }}
@@ -66,24 +67,27 @@ router bgp {{.MyASN}}
   {{- end }}
 {{- if ne .BFDProfile ""}} 
   neighbor {{.Addr}} bfd profile {{.BFDProfile}}
-{{- end -}}
+{{- end }}
 {{- end }}
 {{range $n := .Neighbors -}}
 {{/* no bgp default ipv4-unicast prevents peering if no address families are defined. We declare an ipv4 one for the peer to make the pairing happen */}}
 {{- if eq (len .Advertisements) 0}}
   address-family ipv4 unicast
     neighbor {{$n.Addr}} activate
+    neighbor {{$n.Addr}} route-map {{$n.Addr}}-in in
   exit-address-family
   address-family ipv6 unicast
     neighbor {{$n.Addr}} activate
+    neighbor {{$n.Addr}} route-map {{$n.Addr}}-in in
   exit-address-family
 {{- end}}
 {{- range .Advertisements }}
   address-family {{.Version}} unicast
     neighbor {{$n.Addr}} activate
+    neighbor {{$n.Addr}} route-map {{$n.Addr}}-in in
     network {{.Prefix}}
 	{{- if gt (len .Communities) 0}}
-    neighbor {{$n.Addr}} route-map {{$n.Addr}}-map out
+    neighbor {{$n.Addr}} route-map {{$n.Addr}}-out out
     {{- end}}
   exit-address-family
 {{- end}}
--- a/e2etest/pkg/frr/config/config.go
+++ b/e2etest/pkg/frr/config/config.go
@@ -32,6 +32,7 @@ route-map RMAP permit 10
 set ipv6 next-hop prefer-global
 router bgp {{.ASN}}
   bgp router-id {{.RouterID}}
+  no bgp network import-check
   no bgp ebgp-requires-policy
   no bgp default ipv4-unicast
 {{range .Neighbors }}
@@ -48,6 +49,9 @@ router bgp {{.ASN}}
 {{range .V4Neighbors }}
     neighbor {{.Addr}} next-hop-self
     neighbor {{.Addr}} activate
+    {{- if .ToAdvertise}}
+    network {{.ToAdvertise}}
+    {{- end }}
 {{- end }}
   exit-address-family
 {{- end }}
@@ -57,6 +61,9 @@ router bgp {{.ASN}}
     neighbor {{.Addr}} next-hop-self
     neighbor {{.Addr}} activate
     neighbor {{.Addr}} route-map RMAP in
+    {{- if .ToAdvertise}}
+    network {{.ToAdvertise}}
+    {{- end }}
 {{- end }}
 exit-address-family
 {{- end }}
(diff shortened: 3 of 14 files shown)


internal/bgp/frr/config.go
```diff
@@ -66,7 +69,9 @@ router bgp {{.MyASN}}
   {{- end }}
 {{- if ne .BFDProfile ""}} 
   neighbor {{.Addr}} bfd profile {{.BFDProfile}}
-{{- end -}}
+{{- end }}
+  neighbor {{.Addr}} prefix-list deny-all in
```
msherif1234, line 73: nit v4deny-all

internal/bgp/frr/config.go
```diff
@@ -66,7 +69,9 @@ router bgp {{.MyASN}}
...
+  neighbor {{.Addr}} prefix-list v6deny-all in
```
msherif1234, line 74: this should be done inside address-family block where we apply route-map and based on the version u pick one or the other

russellb:
Lgtm

msherif1234:
so 0.0.0.0/0 and 0::0/0 came to FRR speaker because FRR test containers are host network ? what if we have real deployment and the other side have all kind of routes static or other routing protocols is there a way to block all none BGP routes ?

fedepaol:
> so 0.0.0.0/0 and 0::0/0 came to FRR speaker because FRR test containers are host network ? what if we have real deployment and the other side have all kind of routes static or other routing protocols is there a way to block all none BGP routes ?

0.0.0.0/0 and the other one are a way to block ALL the addresses. By doing this we are blocking all the routes coming from the other side.
If you look at the test, I tried to inject a 32 address, and if you remove this code that route will be added to the host (because frr runs hostnetworked).

e2etest/bgp_test.go
```diff
@@ -946,6 +946,66 @@ var _ = ginkgo.Describe("BGP", func() {
...
+						return fmt.Errorf("Found %s in %s routes", toInject, pod.Name)
+					}
+				}
+			}
+			return nil
+		}, 30*time.Second, 1*time.Second).ShouldNot(HaveOccurred())
+
+	},
+		table.Entry("IPV4", "192.168.10.0/24", "ipv4", "172.16.1.10/32"),
+		table.Entry("IPV6", "fc00:f853:0ccd:e799::/116", "ipv6", "fc00:f853:ccd:e800::1/128"),
```
msherif1234, line 1006: those /32 and 128 are the NIC IP address can we use const and name them accordingly to make obvious ?

msherif1234:
> > so 0.0.0.0/0 and 0::0/0 came to FRR speaker because FRR test containers are host network ? what if we have real deployment and the other side have all kind of routes static or other routing protocols is there a way to block all none BGP routes ?
> 
> 0.0.0.0/0 and the other one are a way to block ALL the addresses. By doing this we are blocking all the routes coming from the other side. If you look at the test, I tried to inject a 32 address, and if you remove this code that route will be added to the host (because frr runs hostnetworked).

so "0" means "*" wildcard in this prefix-list

e2etest/bgp_test.go
fedepaol: They are not the NIC ips. They are random ips that I am injecting routes for.
The corresponding name is the function parameter, I can change that to make it more clear.

internal/bgp/frr/config.go
fedepaol: This will make the template uglier, as we'd need to do it both when there are no advertisements and when we have those, plus iff-ing on the ip family.
I think it's fair to say we always block everything.

e2etest/bgp_test.go
msherif1234: cool I wrote that comment before u replied to other one so its clear to me know

internal/bgp/frr/config.go
msherif1234: ugly ofcourse but it address-family config, not sure what will happen when u pass v4 address to v6 prefix list and other way around to be safe having them under their respective address-family will be less error prune IMO

fedepaol: (I mean, if it works)

fedepaol: I think you are right. The resulting configuration renders like this :

```
 !
 address-family ipv4 unicast
  neighbor fc00:f853:ccd:e793::5 activate
  neighbor fc00:f853:ccd:e793::5 prefix-list v6deny-all in
  neighbor fc00:f853:ccd:e793::6 activate
  neighbor fc00:f853:ccd:e793::6 prefix-list v6deny-all in
 exit-address-family
 !
 address-family ipv6 unicast
  neighbor fc00:f853:ccd:e793::5 activate
  neighbor fc00:f853:ccd:e793::6 activate
 exit-address-family
```

internal/bgp/frr/config.go
```diff
@@ -44,6 +44,9 @@ route-map {{$n.Addr}}-map permit 10
 {{- end }}
 {{- end }}
 
+ip prefix-list deny-all deny 0.0.0.0/0
+ip prefix-list v6deny-all deny ::/0
```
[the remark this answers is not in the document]
fedepaol: I think I tried that and resulted in an error. Let me try again.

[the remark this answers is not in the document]
fedepaol: Yep that yelds:
```
b'line 9: % Unknown command: ip6 prefix-list v6deny-all deny ::/0\n\n'
```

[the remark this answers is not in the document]
cgoncalves: Typo? Should be `ipv6`, not `ip6`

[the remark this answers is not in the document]
fedepaol: Ah, that works! Thanks for spotting it :-)

fedepaol:
@msherif1234 @cgoncalves I think I addressed all comments here.

markkrj:
> 0.0.0.0/0 and the other one are a way to block ALL the addresses. By doing this we are blocking all the routes coming from the other side. If you look at the test, I tried to inject a 32 address, and if you remove this code that route will be added to the host (because frr runs hostnetworked).

If you permit a  comment about this from an outsider, this will work because FRR (and all routers that I know of) assume an implicit deny on prefix-lists (and ACLs in general), but your rule is not getting matched at all... 0.0.0.0/0 matches only this exact prefix (a default route). To match against any prefix, you must use "0.0.0.0/0 le 32".

cgoncalves:
Good point. To make it even more explicit, consider "deny any"
```
centos8.localdomain(config)# ip prefix-list deny-all deny any
centos8.localdomain(config)# ipv6 prefix-list v6deny-all deny any
centos8.localdomain(config)# exit
centos8.localdomain# show running-config 
Building configuration...

Current configuration:
!
frr version 7.5
frr defaults traditional
hostname centos8.localdomain
no ip forwarding
no ipv6 forwarding
no service integrated-vtysh-config
!
ip prefix-list deny-all seq 5 deny any
!
ipv6 prefix-list v6deny-any seq 5 deny any
!
line vty
!
end
```

fedepaol:
Thanks @markkrj and @cgoncalves . I'll use the `any` version, seems more readable to me.

internal/bgp/frr/config.go
```diff
@@ -66,21 +69,24 @@ router bgp {{.MyASN}}
...
   neighbor {{.Addr}} bfd profile {{.BFDProfile}}
-{{- end -}}
+{{- end }}
 {{- end }}
 {{range $n := .Neighbors -}}
 {{/* no bgp default ipv4-unicast prevents peering if no address families are defined. We declare an ipv4 one for the peer to make the pairing happen */}}
 {{- if eq (len .Advertisements) 0}}
   address-family ipv4 unicast
     neighbor {{$n.Addr}} activate
+    neighbor {{.Addr}} prefix-list ipv4deny-all in
```
ton31337, line 79: I'd recommend using route-map for in/out route filtering.

```
route-map {{.Addr}}-in permit {{.index}}
  match ip address prefix-list ipv4deny-all
route-map {{.Addr}}-in permit {{.index}}
  match ipv6 address prefix-list ipv6deny-all
...
neighbor {{.Addr}} route-map {{.Addr}}-in in
neighbor {{.Addr}} route-map {{.Addr}}-out out
```

Just will be more flexible in the future.

fedepaol: will do, thanks!

msherif1234:
I have question/doubt with this rule in place and the other end has BGP redistribute config for any IGP routes will this rule now incorrectly drop those routes ?

fedepaol:
> I have question/doubt with this rule in place and the other end has BGP redistribute config for any IGP routes will this rule now incorrectly drop those routes ?

not sure I understood. On metallb side we must drop all the incoming routes. My understanding is that htey affect only the local instance.

cgoncalves:
FRR-based speakers will deny installing in its RIB any IP prefixes advertised by its BGP neighbors. In other words, it will receive the prefixes but drop them (think of it as an iptables DROP rule).

internal/bgp/frr/config.go
msherif1234: @ton31337 can FRR handle multiple route-map applied in the same direction or they have to be merged into single route-map ?

fedepaol: I am not sure this version looks more readable, I may revert and bring in route-map(s) when we need to deal with more complex scenarios.
Thoughts?

cgoncalves:
Personally, while I agree that @ton31337's suggestion provides more flexibility, it is not something we require today and can be easily changed later. I am also not sure how it would scale with a large number of neighbors as it would result in having a route-map, each with its own prefix-list, for each neighbor.

My preference goes to https://github.com/metallb/metallb/pull/1100#issuecomment-986700928 but I am definitely biased here :sweat_smile:

internal/bgp/frr/config.go
msherif1234: fortunately the existing route-map in the "out" direction so we don't have to merge yet :) so I prefer using route-map I understand the template becoming hard to read but I know from day1 this where it will go I will revive some initial attempt to make it a hierachy of templates not only one as it won't scale with bgp configs

msherif1234:
> Personally, while I agree that @ton31337's suggestion provides more flexibility, it is not something we require today and can be easily changed later. I am also not sure how it would scale with a large number of neighbors as it would result in having a route-map, each with its own prefix-list, for each neighbor.
> 
> My preference goes to [#1100 (comment)](https://github.com/metallb/metallb/pull/1100#issuecomment-986700928) but I am definitely biased here sweat_smile

for scale we probably will need to adopt peer-group configuration

ton31337:
>for scale we probably will need to adopt peer-group configuration

Also, consider using dynamic peers as well (if talking about the scale) :)

internal/bgp/frr/config.go
fedepaol: Let's try to wrap up.
We can have only one route-map per family per direction per neighbor? 
Like here it's allowed having two:

```
route-map {{.Addr}}-in permit {{.index}}
  match ip address prefix-list ipv4deny-all
route-map {{.Addr}}-in permit {{.index}}
  match ipv6 address prefix-list ipv6deny-all
``` 
because they are for different ip families?

And scaling aside, what's the reason for going route-maps now and not later (as this will always be block every route in)?

ton31337: It was just a suggestion :) 

>as this will always be block every route in

What do you mean? Which example?

fedepaol: Was trying to "talk out loudly". What I meant is, what we are trying to achieve here is to always block incoming routes. Due to the nature of metallb, that is really unlikely to change so it's not that bad pinning it to a neighbor (I think).

fedepaol:
Last version, as discussed with @ton31337 this is way more compact.
I also renamed the -in / -out rules to make it more clear.

internal/bgp/frr/testdata/TestBFDProfileAllDefault.golden
```diff
@@ -3,6 +3,9 @@ log file /etc/frr/frr.log informational
 log timestamp precision 3
 hostname dummyhostname
 
+ip prefix-list ipv4deny-all deny any
```
cgoncalves, line 6: Leftovers

fedepaol: Thanks, fixed now

cgoncalves:
> Last version, as discussed with @ton31337 this is way more compact. I also renamed the -in / -out rules to make it more clear.

Could you please help us understand why having a route-map in per neighbor was still preferred over prefix-list? Potential future flexibility over readability?

ton31337:
>Could you please help us understand why having a route-map in per neighbor was still preferred over prefix-list? Potential future flexibility over readability?

@cgoncalves there is no such as _preferred_, it's just more like _universal_. From what I see usually, people prefer using route-maps everywhere and all the logic inside. Because now you have prefix-list and for some reason, you start using route-map, you will end up with two clashing filters (route-map + prefix-list) while you can have only a single one from the beginning. 

I'm not saying it's wrong as it is (with prefix-list), I just suggesting :) Both work fine. 

At least if you decide to drop everything except some routes for example tagged with arbitrary communities or so, route maps must be used. They are just flexible as you said for the future _features_ :)

msherif1234:
> > for scale we probably will need to adopt peer-group configuration
> 
> Also, consider using dynamic peers as well (if talking about the scale) :)

using peer x.x.x.x/24 ? I didn't know FRR support that ? very cool !!

cgoncalves:
LGTM

PR #1110, opened 2021-12-07
BFD metrics: Fix session up metrics determination method

sabinaaledort:
BFD peer has 4 states: up, down, init, or failing.
If we only check the "down" status we state that the session is up while it can be init or failing.


oribon:
lgtm

fedepaol:
LGTM

PR #1111, opened 2021-12-07
E2Etests fix: Use configMapName variable in configmap updater

sabinaaledort:
We should use the configMapName variable instead of hardcoding the configmap name, so we will be able to run the E2E tests
using different MetalLB configmap names.


fedepaol:
LGTM

PR #1113, opened 2021-12-09
FRR: Allow nexthop tracking to resolve via the default route

oribon:
The native implementation could use the default route to peer with
routers outside its network that couldn't be reached otherwise.
We should enable it in FRR as well to support peering with these routers.
https://docs.frrouting.org/en/latest/zebra.html#clicmd-ip-nht-resolve-via-default

### DIFF
--- a/internal/bgp/frr/config.go
+++ b/internal/bgp/frr/config.go
@@ -30,6 +30,8 @@ const configTemplate = `
 log file /etc/frr/frr.log {{.Loglevel}}
 log timestamp precision 3
 hostname {{.Hostname}}
+ip nht resolve-via-default
+ipv6 nht resolve-via-default
 
 {{- range .Routers }}
 {{- range $n := .Neighbors }}
--- a/internal/bgp/frr/testdata/TestBFDProfileAllDefault.golden
+++ b/internal/bgp/frr/testdata/TestBFDProfileAllDefault.golden
@@ -2,6 +2,8 @@
 log file /etc/frr/frr.log informational
 log timestamp precision 3
 hostname dummyhostname
+ip nht resolve-via-default
+ipv6 nht resolve-via-default
 
 
 bfd
--- a/internal/bgp/frr/testdata/TestBFDProfileCornerCases.golden
+++ b/internal/bgp/frr/testdata/TestBFDProfileCornerCases.golden
@@ -2,6 +2,8 @@
 log file /etc/frr/frr.log informational
 log timestamp precision 3
 hostname dummyhostname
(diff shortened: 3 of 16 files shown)


msherif1234:
This is only need for eBGP multi-hop case right ? I don't think its right to unconditionally resolve over default route in other cases ?

fedepaol:
> This is only need for eBGP multi-hop case right ? I don't think its right to unconditionally resolve over default route in other cases ?

I am pretty ignorant here but I think this is a fallback when we don't have explicit routes to reach the neighbor? Also, why would it be limited to eBGP? Can't we have iBGP but multi-hopped (although probably not that common)?

msherif1234:
> > This is only need for eBGP multi-hop case right ? I don't think its right to unconditionally resolve over default route in other cases ?
> 
> I am pretty ignorant here but I think this is a fallback when we don't have explicit routes to reach the neighbor? Also, why would it be limited to eBGP? Can't we have iBGP but multi-hopped (although probably not that common)?

eBGP is the only type "by default" assumes only one hop and send NEXT_HOP in the update messages to peers unlike iBGP so we don't endup with routing loop.

can we get FRR folks eye on this PR

cgoncalves:
> This is only need for eBGP multi-hop case right ? I don't think its right to unconditionally resolve over default route in other cases ?

Not necessarily. Consider a spine-leaf network architecture where the worker nodes (rack servers) need to iBGP peer with the spine switches or datacenter routers but they do not have a specific route in their RIB hence can only be resolved via the default route.

msherif1234:
> > This is only need for eBGP multi-hop case right ? I don't think its right to unconditionally resolve over default route in other cases ?
> 
> Not necessarily. Consider a spine-leaf network architecture where the worker nodes (rack servers) need to iBGP peer with the spine switches or datacenter routers but they do not have a specific route in their RIB hence can only be resolved via the default route.

I thought this works only for passive nbr ?

msherif1234:
> > > This is only need for eBGP multi-hop case right ? I don't think its right to unconditionally resolve over default route in other cases ?
> > 
> > 
> > Not necessarily. Consider a spine-leaf network architecture where the worker nodes (rack servers) need to iBGP peer with the spine switches or datacenter routers but they do not have a specific route in their RIB hence can only be resolved via the default route.
> 
> I thought this works only for passive nbr ?

I doubled checked with frr folks and they think it safe to go with

fedepaol:
So are we all good with this?

fedepaol:
LGTM

PR #1115, opened 2021-12-10
Add set local-preference route-map action to frr config

msherif1234:
BGP Advertisement configuration API allow setting LOCAL_PREF, this PR
extend FRR configuration template to propagate this configuration to FRR

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -58,6 +58,7 @@ const (
 	v4PoolAddresses = "192.168.10.0/24"
 	v6PoolAddresses = "fc00:f853:0ccd:e799::/124"
 	CommunityNoAdv  = "65535:65282" // 0xFFFFFF02: NO_ADVERTISE
+	IPLocalPref     = uint32(300)
 )
 
 var (
@@ -889,7 +890,7 @@ var _ = ginkgo.Describe("BGP", func() {
 					validateFRRPeeredWithNodes(cs, c, ipFamily)
 				}
 
-				svc, _ := createServiceWithBackend(cs, f.Namespace.Name, "external-local-lb", testservice.TrafficPolicyCluster) // Is a sleep required here?
+				svc, _ := createServiceWithBackend(cs, f.Namespace.Name, "external-local-lb", testservice.TrafficPolicyCluster)
 
 				defer func() {
 					err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
@@ -943,6 +944,59 @@ var _ = ginkgo.Describe("BGP", func() {
 						},
 					},
 				}}, "ipv6"))
+
+		table.DescribeTable("configure bgp local-preference and verify it gets propagated",
+			func(poolAddresses []string, ipFamily string, localPref uint32) {
+				configData := config.File{
+					Pools: []config.AddressPool{
+						{
+							Name:      "bgp-test",
+							Protocol:  config.BGP,
+							Addresses: poolAddresses,
+							BGPAdvertisements: []config.BgpAdvertisement{
+								{
+									LocalPref: uint32Ptr(localPref),
+								},
+							},
+						},
+					},
+					Peers: peersForContainers(frrContainers, ipFamily),
+				}
+				for _, c := range frrContainers {
+					pairExternalFRRWithNodes(cs, c, ipFamily)
+				}
+
+				err := configUpdater.Update(configData)
+				framework.ExpectNoError(err)
+
+				for _, c := range frrContainers {
+					validateFRRPeeredWithNodes(cs, c, ipFamily)
+				}
+
+				svc, _ := createServiceWithBackend(cs, f.Namespace.Name, "external-local-lb", testservice.TrafficPolicyCluster)
+
+				defer func() {
+					err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
+					framework.ExpectNoError(err)
+				}()
+
+				allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+				framework.ExpectNoError(err)
+
+				for _, c := range frrContainers {
+					validateService(cs, svc, allNodes.Items, c)
+				}
+				// LocalPref check is only valid for iBGP sessions
+				for _, c := range frrContainers {
+					if c.Name == "frrIBGP" {
+						Eventually(func() error {
+							return frr.RoutesMatchLocalPref(c, ipFamily, localPref)
+						}, 4*time.Minute, 1*time.Second).Should(BeNil())
+					}
+				}
+			},
+			table.Entry("IPV4", []string{v4PoolAddresses}, "ipv4", IPLocalPref),
+			table.Entry("IPV6", []string{v6PoolAddresses}, "ipv6", IPLocalPref))
 	})
 
 	table.DescribeTable("MetalLB FRR rejects any routes advertised by any neighbor", func(addressesRange, pairingIPFamily, toInject string) {
--- a/e2etest/pkg/frr/bgp.go
+++ b/e2etest/pkg/frr/bgp.go
@@ -124,3 +124,27 @@ func ContainsCommunity(exec executor.Executor, community string) error {
 	}
 	return nil
 }
+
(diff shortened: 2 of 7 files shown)


e2etest/bgp_test.go
```diff
@@ -927,6 +927,7 @@ var _ = ginkgo.Describe("BGP", func() {
 							Communities: []string{
 								CommunityNoAdv,
 							},
+							LocalPref: uint32Ptr(300),
```
fedepaol, line 930: Are we able to verify this on the other side inspecting frr?

msherif1234: I couldn’t find a show command to help me with this I just confirmed route map constructed correctly when either options or both in use it’s not always 1:1 also I wanted to mention this an ibgp only feature setting it for ebgp has no effect

cgoncalves: +1 to verify the local preference on the neighbor side. Here's one way to do it where the local preference is set to _300_:

```
bash-5.1# vtysh -c 'show ip bgp json'
{
 "vrfId": 0,
 "vrfName": "default",
 "tableVersion": 1,
 "routerId": "10.89.0.5",
 "defaultLocPrf": 100,
 "localAS": 64512,
 "routes": { "9.9.9.9/32": [
  {
    "valid":true,
    "bestpath":true,
    "pathFrom":"internal",
    "prefix":"9.9.9.9",
    "prefixLen":32,
    "network":"9.9.9.9\/32",
    "metric":0,
    "locPrf":300,
    "weight":0,
    "peerId":"10.89.0.3",
    "path":"",
    "origin":"IGP",
    "nexthops":[
      {
        "ip":"10.89.0.3",
        "hostname":"kind-control-plane",
        "afi":"ipv4",
        "used":true
      }
    ]
  }
] }  }
```

Could you please create another test where only local preference is set (i.e. no BGP communities)?

msherif1234: added test case for localpref

e2etest/bgp_test.go
```diff
@@ -945,6 +947,74 @@ var _ = ginkgo.Describe("BGP", func() {
...
+					framework.ExpectNoError(err)
+				}
+
+				allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+				framework.ExpectNoError(err)
+
+				for _, c := range frrContainers {
+					validateService(cs, svc, allNodes.Items, c)
+					Eventually(func() error {
+						return frr.RoutesMatchedLocalPref(c, IPv4LocalPref, IPv6LocalPref)
```
fedepaol, line 988: I'd rather have the localpref we are matching against passed as a parameter.
Here we match the global constant against the other global constant being set in the table entry, which are incidentally the same. I think it's better to have it more explicit to be the same value..

e2etest/bgp_test.go
```diff
@@ -945,6 +947,74 @@ var _ = ginkgo.Describe("BGP", func() {
...
+					}, 4*time.Minute, 1*time.Second).Should(BeNil())
+				}
+			},
+			table.Entry("IPV4", []config.AddressPool{
```
fedepaol, line 992: See the comment above, this can become like the other tests, where we create AddressPool inside the tests and we pass teh address and the local pref from outside.

e2etest/pkg/frr/bgp.go
```diff
@@ -124,3 +124,24 @@ func ContainsCommunity(exec executor.Executor, community string) error {
 	}
 	return nil
 }
+
+// RoutesMatchedLocalPref check if routes match specific local preference value.
+func RoutesMatchedLocalPref(exec executor.Executor, v4LocalPref, v6LocalPref uint32) error {
```
fedepaol, line 129: We always check only one at the time, I'd pass here the expected localpref and the ip family.

internal/bgp/frr/parse.go
```diff
@@ -49,6 +50,7 @@ type IPInfo struct {
 type FRRRoute struct {
 	Valid    bool   `json:"valid"`
 	PeerID   string `json:"peerId"`
+	LocPrf   uint32 `json:"locPrf"`
```
fedepaol, line 53: nit: LocalPref. No need too abbreviate.

e2etest/pkg/frr/bgp.go
msherif1234: not true in case of dual-stack ?

fedepaol: But the tests that we are adding here are using this one either set v4localpref or v6localpref I think

msherif1234: I will use one value for CI

fedepaol: So, can you change this function to accept only one value and the family of that value?

e2etest/bgp_test.go
```diff
@@ -945,6 +946,55 @@ var _ = ginkgo.Describe("BGP", func() {
...
+				}
+
+				err := configUpdater.Update(configData)
+				framework.ExpectNoError(err)
+
+				for _, c := range frrContainers {
+					validateFRRPeeredWithNodes(cs, c, ipFamily)
+				}
+
+				svc, _ := createServiceWithBackend(cs, f.Namespace.Name, "external-local-lb", testservice.TrafficPolicyCluster) // Is a sleep required here?
```
fedepaol, line 978: nit: remove the comment

e2etest/pkg/frr/bgp.go
```diff
@@ -124,3 +124,27 @@ func ContainsCommunity(exec executor.Executor, community string) error {
 	}
 	return nil
 }
+
+func allroutesMatchLocalPref(routes map[string]bgpfrr.Route, localPref uint32) error {
```
fedepaol, line 128: nit: allRoutes

fedepaol: Also, if this is not exported, makes sense to have it below `RoutesMatchLocalPref`

e2etest/pkg/frr/bgp.go
```diff
@@ -124,3 +124,27 @@ func ContainsCommunity(exec executor.Executor, community string) error {
...
+	for _, route := range routes {
+		if route.LocalPref != 0 && route.LocalPref != localPref {
```
fedepaol, line 130: Doesn't this risk to have false positives?

msherif1234: what do u mean ? localPref 0 check to filter out ebGP session since its not set for them ?

fedepaol: What I mean is, if we set a localPref from outside, and all the routes have LocalPref = 0 this will pass even if it's wrong.

msherif1234: routes will hjave localpref of 0 only for eBGP and for those we need to skip the test. iBGP if u don't set it its default to 100

fedepaol: > routes will hjave localpref of 0 only for eBGP and for those we need to skip the test. iBGP if u don't set it its default to 100

Ok, can we then skip the test for eBGP routes / container and remove the != 0 so we are more explicit in what we are testing?

msherif1234: is there a way in bgp_test to specify ? I coouldn't find away yesterday ?

fedepaol: You are testing against both external frr containers. Picking the ebgp one should make the trick I think

msherif1234: they are both called frrContainers I am using containersNetwork check that should let me run IBGP only

fedepaol: you can filter by name, or have a reference to the ebgp one.

e2etest/bgp_test.go
```diff
@@ -943,6 +944,59 @@ var _ = ginkgo.Describe("BGP", func() {
...
+				}()
+
+				allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+				framework.ExpectNoError(err)
+
+				for _, c := range frrContainers {
+					validateService(cs, svc, allNodes.Items, c)
+				}
+				// LocalPref check is only valid for iBGP sessions
+				if containersNetwork == "host" {
```
fedepaol, line 990: This won't work in CI, because containersNetwork is docker

fedepaol:
LGTM

PR #1116, opened 2021-12-10
ci: Disable fossa github action

russellb:
We tried this github action to trigger license analysis, but it's
generating many false positives. I'm about to set up an alternative to
try out, so just disable this one for now.


fedepaol:
LGTM

PR #1119, opened 2021-12-13
Remove NextHop from advertisements

fedepaol:
The NextHop field is not used anymore, and it's misleading keeping it in the code. Here we remove it from the codebase and from the tests.

### DIFF
--- a/internal/bgp/frr/frr_test.go
+++ b/internal/bgp/frr/frr_test.go
@@ -241,7 +241,6 @@ func TestSingleAdvertisement(t *testing.T) {
 	communities = append(communities, community)
 	adv := &bgp.Advertisement{
 		Prefix:      prefix,
-		NextHop:     net.ParseIP("10.1.1.1"),
 		Communities: communities,
 		LocalPref:   300,
 	}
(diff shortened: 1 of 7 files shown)


internal/bgp/native/native.go
```diff
@@ -49,7 +49,7 @@ type session struct {
 	closed         bool
 	conn           net.Conn
 	actualHoldTime time.Duration
-	defaultNextHop net.IP
+	nextHop        net.IP
```
msherif1234, line 52: if you are removing this field I will remove it from the structure and the api below and just use addr.IP in the update msg, so we will have no reference to nexthop in the native stack

fedepaol: We can't remove it as it's used in sendUpdate below.
See https://github.com/metallb/metallb/pull/1119/files#diff-cd8c067f48fafd2536808963cdc4b62bacf5309bdf396790de744e793a239362R145

msherif1234: I was thinking of getting it internally when its needed instead of passing it around 
```
addr, ok := conn.LocalAddr().(*net.TCPAddr)
	if !ok {
		conn.Close()
		return fmt.Errorf("getting local addr for default nexthop to %q: %s", s.addr, err)
	}
	s.nextHop = addr.IP
```

but I am not really sure if this cleaner or what  u have is cleaner I am fine either way

fedepaol: I understood, but there's another function, different from this, that uses the field and there we don't have addr.IP available.
See https://github.com/metallb/metallb/pull/1119/files#diff-cd8c067f48fafd2536808963cdc4b62bacf5309bdf396790de744e793a239362R126

fedepaol: `func (s *session) sendUpdates() bool {`

msherif1234: you can find the ip value using the code snippet I shared getIP() anyway just a thought I am fine with ur approach just wanted to bounce with u

fedepaol: Now I get it, we could still do it because conn is stored in the session object. However, it doesn't cost too much and I think it is a bit more clear if we have the field named properly.

internal/bgp/frr/frr_test.go
```diff
@@ -241,33 +239,6 @@ func TestSingleAdvertisementNoRouterID(t *testing.T) {
 	testCheckConfigFile(t)
 }
 
-func TestSingleAdvertisementInvalidPrefix(t *testing.T) {
```
oribon, line 244: Why delete this one?

fedepaol: Because it has no value anymore, since the only validation is done against the length of the communities.

oribon: :+1: Do you need to delete the related golden file?

fedepaol: done

msherif1234:
LGTM

sabinaaledort:
Looks good. I think the TestSingleAdvertisementInvalidNextHop test should be removed as well. https://github.com/metallb/metallb/blob/1be4d12f647d257e66e39e9a03da724ae4a07c63/internal/bgp/frr/frr_test.go#L316

cgoncalves:
LGTM. I didn't find any leftovers.

fedepaol:
Thanks!

PR #1120, opened 2021-12-13
E2E: wait for daemons to be available in external frrs

fedepaol:
Some tests fails for not having FRR loading the right configuration right after starting the container. Here we check for the daemons to be ready before giving it away.

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"net"
 	"strconv"
+	"sync"
 	"time"
 
 	"go.universe.tf/metallb/e2etest/pkg/config"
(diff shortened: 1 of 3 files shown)


e2etest/bgp_test.go
```diff
@@ -1167,16 +1170,41 @@ func frrIsPairedOnPods(cs clientset.Interface, n *frrcontainer.FRR, ipFamily str
...
 		g.Go(func() error {
+			toFind := map[string]bool{
+				"zebra":    true,
+				"watchfrr": true,
+				"bgpd":     true,
+				"bfdd":     true,
+			}
 			c, err := frrcontainer.Start(conf.name, conf.nc, conf.rc, containersNetwork, hostIPv4, hostIPv6)
 			if c != nil {
+				err = wait.PollImmediate(time.Second, 5*time.Minute, func() (bool, error) {
```
oribon, line 1187: 5 minutes isn't a bit too long?

fedepaol: That's the worst case scenario, I think I saw occurrences of it failing in minutes in my previous attempts.
If everything works properly, it will be faster. Note that the daemons start, soon or later, it's not a test that may fail.

PR #1122, opened 2021-12-13
E2E: create the external containers only once

fedepaol:
Instead of creating and destroying the containers after each run, we create them only once and we keep them for all the tests.

This has several benefits:
- Reduce the test run time since we don't have to pay the container
  creation each time
- The test suite will clean the container even when we block the test
  execution
- This is likely to reduce / remove the flake that happens at container
  start time

### DIFF
--- a/e2etest/bgp_test.go
+++ b/e2etest/bgp_test.go
@@ -71,10 +71,7 @@ type containerConfig struct {
 	rc   frrconfig.RouterConfig
 }
 
-var _ = ginkgo.Describe("BGP", func() {
-	var configUpdater config.Updater
-	var cs clientset.Interface
-	var f *framework.Framework
+func setupContainers() []*frrcontainer.FRR {
 	ibgpContainerConfig := containerConfig{
 		name: frrIBGP,
 		nc: frrconfig.NeighborConfig{
@@ -100,6 +97,33 @@ var _ = ginkgo.Describe("BGP", func() {
 		},
 	}
 
+	var res []*frrcontainer.FRR
+	var err error
+	if containersNetwork == "host" {
+		if net.ParseIP(hostIPv4) == nil {
+			framework.Fail("Invalid hostIPv4")
+		}
+		if net.ParseIP(hostIPv6) == nil {
+			framework.Fail("Invalid hostIPv6")
+		}
+		res, err = createFRRContainers(ibgpContainerConfig)
+	} else {
+		res, err = createFRRContainers(ibgpContainerConfig, ebgpContainerConfig)
+	}
+	framework.ExpectNoError(err)
+	return res
+}
+
+func tearDownContainers(containers []*frrcontainer.FRR) {
+	err := stopFRRContainers(containers)
+	framework.ExpectNoError(err)
+}
+
+var _ = ginkgo.Describe("BGP", func() {
+	var configUpdater config.Updater
+	var cs clientset.Interface
+	var f *framework.Framework
+
 	ginkgo.AfterEach(func() {
 		if ginkgo.CurrentGinkgoTestDescription().Failed {
 			for _, c := range frrContainers {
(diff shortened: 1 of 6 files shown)


e2etest/bgp_test.go
```diff
@@ -39,6 +39,7 @@ import (
 	dto "github.com/prometheus/client_model/go"
 	"go.universe.tf/metallb/e2etest/pkg/frr"
 	frrconfig "go.universe.tf/metallb/e2etest/pkg/frr/config"
+	"go.universe.tf/metallb/e2etest/pkg/frr/container"
```
oribon, line 42: double import

e2etest/pkg/frr/container/container.go
```diff
@@ -36,15 +37,23 @@ type FRR struct {
 	Ipv6           string
 }
 
+type Config struct {
```
oribon, line 40: This seems pretty similar to the containerConfig struct we already have. WDYT about combining them here and have the Start func receive only that?

e2etest/bgp_test.go
```diff
@@ -86,6 +86,11 @@ var _ = ginkgo.Describe("BGP", func() {
 			BGPPort:  179,
 			Password: "ibgp-test",
 		},
+		cc: container.Config{
```
oribon, line 89: Feels a bit confusing having container.Config inside containerConfig, combining the structs should solve this.

e2etest/e2etest_suite_test.go
```diff
@@ -77,6 +80,8 @@ func handleFlags() {
 	flag.BoolVar(&skipDockerCmd, "skip-docker", false, "set this to true if the BGP daemon is running on the host instead of in a container")
 	flag.StringVar(&ipv4ServiceRange, "ipv4-service-range", "0", "a range of IPv4 addresses for MetalLB to use when running in layer2 mode")
 	flag.StringVar(&ipv6ServiceRange, "ipv6-service-range", "0", "a range of IPv6 addresses for MetalLB to use when running in layer2 mode")
+	flag.StringVar(&ipv4ForContainers, "ips-for-containers-v4", "0", "a range of IPv4 addresses available for containers")
+	flag.StringVar(&ipv6ForContainers, "ips-for-containers-v6", "0", "a range of IPv4 addresses available for containers")
```
oribon, line 84: Should be `of IPv6...`

e2etest/e2etest_suite_test.go
```diff
@@ -77,6 +80,8 @@ func handleFlags() {
 	flag.BoolVar(&skipDockerCmd, "skip-docker", false, "set this to true if the BGP daemon is running on the host instead of in a container")
 	flag.StringVar(&ipv4ServiceRange, "ipv4-service-range", "0", "a range of IPv4 addresses for MetalLB to use when running in layer2 mode")
 	flag.StringVar(&ipv6ServiceRange, "ipv6-service-range", "0", "a range of IPv6 addresses for MetalLB to use when running in layer2 mode")
+	flag.StringVar(&ipv4ForContainers, "ips-for-containers-v4", "0", "a range of IPv4 addresses available for containers")
```
oribon, line 83: The code expects a comma separated list of IPs and not a range

e2etest/pkg/frr/container/container.go
```diff
@@ -59,9 +68,16 @@ func Start(name string, nc config.NeighborConfig, rc config.RouterConfig, networ
 		RouterConfig:   rc,
 	}
 
-	if network == hostNetwork {
-		frr.Ipv4 = hostipv4
-		frr.Ipv6 = hostipv6
+	if cc.Network == hostNetwork {
+		if net.ParseIP(cc.HostIPV4) == nil {
+			return nil, errors.New("Invalid hostIPv4")
```
oribon, line 73: Can you format this with the hostIPv4?

e2etest/pkg/frr/container/container.go
fedepaol: Not sure I got it right. containerConfig is a wrapper to nc / rc (and now cc).
I initially thought to move it here (and have `Start` accepting it, as you said), but I wasn't sure why it wasn't done in the first place, so I kept it consistent.
I'll move it.

oribon:
LGTM

PR #1123, opened 2021-12-13
E2E: dump frr containers in pods only if running in frr mode

fedepaol:
When a test fails, trying to dump the frr information in native mode will result in an error. Here we check the len of containers and skip if there's only one container, assuming MetalLB is running in legacy mode.

### DIFF
--- a/e2etest/bgptests/bgp.go
+++ b/e2etest/bgptests/bgp.go
@@ -84,6 +84,9 @@ var _ = ginkgo.Describe("BGP", func() {
 			speakerPods, err := metallb.SpeakerPods(cs)
 			framework.ExpectNoError(err)
 			for _, pod := range speakerPods {
+				if len(pod.Spec.Containers) == 1 { // we dump only in case of frr
+					break
+				}
 				podExec := executor.ForPod(pod.Namespace, pod.Name, "frr")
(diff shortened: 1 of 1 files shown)


e2etest/bgp_test.go
```diff
@@ -138,6 +138,9 @@ var _ = ginkgo.Describe("BGP", func() {
 
 			speakerPods := getSpeakerPods(cs)
 			for _, pod := range speakerPods {
+				if len(pod.Spec.Containers) == 1 { // we bump only in case of frr
```
oribon, line 141: nit: bump -> dump

e2etest/bgptests/bgp.go
```diff
@@ -67,6 +67,9 @@ var _ = ginkgo.Describe("BGP", func() {
 			speakerPods, err := metallb.SpeakerPods(cs)
 			framework.ExpectNoError(err)
 			for _, pod := range speakerPods {
+				if len(pod.Spec.Containers) == 1 { // we dump only in case of frr
+					continue
```
cgoncalves, line 71: Mixed BGP backends are not supported. We could exit (break) the for-loop on the first speaker pod if its number of containers is equal to 1 (native mode).

msherif1234:
dump instead of bump in PR subject too

PR #1124, opened 2021-12-14
Add an example configuration for metallb.universe.tf/allow-shared-ip

alphabet5:
I think an example of tagging could be useful in the documentation here. Specifically to clarify the "key" that is used in the annotation. If it's not useful or unnecessary feel free to delete this.

I also thought it might help clarify the following section on separate services for TCP/UDP.


website/content/usage/_index.md
```diff
@@ -174,6 +174,46 @@ There are two main reasons to colocate services in this fashion: to
 work around a Kubernetes limitation, and to work with limited IP
 addresses.
 
+Example configuration for two services you want to share the same ip address.
```
fedepaol, line 177:
```suggestion
Here is an example configuration of two services that share the same ip address:
```

fedepaol:
Thanks @alphabet5 , added a small nit, the addition makes sense.

fedepaol:
LGTM, thanks!

PR #1126, opened 2021-12-15
E2E tests: Fix MetalLB configmap exist check in BeforeSuite

sabinaaledort:
If we are running in operator mode the MetalLB configmap can be found in the cluster because the operator is managing it
and not the tests. In such case BeforeSuite shouldn't fail.

This change allows us running the E2E tests more than once in a cluster where MetalLB was deploying through operator.
With the previous validation method a second run of the tests was failing as the configmap exist.

### DIFF
--- a/e2etest/e2etest_suite_test.go
+++ b/e2etest/e2etest_suite_test.go
@@ -145,17 +145,19 @@ var _ = ginkgo.BeforeSuite(func() {
 	cs, err := framework.LoadClientset()
 	framework.ExpectNoError(err)
 
-	_, err = cs.CoreV1().ConfigMaps(testNameSpace).Get(context.TODO(), configMapName, metav1.GetOptions{})
-	framework.ExpectEqual(errors.IsNotFound(err), true)
+	if !useOperator {
+		_, err = cs.CoreV1().ConfigMaps(testNameSpace).Get(context.TODO(), configMapName, metav1.GetOptions{})
+		framework.ExpectEqual(errors.IsNotFound(err), true)
 
-	// Init empty MetalLB ConfigMap.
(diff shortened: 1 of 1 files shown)


e2etest/e2etest_suite_test.go
```diff
@@ -146,16 +146,19 @@ var _ = ginkgo.BeforeSuite(func() {
...
-
-	// Init empty MetalLB ConfigMap.
-	_, err = cs.CoreV1().ConfigMaps(testNameSpace).Create(context.TODO(), &v1.ConfigMap{
-		ObjectMeta: metav1.ObjectMeta{
-			Name:      configMapName,
-			Namespace: testNameSpace,
-		},
-	}, metav1.CreateOptions{})
-	framework.ExpectNoError(err)
+	// When running in operator mode the configmap can be found in the cluster which is a valid state.
```
fedepaol, line 149: Can we just check if we are running in operator mode and skip?

sabinaaledort: we need to create the configmap if that's the first run. it's not created in the deployment

fedepaol: Even if we are running in operator mode?
What I am suggesting here is to condition the notFoundCheck / configmap creation only when we are not running in operator mode.

sabinaaledort: yeah i get it, but if we deploy metallb with the operator, what would create the configmap?

fedepaol: The operator?

sabinaaledort: ok sorry i get it now

fedepaol:
LGTM

PR #1130, opened 2021-12-16
Check aggregation length against the wider cidr generated from range

fedepaol:
If we specify addresses in a pool by range (and not by cidr), the check
that the lenght of the cidr is bigger than the aggregation lenght is
performed against all the possible cidrs that range translates to, and
it is misleading. The check makes sense when we declare a cidr, because
we know we won that whole cidr, but in the other case we may be using a
subset of a cidr for other reasons, and the check here leads to false
negatives.

A scenario like the one in the test that is being added should not fail:
3.3.3.2-3.3.3.254 with aggregation length 26

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -424,21 +424,23 @@ func parseAddressPool(p addressPool, bgpCommunities map[string]uint32) (*Pool, e
 	if len(p.Addresses) == 0 {
 		return nil, errors.New("pool has no prefixes defined")
 	}
+
+	cidrsPerAddresses := map[string][]*net.IPNet{}
 	for _, cidr := range p.Addresses {
 		nets, err := ParseCIDR(cidr)
 		if err != nil {
 			return nil, fmt.Errorf("invalid CIDR %q in pool %q: %s", cidr, p.Name, err)
 		}
 		ret.CIDR = append(ret.CIDR, nets...)
+		cidrsPerAddresses[cidr] = nets
 	}
-
 	switch ret.Protocol {
 	case Layer2:
 		if len(p.BGPAdvertisements) > 0 {
 			return nil, errors.New("cannot have bgp-advertisements configuration element in a layer2 address pool")
 		}
 	case BGP:
-		ads, err := parseBGPAdvertisements(p.BGPAdvertisements, ret.CIDR, bgpCommunities)
+		ads, err := parseBGPAdvertisements(p.BGPAdvertisements, cidrsPerAddresses, bgpCommunities)
 		if err != nil {
 			return nil, fmt.Errorf("parsing BGP communities: %s", err)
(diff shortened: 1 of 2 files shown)


internal/config/config.go
```diff
@@ -424,21 +424,26 @@ func parseAddressPool(p addressPool, bgpCommunities map[string]uint32) (*Pool, e
 	if len(p.Addresses) == 0 {
 		return nil, errors.New("pool has no prefixes defined")
 	}
+
```
russellb, line 427: s/won/own/ in the commit msg

russellb (commented):
small typo. otherwise lgtm

msherif1234:
its very odd that we have check for one addresspool fmt but not the other, look like a kludge to me with either remove the check or find a better way to do it

internal/config/config.go
```diff
@@ -525,6 +530,7 @@ func parseBGPAdvertisements(ads []bgpAdvertisement, cidrs []*net.IPNet, communit
 			if cidr.IP.To4() == nil {
 				maxLength = ad.AggregationLengthV6
 			}
+
```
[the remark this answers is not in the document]
fedepaol: Because they are added all of them to the same slice

fedepaol:
> its very odd that we have check for one addresspool fmt but not the other, look like a kludge to me, we either remove the check or find a better way to do it

Yep, but I think adding the check when specifying a cidr has some value. Another thing we can do is, when we have multiple cidrs to filter against the largest one. It will be more correct, probably as much as clunky.

internal/config/config.go
oribon: Also lenght -> length

fedepaol: Fixed (it's one of those things I never get right)

internal/config/config.go
```diff
@@ -629,6 +637,20 @@ func cidrContainsCIDR(outer, inner *net.IPNet) bool {
...
 
+func lowestMask(cidrs []*net.IPNet) int {
+	if len(cidrs) == 0 {
+		return 0
+	}
+	lowest, _ := cidrs[0].Mask.Size()
+	for _, c := range cidrs {
+		s, _ := c.Mask.Size()
+		if lowest > s {
+			lowest, _ = c.Mask.Size()
```
oribon, line 648: Why not `lowest = s`?

fedepaol: `internal/config/config.go:648:11: assignment mismatch: 1 variable but c.Mask.Size returns 2 values`

fedepaol: Sorry, now I get it. You are right.

fedepaol: Fixed

oribon:
lgtm

PR #1133, opened 2021-12-20
Add helm lanes to github actions

fedepaol:
We have helm lanes on circleci, we want to make gh actions on par so we can shut down circleci.


oribon:
lgtm

PR #1134, opened 2021-12-20
Remove e2e jobs from circleci

fedepaol:
After the successful migration to gh actions, we can remove the e2e lanes from circleci, leaving only the promotion jobs.


oribon:
lgtm

PR #1138, opened 2021-12-21
CI: extend the kind logs with the deployment type

fedepaol:
Now that we have also the deployment as another dimension for lanes, we need to embed it into the names of the logs otherwise they get overwritten.


oribon:
lgtm

PR #1140, opened 2021-12-21
E2E: Fix crashinfo dump

oribon:
Running `ls` and piping to `head -n 1` will always(?)
result in a 0 exit code.
Refactoring to check if crashlogs exist and iterating
over them one by one.


fedepaol:
LGTM

PR #1141, opened 2021-12-21
E2E: Combine FRR dump errors

oribon:
This change will enable all dumping steps
to finish before returning from the func,
which will give us more logs in case some
step fails before the others.


fedepaol:
lgtm

PR #1142, opened 2021-12-21
FRR config: deny is added only when announcing an address

fedepaol:
If we have a peer but no advertisements, the deny route map is not added and a peer may inject routes from outside.


oribon:
lgtm

PR #1143, opened 2021-12-22
E2E: Verify configuration in FRR containers

### LINKED ISSUE #1047: FRR E2E Tests: inspect the frr configuration to see if it matches the desired one
Right now e2e tests handle metallb as a black box, going to inspect only the content of the external FRR containers.
However, there are some parameters (i.e. password, timers) that are not sent on the wire and thus not available for inspection on the other peer.

We unit tests the various segments, but we don't really know if a value we set in config ends up rendered correctly.
It would be nice to have a set of "FRR" tagged e2e tests that do that.

zeeke:
Fixes #1047 

Adding a single test to inspect a full BFD configuration. Multiple configurations tests are already in place in unit tests (see internal/bgb/frr package).

~~The test also connect to a BGP peer and publish a serivce, that it may be redundant.~~
Update: test checks frr configuration is correct without connecting to peers, as it result in a faster test run.

Signed-off-by: Andrea Panattoni <apanatto@redhat.com>

### DIFF
--- a/e2etest/pkg/service/create.go
+++ b/e2etest/pkg/service/create.go
@@ -3,10 +3,12 @@
 package service
 
 import (
+	"context"
 	"fmt"
 
 	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/kubernetes/test/e2e/framework"
@@ -34,6 +36,11 @@ func CreateWithBackend(cs clientset.Interface, namespace string, jigName string,
 	})
 	framework.ExpectNoError(err)
 	return svc, jig
+}
+
+func Delete(cs clientset.Interface, svc *corev1.Service) {
+	err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
+	framework.ExpectNoError(err)
 }
 
 func tweakServicePort(svc *v1.Service) {
--- a/e2etest/bgptests/bgp.go
+++ b/e2etest/bgptests/bgp.go
@@ -21,6 +21,7 @@ package bgptests
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"go.universe.tf/metallb/e2etest/pkg/config"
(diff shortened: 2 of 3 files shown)


fedepaol:
Can we also add a test that validates the following parameters on the bgp side:

```
	HoldTime      string         `yaml:"hold-time"`
	KeepaliveTime string         `yaml:"keepalive-time"`
	RouterID      string         `yaml:"router-id"`
```

I think they are not propagated.

e2etest/bgp_test.go
```diff
@@ -1103,6 +1126,11 @@ func createServiceWithBackend(cs clientset.Interface, namespace string, jigName
 	return svc, jig
 }
 
+func deleteService(cs clientset.Interface, svc *corev1.Service) {
```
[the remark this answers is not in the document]
zeeke: Sure

zeeke:
> Can we also add a test that validates the following parameters on the bgp side:
> 
> ```
> 	HoldTime      string         `yaml:"hold-time"`
> 	KeepaliveTime string         `yaml:"keepalive-time"`
> 	RouterID      string         `yaml:"router-id"`
> ```
> 
> I think they are not propagated.

As discussed offline, current FRR implementation does not support different RouterID for each peer.
Further investigation on how to address this difference with native BGP implementation is needed (eg. #1150).

e2etest/bgp_test.go
```diff
@@ -1151,8 +1152,82 @@ var _ = ginkgo.Describe("BGP", func() {
...
+
+			speakerPods := getSpeakerPods(cs)
+			for _, pod := range speakerPods {
+				podExecutor := executor.ForPod(pod.Namespace, pod.Name, "frr")
+
+				Eventually(func() string {
+					// We need to assert against the output of the command as a bare string, as
+					// there is no json version of the command.
+					cfgStr, err := podExecutor.Exec("vtysh", "-c", "show running-config")
+					framework.ExpectNoError(err)
```
oribon, line 1197: Can you return the err instead of failing the test? That way we won't fail if for some reason the Exec failed that would have worked given another iteration of the Eventually.
```
if err != nil {
	return err.Error()
}
```

e2etest/bgp_test.go
```diff
@@ -1151,8 +1152,82 @@ var _ = ginkgo.Describe("BGP", func() {
...
+				)
+			}
+		})
+	})
+
 })
 
+// substringCount creates a Gomega transform function that
+// counts the number of occurrences in the subject string.
+func substringCount(substr string) interface{} {
```
oribon, line 1225: Why does this return interface{}?

e2etest/bgp_test.go
zeeke: You are right! Going to fix it

e2etest/bgp_test.go
zeeke: It's what `gomega.WithTransform(...)` expects to receive. I'm not sure a signature like:

```
func substringCount(substr string) func(action string) int {
    ...
```

would be more readable or it can be reused someway in other point.
Do you spot any good point on this?

oribon: tbh not sure what's better (or even declaring it as an anonymous only where it's used). I guess it's fine if we keep it this way.

e2etest/bgp_test.go
```diff
@@ -1183,8 +1152,85 @@ var _ = ginkgo.Describe("BGP", func() {
...
+
+			configData := config.File{
+				Pools: []config.AddressPool{
+					{
+						Name:      "bgp-test",
+						Protocol:  config.BGP,
+						Addresses: []string{v4PoolAddresses},
+					},
+				},
+				Peers:       peersForContainers(frrContainers, "ipv4"),
```
oribon, line 1175: Will this work for ipv6? (we can also wait for the ci to tell us)

zeeke: Yes, the test should work because it does not make assertion on BGP connectivity with peers. This way it focus on the configuration rendering and it won't break for any other reasons. As a plus, it should be much quicker

e2etest/pkg/service/create.go
```diff
@@ -36,6 +38,11 @@ func CreateWithBackend(cs clientset.Interface, namespace string, jigName string,
 	return svc, jig
 }
 
+func DeleteService(cs clientset.Interface, svc *corev1.Service) {
```
oribon, line 41: Small nit: can you rename this `Delete` instead? We usually avoid repeating the package name in a func (calling service.DeleteService).

oribon (commented):
Other than than that, lgtm

fedepaol:
LGTM, thanks!

PR #1145, opened 2021-12-22
E2E: Quote gingko parameters

zeeke:
When focusing on a single test case, it's useful to specify it
by name and it often contains blank spaces, so quotes are needed.

Even if dots can be used instead of blank spaces as workaround, it can be very
tricky to understand where the problem is, as mentioned in [#1047](https://github.com/metallb/metallb/issues/1047#issuecomment-997974875)


fedepaol:
LGTM

PR #1149, opened 2021-12-23
E2E: provide empty ips for the containers when the ranges are passed

fedepaol:
When the services ipv4 / v6 ranges are injected from outside, we don't
have to reserve a specific ip to the containers, so we must not set the
"ips-for-containers*" parmaeters and let docker assign custom ips.


zeeke:
From a python code perspective, it looks good to me.

PR #1150, opened 2021-12-24
FRR mode: fail in unsupported scenarios

fedepaol:
Currently FRR does not allow to do two different things:

- Having different myAsns for different bgp peers
- Having different routerids for different peers

The result is misleading, because the rendered configuration will consume only the last added myasn (or routerid) of the list, resulting in a different behaviour from what the user is expecting from the api. Here, we reject the configuration in such cases, and we remove the routerid setting from the tests.

### DIFF
--- a/e2etest/pkg/config/parse.go
+++ b/e2etest/pkg/config/parse.go
@@ -27,7 +27,6 @@ type Peer struct {
 	SrcAddr       string         `yaml:"source-address,omitempty"`
 	Port          uint16         `yaml:"peer-port,omitempty"`
 	HoldTime      string         `yaml:"hold-time,omitempty"`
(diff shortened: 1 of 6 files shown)


msherif1234:
we are doing this via webhooks since CR is where we want to be why we add he additional effort to block those configs

fedepaol:
> we are doing this via webhooks since CR is where we want to be why we add he additional effort to block those configs ? I am sure I have myASN now router-id I had originally but been asked to remove from webhoook I could bring it back

Because not all the users of metallb use the operator.
Re: routerID, the thing is, we never noticed it wasn't working as expected, so at this point in time it's safer to block it.

oribon:
lgtm

cgoncalves:
LGTM

PR #1152, opened 2022-01-04
Update issue page with Weave NO_MASQ_LOCAL default

HeroCC:
`NO_MASQ_LOCAL` is enabled by default (https://github.com/weaveworks/weave/pull/3756) as of weave 2.7, so update the docs to reflect this change


russellb:
thanks!

PR #1160, opened 2022-01-10
Make the tests timeout longer

fedepaol:
Some CI runs failed because they hit the 40m limit (especially dual stack tests), here we make the timeout longer.


russellb:
lgtm

PR #1161, opened 2022-01-11
Refactor uint32ptr / boolptr in a package

fedepaol:
Many implementations of the same function were scattered across different test packages. Here we collect all of them in a package and use it.

### DIFF
--- a/internal/pointer/pointer.go
+++ b/internal/pointer/pointer.go
@@ -1,0 +1,23 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package pointer
+
+func Uint32Ptr(n uint32) *uint32 {
+	return &n
+}
+
+func Int32Ptr(n int32) *int32 {
+	return &n
+}
+
+func BoolPtr(b bool) *bool {
+	return &b
+}
+
+func StrPtr(s string) *string {
+	return &s
+}
+
+func IntPtr(i int) *int {
+	return &i
+}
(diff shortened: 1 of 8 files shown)


msherif1234:
LGTM

liornoy:
great PR for a more organized code base, I myself needed int64Ptr and will use this pointer package for it.

internal/pointer/pointer.go
```diff
@@ -0,0 +1,15 @@
...
+	return &n
+}
+
+func BoolPtr(b bool) *bool {
+	return &b
+}
+
+func StrPtr(s string) *string {
+	return &s
+}
```
msherif1234, line 15: is it safer to allocate via new() and assign to passed in value instead of returning passed in var ptr ? in the current use where its read only I think should be ok but if we thing about using this pkg, but be better the suggested way what do u think ?

fedepaol: That's the same. Golang escape analysis will put the variable on the heap, so the behaviour will be identical (and I think this is way more readable)

msherif1234: ok other language won't do this magic "c" so if go can do that is great

fedepaol: Yeah with C this would explode.

oribon:
Was this left out of the package on purpose? https://github.com/metallb/metallb/blob/35bbb235d0e7f176856166dc16c997a09367b62b/internal/config/validate_test.go#L233

oribon:
lgtm

PR #1164, opened 2022-01-12
Remove namespace from PodSecurityPolicy definitions

spdfnet:
Remove namespace from PodSecurityPolicy definitions

PSP are cluster resources and the _namespace_ definition is ignored by the API anyway.

### DIFF
--- a/manifests/metallb-frr.yaml
+++ b/manifests/metallb-frr.yaml
@@ -4,7 +4,6 @@ metadata:
   labels:
     app: metallb
   name: controller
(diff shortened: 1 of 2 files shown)


fedepaol:
Hi @spdfnet , thanks for the PR! Mind checking also the helm charts, if they suffer from the same issue?

spdfnet:
Hello there, the helm charts do not have these ✓

fedepaol:
thanks for checking, LGTM

PR #1165, opened 2022-01-13
use home dir for frr volume and replace get_service_range method

### LINKED ISSUE #1162: dev-env: bgp config is not loaded into BGPd process
While trying to bringup metallb environment using [dev-env](https://github.com/metallb/metallb/tree/main/dev-env) with `inv dev-env -p bgp -b frr -i ipv4` command, it brings up metallb controller, speaker pods and frr router container, but frr container fails to load default `/etc/frr/bgpd.conf `file content mounted by tasks.py. because of this none of the bgp speakers are able to peer with frr router and connect request fails with `getsockopt: connection refused` error.

To workaround this we may have to either manually load bgpd.conf content via vtysh conf command or run this logic as similar to this [one](https://github.com/metallb/metallb/blob/main/e2etest/pkg/frr/container/container.go#L313-L329) done for e2e tests. 

The `inv dev-env --protocol layer2` command also fails with below error and it might need tasks.py script update.
```
Configuring MetalLB with a layer 2 test environment
Traceback (most recent call last):
  File "/usr/local/bin/inv", line 8, in <module>
    sys.exit(program.run())
  File "/usr/local/lib/python3.9/site-packages/invoke/program.py", line 384, in run
    self.execute()
  File "/usr/local/lib/python3.9/site-packages/invoke/program.py", line 566, in execute
    executor.execute(*self.tasks)
  File "/usr/local/lib/python3.9/site-packages/invoke/executor.py", line 129, in execute
    result = call.task(*args, **call.kwargs)
  File "/usr/local/lib/python3.9/site-packages/invoke/tasks.py", line 127, in __call__
    result = self.body(*args, **kwargs)
  File "/mnt/metallb/tasks.py", line 361, in dev_env
    layer2_dev_env()
  File "/mnt/metallb/tasks.py", line 373, in layer2_dev_env
    "SERVICE_V4_RANGE", get_service_range(4))
NameError: name 'get_service_range' is not defined
```

pperiyasamy:
The PR attempts to fix #1162.

This uses home directory for frr volume mount because nested volume mount doesn't work on a VM running in virtual box. 
It also removes usage of get_service_range method.


tasks.py
```diff
@@ -400,7 +400,7 @@ def bgp_dev_env(ip_family):
     try:
         # sudo because past docker runs will have changed ownership of this dir
         run('sudo rm -rf "%s"' % frr_volume_dir)
-        os.mkdir(frr_volume_dir)
+        os.makedirs(frr_volume_dir, exist_ok=True)
```
fedepaol, line 403: Why we need this change?

tasks.py
```diff
@@ -383,7 +383,7 @@ def layer2_dev_env():
 # See dev-env/bgp/README.md for some more information.
 def bgp_dev_env(ip_family):
     dev_env_dir = os.getcwd() + "/dev-env/bgp"
-    frr_volume_dir = dev_env_dir + "/frr-volume"
+    frr_volume_dir = os.path.expanduser('~') + "/.metallb/frr-volume"
```
fedepaol, line 386: I'd rather provide an alternative location via a parameter, than dropping the volume on an arbitrary location outside of the checkout

fedepaol:
I'd split the changes in two commits, as they belong to separate things.
Also, I'd extend the commits following [this guide](https://chris.beams.io/posts/git-commit/) we refer in the [contributor's guidelines](https://metallb.universe.tf/community/#commit-messages)

tasks.py
pperiyasamy: ok, sure, will take alternative location via a parameter and by default let it use the current checkout directory.

tasks.py
pperiyasamy: I've used this to create nested directory and this may not be needed if its set via parameter.

pperiyasamy:
> I'd split the changes in two commits, as they belong to separate things. Also, I'd extend the commits following [this guide](https://chris.beams.io/posts/git-commit/) we refer in the [contributor's guidelines](https://metallb.universe.tf/community/#commit-messages)

@fedepaol Thanks for the info, will split this into two commits and follow the commit msg guidelines.

tasks.py
pperiyasamy: this change is reverted now.

tasks.py
```diff
@@ -264,12 +264,15 @@ def validate_kind_version():
...
                 "Supported: 'native' (default), 'frr'",
+    "frr_volume_dir": "FRR router config directory to be mounted inside frr container. "
+                      "Default: ./dev-env/bgp/frr-volume",
     "log_level": "Log level for the controller and the speaker."
                 "Default: info, Supported: 'all', 'debug', 'info', 'warn', 'error' or 'none'",
     "helm_install": "Optional install MetalLB via helm chart instead of manifests."
                 "Default: False."
 })
 def dev_env(ctx, architecture="amd64", name="kind", protocol=None,
+        frr_volume_dir=os.getcwd() + "/dev-env/bgp/frr-volume",
```
fedepaol, line 275: I would do the other way around, interpreting this as "override" and leaving it as "" by default, and checking below if the override is "" or not. 
This is because here you make assumption about `dev_env_dir = os.getcwd() + "/dev-env/bgp"` , which we don't know if it may change in the future

pperiyasamy: done.

fedepaol:
LGTM, waiting for CI

pperiyasamy:
Thanks @fedepaol !

PR #1166, opened 2022-01-13
Controller: enqueue the failed configmap till services update

msherif1234:
When edit configmap for existing service, the config was rejected and
not re-enqueued for reprocess when the service update takes place,
causing the service to stay in pending state till it get deleted
by k8s.


fedepaol:
LGTM, waiting for CI

PR #1167, opened 2022-01-14
FRR mode: allow setting routerid if all equal

fedepaol:
To overcome the current frr limitation, instead of failing if
routerid is set, we enable a more permissive logic to allow
setting routerid only if all the routerids are equal.

### DIFF
--- a/e2etest/bgptests/bgp.go
+++ b/e2etest/bgptests/bgp.go
@@ -414,6 +414,31 @@ var _ = ginkgo.Describe("BGP", func() {
 			),
 		)
 	})
+	table.DescribeTable("configure peers with routerid and validate external containers are paired with nodes", func(ipFamily ipfamily.Family) {
+		ginkgo.By("configure peer")
+
+		configData := config.File{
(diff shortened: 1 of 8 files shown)


fedepaol:
cc @msherif1234

msherif1234:
Can u pls add the e2e parts too and just set one constant value instead of two so all the hooks will be in place for future use and e2e.

fedepaol:
> Can u pls add the e2e parts too and just set one constant value instead of two so all the hooks will be in place for future use and e2e.

Not sure what you mean by constant value.

Good point about RouterID, I'll add a test that sets the RouterID to be the same and check if that gets propagated.

msherif1234:
> > Can u pls add the e2e parts too and just set one constant value instead of two so all the hooks will be in place for future use and e2e.
> 
> Not sure what you mean by constant value.
> 
> Good point about RouterID, I'll add a test that sets the RouterID to be the same and check if that gets propagated.
what I meant is what u are doing adding the same rtrid for both peers so ur verification as well as the webhook CI will pass

PR #1169, opened 2022-01-16
Speaker: Fix: Remove items from work queue after sync 

sabinaaledort:
After doing a successful sync of an item it should be removed from the work queue.

Doing this mainly for deleted services. So far, after deleting a service the service item was left in the queue
(because "SyncStateSuccess" is not doing anything) and once created again (for example in the E2E tests), each update
was taking longer and longer time based on the exponential rate limiter.

### DIFF
--- a/internal/k8s/k8s.go
+++ b/internal/k8s/k8s.go
@@ -443,8 +443,7 @@ func (c *Client) Run(stopCh <-chan struct{}) error {
 		updates.Inc()
 		st := c.sync(key)
 		switch st {
-		case SyncStateSuccess:
-		case SyncStateErrorNoRetry:
+		case SyncStateErrorNoRetry, SyncStateSuccess:
 			c.queue.Forget(key)
 		case SyncStateError:
 			updateErrors.Inc()


internal/k8s/k8s.go
```diff
@@ -444,7 +446,7 @@ func (c *Client) Run(stopCh <-chan struct{}) error {
 		st := c.sync(key)
 		switch st {
 		case SyncStateSuccess:
-		case SyncStateErrorNoRetry:
+		case SyncStateErrorNoRetry, SyncStateDeleteSuccess:
```
fedepaol, line 449: The previous version was returning SyncStateSuccess, so I'd expect `c.queue.Forget` to be called even in that scenario.
Can you explain why this version works and the previous one wasn't?

fedepaol: Ok, cleared offline on the fact that go has no cascading cases.
Now the question is, why do we want this only for successful deletes and not for all `SyncStateSuccess` ?

I think it's fair to say that if we process an element successfully, we can forget about it. Also, with fallthrough you can enable cascading cases: https://stackoverflow.com/questions/40821855/do-go-switch-cases-fallthrough-or-not

internal/k8s/k8s.go
```diff
@@ -443,8 +443,7 @@ func (c *Client) Run(stopCh <-chan struct{}) error {
 		updates.Inc()
 		st := c.sync(key)
 		switch st {
-		case SyncStateSuccess:
-		case SyncStateErrorNoRetry:
+		case SyncStateErrorNoRetry, SyncStateSuccess:
```
fedepaol, line 446: typo in the commit title: itmes

fedepaol:
LGTM

PR #1172, opened 2022-01-18
FRR: fix communities and local preferences

fedepaol:
The current communities / local pref handling suffers of two issues:

- the route-map is added always with the same id, basically overriding
  all the communities and leaving the last one as the only available
- the route-map is not filtered per prefix. MetalLB api bounds the community and the local preference to the address pool, which means that they must be a property of the announcement.

Here we make a change in the template so that every route-map entry with the same name gets a different id, with the extra complication that we can't bind a route-map entry to a given ip, so it's necessary to collect the ips related to a community (or to a local pref) under a prefix-list.

E2E tests are modified to validate that only the ips coming from the right pool are advertised with the given community / local preference.

This is how the config starting from pools look like before the fix:

```
 address-pools:
    - protocol: bgp
      name: bgp-with-community
      addresses:
      - 192.168.10.0/24
      bgp-advertisements:
      - communities:
        - 65535:65282
    - protocol: bgp
      name: bgp-with-no-community
      addresses:
      - 192.168.16.0/24
    bfd-profiles: []
```

```
Current configuration:
!
frr version 7.5.1_git
frr defaults traditional
hostname kind-worker
log timestamp precision 3
no ipv6 forwarding
!
router bgp 64512
 no bgp ebgp-requires-policy
 no bgp default ipv4-unicast
 no bgp network import-check
 neighbor 172.18.0.6 remote-as 64513
 neighbor 172.18.0.6 timers 60 180
 neighbor 172.18.0.7 remote-as 64512
 neighbor 172.18.0.7 password ibgp-test
 neighbor 172.18.0.7 timers 120 360
 neighbor 172.30.0.2 remote-as 64513
 neighbor 172.30.0.2 password ebgp-test
 neighbor 172.30.0.2 port 180
 neighbor 172.30.0.2 ebgp-multihop 255
 neighbor 172.30.0.2 timers 30 90
 neighbor 172.30.0.3 remote-as 64512
 neighbor 172.30.0.3 password ibgp-test
 neighbor 172.30.0.3 port 180
 neighbor 172.30.0.3 timers 180 540
 !
 address-family ipv4 unicast
  network 192.168.10.0/32
  network 192.168.10.1/32
  network 192.168.16.0/32
  neighbor 172.18.0.6 activate
  neighbor 172.18.0.6 route-map 172.18.0.6-in in
  neighbor 172.18.0.6 route-map 172.18.0.6-out out
  neighbor 172.18.0.7 activate
  neighbor 172.18.0.7 route-map 172.18.0.7-in in
  neighbor 172.18.0.7 route-map 172.18.0.7-out out
  neighbor 172.30.0.2 activate
  neighbor 172.30.0.2 route-map 172.30.0.2-in in
  neighbor 172.30.0.2 route-map 172.30.0.2-out out
  neighbor 172.30.0.3 activate
  neighbor 172.30.0.3 route-map 172.30.0.3-in in
  neighbor 172.30.0.3 route-map 172.30.0.3-out out
 exit-address-family
!
route-map 172.18.0.7-in deny 20
!
route-map 172.30.0.3-in deny 20
!
route-map 172.18.0.6-in deny 20
!
route-map 172.30.0.2-in deny 20
!
route-map 172.18.0.7-out permit 10
 set community no-advertise additive
!
route-map 172.30.0.3-out permit 10
 set community no-advertise additive
!
route-map 172.18.0.6-out permit 10
 set community no-advertise additive
!
route-map 172.30.0.2-out permit 10
 set community no-advertise additive
!
ip nht resolve-via-default
!
ipv6 nht resolve-via-default
!
line vty
!
end
```

### DIFF
--- a/e2etest/bgptests/bgp.go
+++ b/e2etest/bgptests/bgp.go
@@ -841,10 +841,31 @@ var _ = ginkgo.Describe("BGP", func() {
 			table.Entry("IPV6", ipfamily.IPv6))
 
 		table.DescribeTable("configure bgp community and verify it gets propagated",
-			func(addressPools []config.AddressPool, ipFamily ipfamily.Family) {
+			func(rangeWithCommunity string, rangeWithoutCommunity string, ipFamily ipfamily.Family) {
 				configData := config.File{
 					Peers: metallb.PeersForContainers(FRRContainers, ipFamily),
-					Pools: addressPools,
+					Pools: []config.AddressPool{
+						{
+							Name:     "bgp-with-community",
+							Protocol: config.BGP,
+							Addresses: []string{
+								rangeWithCommunity,
+							},
+							BGPAdvertisements: []config.BgpAdvertisement{
+								{
+									Communities: []string{
+										CommunityNoAdv,
+									},
+								},
+							},
+						},
+						{
+							Name:     "bgp-with-no-community",
+							Protocol: config.BGP,
+							Addresses: []string{
+								rangeWithoutCommunity,
+							},
+						}},
 				}
 				for _, c := range FRRContainers {
 					err := frrcontainer.PairWithNodes(cs, c, ipFamily)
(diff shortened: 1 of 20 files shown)


msherif1234:
hmm what is the point of those multiple set in the route-map won't the result be the final set with something like localpref even for the commuity since all set the same thing the last will overwrite ?

fedepaol:
> hmm what is the point of those multiple set in the route-map won't the result be the final set with something like localpref even for the commuity since all set the same thing the last will overwrite ? for commuity code be one rule with route-map RMAP0:2222 permit 1 set community 0:2222 0:3333 0:44444 additive set local-preference 100 200 300
> 

Your example is how it was rendered before this change.

There are two main differences:
- the usage of a different id for each entry, which will avoid the various route maps squashing
- the addition of a match section on the rules

Each rule is named after the neighbor, and we add entries matching the IPs depending on the address pool the ip belongs to.
Same for local pref. 
So they are gonna be applied only if the ip is in the prefix list, which makes the whole thing work as intended. We have different prefix lists for different communities and for different local preferences.

Probably https://github.com/metallb/metallb/pull/1172/files#diff-ca8f31c1976ff57d31792d0304594932a61322e8328de154fcefb10f8cde74a3 provides a good example on how the template is now rendered.

> not following the use of prefixlist here if route-map is applied to specific nbr

Because [FRR](http://docs.frrouting.org/en/latest/routemap.html?highlight=route-map#clicmd-match-ip-address-prefix-list-PREFIX_LIST) do not allow to set a route-map entry to a specific prefix (or I couldn't find how to do so). It only allows to do so with prefix lists. What's more, you need to have a v4 and a v6 prefix list because the match condition requires that.

Hope it's clearer.

msherif1234:
is it only me who see some redundancy in route-map rules for example u pointed to in the UT
```
route-map 10.2.2.254-out permit 1
  match ip address prefix-list 1111:2222-v4prefixes
  set community 1111:2222 additive
```

and

```
route-map 10.2.2.254-out permit 5
  match ip address prefix-list 1111:2222-v4prefixes
  set community 1111:2222 additive
```

fedepaol:
> 

This may be a bug in the implementation I did not notice. Hope the general logic makes sense to you. Does it?

fedepaol:
Looks much better now, thanks @msherif1234 for spotting that.

internal/bgp/frr/config.go
```diff
@@ -33,20 +33,53 @@ hostname {{.Hostname}}
...
+{{- end}}
+{{- range $localPref, $prefixes := .PrefixesV4ForLocalPref }}
+{{- range $p := $prefixes.Elements }}
+ip prefix-list {{$localPref}}-v4localpref-prefixes permit {{$p}}
+{{- end}}
+{{- end}}
+{{- range $localPref, $prefixes := .PrefixesV6ForLocalPref }}
+{{- range $p := $prefixes.Elements }}
+ip prefix-list {{$localPref}}-v6localpref-prefixes permit {{$p}}
+{{- end}}
```
msherif1234, line 54: we are creating multiple identical prefixlist one for local and another with community why not create on prefix list and use it for both rules u can have multiple route-map rules match against the same match criteria

internal/bgp/frr/config.go
```diff
@@ -33,20 +33,53 @@ hostname {{.Hostname}}
...
 {{- end }}
+{{- range $community, $prefixes := $.PrefixesV6ForCommunity }}
+route-map {{$n.Addr}}-out permit {{ counter $n.Addr }}
+  match ipv6 address prefix-list {{$community}}-v6prefixes
+  set community {{$community}} additive
 {{- end }}
-route-map {{$n.Addr}}-in deny 20
+route-map {{$n.Addr}}-out permit {{ counter $n.Addr }}
 {{- end }}
 {{- end }}
```
msherif1234, line 85: I knew from day 1 this template will hard to maintain and indeed its getting there now and we still have more config to comes,  I proposed breaking this to sub templates do u think its time to refactor this I have some code for that but its way out of date should we consider this for this release or wait to the next release ?

fedepaol: I agree, I am fine with both. We have enough coverage now that should make us confident enough in doing the refactor.
Maybe I'd try to reach a stable point, cut a release and then refactor.

internal/bgp/frr/config.go
fedepaol: That's because you may have one addresspool with a localPref and one addresspool with community (and one with both).
So the lists must be separate.

See for example https://github.com/metallb/metallb/pull/1172/files#diff-ca8f31c1976ff57d31792d0304594932a61322e8328de154fcefb10f8cde74a3R7 , where one advertisement has the community and the other one has both the community and the local pref:

```
ip prefix-list 1111:2222-v4prefixes permit 172.16.1.10/24
ip prefix-list 1111:2222-v4prefixes permit 172.16.1.11/24
ip prefix-list 2-v4localpref-prefixes permit 172.16.1.11/24
```

The .11 goes to both lists, but the 10 goes only to the community one.

msherif1234: the way I see this we have prefix lst for every prefix we adv
```
ip prefix-list 172.16.1.10-v4prefixes permit 172.16.1.10/24
ip prefix-list 172.16.1.11-v4prefixes permit 172.16.1.11/24
```
then
```
route-map ..... permit 1
match 172.16.1.10-v4prefixes 
if (communit) 
  set community ...
if (locPref)
  set localPref ...  
 ```

Comment (fedepaol) [reply internal/bgp/frr/config.go:54]: We have Advertisements under neighbors, so that will result in duplicating the access rules again, unless drastically changing the template.

Even so, I see both solutions are equivalent. In one case we have multiple prefixes (one for each address), in the other we have multiple rules, one for each community / local pref.

We could do an extra optimization and squash prefixes that are the same, but that would make things even more complex and not sure it's worth the effort.

Comment (msherif1234) [reply internal/bgp/frr/config.go:54]: ok lets hold on any optimization for now till the 1st release out then we can refactor this further

Comment (msherif1234) [reply internal/bgp/frr/config.go:54]: at some point in the future we should refrain from adding those route-map under the cover and have CR for the user to add route-map rules directly

File: e2etest/bgptests/bgp.go [hunk 3]
```diff
@@ -907,57 +936,87 @@ var _ = ginkgo.Describe("BGP", func() {
...
+					},
+				}},
+				"fc00:f853:ccd:e799::1",
+				"fc00:f853:ccd:e799::2",
+				"fc00:f853:ccd:e799::20",
+				ipfamily.IPv6))
 
 		table.DescribeTable("configure bgp local-preference and verify it gets propagated",
-			func(poolAddresses []string, ipFamily ipfamily.Family, localPref uint32) {
+			func(addressesWithPref, addressesNoPref, ipWithLocalPref, ipNoLocalPref string, family ipfamily.Family, localPref uint32) {
```
oribon, line 952: Would it add value having 2 ips with LocalPref like you did with the communities?

internal/bgp/frr/config.go
```diff
@@ -289,3 +339,26 @@ func debouncer(body func(config *frrConfig),
...
+
+func newSet() Set {
+	return map[string]struct{}{}
+}
+
+func (s Set) Add(item string) {
+	s[item] = struct{}{}
+}
+
+func (s Set) Has(item string) bool {
```
oribon, line 353: I think this isn't used as well.

internal/bgp/frr/frr.go
```diff
@@ -261,23 +267,69 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
...
+
 			neighbor.Advertisements = append(neighbor.Advertisements, &advConfig)
 		}
 	}
 
 	return config, nil
 }
 
+func addPrefixForLocalPref(config *frrConfig, ipFamily, prefix string, localPref uint32) {
+	if ipFamily == "ipv4" {
```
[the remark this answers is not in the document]
fedepaol: I was torn about this, because here Version is also what is gonna be rendered in the template, which only incidentally matches with the constant associated to ipfamily.IPv4 / v6, but I think I can accomodate the change.

internal/bgp/frr/frr.go
```diff
@@ -261,23 +267,69 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
 			} else if adv.Prefix.IP.To16() != nil {
 				version = "ipv6"
 			}
+			v4CommunityRuleIDs := map[string]int{}
+			v6CommunityRuleIDs := map[string]int{}
```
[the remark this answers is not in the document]
fedepaol: Yep they are residual from the initial version where instead of having a counter I had ruleIds mapped to communities.

e2etest/bgptests/bgp.go
fedepaol: I just wanted to ensure the mechanism did not break with multiple rules, but this won't hurt and it won't change the tests running time.

e2etest/bgptests/bgp.go
```diff
@@ -858,29 +858,47 @@ var _ = ginkgo.Describe("BGP", func() {
...
 				framework.ExpectNoError(err)
 
 				for _, c := range FRRContainers {
-					validateService(cs, svc, allNodes.Items, c)
-					Eventually(func() error {
-						return frr.ContainsCommunity(c, "no-advertise")
-					}, 4*time.Minute, 1*time.Second).Should(BeNil())
+					validateService(cs, svcCommunity, allNodes.Items, c)
+					validateService(cs, svcNoCommunity, allNodes.Items, c)
+					routes, err := frr.RoutesForCommunity(c, "no-advertise", ipFamily)
```
sabinaaledort, line 886: missing validate for svcCommunity1

e2etest/bgptests/bgp.go
```diff
@@ -858,29 +858,47 @@ var _ = ginkgo.Describe("BGP", func() {
...
+					framework.ExpectNoError(err)
+					if _, ok := routes[ipNoCommunity]; ok {
+						framework.Failf("found %s route for community no-advertise", ipNoCommunity)
+					}
+					if _, ok := routes[ipWithCommunity1]; !ok {
+						framework.Failf("%s route not found for community no-advertise", ipWithCommunity)
```
sabinaaledort, line 892: ipWithCommunity -> ipWithCommunity1

cgoncalves:
lgtm

The e2e (dual, frr, helm) CI run failed on a L2 test, unrelated to this change.

oribon:
nice! lgtm

PR #1174, opened 2022-01-19
Controller: custom annotation for pinning ips (single and dual stack)

### LINKED ISSUE #1155: Custom annotation for pinning ips (single and dual stack)
The loadbalancerip field is gonna be deprecated (see https://github.com/kubernetes/kubernetes/pull/107235)

That has two implications:
- we need to have an alternative way to pin to a specific ip (the suggested way is to implement custom annotations on the service, see https://github.com/kubernetes/enhancements/pull/1992)
- that will enable us to pin both v6 and v4 addresses to the service (currently, only one is possible)

So, this issue is about having a custom annotation handled by metallb, that declares the user's will to use a specific ip (or ips).
See also https://github.com/kubernetes/enhancements/pull/1992#issuecomment-1000113208 for reference.

pperiyasamy:
This PR aims to fix #1155 issue by adding new custom annotation field `metallb.universe.tf/loadBalancerIPs `which can be used to request specific IPs (single and dual stack). For now this would coexist with `svc.Spec.LoadBalancerIP `field. when both fieds are specified, then annotation field takes higher precedence than spec field.

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -211,12 +211,88 @@ func TestControllerMutation(t *testing.T) {
 		},
 
 		{
+			desc: "request specific IP via custom annotation",
+			in: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						annotationLoadBalancerIPs: "1.2.3.1",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					ClusterIPs: []string{"1.2.3.4"},
+					Type:       "LoadBalancer",
+				},
+			},
+			want: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						annotationLoadBalancerIPs: "1.2.3.1",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					ClusterIPs: []string{"1.2.3.4"},
+					Type:       "LoadBalancer",
+				},
+				Status: statusAssigned([]string{"1.2.3.1"}),
+			},
+		},
+
+		{
+			desc: "request IP from both svc.spec.LoadBalancerIP and custom annotation",
+			in: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						annotationLoadBalancerIPs: "1.2.3.1",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					ClusterIPs:     []string{"1.2.3.4"},
+					Type:           "LoadBalancer",
+					LoadBalancerIP: "1.2.3.2",
+				},
+			},
+			wantErr: true,
+		},
+
+		{
 			desc: "request invalid IP",
 			in: &v1.Service{
 				Spec: v1.ServiceSpec{
 					Type:           "LoadBalancer",
 					LoadBalancerIP: "please sir may I have an IP address thank you",
 					ClusterIPs:     []string{"1.2.3.4"},
+				},
+			},
+			wantErr: true,
+		},
+
+		{
+			desc: "request invalid IP via custom annotation",
+			in: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						annotationLoadBalancerIPs: "please sir may I have an IP address thank you",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					ClusterIPs: []string{"1.2.3.4"},
+					Type:       "LoadBalancer",
+				},
+			},
+			wantErr: true,
+		},
+
+		{
+			desc: "request two IPs from same ip family via custom annotation",
+			in: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						annotationLoadBalancerIPs: "1.2.3.1,1.2.3.2",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					ClusterIPs: []string{"1.2.3.4"},
+					Type:       "LoadBalancer",
 				},
 			},
 			wantErr: true,
@@ -440,12 +516,44 @@ func TestControllerMutation(t *testing.T) {
 		},
 
 		{
+			desc: "request IP from wrong ip-family (ipv4) via custom annotation",
+			in: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						annotationLoadBalancerIPs: "1.2.3.1",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					ClusterIPs: []string{"3000::1"},
+					Type:       "LoadBalancer",
+				},
+			},
+			wantErr: true,
+		},
+
+		{
 			desc: "request IP from wrong ip-family (ipv6)",
 			in: &v1.Service{
 				Spec: v1.ServiceSpec{
 					Type:           "LoadBalancer",
 					LoadBalancerIP: "1000::",
 					ClusterIPs:     []string{"1.2.3.4"},
+				},
+			},
+			wantErr: true,
+		},
+
+		{
+			desc: "request IP from wrong ip-family (ipv6) via custom annotation",
+			in: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						annotationLoadBalancerIPs: "1000::",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					ClusterIPs: []string{"1.2.3.4"},
+					Type:       "LoadBalancer",
 				},
 			},
 			wantErr: true,
@@ -542,6 +650,47 @@ func TestControllerMutation(t *testing.T) {
 				},
 			},
 			wantErr: true,
+		},
+		{
+			desc: "request specific loadbalancer IPv4 address via custom annotation for dual-stack config",
+			in: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						annotationLoadBalancerIPs: "1.2.3.1",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					ClusterIPs: []string{"1.2.3.4", "3000::1"},
+					Type:       "LoadBalancer",
+				},
+			},
+			wantErr: true,
+		},
+		{
+			desc: "request dual-stack loadbalancer IPs via custom annotation",
+			in: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						annotationLoadBalancerIPs: "1.2.3.0,1000::",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					ClusterIPs: []string{"1.2.3.4", "3000::1"},
+					Type:       "LoadBalancer",
+				},
+			},
+			want: &v1.Service{
+				ObjectMeta: metav1.ObjectMeta{
+					Annotations: map[string]string{
+						annotationLoadBalancerIPs: "1.2.3.0,1000::",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4", "3000::1"},
+				},
+				Status: statusAssigned([]string{"1.2.3.0", "1000::"}),
+			},
 		},
 		{
 			desc: "request dual-stack loadbalancer with invalid ingress",
--- a/controller/service.go
+++ b/controller/service.go
@@ -17,6 +17,9 @@ package main
 import (
 	"fmt"
 	"net"
+	"reflect"
+	"sort"
+	"strings"
 
 	"github.com/go-kit/kit/log"
 	"github.com/go-kit/kit/log/level"
@@ -27,7 +30,8 @@ import (
 )
 
 const (
-	annotationAddressPool = "metallb.universe.tf/address-pool"
+	annotationAddressPool     = "metallb.universe.tf/address-pool"
+	annotationLoadBalancerIPs = "metallb.universe.tf/loadBalancerIPs"
 )
 
 func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service) bool {
@@ -60,12 +64,10 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 			lbIPs = append(lbIPs, net.ParseIP(ip))
 		}
 	}
-
-	lbIPsIPFamily := ipfamily.Unknown
 	if len(lbIPs) == 0 {
 		c.clearServiceState(key, svc)
 	} else {
-		lbIPsIPFamily, err = ipfamily.ForAddressesIPs(lbIPs)
+		lbIPsIPFamily, err := ipfamily.ForAddressesIPs(lbIPs)
 		if err != nil {
 			level.Error(l).Log("event", "clearAssignment", "reason", "nolbIPsIPFamily", "msg", "Failed to retrieve lbIPs family")
 			c.client.Errorf(svc, "nolbIPsIPFamily", "Failed to retrieve LBIPs IPFamily for %q: %s", lbIPs, err)
@@ -106,20 +108,19 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 			c.clearServiceState(key, svc)
 			lbIPs = []net.IP{}
 		}
-		// User set or changed the desired LB IP, nuke the
-		// state. allocateIP will pay attention to LoadBalancerIP and try
+		// User set or changed the desired LB IP(s), nuke the
+		// state. allocateIP will pay attention to LoadBalancerIP(s) and try
 		// to meet the user's demands.
-		if svc.Spec.LoadBalancerIP != "" {
-			if lbIPsIPFamily == ipfamily.DualStack {
-				level.Error(l).Log("event", "loadbalancerIP", "error", "DualStack", "msg", "user requested loadbalancer IP for dualstack address family")
-				c.client.Errorf(svc, "LoadBalancerFailed", "Failed loadbalancer IP for dual-stack")
-				return true
-			}
-			if len(lbIPs) != 0 && svc.Spec.LoadBalancerIP != lbIPs[0].String() {
-				level.Info(l).Log("event", "clearAssignment", "reason", "differentIPRequested", "msg", "user requested a different IP than the one currently assigned")
-				c.clearServiceState(key, svc)
-				lbIPs = []net.IP{}
-			}
+		desiredLbIPs, _, err := getDesiredLbIPs(svc)
+		if err != nil {
+			level.Error(l).Log("event", "loadbalancerIP", "error", err, "msg", "invalid requested loadbalancer IPs")
+			c.client.Errorf(svc, "LoadBalancerFailed", "invalid requested loadbalancer IPs: %s", err)
+			return true
+		}
+		if len(desiredLbIPs) > 0 && !isEqualIPs(lbIPs, desiredLbIPs) {
+			level.Info(l).Log("event", "clearAssignment", "reason", "differentIPRequested", "msg", "user requested a different IP than the one currently assigned")
+			c.clearServiceState(key, svc)
+			lbIPs = []net.IP{}
 		}
 	}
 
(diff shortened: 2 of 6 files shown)


controller/controller_test.go
```diff
@@ -543,6 +663,32 @@ func TestControllerMutation(t *testing.T) {
...
+						annotationLoadBalancerIPs: "1.2.3.0,1000::",
+					},
+				},
+				Spec: v1.ServiceSpec{
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4", "3000::1"},
+				},
+				Status: statusAssigned([]string{"1.2.3.0", "1000::"}),
+			},
+		},
```
msherif1234, line 691: what if u have only one IP and we are in dual-stack mode would that flag an error ? can we add test coverage for this case too ?

controller/service.go
```diff
@@ -214,3 +260,37 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error)
...
+		ip := net.ParseIP(desiredLbIP)
+		if ip == nil {
+			return nil, fmt.Errorf("invalid %s: %q", annotationLoadBalancerIPs, lbIPsStr)
+		}
+		lbIPs = append(lbIPs, ip)
+	}
+	return lbIPs, nil
+}
+
+func (c *controller) isLbIPsEqual(ipsA []net.IP, ipsB []net.IP) (bool, error) {
```
msherif1234, line 281: why not using reflect.DeepEqual ?

controller/service.go
```diff
@@ -121,6 +125,26 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
...
 		}
+		// annotationLoadBalancerIPs takes higher precedence over svc.Spec.LoadBalancerIP
+		desiredLbIPs, err := c.getDesiredLbIPs(svc)
+		if err != nil {
+			level.Error(l).Log("event", "loadbalancerIP", "error", err, "msg", "invalid requested loadbalancer IPs")
+			c.client.Errorf(svc, "LoadBalancerFailed", "invalid requested loadbalancer IPs: %s", err)
+			return true
+		}
+		if len(desiredLbIPs) > 0 {
+			ok, err := c.isLbIPsEqual(lbIPs, desiredLbIPs)
```
msherif1234, line 136: can u use reflect.DeepEqual() instead of this new custom API ?

msherif1234:
we probably need to add something here https://github.com/metallb/metallb/blob/main/website/content/usage/_index.md about this new annotation and that fact that loadbalancerIP will be deprecated

controller/service.go
```diff
@@ -109,6 +111,8 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 		// User set or changed the desired LB IP, nuke the
 		// state. allocateIP will pay attention to LoadBalancerIP and try
 		// to meet the user's demands.
+		// TODO: remove check for svc.Spec.LoadBalancerIP when this field is removed
+		// from service spec.
```
msherif1234, line 115: since the annotation takes precedence over loadbalancerIp then there is no need to do both in this case and overwrite with annotation u can do the annotation 1st and skip loadbalancer check if len(desiredLbIps) > 0

controller/controller_test.go
pperiyasamy: yes, it must throw an error as per the existing LoadBalancerIP unit test. added a test to cover that too.

controller/service.go
pperiyasamy: Good find ! changed it.

controller/service.go
pperiyasamy: at this point both lbIPs are unsorted, so couldn't use reflect.DeepEqual.

pperiyasamy:
> we probably need to add something here https://github.com/metallb/metallb/blob/main/website/content/usage/_index.md about this new annotation and that fact that loadbalancerIP will be deprecated

done. please check.

controller/service.go
```diff
@@ -106,10 +108,31 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
...
-		// state. allocateIP will pay attention to LoadBalancerIP and try
-		// to meet the user's demands.
-		if svc.Spec.LoadBalancerIP != "" {
+		// User set or changed the desired LB IP(s), nuke the
+		// state. allocateIP will pay attention to LoadBalancerIP(s) and try
+		// to meet the user's demands. annotationLoadBalancerIPs takes higher
+		// precedence over svc.Spec.LoadBalancerIP.
+		// TODO: remove check for svc.Spec.LoadBalancerIP when this field is removed
+		// from service spec.
+		desiredLbIPs, err := c.getDesiredLbIPs(svc)
```
fedepaol, line 117: Can you have the LoadBalancerIP logic embedded in `getDesiredLbIPs` ? I think it will remove duplication and in general it will express what `getDesiredIPs` means.
So if the user set the annotation use that, otherwise fallback to Status.LoadBalancerIP .

controller/service.go
```diff
@@ -106,10 +108,31 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
...
+		if err != nil {
```
fedepaol, line 118: What's missing here is a check on ipFamily. If the user puts the wrong ip family compared to the service's one, we must fail.
I'd suggest incorporating that check to have `getDesiredLbIPs` return a slice and always check ipFamily. That will work also with the loadbalancerip type.

controller/service.go
```diff
@@ -214,3 +257,37 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error)
...
 }
+
+func (c *controller) getDesiredLbIPs(svc *v1.Service) ([]net.IP, error) {
+	lbIPsStr := svc.Annotations[annotationLoadBalancerIPs]
+	if lbIPsStr == "" {
+		return nil, nil
+	}
+	var lbIPs []net.IP
+	desiredLbIPs := strings.Split(lbIPsStr, ",")
+	for _, desiredLbIP := range desiredLbIPs {
```
fedepaol, line 268: We must ensure that if the ips are two, one is v4 and the other is v6.
Also, we must fail if there are more than 2.

controller/service.go
```diff
@@ -214,3 +257,37 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error)
 	// Okay, in that case just bruteforce across all pools.
 	return c.ips.Allocate(key, serviceIPFamily, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc))
 }
+
+func (c *controller) getDesiredLbIPs(svc *v1.Service) ([]net.IP, error) {
```
fedepaol, line 261: No need to be a method of controller.
This can be a pure function with no state.

controller/service.go
```diff
@@ -214,3 +257,37 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error)
...
+}
+
+func (c *controller) isLbIPsEqual(ipsA []net.IP, ipsB []net.IP) (bool, error) {
+	if len(ipsA) > 2 || len(ipsB) > 2 {
+		return false, fmt.Errorf("unsupported number of load balancer IPs")
+	}
+	if len(ipsA) != len(ipsB) {
+		return false, nil
+	}
+	if len(ipsA) == 2 && (ipsA[0].String() == ipsB[0].String() || ipsA[0].String() == ipsB[1].String()) &&
```
fedepaol, line 285: Can you just sort? It will be easier to understand.

controller/service.go
```diff
@@ -214,3 +257,37 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error)
...
+		ip := net.ParseIP(desiredLbIP)
+		if ip == nil {
+			return nil, fmt.Errorf("invalid %s: %q", annotationLoadBalancerIPs, lbIPsStr)
+		}
+		lbIPs = append(lbIPs, ip)
+	}
+	return lbIPs, nil
+}
+
+func (c *controller) isLbIPsEqual(ipsA []net.IP, ipsB []net.IP) (bool, error) {
```
fedepaol, line 278: Same, no need to be a method.

website/content/usage/_index.md
```diff
@@ -20,6 +20,11 @@ address, or if the address is already in use by another service,
 assignment will fail and MetalLB will log a warning event visible in
 `kubectl describe service <service name>`.
 
+The `spec.loadBalancerIP` parameter is going to be deprecated soon, so
```
fedepaol, line 23: I'd reword this to:

"We both support `spec.loadBalancerIP` and a custom `metallb.universe.tf/loadBalancerIPs`.
`metallb.universe.tf/loadBalancerIPs` also supporst a comma separated list of IPs to be used in case of Dual Stack services.

Please note that `spec.LoadBalancerIP` is planned to be deprecated in [k8s apis](https://github.com/kubernetes/kubernetes/pull/107235).
"

controller/service.go
```diff
@@ -214,3 +257,37 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error)
...
+	if len(ipsA) > 2 || len(ipsB) > 2 {
```
fedepaol, line 279: This should be done in the `getDesiredLbIPs` function

controller/service.go
fedepaol: Also, rename as `ipsAreEqual` because that's what the function do.

controller/service.go
fedepaol, line 285: Also, you can use the `Equal` method of net.IP, no need to convert to string.

controller/service.go
```diff
@@ -185,8 +208,24 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error)
 		return nil, err
 	}
 
-	// If the user asked for a specific IP, try that.
-	if svc.Spec.LoadBalancerIP != "" {
+	var lbIPs []net.IP
+	// annotationLoadBalancerIPs takes higher precedence over svc.Spec.LoadBalancerIP
+	// TODO: remove svc.Spec.LoadBalancerIP check when this field is removed
```
fedepaol, line 213: Same, no need for TODO

controller/service.go
```diff
@@ -185,8 +208,24 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error)
...
+	// from service spec.
+	desiredLbIPs, err := c.getDesiredLbIPs(svc)
```
fedepaol, line 215: Same as above, incorporate `spec.LoadBalancerIP` in getDesiredLBIPs

controller/service.go
```diff
@@ -185,8 +208,24 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error)
...
+	if err != nil {
+		return nil, err
+	}
+	if len(desiredLbIPs) > 0 {
+		lbipFamily, err := ipfamily.ForAddressesIPs(desiredLbIPs)
```
fedepaol, line 220: nit: lbIPFamily. Also, rename to `desiredLbIPFamily`

fedepaol:
Left a couple of comments, I'd also add an e2e test to make sure this works on the happy path.

controller/service.go
```diff
@@ -106,10 +108,31 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
...
 		}
-		// User set or changed the desired LB IP, nuke the
-		// state. allocateIP will pay attention to LoadBalancerIP and try
-		// to meet the user's demands.
-		if svc.Spec.LoadBalancerIP != "" {
+		// User set or changed the desired LB IP(s), nuke the
+		// state. allocateIP will pay attention to LoadBalancerIP(s) and try
+		// to meet the user's demands. annotationLoadBalancerIPs takes higher
+		// precedence over svc.Spec.LoadBalancerIP.
+		// TODO: remove check for svc.Spec.LoadBalancerIP when this field is removed
```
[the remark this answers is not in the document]
pperiyasamy: yes, done

controller/service.go
pperiyasamy: Yes! this is an elegant approach. made the changes now.

controller/service.go
pperiyasamy: done

controller/service.go
pperiyasamy: done

controller/service.go
pperiyasamy: done

controller/service.go
pperiyasamy: yes, done.

controller/service.go
pperiyasamy: of course, done

controller/service.go
pperiyasamy: The `ipfamily.ForAddressesIPs` method call already does the check, so we may not need an explicit check.

controller/service.go
pperiyasamy: removed this method and using `sort` and `reflect.DeepEqual `now

controller/service.go
pperiyasamy: this is no more applicable

controller/service.go
pperiyasamy: done

pperiyasamy: ah! correct, anyway this is no more needed due to using `reflect.DeepEqual`

website/content/usage/_index.md
pperiyasamy: done

e2etest/bgptests/bgp.go
```diff
@@ -174,6 +174,11 @@ var _ = ginkgo.Describe("BGP", func() {
 				testservice.TrafficPolicyCluster(svc)
 				testservice.ForceV6(svc)
 			}),
+		table.Entry("IPV4 - ExternalTrafficPolicyCluster - request IPv4 via custom annotation", ipfamily.IPv4, "ExternalTrafficPolicyCluster", []string{v4PoolAddresses},
+			func(svc *corev1.Service) {
+				testservice.TrafficPolicyCluster(svc)
+				testservice.WithSpecificIP(svc, "192.168.10.100")
+			}),
```
msherif1234, line 181: pls add IPv6 and dual-stack coverage

msherif1234:
you need to organize your commits I woudl suggest two commits one for the controller and controller-unittest and another for e2e coverage with proper title and descriptions

website/content/usage/_index.md
```diff
@@ -20,6 +20,12 @@ address, or if the address is already in use by another service,
 assignment will fail and MetalLB will log a warning event visible in
 `kubectl describe service <service name>`.
 
+We both support `spec.loadBalancerIP` and a custom `metallb.universe.tf/loadBalancerIPs`
```
msherif1234, line 23: metalLB supports `spec.loadBalancerIP` and a custom `metallb.universe.tf/loadBalancerIPs`

e2etest/pkg/service/tweak.go
```diff
@@ -25,3 +25,9 @@ func DualStack(svc *corev1.Service) {
 func TrafficPolicyCluster(svc *corev1.Service) {
 	svc.Spec.ExternalTrafficPolicy = corev1.ServiceExternalTrafficPolicyTypeCluster
 }
+
+func WithSpecificIP(svc *corev1.Service, ip string) {
```
msherif1234, line 29: this can't be used to return multiple IPs to test dual-stack

e2etest/bgptests/bgp.go
pperiyasamy: done

e2etest/pkg/service/tweak.go
pperiyasamy: yes, changed it to var-args now.

website/content/usage/_index.md
pperiyasamy: done

msherif1234:
pls take out the doc update from the 1st commit and have in its own commit

pperiyasamy:
> pls take out the doc update from the 1st commit and have in its own commit

yes @msherif1234 , done.

e2etest/bgptests/bgp.go
```diff
@@ -174,6 +174,17 @@ var _ = ginkgo.Describe("BGP", func() {
...
+		table.Entry("IPV4 - ExternalTrafficPolicyCluster - request IPv4 via custom annotation", ipfamily.IPv4, "ExternalTrafficPolicyCluster", []string{v4PoolAddresses},
+			func(svc *corev1.Service) {
+				testservice.TrafficPolicyCluster(svc)
+				testservice.WithSpecificIPs(svc, "192.168.10.100")
+			}),
+		table.Entry("DUALSTACK - ExternalTrafficPolicyCluster - request Dual Stack via custom annotation", ipfamily.DualStack, "ExternalTrafficPolicyCluster", []string{v4PoolAddresses, v6PoolAddresses},
+			func(svc *corev1.Service) {
+				testservice.TrafficPolicyCluster(svc)
+				testservice.DualStack(svc)
+				testservice.WithSpecificIPs(svc, "192.168.10.100", "fc00:f853:0ccd:e799::100")
```
fedepaol, line 186: Here we don't have control if the test will actually receive the ips we are asking for.
We only rely on the existing tests, which make sure the ips assigned to the service will work.

e2etest/pkg/service/tweak.go
```diff
@@ -25,3 +29,19 @@ func DualStack(svc *corev1.Service) {
 func TrafficPolicyCluster(svc *corev1.Service) {
 	svc.Spec.ExternalTrafficPolicy = corev1.ServiceExternalTrafficPolicyTypeCluster
 }
+
+func WithSpecificIPs(svc *corev1.Service, ips ...string) {
+	if len(ips) == 0 {
+		return
+	}
+	var allIPs string
+	for _, ip := range ips {
```
fedepaol, line 38: you can use `strings.Join`

e2etest/pkg/service/tweak.go
```diff
@@ -25,3 +29,19 @@ func DualStack(svc *corev1.Service) {
...
+		allIPs += ip + ","
+	}
+	if len(allIPs) > 0 {
+		allIPs = strings.TrimRight(allIPs, ",")
+	}
+	svc.ObjectMeta.Annotations = map[string]string{
```
fedepaol, line 44: ObjectMeta is embedded, you can use svc.Annotations

pperiyasamy:
@fedepaol I've addressed most of your latest review comments, please have a look now.

e2etest/bgptests/bgp.go
```diff
@@ -126,6 +126,19 @@ var _ = ginkgo.Describe("BGP", func() {
 
 			svc, _ := testservice.CreateWithBackend(cs, f.Namespace.Name, "external-local-lb", tweak)
 			defer testservice.Delete(cs, svc)
+			desiredLbIPs := svc.Annotations["metallb.universe.tf/loadBalancerIPs"]
+			if desiredLbIPs != "" {
+				ingressPts := svc.Status.LoadBalancer.Ingress
+				framework.ExpectNotEqual(len(ingressPts), 0)
```
fedepaol, line 132: no need for this check as you are comparing below.

e2etest/bgptests/bgp.go
```diff
@@ -126,6 +126,19 @@ var _ = ginkgo.Describe("BGP", func() {
...
+				assignedIPs := strings.Join(func(ingresses []corev1.LoadBalancerIngress) []string {
```
fedepaol, line 133: This is a bit hard to follow. I'd just produce the []string out of svc.Status.LoadBalancer.Ingress and use it to feed strings.Join, or even better, do that in a `ValidateDesiredLB(svc)` func under the `service` package .

controller/service.go
```diff
@@ -185,21 +186,20 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error)
...
-			return nil, fmt.Errorf("requested spec.loadBalancerIP %q does not match the ipFamily of the service", svc.Spec.LoadBalancerIP)
+	desiredLbIPs, desiredLbIPFamily, err := getDesiredLbIPs(svc)
+	if err != nil {
+		return nil, err
+	}
+
+	// If the user asked for a specific IPs, try that.
+	if len(desiredLbIPs) > 0 {
+		if serviceIPFamily != desiredLbIPFamily {
+			return nil, fmt.Errorf("requested loadBalancer IP %q does not match the ipFamily of the service", desiredLbIPs)
```
fedepaol, line 197: nit: loadBalancer IP(s)

pperiyasamy: done

e2etest/bgptests/bgp.go
pperiyasamy: sure, removed nested function now and also factored out this whole logic into `ValidateDesiredLB` function in` validate.go`

e2etest/bgptests/bgp.go
pperiyasamy: yes, done.

fedepaol:
Looks great, thanks!

pperiyasamy:
Thanks @fedepaol and @msherif1234 for all the feedback !

PR #1176, opened 2022-01-20
E2E: Fix operator mode

oribon:
Calling framework.NewDefaultFramework() inside BeforeSuite results in:
```
You may only call BeforeEach from within a Describe, Context or When
```
Changing that to get the ClientConfig in a different way.


fedepaol:
LGTM

PR #1177, opened 2022-01-20
FRR: Provide better feedback for FRR reloading

### LINKED ISSUE #1046: Provide a better feedback in case of failed frr reloading
Today the real configuration reloading is lost behind a sighup call.
It would be ideal to provide a better feedback to the user.
Some ideas to understand if the configuration is failed are:
- some kind of versioning / timestamp between the desired and applied configuration. If they drift, there has been an error
- a sync call (more rich than sighup), that returns in case of failure (this was discarded originally, but we may want to take it back)

liornoy:
This PR introduces better feedback for FRR reloading.
The problem is described more at this [issue](https://github.com/metallb/metallb/issues/1046)
fix #1046 

More details about the implementation are in the commit comment.


fedepaol:
If I got this right, after a reload, the debouncer signals the reloader that waits an arbitrary time and checks if the file is newer than that. One point of concern is the amount of timed synchronization between the debouncer and the reloader, and the moving parts.

On example is, multiple updates, `lastTriggeredTimeStamp` changes but the timer is not set, effectively giving the reload procedure less time.

I'd try to keep the truth of a successful / unsuccessful reload as reliable as we can, and the reloader script is the one in control.
Something that comes to my mind is, we could have two files, one for successful reloads and one for failed reloads (or one file with the result).

Then, it would be just a matter of polling the file(s) (and checking if it changed from the last time we polled), and emitting the succeed / failed log and event. We could still miss events, if the polling interval is coarser than the reloading, but that is not an issue as we care about  the last state.

WDYT?

e2etest/bgptests/bgp.go
```diff
@@ -1081,6 +1083,46 @@ var _ = ginkgo.Describe("BGP", func() {
 		table.Entry("IPV6", "fc00:f853:0ccd:e799::/116", "fc00:f853:ccd:e800::1/128", ipfamily.IPv6),
 	)
 
+	ginkgo.Context("FRR validate reload feedback", func() {
+		ginkgo.It("should log a reload-validate success in speaker container", func() {
```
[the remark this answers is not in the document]
liornoy: Yes, sure.

liornoy:
> If I got this right, after a reload, the debouncer signals the reloader that waits an arbitrary time and checks if the file is newer than that. One point of concern is the amount of timed synchronization between the debouncer and the reloader, and the moving parts.
> 
> On example is, multiple updates, `lastTriggeredTimeStamp` changes but the timer is not set, effectively giving the reload procedure less time.
> 
> I'd try to keep the truth of a successful / unsuccessful reload as reliable as we can, and the reloader script is the one in control. Something that comes to my mind is, we could have two files, one for successful reloads and one for failed reloads (or one file with the result).
> 
> Then, it would be just a matter of polling the file(s) (and checking if it changed from the last time we polled), and emitting the succeed / failed log and event. We could still miss events, if the polling interval is coarser than the reloading, but that is not an issue as we care about the last state.
> 
> WDYT?

I didn't understand how the example you gave is an example of a risky situation. as on every call for the debouncer it echoes the signal and set a new `lastTriggeredTimeStamp` value, and when you do just one update, the timer initially is not set so it waits the 500 milliseconds before starting the reload, and yet when I tested it, the reload finished later, making 2 seconds difference between the `lastTriggeredTimeStamp` and what was on the time stamp on the .status file.

Moreover, regarding the implementation you suggested, it is identical to my first design of the solution, and I re-do that because it doesn't work in the following scenario:
load good config -> reloader success -> file change -> validation sees file changed -> log reload success
BUT:
load good config -> reloader success -> file change -> load bad config -> reloader fail -> fail unchanged -> validation still sees file changed (because of first loading) -> log reload success 

meaning by this method the last state was a failure but it saw a change from the last time so it thinks it was a success. 
this is why I went to a timestamp approach.
let me know if this is what you meant or am I missing something.

fedepaol:
> 



> > If I got this right, after a reload, the debouncer signals the reloader that waits an arbitrary time and checks if the file is newer than that. One point of concern is the amount of timed synchronization between the debouncer and the reloader, and the moving parts.
> > On example is, multiple updates, `lastTriggeredTimeStamp` changes but the timer is not set, effectively giving the reload procedure less time.
> > I'd try to keep the truth of a successful / unsuccessful reload as reliable as we can, and the reloader script is the one in control. Something that comes to my mind is, we could have two files, one for successful reloads and one for failed reloads (or one file with the result).
> > Then, it would be just a matter of polling the file(s) (and checking if it changed from the last time we polled), and emitting the succeed / failed log and event. We could still miss events, if the polling interval is coarser than the reloading, but that is not an issue as we care about the last state.
> > WDYT?
> 
> I didn't understand how the example you gave is an example of a risky situation. as on every call for the debouncer it echoes the signal and set a new `lastTriggeredTimeStamp` value, and when you do just one update, the timer initially is not set so it waits the 500 milliseconds before starting the reload, and yet when I tested it, the reload finished later, making 2 seconds difference between the `lastTriggeredTimeStamp` and what was on the time stamp on the .status file.
> 

I missed the reset part, my bad. In any case, I'd prefer going on a more straightforward solution. See my answer below.

> Moreover, regarding the implementation you suggested, it is identical to my first design of the solution, and I re-do that because it doesn't work in the following scenario: load good config -> reloader success -> file change -> validation sees file changed -> log reload success BUT: load good config -> reloader success -> file change -> load bad config -> reloader fail -> fail unchanged -> validation still sees file changed (because of first loading) -> log reload success
> 
> meaning by this method the last state was a failure but it saw a change from the last time so it thinks it was a success. this is why I went to a timestamp approach. let me know if this is what you meant or am I missing something.

I meant to track also failed reloads, so you'd update the file even in case of bad config (with a different file or with a couple timestamp, result as content). In that scenario you always know the state of the last update, can compare to the last one you notified and in case notify the update.

e2etest/bgptests/bgp.go
```diff
@@ -1060,6 +1062,46 @@ var _ = ginkgo.Describe("BGP", func() {
 		table.Entry("IPV6", "fc00:f853:0ccd:e799::/116", "fc00:f853:ccd:e800::1/128", ipfamily.IPv6),
 	)
 
+	ginkgo.FContext("FRR validate reload feedback", func() {
```
fedepaol, line 1065: FContext

e2etest/pkg/k8s/pods.go
```diff
@@ -0,0 +1,30 @@
...
+import (
+	"bytes"
+	"context"
+	"io"
+
+	corev1 "k8s.io/api/core/v1"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+func GetPodLogs(cs clientset.Interface, pod *corev1.Pod, podLogOpts corev1.PodLogOptions) (string, error) {
```
fedepaol, line 14: Just "PodLogs", also I'd just have the containerName as parameter, and hardcode the # of lines

internal/bgp/frr/frr.go
```diff
@@ -363,11 +365,65 @@ func NewSessionManager(l log.Logger) *sessionManager {
...
+	lastReloadStatus := strings.Fields(string(bytes))
+
+	if len(lastReloadStatus) == 1 {
+		return
+	}
+
+	timeStamp, status := lastReloadStatus[0], lastReloadStatus[1]
+
+	currReloadTimeStamp, err := strconv.Atoi(timeStamp)
+
```
fedepaol, line 406: no need for this line

internal/bgp/frr/frr.go
```diff
@@ -363,11 +365,65 @@ func NewSessionManager(l log.Logger) *sessionManager {
...
+func validateReload(l log.Logger, prevReloadTimeStamp *int) {
+	bytes, err := os.ReadFile(statusFileName)
+	if err != nil {
+		_ = level.Error(l).Log("op", "reload-validate", "error", err, "cause", "readFile", "fileName", statusFileName)
+		return
+	}
+
+	lastReloadStatus := strings.Fields(string(bytes))
+
+	if len(lastReloadStatus) == 1 {
```
fedepaol, line 399: Can we make this simpler, and just assume that if the file is not there, it's because we did not create it yet?
That will make the entry point below simpler

e2etest/bgptests/bgp.go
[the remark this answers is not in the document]
fedepaol: Wondering if we are able to force a failed reload somehow..

fedepaol:
Looks good overall, left a couple of comments.

e2etest/bgptests/bgp.go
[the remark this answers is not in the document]
liornoy: Me too. I've tried a couple of times with no success. it's is very robust and goes through many verifications before getting to the reloader.sh script.

e2etest/pkg/k8s/pods.go
```diff
@@ -0,0 +1,42 @@
...
+		return "", err
+	}
+	str := buf.String()
+
+	return str, nil
+}
+
+func int64Ptr(n int64) *int64 {
+	return &n
+}
```
liornoy, line 42: This will be moved to metallb/internal/pointer package after #1161 will merge

internal/bgp/frr/frr.go
liornoy: simpler is better. done

internal/bgp/frr/frr.go
liornoy: deleted

e2etest/pkg/k8s/pods.go
liornoy: fixed

e2etest/bgptests/bgp.go
liornoy: fixed

e2etest/pkg/k8s/pods.go
```diff
@@ -0,0 +1,39 @@
...
+		return "", err
+	}
+	defer podLogs.Close()
+
+	buf := new(bytes.Buffer)
+	_, err = io.Copy(buf, podLogs)
+	if err != nil {
+		return "", err
+	}
+	str := buf.String()
```
fedepaol, line 36: add one empty line hereand remove the one below

internal/bgp/frr/frr.go
```diff
@@ -368,11 +371,62 @@ func NewSessionManager(l log.Logger, logLevel logging.Level) *sessionManager {
...
+
+	ticker := time.NewTicker(tickerIntervals)
+	go func() {
+		for range ticker.C {
+			validateReload(l, &prevReloadTimeStamp)
+		}
+	}()
+}
+
+var statusFileName = "/etc/frr_reloader/.status"
```
fedepaol, line 394: this can be a const I think

fedepaol:
A few nits, then it's ready to go. Thanks!
Did you try to force the failure (by messing up with the frr file for example) to see it in action?

internal/bgp/frr/frr.go
```diff
@@ -368,11 +371,62 @@ func NewSessionManager(l log.Logger, logLevel logging.Level) *sessionManager {
...
+			_ = level.Error(l).Log("op", "reload-validate", "error", err, "cause", "readFile", "fileName", statusFileName)
+		}
+		return
+	}
+
+	lastReloadStatus := strings.Fields(string(bytes))
+
+	if len(lastReloadStatus) != 2 {
+		_ = level.Error(l).Log("op", "reload-validate", "error", err, "cause", "Fields", "bytes", string(bytes))
+
```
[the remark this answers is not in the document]
liornoy: done

e2etest/pkg/k8s/pods.go
liornoy: done

internal/bgp/frr/frr.go
liornoy: agree, switched to const

liornoy:
> A few nits, then it's ready to go. Thanks! Did you try to force the failure (by messing up with the frr file for example) to see it in action?

yes, and it's emitting the failure log line

fedepaol:
LGTM

PR #1179, opened 2022-01-20
E2E: make bfd profile name lowercase

fedepaol:
The e2e tests work in configmap mode, but do not in operatormode because uppercase is not allowed in CRs names.


zeeke:
lgtm

PR #1180, opened 2022-01-21
Forward speaker log level to FRR configuration in order to improve debuggability

zeeke:
Hi,

To allow debugging of FRR speaker container, it would be useful to forward the speaker `--log-level` parameter to the FRR configuration. Not sure about the design in forwarding the logLevel parameters from `main()` to `frr.NewSessionManager(...)`, feedback is wellcome.

It would also useful to add lines like these to the configuration, in order to debug specific protocols:

```
debug zebra events
debug zebra nht
debug zebra kernel
debug zebra rib
debug zebra nexthop
debug bgp bfd
debug bgp neighbor-events
debug bgp updates
debug bgp keepalives
debug bgp nht
debug bgp zebra
debug bfd distributed
debug bfd network
debug bfd peer
debug bfd zebra
```

To achieve that, the following options come to my mind:
a. Add all that configuration if `--log-level` is set do debug, or if condition like that is met.
b. Allow the user to inject custom configuration snippets at the bottom of the FRR configuration file. It can be done either through the ConfigMap (or CR in the future), or with an environment variable. This option can become handy when facing unexpected scenarios.
c. Add a speaker parameters like `--debug-frr-protocols <comma-separated-list-of-protocols>` (like `--debug-frr-protocols bfd, bgp`) ot enable protocol specific debug commands.

Please, tell me if I'm over-engineered this last feature. Maybe it can be much simpler.

Regards

### DIFF
--- a/internal/bgp/frr/config.go
+++ b/internal/bgp/frr/config.go
@@ -31,6 +31,21 @@ var reloaderPidFileName = "/etc/frr_reloader/reloader.pid"
 const configTemplate = `
 log file /etc/frr/frr.log {{.Loglevel}}
 log timestamp precision 3
+{{- if eq .Loglevel "debugging" }}
+debug zebra events
+debug zebra nht
+debug zebra kernel
+debug zebra rib
+debug zebra nexthop
+debug bgp neighbor-events
+debug bgp updates
+debug bgp keepalives
+debug bgp nht
+debug bgp zebra
+debug bfd network
+debug bfd peer
+debug bfd zebra
+{{- end }}
 hostname {{.Hostname}}
 ip nht resolve-via-default
 ipv6 nht resolve-via-default
--- a/internal/bgp/frr/frr_bfd_test.go
+++ b/internal/bgp/frr/frr_bfd_test.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/go-kit/kit/log"
 	"go.universe.tf/metallb/internal/config"
+	"go.universe.tf/metallb/internal/logging"
 )
 
 func TestBFDProfileNoSessions(t *testing.T) {
@@ -37,7 +38,7 @@ func TestBFDProfileNoSessions(t *testing.T) {
 		},
 	}
 	l := log.NewNopLogger()
-	sessionManager := NewSessionManager(l)
+	sessionManager := NewSessionManager(l, logging.LevelInfo)
 	defer close(sessionManager.reloadConfig)
 
 	err := sessionManager.SyncBFDProfiles(pp)
@@ -69,7 +70,7 @@ func TestBFDProfileCornerCases(t *testing.T) {
 	}
 
 	l := log.NewNopLogger()
-	sessionManager := NewSessionManager(l)
+	sessionManager := NewSessionManager(l, logging.LevelInfo)
 	defer close(sessionManager.reloadConfig)
 
 	err := sessionManager.SyncBFDProfiles(pp)
@@ -111,7 +112,7 @@ func TestBFDWithSession(t *testing.T) {
 	}
 
 	l := log.NewNopLogger()
-	sessionManager := NewSessionManager(l)
+	sessionManager := NewSessionManager(l, logging.LevelInfo)
 	defer close(sessionManager.reloadConfig)
 
 	err := sessionManager.SyncBFDProfiles(pp)
@@ -152,7 +153,7 @@ func TestBFDProfileAllDefault(t *testing.T) {
 	}
 
 	l := log.NewNopLogger()
-	sessionManager := NewSessionManager(l)
+	sessionManager := NewSessionManager(l, logging.LevelInfo)
 	defer close(sessionManager.reloadConfig)
 
 	err := sessionManager.SyncBFDProfiles(pp)
--- a/internal/bgp/frr/frr_test.go
+++ b/internal/bgp/frr/frr_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/pkg/errors"
 	"go.universe.tf/metallb/internal/bgp"
 	"go.universe.tf/metallb/internal/config"
+	"go.universe.tf/metallb/internal/logging"
 	"k8s.io/apimachinery/pkg/util/wait"
 )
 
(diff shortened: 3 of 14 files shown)


internal/bgp/frr/frr_bfd_test.go
```diff
@@ -152,7 +152,7 @@ func TestBFDProfileAllDefault(t *testing.T) {
 	}
 
 	l := log.NewNopLogger()
-	sessionManager := NewSessionManager(l)
+	sessionManager := NewSessionManager(l, "info")
```
liornoy, line 155: Is there a reason why not use `logging.LevelInfo` here?

internal/bgp/frr/frr_test.go
```diff
@@ -520,3 +521,19 @@ func TestTwoAdvertisementsTwoSessions(t *testing.T) {
 
 	testCheckConfigFile(t)
 }
+
+func TestLoggingConfiguration(t *testing.T) {
+	testSetup(t)
+
+	l := log.NewNopLogger()
+	sessionManager := NewSessionManager(l, "warning")
```
liornoy, line 529: > and `logging.LevelWarn` here?

internal/bgp/frr/testdata/TestLoggingConfiguration.golden
```diff
@@ -0,0 +1,9 @@
+
+log file /etc/frr/frr.log informational
+log stdout informational
```
liornoy, line 3: shouldn't it be a warning here instead of informational? as you configured the log level to the warning in the session manager of this test?

zeeke: you are definitely right! it's getting the default value. fixing it

cgoncalves:
> a. Add all that configuration if `--log-level` is set do debug, or if condition like that is met.

Yes.

> b. Allow the user to inject custom configuration snippets at the bottom of the FRR configuration file. It can be done either through the ConfigMap (or CR in the future), or with an environment variable. This option can become handy when facing unexpected scenarios.

Extra care is needed when allowing users to add arbitrary configurations (security reasons, project/product support levels, etc). There might be some other non-logging options that users would like to have more control of by means of arbitrary FRR configurations. I think that should be discussed separately from this PR.

> c. Add a speaker parameters like `--debug-frr-protocols <comma-separated-list-of-protocols>` (like `--debug-frr-protocols bfd, bgp`) ot enable protocol specific debug commands.

Zebra and BGP are core functions of any FRR-based MetalLB BGP deployment. BFD may or may not be used, and will only output debug level messages if enabled by at least one of its peers. I am in favor of enabling all of the abovementioned.

zeeke:
> Zebra and BGP are core functions of any FRR-based MetalLB BGP deployment. BFD may or may not be used, and will only output debug level messages if enabled by at least one of its peers. I am in favor of enabling all of the abovementioned.

Agree, simple is better!

cgoncalves:
Upgrading to FRR 8.x is a sensitive topic that was considered not so long ago when adding initial FRR support and later BFD support. While upgrading can easily be done upstream, we have to consider the possible ramifications it may have on vendors who may not so easily offer the latest and greatest FRR. In this case and in my opinion, upgrading FRR so that those two debug options can be enabled does not outweigh bumping the minimum required FRR version. Happy to discuss this further!

zeeke:
Again, I agree.

liornoy:
LGTM

internal/bgp/frr/frr_test.go
```diff
@@ -520,3 +521,35 @@ func TestTwoAdvertisementsTwoSessions(t *testing.T) {
...
 }
+
+func TestLoggingConfiguration(t *testing.T) {
+	testSetup(t)
+
+	l := log.NewNopLogger()
+	sessionManager := NewSessionManager(l, logging.LevelWarn)
+	defer close(sessionManager.reloadConfig)
+
+	config, err := sessionManager.createConfig()
```
fedepaol, line 532: how about creating a dumb session instead of manually doing this? It would be more aligned to how this is really used (instead of writing manually to the channel)

speaker/main.go
```diff
@@ -72,7 +71,7 @@ func main() {
 		mlSecret        = flag.String("ml-secret-key", os.Getenv("METALLB_ML_SECRET_KEY"), "Secret key for MemberList (fast dead node detection)")
 		myNode          = flag.String("node-name", os.Getenv("METALLB_NODE_NAME"), "name of this Kubernetes node (spec.nodeName)")
 		port            = flag.Int("port", 7472, "HTTP listening port")
-		logLevel        = flag.String("log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", strings.Join(logging.Levels, ", ")))
+		logLevel        = flag.String("log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", logging.Levels.String()))
```
fedepaol, line 74: Not sure I like this more or less than it was before. I'd probably add a custom function to the logging package as opposed to convert the slice to a string. WDYT?

fedepaol:
@zeeke not sure you missed the last comments

internal/bgp/frr/frr_test.go
zeeke: Good point. It would be nearer to the real usage of SessionManager. Anyway, the bare minimum code to get it a session is:

```
	session1, err := sessionManager.NewSession(l, "10.2.2.255:179", net.ParseIP("10.1.1.254"), 0, net.ParseIP("10.1.1.254"), 0, time.Second, time.Second, "", "", "", true)
	if err != nil {
		t.Fatalf("Could not create session: %s", err)
	}
	defer session1.Close()
```

as the addresses need to be good address.
That code, with the counterpart in the golden file, distract a little from the purpose of the test, that is the loglevel value.

Also, even if `sessionManager.reloadConfig` is a private field, it's already used by tests for closing the channel. I don't know if it's worse than a dumb session.

speaker/main.go
zeeke: TBH, I'm not very happy with that code: in order to get the same string as before, I added a String() method to the []Level type, adding a custom type as well (`logging.levelSlice`). It's probably over engineering, but since Levels is not a []string, it cannot be passed to strings.Join(...).

We can go for a fmt.Sprintf(...) of levels Slice, getting a little different output, if it's not a big deal.

fedepaol:
LGTM

internal/bgp/frr/config.go
```diff
@@ -30,7 +30,24 @@ var reloaderPidFileName = "/etc/frr_reloader/reloader.pid"
 // should also be considered.
 const configTemplate = `
 log file /etc/frr/frr.log {{.Loglevel}}
+log stdout {{.Loglevel}}
```
fedepaol, line 33: What is the reason for this?

fedepaol:
@zeeke looks like all the frr lanes are failing, mind having a look?

zeeke:
Sure, just pushed.

Do you mind consider [the comment about the parameters help string](https://github.com/metallb/metallb/pull/1180#discussion_r796915665)?

speaker/main.go
fedepaol: Fair enough, let's keep it as it is.

fedepaol:
LGTM

PR #1186, opened 2022-01-24
E2E tests: Fix operator mode AddressPool conversion function

sabinaaledort:
1. The BgpAdvertisements length should be zero to avoid empty elements in the slice.

2. The Communities length should be set before copying the slice as copy function set's the number of copied
elements to the minimum of len(dst) and len(src) which is zero in this case.


fedepaol:
LGTM, let's wait for https://github.com/metallb/metallb/pull/1188 to have clean unit tests

fedepaol:
Failed unit test being addressed in https://github.com/metallb/metallb/pull/1188, not related to this change.

fedepaol:
LGTM

PR #1187, opened 2022-01-24
AddressPool validation: Add BGPAdvertisement validation to avoid duplication

sabinaaledort:
When configuring an addressPool with two same BGPAdvertisements, an error should be produced.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -589,7 +589,7 @@ func ParseCommunity(c string) (uint32, error) {
 	}
 	b, err := strconv.ParseUint(fs[1], 10, 16)
 	if err != nil {
-		return 0, fmt.Errorf("invalid second section of community %q: %s", fs[0], err)
+		return 0, fmt.Errorf("invalid second section of community %q: %s", fs[1], err)
 	}
 
 	return (uint32(a) << 16) + uint32(b), nil
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -506,8 +506,18 @@ func parseBGPAdvertisements(ads []bgpAdvertisement, cidrsPerAddresses map[string
 		}, nil
 	}
 
+	err := validateDuplicateAdvertisements(ads)
+	if err != nil {
+		return nil, err
+	}
+
 	var ret []*BGPAdvertisement
 	for _, rawAd := range ads {
+		err := validateDuplicateCommunities(rawAd.Communities)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


internal/config/config.go
```diff
@@ -667,3 +677,25 @@ func bfdIntFromConfig(value *uint32, min, max uint32) (*uint32, error) {
 	}
 	return value, nil
 }
+
+func validateDuplicateAdvertisements(ads []bgpAdvertisement) error {
+	for i := 0; i < len(ads); i++ {
+		for j := i + 1; j < len(ads); j++ {
+			if reflect.DeepEqual(ads[i], ads[j]) {
+				return errors.New("duplicate definition of bgpadvertisement")
```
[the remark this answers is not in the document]
sabinaaledort: I'm not sure about it because it means printing nils for some of the advertisement fields.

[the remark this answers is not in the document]
fedepaol: I don't think it is an issue as long as we do not panic.

internal/config/config.go
```diff
@@ -506,8 +506,18 @@ func parseBGPAdvertisements(ads []bgpAdvertisement, cidrsPerAddresses map[string
 		}, nil
 	}
 
+	err := validateDuplicateAdvertisements(ads)
```
fedepaol, line 509: Can you change the wrapping error (from https://github.com/metallb/metallb/blob/19d5b67b8276343603afc95a7f2a62258ab8eeac/internal/config/config.go#L450 ) so it mentions the addresspool it's failing? Otherwise we'll just say "community x" or "advertisement y", but if there are multiple address pools it would still be hard to find which one.

fedepaol:
@sabinaaledort can you fix the commit messages?
The subject is truncated and part of it is moved to the body. Also, some of the commits are without the body.

internal/config/config.go
sabinaaledort: here i can't do that because the name of the addresspool is not passed to the function. i can pass the addresspool name to the function but i don't think it's needed because the error here will be anyway wrapped by this one that's pointing to the addresspool number. 
https://github.com/metallb/metallb/blob/35bbb235d0e7f176856166dc16c997a09367b62b/internal/config/config.go#L310

fedepaol: Ah cool

fedepaol:
LGTM, thanks!

PR #1188, opened 2022-01-24
Convert maps to sorted slices for prefixes

fedepaol:
Using maps to render the template works but has one drawback. When
iterating over maps, frr does not guarantee order.
This has two major drawbacks: the unit tests with goldenfile are not
stable, but more importantly, we risk to produce different rendering
from the same config, causing un-necessary reloads.

Here we still use maps (and sets) to guarantee there are no
duplicates, but before rendering we convert the map to a sorted slice
to make the output stable.

### DIFF
--- a/internal/bgp/frr/config.go
+++ b/internal/bgp/frr/config.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"sort"
 	"strconv"
 	"syscall"
 	"text/template"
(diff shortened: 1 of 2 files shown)


liornoy:
After reading through this PR I have only 1 comment which is not critical.
At `internal/bgp/frr/config.go` file you use `pp` as var name for communityPrefixes and localPrefPrefixes which got me a little confused.
'p' for prefix I understand
but 'pp' I don't know what it stands for. maybe it is a convention I am not familiar with.
Apart from that, it looks good.

internal/bgp/frr/config.go
```diff
@@ -351,10 +361,14 @@ func (s stringSet) Add(item string) {
...
 
+// Elements returns the sorted slice of elements in the set.
 func (s stringSet) Elements() []string {
 	res := []string{}
 	for k := range s {
-		res = append(res, k)
+		i := sort.SearchStrings(res, k)
+		res = append(res, "")
+		copy(res[i+1:], res[i:])
+		res[i] = k
```
oribon, line 371: Is this more effective than just sticking the keys in a slice and sorting it in the end? IMO the other way is a bit more readable.

fedepaol: Good point, and I don't think it is. I just went straigth to "insert into sorted slice" and found a solution for that.
I don't think complexity wise it's different, and at this point we should not care about it if readability is improved.

fedepaol:
> > After reading through this PR I have only 1 comment which is not critical. At `internal/bgp/frr/config.go` file you use `pp` as var name for communityPrefixes and localPrefPrefixes which got me a little confused. 'p' for prefix I understand but 'pp' I don't know what it stands for. maybe it is a convention I am not familiar with. Apart from that, it looks good.
> 
> I meant that as for "prefixes", (as pp in plural) given the locality, there should not be much ambiguity (all the scope enters in 4 lines). I think I even read that somewhere, but let me change that to prefixes so it's clearer.

ended up keeping the short names, pc and pl. Using prefixes.Prefixes is bloating the template.

internal/bgp/frr/config.go
fedepaol: Done

internal/bgp/frr/config.go
```diff
@@ -34,51 +35,50 @@ hostname {{.Hostname}}
...
-  set local-preference {{$localPref}}
+  match ip address prefix-list {{.LocalPreference}}-v4localpref-prefixes
+  set local-preference {{.LocalPreference}}
 {{- end }}
-{{- range $localPref, $prefixes := $.PrefixesV6ForLocalPref }}
+{{- range $.PrefixesV6ForLocalPref }}
 route-map {{$n.Addr}}-out permit {{counter $n.Addr}}
-  match ipv6 address prefix-list {{$localPref}}-v6localpref-prefixes
-  set local-preference {{$localPref}}
+  match ip address prefix-list {{.LocalPreference}}-v6localpref-prefixes
```
oribon, line 70: `ipv6` dropped intentionally?

fedepaol: nope, magic of copy / paste

fedepaol: done

oribon:
lgtm

PR #1189, opened 2022-01-24
E2E: Run e2e also in operator mode

### LINKED ISSUE #1185: Add a new lane for deploying metallb via the operator and running e2e in operator mode
No need to have the full coverage.
A simple lane with ipv4 + frr that deploys metallb via the operator using the manifests and run e2e to make sure operator mode is solid.

pperiyasamy:
This makes e2e also run in operator mode to ensure MetalLB CI setup is
working fine when its deployed via its operator.

Fixes #1185

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -131,3 +131,74 @@ jobs:
         with:
           name: kind-logs-${{ matrix.ip-family }}-${{ matrix.bgp-type }}-${{ matrix.deployment}}
           path: /tmp/kind_logs/
+
+  e2e-use-operator:
+    runs-on: ubuntu-20.04
+    defaults:
+      run:
+        shell: bash
+        working-directory: metallb
+    strategy:
+      fail-fast: false
+    steps:
+      - name: Checkout Metal LB Operator
+        uses: actions/checkout@v2
+        with:
+          repository: metallb/metallb-operator
+          path: metallboperator
+          ref: e84470bf3644e6444169abb65483bb1a60c957af
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.16.3'
+
+      - name: Checkout MetalLB
+        uses: actions/checkout@v2
+        with:
+          path: metallb
+          fetch-depth: 0
+
+      - name: Install Dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install python3-pip arping ndisc6
+          sudo pip3 install -r ./dev-env/requirements.txt
+          GO111MODULE="on" go get sigs.k8s.io/kind@v0.11.1
+
+      - name: Build MetalLB Images
+        run: |
+          inv build
+
+      - name: Create multi-node K8s Kind Cluster
+        run: |
+          cd ${GITHUB_WORKSPACE}/metallboperator
+          ./hack/kind-multi-node-cluster-without-registry.sh
+          kind load docker-image quay.io/metallb/speaker:dev-amd64
+          kind load docker-image quay.io/metallb/controller:dev-amd64
+          export KUBECONFIG=${HOME}/.kube/config
+
+      - name: Deploy Metal LB Operator
+        run: |
+          cd ${GITHUB_WORKSPACE}/metallboperator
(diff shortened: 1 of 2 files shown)


pperiyasamy:
I just realized there is no need of another job for this e2e test in operator mode, will make the change to use the existing e2e job itself.

fedepaol:
> I just realized there is no need of another job for this e2e test in operator mode, will make the change to use the existing e2e job itself.

We still want to make sure that metallb works with regular deployment, so the new job makes sense and its only purpouse is to validate early we did not break the operator (or its test)

sabinaaledort:
nice! great setup, looks good. let's bump metallb-operator commit to latest and run CI?

.github/workflows/ci.yaml
```diff
@@ -131,3 +132,96 @@ jobs:
...
+        uses: actions/checkout@v2
+        with:
+          path: metallb
+          fetch-depth: 0
+
+      - name: Install Dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install python3-pip arping ndisc6
+          sudo pip3 install invoke semver pyyaml
```
sabinaaledort, line 185: this can be done using with 
`pip3 install --user -r ./dev-env/requirements.txt` the in case we add more dependencies.

.github/workflows/ci.yaml
```diff
@@ -131,3 +132,96 @@ jobs:
...
+    runs-on: ubuntu-20.04
+    defaults:
+      run:
+        shell: bash
+        working-directory: metallb
+    env:
+      metallb_operator_image: "metallb-operator:ci"
+    strategy:
+      fail-fast: false
+    steps:
```
fedepaol, line 146: No need to rebuild.
I think you can save all the steps and just deploy using the single manifest: https://github.com/metallb/metallb-operator/blob/main/bin/metallb-operator-with-webhooks.yaml

sabinaaledort: +1, right! but still need the sed part:
```
sed -i 's/quay.io\/metallb\/speaker:main/quay.io\/metallb\/speaker:dev-amd64/g' bin/metallb-operator-with-webhooks.yaml
sed -i 's/quay.io\/metallb\/controller:main/quay.io\/metallb\/controller:dev-amd64/g' bin/metallb-operator-with-webhooks.yaml
```

.github/workflows/ci.yaml
pperiyasamy: done

.github/workflows/ci.yaml
```diff
@@ -131,3 +132,71 @@ jobs:
...
+        with:
+          repository: metallb/metallb-operator
+          path: metallboperator
+          ref: d5836e27522a54c3e022fde53c1f52a1446059c4
+
+      - uses: actions/setup-go@v2
+        with:
+          go-version: '1.16.3'
+
+      - name: Checkout MetalLB
```
fedepaol, line 156: Do we need to do this? I'd expect MetalLB to be already available.

pperiyasamy: This is the first time metallb being checked out. do you want to do something like  
[    steps:
      - name: Checkout
        uses: actions/checkout@v2
        with:
          fetch-depth: 0](https://github.com/metallb/metallb/blob/31080d6f4e06f7d579127034b8ecc28e5ca702d4/.github/workflows/ci.yaml#L85-L89)
at the start of the job ?

.github/workflows/ci.yaml
```diff
@@ -119,6 +119,7 @@ jobs:
           if [ "${{ matrix.ip-family }}" == "ipv6" ] && [ ${{ matrix.bgp-type }} == "native" ]; then SKIP="$SKIP|BGP"; fi
           echo "Skipping $SKIP"
           sudo -E env "PATH=$PATH" inv e2etest --skip $SKIP -e /tmp/kind_logs $HELM_FLAGS
+          sudo -E env "PATH=$PATH" inv dev-env-cleanup
```
fedepaol, line 122: why we need this? It's CI, so it will be thrown away

.github/workflows/ci.yaml
fedepaol: Ah ignore me, I just noticed the other job does that as well

fedepaol:
LGTM

PR #1191, opened 2022-01-24
E2E: override the speaker / controller selector

fedepaol:
In some deployments it's possible to have a different speaker / controller label selector. Here we add a way to override them in tests so we can run e2e against those deployments.


msherif1234:
lgtm

PR #1198, opened 2022-01-26
Repeat check for frr config file

zeeke:
Hi,

I discovered a flaky unit test caused by the [asynchronous nature of the debouncer](https://github.com/metallb/metallb/blob/46675f03d788549897ff59c6b3b268c162a394d0/internal/bgp/frr/config.go#L328).
Here the symptom:
https://github.com/metallb/metallb/runs/4920413290?check_suite_focus=true

The check is done against an intermediate configuration file generated during the test, as `TestTwoAdvertisementsTwoSessions` triggers the configuration render twice.

This PR make the assertion to be repeated using a poll interval, so if spurious config file should be discarder.

Please, let me know if there is a better way to work around this edge case.

### DIFF
--- a/internal/bgp/frr/frr_test.go
+++ b/internal/bgp/frr/frr_test.go
@@ -4,6 +4,7 @@ package frr
 
 import (
 	"flag"
+	"fmt"
 	"net"
 	"os"
 	"os/exec"
@@ -13,7 +14,6 @@ import (
 	"time"
 
 	"github.com/go-kit/kit/log"
-	"github.com/pkg/errors"
 	"go.universe.tf/metallb/internal/bgp"
 	"go.universe.tf/metallb/internal/config"
 	"k8s.io/apimachinery/pkg/util/wait"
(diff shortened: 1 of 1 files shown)


liornoy:
I had this unit test flake on my PR also, so I'm glad you filed this fix and it's looking good IMO

internal/bgp/frr/frr_test.go
```diff
@@ -31,9 +32,27 @@ func testOsHostname() (string, error) {
 
 func testCompareFiles(t *testing.T, configFile, goldenFile string) {
 	cmd := exec.Command("diff", configFile, goldenFile)
-	output, err := cmd.Output()
+	var lastError error
+
+	// Try comparing files multiple times because tests can generate more than one configuration
+	err := wait.PollImmediate(50*time.Millisecond, 2*time.Second, func() (bool, error) {
```
fedepaol, line 38: Can we embed the check for the file presence here too?
Or, let's put the other way: with this, do we still need to wait for the file to be present?

internal/bgp/frr/frr_test.go
```diff
@@ -31,9 +32,27 @@ func testOsHostname() (string, error) {
...
+			lastError = fmt.Errorf("command %s returned error: %s\n%s", cmd.String(), err, output)
+			return false, nil
+		}
+
+		return true, nil
+	})
+
 	if err != nil {
-		t.Fatalf("command %s returned error: %s\n%s", cmd.String(), err, output)
+		t.Fatalf("Failed to compare configfiles %s, %s, err %v", configFile, goldenFile, err)
```
fedepaol, line 51: this won't dump lastError because fatal will halt the execution I believe?

fedepaol:
Another note: I know that sleeping is not elegant, but I think we'd also risk to produce incomplete golden files for the same reason (and in that case, we won't have anything to compare against).
Would it make sense to put a sleep there, with a comment explaining why?

internal/bgp/frr/frr_test.go
zeeke: Good point! going to remove the file presence check, as it is embedded in the diff command

fedepaol:
typo in the commit message: "asynchronousl"

internal/bgp/frr/frr_test.go
```diff
@@ -30,15 +30,38 @@ func testOsHostname() (string, error) {
...
+		}
+
+		return true, nil
+	})
+
+	// err can only be a ErrWaitTimeout, as the check function always return nil errors.
+	// So lastError is always set
 	if err != nil {
-		t.Fatalf("command %s returned error: %s\n%s", cmd.String(), err, output)
+		t.Fatalf("failed to compare configfiles %s, %s using poll interval\nlast error: %v", configFile, goldenFile, lastError)
```
fedepaol, line 53: I'd remove "using poll interval" from the log

fedepaol:
LGTM, just the commit message needs to be fixed

PR #1201, opened 2022-01-27
FRR route-rules: add on-match next

fedepaol:
When we have both communities and localpref advertisement, the community advertisement is skipped because frr stops at the first match.
Here we add on-match next to let it continue.

Also, refactoring advertisement e2e so we can have mixed scenarios.

### DIFF
--- a/e2etest/bgptests/bgp.go
+++ b/e2etest/bgptests/bgp.go
@@ -840,30 +840,24 @@ var _ = ginkgo.Describe("BGP", func() {
 			table.Entry("IPV4", ipfamily.IPv4),
 			table.Entry("IPV6", ipfamily.IPv6))
 
-		table.DescribeTable("configure bgp community and verify it gets propagated",
-			func(rangeWithCommunity string, rangeWithoutCommunity string, ipFamily ipfamily.Family) {
+		table.DescribeTable("configure bgp advertisement and verify it gets propagated",
+			func(rangeWithAdvertisement string, rangeWithoutAdvertisement string, advertisement config.BgpAdvertisement, ipFamily ipfamily.Family) {
 				configData := config.File{
 					Peers: metallb.PeersForContainers(FRRContainers, ipFamily),
 					Pools: []config.AddressPool{
 						{
-							Name:     "bgp-with-community",
+							Name:     "bgp-with-advertisement",
 							Protocol: config.BGP,
 							Addresses: []string{
-								rangeWithCommunity,
+								rangeWithAdvertisement,
 							},
-							BGPAdvertisements: []config.BgpAdvertisement{
-								{
-									Communities: []string{
-										CommunityNoAdv,
-									},
-								},
-							},
+							BGPAdvertisements: []config.BgpAdvertisement{advertisement},
 						},
 						{
-							Name:     "bgp-with-no-community",
+							Name:     "bgp-with-no-advertisement",
 							Protocol: config.BGP,
 							Addresses: []string{
-								rangeWithoutCommunity,
+								rangeWithoutAdvertisement,
 							},
 						}},
 				}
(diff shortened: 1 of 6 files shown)


e2etest/bgptests/bgp.go
```diff
@@ -879,145 +875,122 @@ var _ = ginkgo.Describe("BGP", func() {
...
+							}
+							if localPrefix == *advertisement.LocalPref {
+								return fmt.Errorf("%s %s matching local pref", c.Name, ipNoAdvertisement)
+							}
 
-				for _, c := range FRRContainers {
-					validateFRRPeeredWithNodes(cs, c, family)
+						}
+						return nil
+					}, 1*time.Minute, 1*time.Second).Should(BeNil())
```
sabinaaledort, line 953: 1 min for this validation is enough?

e2etest/bgptests/bgp.go
```diff
@@ -879,145 +875,122 @@ var _ = ginkgo.Describe("BGP", func() {
...
-				ipWithLocalPref1, err := config.GetIPFromRangeByIndex(rangeWithPref, 1)
-				framework.ExpectNoError(err)
-				ipNoLocalPref, err := config.GetIPFromRangeByIndex(rangeNoPref, 0)
-				framework.ExpectNoError(err)
-				svcLocalPref, _ := testservice.CreateWithBackend(cs, f.Namespace.Name, "svc-local-pref",
-					func(s *corev1.Service) {
-						s.Spec.LoadBalancerIP = ipWithLocalPref
+			},
+			table.Entry("IPV4 - community and localpref",
+				"192.168.10.0/24",
```
sabinaaledort, line 958: can use the v4PoolAddresses

e2etest/bgptests/bgp.go
```diff
@@ -879,145 +875,122 @@ var _ = ginkgo.Describe("BGP", func() {
...
-				framework.ExpectNoError(err)
-
-				for _, c := range FRRContainers {
-					validateService(cs, svcLocalPref, allNodes.Items, c)
-					validateService(cs, svcLocalPref1, allNodes.Items, c)
-					validateService(cs, svcNoLocalPref, allNodes.Items, c)
-				}
+					LocalPref: uint32Ptr(50),
+				},
+				ipfamily.IPv6))
```
sabinaaledort, line 992: is that intentional that there are no ipv6 localpref only and community only tests?

e2etest/bgptests/bgp.go
fedepaol: I think it's clearer to have both of them explicit here

e2etest/bgptests/bgp.go
fedepaol: Looks like. Routes are already set and we run validateService before

e2etest/bgptests/bgp.go
fedepaol: Yes, I wanted to validate the mechanism with ipv4 but not add too much overhead and just make sure the most complex combination works with v6

msherif1234:
@fedepaol r u sure about this ? in the examples they should on-match next when call action was used ?
here is the example I saw
```
! Set local-pref as requested
route-map rm-prefmod permit 10
match community cm-prefmod-100
set local-preference 100
route-map rm-prefmod permit 20
match community cm-prefmod-200
set local-preference 200
route-map rm-prefmod permit 30
match community cm-prefmod-300
set local-preference 300
route-map rm-prefmod permit 40
match community cm-prefmod-400
set local-preference 400
```

which didn't have on-match next while this part has it
```
route-map rm-community-in permit 10
description check for blackholing, no point continuing if it matches.
match community cm-blackhole
call rm-blackhole
route-map rm-community-in permit 20
match community cm-set-no-export
call rm-no-export
on-match next
route-map rm-communit
```

fedepaol:
> @fedepaol r u sure about this ? in the examples they should on-match next when call action was used ? here is the example I saw
> 
> ```
> ! Set local-pref as requested
> route-map rm-prefmod permit 10
> match community cm-prefmod-100
> set local-preference 100
> route-map rm-prefmod permit 20
> match community cm-prefmod-200
> set local-preference 200
> route-map rm-prefmod permit 30
> match community cm-prefmod-300
> set local-preference 300
> route-map rm-prefmod permit 40
> match community cm-prefmod-400
> set local-preference 400
> ```
> 
> which didn't have on-match next while this part has it
> 
> ```
> route-map rm-community-in permit 10
> description check for blackholing, no point continuing if it matches.
> match community cm-blackhole
> call rm-blackhole
> route-map rm-community-in permit 20
> match community cm-set-no-export
> call rm-no-export
> on-match next
> route-map rm-communit
> ```

yep, I got to this because I spotted a conversation on the [frr slack](https://frrouting.slack.com/archives/C58SZTP39/p1643110265058000). The test I added with both localpref and community won't work without the change.

It's not super clear from the docs, but it's described there:
http://docs.frrouting.org/en/latest/routemap.html?highlight=route-map

<p>To summarise the above:</p>
<div class="wy-table-responsive">

  | Match | No Match
-- | -- | --
Permit | action | cont
Deny | deny | cont

</div>


Route-map stops at the first match by default.

msherif1234:
> > @fedepaol r u sure about this ? in the examples they should on-match next when call action was used ? here is the example I saw
> > ```
> > ! Set local-pref as requested
> > route-map rm-prefmod permit 10
> > match community cm-prefmod-100
> > set local-preference 100
> > route-map rm-prefmod permit 20
> > match community cm-prefmod-200
> > set local-preference 200
> > route-map rm-prefmod permit 30
> > match community cm-prefmod-300
> > set local-preference 300
> > route-map rm-prefmod permit 40
> > match community cm-prefmod-400
> > set local-preference 400
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > which didn't have on-match next while this part has it
> > ```
> > route-map rm-community-in permit 10
> > description check for blackholing, no point continuing if it matches.
> > match community cm-blackhole
> > call rm-blackhole
> > route-map rm-community-in permit 20
> > match community cm-set-no-export
> > call rm-no-export
> > on-match next
> > route-map rm-communit
> > ```
> 
> yep, I got to this because I spotted a conversation on the [frr slack](https://frrouting.slack.com/archives/C58SZTP39/p1643110265058000). The test I added with both localpref and community won't work without the change.
> 
> It's not super clear from the docs, but it's described there: http://docs.frrouting.org/en/latest/routemap.html?highlight=route-map
> 
> To summarise the above:
> 
>  	Match	No Match
> Permit	action	cont
> Deny	deny	cont
> Route-map stops at the first match by default.

I think that makes sense its our prob that we spread the actions of the same match to multiple rules with different match rules names

fedepaol:
> I think that makes sense its our prob that we spread the actions of the same match to multiple rules with different match rules names

Yep, with the subtle thing that the match is incidentally the same because we have ips with both the localpref and the community. That's because we need to reconcile the metallb api with the frr one and not always it is straightforward.

e2etest/bgptests/bgp.go
cgoncalves: I agree with Sabina here. Are we confident we won't ever regress for IPv6 with community/local preference only?
What is the additional overhead? Test code length? Is it test execution time (how much)?

Also more or less in line, we are not testing the full parameter setting for bgp-advertisements (localPref + communities + aggregation-length). Am I missing something or aggregation length is not being e2e tested?

fedepaol:
@msherif1234 just double checking, does this lgty?

cgoncalves (commented):
Just a few minor comments for the test code part, otherwise LGTM.

e2etest/bgptests/bgp.go
fedepaol: My reasoning was, the rendering part was covered by v4 and the "functional part" was covered by the v4 tests. I wanted to save some test running time, but now that you insist I even believe there are no lanes that run v4 and v6 together (dual runs v4 + explicit DUAL v6 tests). Let me extend that.

Re: aggr length, it's not being tested (I suppose because its a feature that does not depend on the bgp implementation, but better being on the safe side). Let me check if I can add it here.

msherif1234:
> @msherif1234 just double checking, does this lgty?

yes LGTM I looked only at the frr config side

e2etest/bgptests/bgp.go
fedepaol: @cgoncalves adding proper aggregation length tests in this context would require a bit of rethinking on how we represent routes in e2e tests. I'd defer this to another pr so we can push this part we know is broken.

fedepaol: Added the v6 tests

cgoncalves: Sure, the aggregation length tests can definitely be addressed outside this PR.
Thanks for the IPv6 tests!

cgoncalves (approved):
LGTM

Issue #1195, opened 2022-01-26
E2E test flake: metric metallb_speaker_announced not found


sabinaaledort:
The following test is failing in CI from time to time:

```
2022-01-23T08:02:49.0120353Z [91m[1m• Failure [124.416 seconds][0m
2022-01-23T08:02:49.0120967Z L2
2022-01-23T08:02:49.0122088Z [90m/home/runner/work/metallb/metallb/e2etest/l2tests/l2.go:56[0m
2022-01-23T08:02:49.0122435Z   metrics
2022-01-23T08:02:49.0122912Z   [90m/home/runner/work/metallb/metallb/e2etest/l2tests/l2.go:330[0m
2022-01-23T08:02:49.0123326Z     should be exposed by the controller
2022-01-23T08:02:49.0123898Z     [90m/home/runner/work/metallb/metallb/e2etest/l2tests/l2.go:349[0m
2022-01-23T08:02:49.0124422Z       [91m[1mIPV4 - Checking service [It][0m
2022-01-23T08:02:49.0125044Z       [90m/home/runner/work/metallb/metallb/e2etest/l2tests/l2.go:468[0m
2022-01-23T08:02:49.0125251Z 
2022-01-23T08:02:49.0125810Z       [91mJan 23 08:02:48.382: Timed out after 120.001s.
2022-01-23T08:02:49.0126110Z       Expected
2022-01-23T08:02:49.0126603Z           <*errors.errorString | 0xc00062d980>: {
2022-01-23T08:02:49.0127180Z               s: "metric metallb_speaker_announced not found",
2022-01-23T08:02:49.0127475Z           }
2022-01-23T08:02:49.0127791Z       to be nil[0m
2022-01-23T08:02:49.0127934Z 
2022-01-23T08:02:49.0128355Z       /home/runner/work/metallb/metallb/e2etest/l2tests/l2.go:456
```

sabinaaledort:
I'll work on this.

fedepaol:
thanks!

fedepaol:
Closing this, we are not seeing the flake anymore

PR #1203, opened 2022-01-30
L2 E2E tests: Flake fix: metrics validation of wrong speaker 

sabinaaledort:
In this test sometimes we get the wrong node when looking for the advertising node of a service.
A newly created service can get the same IP the service from the previous test got (since it was already deleted).
While the previous service is down it can take time for the running host ARP table to be updated.

The solution for such situation is to look for the advertising node of a service inside an eventually.

Fixing issue https://github.com/metallb/metallb/issues/1195.

An example of the failing test can be found here:
https://github.com/metallb/metallb/runs/4911555260?check_suite_focus=true

### DIFF
--- a/e2etest/l2tests/l2.go
+++ b/e2etest/l2tests/l2.go
@@ -428,13 +428,21 @@ var _ = ginkgo.Describe("L2", func() {
 			allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
 			framework.ExpectNoError(err)
 
-			advNode, err := advertisingNodeFromMAC(allNodes.Items, ingressIP, executor.Host)
-			framework.ExpectNoError(err)
-			advSpeaker, ok := speakerPods[advNode.Name]
-			framework.ExpectEqual(ok, true, fmt.Sprintf("could not find speaker pod on announcing node %s", advNode.Name))
-			delete(speakerPods, advSpeaker.Spec.NodeName)
-
+			var advNode *corev1.Node
+			var advSpeaker *corev1.Pod
 			gomega.Eventually(func() error {
+				var ok bool
+
+				advNode, err = advertisingNodeFromMAC(allNodes.Items, ingressIP, executor.Host)
+				if err != nil {
+					return err
+				}
+
+				advSpeaker, ok = speakerPods[advNode.Name]
+				if !ok {
+					return fmt.Errorf("could not find speaker pod on announcing node %s", advNode.Name)
+				}
+
 				speakerMetrics, err := metrics.ForPod(controllerPod, advSpeaker, metallb.Namespace)
 				if err != nil {
 					return err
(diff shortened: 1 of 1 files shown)


fedepaol:
Typo in the first commit "For exmaple:"

e2etest/l2tests/l2.go
```diff
@@ -434,7 +435,7 @@ var _ = ginkgo.Describe("L2", func() {
 
 				err = metrics.ValidateGaugeValue(1, "metallb_speaker_announced", map[string]string{"node": advSpeaker.Spec.NodeName, "protocol": "layer2", "service": fmt.Sprintf("%s/%s", f.Namespace.Name, svc.Name)}, speakerMetrics)
 				if err != nil {
-					return err
+					return pkgerrors.Wrapf(err, "failed to validate metric metallb_speaker_announced for labels: node: %s, protocol: %s, service: %s speakerMetrics: %s", advSpeaker.Spec.NodeName, "layer2", fmt.Sprintf("%s/%s", f.Namespace.Name, svc.Name), speakerMetrics)
```
fedepaol, line 438: I believe you can hardcode the protocol. Also, I believe the metrics are already printed as result of kubectl (but it won't hurt having them here). Did you check how the output renders?
I am afraid ginkgo will cut the error at some point.

test
```diff
@@ -0,0 +1 @@
+3
```
fedepaol, line 1: extra file?

sabinaaledort: just for testing

e2etest/l2tests/l2.go
sabinaaledort: the kubectl metrics are printed and i see the specific metric there, but the parsed map is not printed and i suspect the parse function, that's why i want to print it. i tried testing the parse function for the specific output i get from kubectl with unit tests but didn't find anything. 
i printed this whole thing, it's not too long so i don't think ginkgo will cut it.

sabinaaledort: sorry, you were right, ginkgo cuts the error.

e2etest/l2tests/l2.go
```diff
@@ -464,6 +468,12 @@ var _ = ginkgo.Describe("L2", func() {
 			}, 2*time.Minute, 1*time.Second).Should(gomega.BeNil())
 
 			// Negative - validate that the other speakers don't publish layer2 metrics
+			advNode, err := advertisingNodeFromMAC(allNodes.Items, ingressIP, executor.Host)
```
fedepaol, line 471: You can avoid repeating the last round of the loop by having advNode and advSpeaker outside and not inited inside the eventually.

sabinaaledort: As far as I know and tried I can't do that because they are set inside a function in the eventually.

fedepaol:
LGTM

PR #1214, opened 2022-02-08
BFDProfile: Fix: Deleted profile not being updated in speakers

sabinaaledort:
If a BFD profile was deleted and there are no more configured BFD profiles (the length of profiles is 0) the sync function should be called to update the config and not skipped.

### DIFF
--- a/e2etest/bgptests/bgp.go
+++ b/e2etest/bgptests/bgp.go
@@ -1136,6 +1136,43 @@ var _ = ginkgo.Describe("BGP", func() {
 					),
 				)
 			}
+
+			ginkgo.By("validating speakers running-config is updated after deleting the BFD profile")
+			configData.BFDProfiles = nil
+			for i := range configData.Peers {
+				configData.Peers[i].BFDProfile = ""
+			}
+
+			err = ConfigUpdater.Update(configData)
+			framework.ExpectNoError(err)
+
+			for _, pod := range speakerPods {
+				podExecutor := executor.ForPod(pod.Namespace, pod.Name, "frr")
+
+				Eventually(func() string {
+					// We need to assert against the output of the command as a bare string, as
+					// there is no json version of the command.
+					cfgStr, err := podExecutor.Exec("vtysh", "-c", "show running-config")
+					if err != nil {
+						return err.Error()
+					}
+
+					return cfgStr
+				}, 1*time.Minute).Should(
+					And(
+						ContainSubstring("log file /etc/frr/frr.log informational"),
+						WithTransform(substringCount("\n profile fullbfdprofile1"), Equal(0)),
+						WithTransform(func(cfgStr string) bool {
+							for _, c := range FRRContainers {
+								if strings.Contains(cfgStr, fmt.Sprintf("neighbor %s bfd", c.Ipv4)) {
+									return true
+								}
+							}
+							return false
+						}, Equal(false)),
+					), "Failed to update speakers after deleting the BFD profile",
+				)
+			}
 		})
 	})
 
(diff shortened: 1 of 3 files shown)


cgoncalves:
Thanks for the fix!
Would it be possible to add test coverage? E2E would be great, if possible.

fedepaol:
> Thanks for the fix! Would it be possible to add test coverage? E2E would be great, if possible.

+1 , if unit tests are too narrow to cover this, another good place is to extend this test https://github.com/metallb/metallb/blob/31080d6f4e06f7d579127034b8ecc28e5ca702d4/e2etest/bgptests/bgp.go#L1076 by removing the profile and inspecting the file again.

e2etest/bgptests/bgp.go
```diff
@@ -1136,6 +1136,35 @@ var _ = ginkgo.Describe("BGP", func() {
...
+					cfgStr, err := podExecutor.Exec("vtysh", "-c", "show running-config")
+					if err != nil {
+						return err.Error()
+					}
+
+					return cfgStr
+				}, 1*time.Minute).Should(
+					And(
+						ContainSubstring("log file /etc/frr/frr.log informational"),
+						WithTransform(substringCount("\n profile fullbfdprofile1"), Equal(0)),
```
cgoncalves, line 1164: Can we also validate that the BGP peer does not have BFD enabled with the default BFD profile?

This is, should not have `neighbor $NEIGHBOR_IP bfd`

cgoncalves (changes requested):

e2etest/bgptests/bgp.go
```diff
@@ -1136,6 +1136,43 @@ var _ = ginkgo.Describe("BGP", func() {
 					),
 				)
 			}
+
+			ginkgo.By("validating speakers running-config is updated after deleting the BFD profile")
```
pperiyasamy, line 1140: should we also validate if peers are referring to an existing BFD profile ?

e2etest/bgptests/bgp.go
sabinaaledort: I added the test.
It seems we found a bug in FRR. The BFD profile is removed from the FRR config but the BGP peers still have BFD enabled with: `neighbor $NEIGHBOR_IP bfd`.

fedepaol: Can you check that the renderer file does not contain them (i.e. it's an frr bug and not a bug on our side)?

e2etest/bgptests/bgp.go
sabinaaledort: if the BFD profile doesn't exist we produce an error [here](https://github.com/metallb/metallb/blob/81a0de8f7a8e2d073590d77e55065f14df97d698/internal/config/config.go#L282) which is printed to the speakers log, i think checking the speakers log can be a bit tricky.

e2etest/bgptests/bgp.go
sabinaaledort: I checked the rendered file, it doesn't have BFD configuration at all.
We tested it manually as well and got the same result.

e2etest/bgptests/bgp.go
pperiyasamy: oh! no, both states are just used to remove the entry from queue, looks this test is not feasible.

pperiyasamy: we may use `metallb_k8s_client_update_errors_total` metric if that throws `SyncStateError`, but i'm not sure if this PR intended to do this.

liornoy: @sabinaaledort If you are interested in checking speakers logs you can use what I wrote on my PR
https://github.com/metallb/metallb/pull/1177/files#diff-dcd1cc413d712c08cb1ff01c2fb08dcaa495e4d2ed52e75d2367fc6b6da7ca82

fedepaol:
@sabinaaledort this is on hold until frr is fixed right?

sabinaaledort:
> @sabinaaledort this is on hold until frr is fixed right?

yes.

PR #1220, opened 2022-02-15
Remove the control for aggregation length v6 in native mode

fedepaol:
In the operator crds we default the value to 128, so running the
operator in native mode will fail when we have advertisements. We still
check for not having v6 addresses in native mode, so the users will still
know if something is not working.


russellb:
Lgtm

Issue #1221, opened 2022-02-15
frr is enabled by default in chart 0.12.0


vchrisb:
The latest release of the metallb helm chart does add frr support.
`speaker.frr.enabled` is `true` by default which is challenging:
* frr is experimental
* updating the helm chart to `0.12.0` will switch existing installations to frr, potentially breaking things: for me it was #1216 but there are other limitations pointed out in the release notes.

please consider making `frr` disabled by default, at least while it is experimental.

jhoos:
In addition to it defaulting to true, explicitly setting it to false does not remove the frr-specific init containers from the daemonset; perhaps it should?

fedepaol:
Thanks @vchrisb @jhoos , definitely.
We are gonna fix this.

sabinaaledort:
@jhoos Hey, i just deployed MetalLB with the helm chart and set `speaker.frr.enabled` to `false`. I don't see the FRR init containers in the daemonset. Can you share more details on how you did the deployment?

fedepaol:
We just released 0.12.1. Will close this.
With the info provided by @jhoos we'll file a new one.
Thanks!

sabinaaledort:
> > @jhoos Hey, i just deployed MetalLB with the helm chart and set `speaker.frr.enabled` to `false`. I don't see the FRR init containers in the daemonset. Can you share more details on how you did the deployment?
> 
> I tried it again with 0.12.1, and the init containers are disabled now. Not sure why they weren't the first time I tried it; best guess is that for whatever reason my change of `frr.enabled` from true to false didn't stick.
> 
> Anyhow it's working as it should, so disregard me. :)

Thanks for the update!

PR #1222, opened 2022-02-16
Helm chart: disable FRR mode by default

sabinaaledort:
As the FRR mode is still experimental we set the helm chart default to disable FRR (can be optionally enabled).

Fixing issue https://github.com/metallb/metallb/issues/1221.


fedepaol:
LGTM

PR #1223, opened 2022-02-16
Cancel previous jobs when overriding a PR

fedepaol:
When overriding a PR, previous jobs still run, and eventually filling the jobs available. Since the PR was overridden, there is no point in waiting them.

Here we use https://github.com/marketplace/actions/cancel-workflow-action to cancel them.


sabinaaledort:
Thank you!! Much needed!!
/lgtm

rata (approved):
If this works with forks, it of course LGTM!

PR #1227, opened 2022-02-16
Speaker metrics: Bug fix: wrong label 

sabinaaledort:
The speaker metrics wasn't updated with deleted services because we were using the wrong label to delete it.


e2etest/l2tests/l2.go
```diff
@@ -469,6 +472,23 @@ var _ = ginkgo.Describe("L2", func() {
...
+					return err
+				}
+
+				err = metrics.ValidateGaugeValue(1, "metallb_speaker_announced", map[string]string{"node": advSpeaker.Spec.NodeName, "protocol": "layer2", "service": fmt.Sprintf("%s/%s", f.Namespace.Name, svc.Name)}, speakerMetrics)
+				if err == nil {
+					return fmt.Errorf("metallb_speaker_announced present in node: %s", advSpeaker.Spec.NodeName)
+				}
+
+				return nil
+			}, 1*time.Minute, 1*time.Second).Should(gomega.BeNil())
```
msherif1234, line 491: do we really need eventually to confirm we don't get metric after the svc is deleted ?

e2etest/l2tests/l2.go
```diff
@@ -392,8 +392,11 @@ var _ = ginkgo.Describe("L2", func() {
 			ginkgo.By("creating a service")
 			svc, _ := service.CreateWithBackend(cs, f.Namespace.Name, "external-local-lb", testservice.TrafficPolicyCluster)
 			defer func() {
-				err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
-				framework.ExpectNoError(err)
+				_, err := cs.CoreV1().Services(svc.Namespace).Get(context.TODO(), svc.Name, metav1.GetOptions{})
```
fedepaol, line 395: Why you need this?

speaker/main.go
```diff
@@ -339,12 +339,16 @@ func (c *controller) deleteBalancer(l log.Logger, name, reason string) k8s.SyncS
...
+		ok := announcing.Delete(prometheus.Labels{
 			"protocol": string(proto),
 			"service":  name,
 			"node":     c.myNode,
-			"ips":      ip.String(),
+			"ip":       ip.String(),
 		})
+		if !ok {
+			level.Error(l).Log("op", "deleteBalancer", "error", "failed to delete service metric", "service", name, "ip", ip.String())
+			return k8s.SyncStateError
```
fedepaol, line 350: I wouldn't return SyncStateError for this, the sync went fine, we just did not remove the metrics.
The log is enough I think.

e2etest/l2tests/l2.go
sabinaaledort: Because i added in the end of the test the deletion of the service to validate the metrics. Once the defer is called it can't find the service (as it was already deleted) and the test is failing. I didn't want to remove the defer since the test can fail before getting to the deletion part, in such state we need the defer to delete the service.

fedepaol: Then just ignore the error if it's not found, I think it will be clearer.

fedepaol:
LGTM

PR #1229, opened 2022-02-17
Fix my email in the README

rata:
This PR is quite simple:
 * It fixes my email (kinvolk.com is not a valid domain, it is kinvolk.io!)
 * It changes the advice to mail all of the maintainers instead (more reasons in the commit, but it seems just simpler and in case someone is AFK).


fedepaol:
LGTM, thanks!

PR #1230, opened 2022-02-17
When services share IP, modifying any service should fail instead of allocating new svc IP

msherif1234:
Users prefer seeing errors when modify services that are sharing the sameIP
using "allow-shared-ip" annotation than getting newly allocated serviceIP


fedepaol:
LGTM

PR #1236, opened 2022-02-22
E2E: add a test that checks overlapping ip services in bgp

fedepaol:
We have a test that validate overlapping ips in l2, but we want to ensure it works also with bgp.

### DIFF
--- a/e2etest/bgptests/bgp.go
+++ b/e2etest/bgptests/bgp.go
@@ -31,6 +31,7 @@ import (
 	"go.universe.tf/metallb/e2etest/pkg/k8s"
 	"go.universe.tf/metallb/e2etest/pkg/metallb"
 	"go.universe.tf/metallb/e2etest/pkg/metrics"
+	"go.universe.tf/metallb/e2etest/pkg/service"
 	metallbconfig "go.universe.tf/metallb/internal/config"
 	"go.universe.tf/metallb/internal/pointer"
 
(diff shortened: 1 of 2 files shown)


liornoy:
Looks good. my only concern is that the test description is not descriptive enough:
the sentence "A service of protocol load balancer should work with"
don't tell that its purpose is to test overlapping ip services.
maybe add a more specific description?
This way the output in case of failed test will ease the understanding of what
the scenario it failed.

liornoy:
Looks good. my only concern is that the test description:
"A service of protocol load balancer should work with"
don't tell that its purpose is to test overlapping ip services.
maybe add a more specific description?
This way the output in case of failed test will ease the understanding of what
the scenario it failed.

fedepaol:
> Looks good. my only concern is that the test description: "A service of protocol load balancer should work with" don't tell that its purpose is to test overlapping ip services. maybe add a more specific description? This way the output in case of failed test will ease the understanding of what the scenario it failed.

oh, good catch! That was a copy - paste :P 
I'll fix that (and also I'll fix the FEntry which will mae the tests run only one :D )

e2etest/bgptests/bgp.go
```diff
@@ -208,6 +209,71 @@ var _ = ginkgo.Describe("BGP", func() {
...
+		defer testservice.Delete(cs, svc1)
+
+		validateDesiredLB(svc)
+		validateDesiredLB(svc1)
+
+		for _, c := range FRRContainers {
+			validateService(cs, svc, allNodes.Items, c)
+			validateService(cs, svc1, allNodes.Items, c)
+		}
+		time.Sleep(time.Minute)
```
sabinaaledort, line 271: i think you forgot to remove this sleep?

fedepaol: Good catch :-)

e2etest/pkg/service/create.go
```diff
@@ -43,17 +56,15 @@ func Delete(cs clientset.Interface, svc *corev1.Service) {
 	framework.ExpectNoError(err)
 }
 
-func tweakServicePort(svc *v1.Service) {
-	if TestServicePort != 80 {
-		// if TestServicePort is non default, then change service spec.
-		svc.Spec.Ports[0].TargetPort = intstr.FromInt(int(TestServicePort))
+func TweakServicePort(port int) func(svc *v1.Service) {
+	return func(svc *v1.Service) {
+		svc.Spec.Ports[0].TargetPort = intstr.FromInt(int(port))
```
liornoy, line 61: port is already int. Is the casting necessary here?

fedepaol: Nope, fixed

e2etest/bgptests/bgp.go
fedepaol: removed

liornoy:
Can you please explain why are we tweaking the second service and why do we need it?

fedepaol:
> Can you please explain why do we need to tweak the second service?

We tweak both services to add the annotation AND to have different ports. That's the requisite to have multiple services using the same ip

liornoy:
> > Can you please explain why do we need to tweak the second service?
> 
> We tweak both services to add the annotation AND to have different ports. That's the requisite to have multiple services using the same ip

I am referring to the other tweaks - TweakServicePort and tweakRCPort which are applied only for the second service with port != 0

fedepaol:
> > > Can you please explain why do we need to tweak the second service?
> > 
> > 
> > We tweak both services to add the annotation AND to have different ports. That's the requisite to have multiple services using the same ip
> 
> I am referring to the other tweaks - TweakServicePort and tweakRCPort which are applied only for the second service with port != 0

Not sure I follow. Both services are called with CreateWithBackendPort, which in turn calls the two functions you mentioned.

liornoy:
> > > > Can you please explain why do we need to tweak the second service?
> > > 
> > > 
> > > We tweak both services to add the annotation AND to have different ports. That's the requisite to have multiple services using the same ip
> > 
> > 
> > I am referring to the other tweaks - TweakServicePort and tweakRCPort which are applied only for the second service with port != 0
> 
> Not sure I follow. Both services are called with CreateWithBackendPort, which in turn calls the two functions you mentioned.

Sorry, I thought they are called with CreateWithBackend. 
The logic there is confusing me. 
TestServicePort is a global var that can only change by go test flags.
The default value is 80 - in this case, it is calling CreateWithBackendPort with port 0, and not doing the second round of tweaks. it's not clear to me what the intention is there.

fedepaol:
> > > > > Can you please explain why do we need to tweak the second service?
> > > > 
> > > > 
> > > > We tweak both services to add the annotation AND to have different ports. That's the requisite to have multiple services using the same ip
> > > 
> > > 
> > > I am referring to the other tweaks - TweakServicePort and tweakRCPort which are applied only for the second service with port != 0
> > 
> > 
> > Not sure I follow. Both services are called with CreateWithBackendPort, which in turn calls the two functions you mentioned.
> 
> Sorry, I thought they are called with CreateWithBackend. The logic there is confusing me. TestServicePort is a global var that can only change by go test flags. The default value is 80 - in this case, it is calling CreateWithBackendPort with port 0, and not doing the second round of tweaks. it's not clear to me what the intention is there.

Yep, I tried to preserve the existing logic and avoid changing too much, resulting in this. I'll try to rework it to make it more clear, thanks for the feedback.

fedepaol:
> > > > > > Can you please explain why do we need to tweak the second service?
> > > > > 
> > > > > 
> > > > > We tweak both services to add the annotation AND to have different ports. That's the requisite to have multiple services using the same ip
> > > > 
> > > > 
> > > > I am referring to the other tweaks - TweakServicePort and tweakRCPort which are applied only for the second service with port != 0
> > > 
> > > 
> > > Not sure I follow. Both services are called with CreateWithBackendPort, which in turn calls the two functions you mentioned.
> > 
> > 
> > Sorry, I thought they are called with CreateWithBackend. The logic there is confusing me. TestServicePort is a global var that can only change by go test flags. The default value is 80 - in this case, it is calling CreateWithBackendPort with port 0, and not doing the second round of tweaks. it's not clear to me what the intention is there.
> 
> Yep, I tried to preserve the existing logic and avoid changing too much, resulting in this. I'll try to rework it to make it more clear, thanks for the feedback.

@liornoy done, we now always force the port (even though 80 is what's being set inside by default). I think it's more clear now and less clunky

liornoy:
Thank you. Indeed more clear.
Also please rename TweakServicePort -> tweakServicePort 
Regarding the `tweakServicePort` and `tweakRCPort` maybe they could get both arguments instead of returning a function.
i.e. : 
```
	svc, err = jig.CreateLoadBalancerService(timeout, func(svc *corev1.Service) {
		tweakServicePort(svc, port)
		for _, f := range tweak {
			f(svc)
		}
	})
	.
	.
	.
      func tweakServicePort(svc *v1.Service, port int) {
	      svc.Spec.Ports[0].TargetPort = intstr.FromInt(port)
      }
```
Or are there extra advantages to having the functions return a function?

fedepaol:
> Or are there extra advantages to having the functions return a function?

Not really, now that they do not have to consume the global parameter there's no need to use closures. Will change it

liornoy:
lgtm

Issue #1231, opened 2022-02-18
Using Metallb version 0.12.0+ is problematic when IPv4/IPv6 dual-stack networking is disabled


SunnyDou:
# Version

MetalLB: 0.12.0+
Kubernetes: 1.19.16
network addon: flannel
kube-proxy: ipvs
mode: layer2

# Details

I found that when I was using metallb 0.12.0+, SVC could not be assigned to EXTERNAL-IP. The following error is reported in the controller log:
```
{"caller":"level.go:63","event":"clearAssignment","level":"info","msg":"No NoClusterIPs ClusterIPs ", "" reason" : ""," service ":" default/nginx, "" ts" : "the 2022-02-18 T05:59:37. 053567956 z"}
```

ClusterIP->ClusterIPs was modified from 0.12.0 in controller service.go.The reason is that dual-Stack is supported by a [new feature](https://github.com/metallb/metallb/pull/1065).

However, as of 1.21, Kubernetes clusters have enabled IPv4/IPv6 dual stack networks by default, allowing both IPv4 and IPv6 addresses to be allocated.

If I am using Kubernetes 1.21 or later, IPv4/IPv6 Dual Stack Networks is not enabled,Then I get the error I described earlier.

Is this a Bug? I think this option should be configured as a configuration item.

fedepaol:
Hi @SunnyDou, 1.19 is not supported anymore, and MetalLB follows the same logic ensuring compatibility with supported k8s versions (and the last 3 are up to 1.21, see https://endoflife.date/kubernetes).
So, I am keen to say this is not supported, however I'll try to bring this up with the other maintainers to hear their opinion.
For now I'll leave this open.

SunnyDou:
Hi @fedepaol, thank you for your reply. Now I know why. But I'm still looking forward to retaining support for older features or deprecating notifications.

fedepaol:
@SunnyDou we'll try to provide a fix for this.
Also, could you please elaborate, my understanding is that this is happening for 1.19 but you say it's happening also on 1.21+ with dual stack disabled. Can you confirm that?

SunnyDou:
[...]

PR #1238, opened 2022-02-22
Fallback to clusterip to determine the service ipfamily

fedepaol:
K8s versions prior to 1.20 do not have ClusterIPs set, so we fallback to clusterip (singular) to determine the ip family.

Tested locally with 
`inv dev-env --node-img=kindest/node:v1.19.11@sha256:07db187ae84b4b7de440a73886f008cf903fcf5764ba8106a9fd5243d6f32729` as the CI lanes are running against 1.21.

### DIFF
--- a/controller/service.go
+++ b/controller/service.go
@@ -51,7 +51,7 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 
 	// If the ClusterIPs is malformed or not set we can't determine the
 	// ipFamily to use.
-	if len(svc.Spec.ClusterIPs) == 0 {
+	if len(svc.Spec.ClusterIPs) == 0 && svc.Spec.ClusterIP == "" {
 		level.Info(l).Log("event", "clearAssignment", "reason", "noClusterIPs", "msg", "No ClusterIPs")
 		c.clearServiceState(key, svc)
 		return true
@@ -75,10 +75,10 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 			c.client.Errorf(svc, "nolbIPsIPFamily", "Failed to retrieve LBIPs IPFamily for %q: %s", lbIPs, err)
 			return true
 		}
-		clusterIPsIPFamily, err := ipfamily.ForAddresses(svc.Spec.ClusterIPs)
+		clusterIPsIPFamily, err := ipfamily.ForService(svc)
 		if err != nil {
 			level.Error(l).Log("event", "clearAssignment", "reason", "noclusterIPsIPFamily", "msg", "Failed to retrieve clusterIPs family")
-			c.client.Errorf(svc, "noclusterIPsIPFamily", "Failed to retrieve ClusterIPs IPFamily for %q: %s", svc.Spec.ClusterIPs, err)
+			c.client.Errorf(svc, "noclusterIPsIPFamily", "Failed to retrieve ClusterIPs IPFamily for %q %s: %s", svc.Spec.ClusterIPs, svc.Spec.ClusterIP, err)
 			return true
 		}
 		// Clear the lbIP if it has a different ipFamily compared to the clusterIP.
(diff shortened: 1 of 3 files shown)


fedepaol:
Fixes https://github.com/metallb/metallb/issues/1231

rata (approved):
LGTM, thanks a lot @fedepaol 

I haven't tested this, but what you did seems enough :)

msherif1234:
I feel if we are in this mode we should fail any attempt to provision svc with dual stack ?

fedepaol:
> I feel if we are in this mode we should fail any attempt to provision svc with dual stack ?

That will come for free as in 1.19 there's no dual stack fields, which is exactly the reason why we need to fallback to ClusterIP

msherif1234:
> > I feel if we are in this mode we should fail any attempt to provision svc with dual stack ?
> 
> That will come for free as in 1.19 there's no dual stack fields, which is exactly the reason why we need to fallback to ClusterIP

we have annotations that someone mistakenly can use with older k8s right ?

fedepaol:
> > > > > I feel if we are in this mode we should fail any attempt to provision svc with dual stack ?
> > > > 
> > > > 
> > > > That will come for free as in 1.19 there's no dual stack fields, which is exactly the reason why we need to fallback to ClusterIP
> > > 
> > > 
> > > we have annotations that someone mistakenly can use with older k8s right ?
> > 
> > 
> > Yes, and that would fail because the ip families would not match
> 
> maybe worth adding testcase to make sure this should fail if it not already there

We don't have 1.19 in our CI lanes (and I don't think it's worth adding a lane for a k8s version that is not supported)

msherif1234:
> > > > > > I feel if we are in this mode we should fail any attempt to provision svc with dual stack ?
> > > > > 
> > > > > 
> > > > > That will come for free as in 1.19 there's no dual stack fields, which is exactly the reason why we need to fallback to ClusterIP
> > > > 
> > > > 
> > > > we have annotations that someone mistakenly can use with older k8s right ?
> > > 
> > > 
> > > Yes, and that would fail because the ip families would not match
> > 
> > 
> > maybe worth adding testcase to make sure this should fail if it not already there
> 
> We don't have 1.19 in our CI lanes (and I don't think it's worth adding a lane for a k8s version that is not supported)

I was asking about unit-test test case

fedepaol:
> > > > > > > I feel if we are in this mode we should fail any attempt to provision svc with dual stack ?
> > > > > > 
> > > > > > 
> > > > > > That will come for free as in 1.19 there's no dual stack fields, which is exactly the reason why we need to fallback to ClusterIP
> > > > > 
> > > > > 
> > > > > we have annotations that someone mistakenly can use with older k8s right ?
> > > > 
> > > > 
> > > > Yes, and that would fail because the ip families would not match
> > > 
> > > 
> > > maybe worth adding testcase to make sure this should fail if it not already there
> > 
> > 
> > We don't have 1.19 in our CI lanes (and I don't think it's worth adding a lane for a k8s version that is not supported)
> 
> I was asking about unit-test test case

Ah then yes, I can add it, and makes sense, you are right.

controller/service.go
```diff
@@ -260,3 +260,12 @@ func isEqualIPs(ipsA, ipsB []net.IP) bool {
 	})
 	return reflect.DeepEqual(ipsA, ipsB)
 }
+
```
msherif1234, line 263: shouldn't we have this function in ipfamily package ?

fedepaol: The reason why I put this here was that the ipfamily package is (was) about ips and families and independent from k8s concepts, so it's a bit borderline and felt more like a local util function.
I don't have a strong opinion though, `ipfamily.ForService` sounds as much as good. Will move it.

rata: heh, I started a similar comment, then realized the ipFamily pkg was agnostic of k8s and decided it was fine too.

I am all for `ipfamily.ForService`, but if you do it and decide it doesn't fit, I'm ok with how the code currently is too :)

fedepaol: I just pushed the change :-)

rata (approved):
LGTM, thanks!

PR #1240, opened 2022-02-24
Dev-env: force debug logs in the pods

fedepaol:
When running dev-env, force debug logs in speaker and controller to ease the debug.


zeeke:
LGTM

PR #1252, opened 2022-03-08
Add community CRD

liornoy:
In BGP, "community" is an optional attribute that can be passed to the
other BGP peers as an attached tag, that govern a certain behavior.
for more info, you can read [here](https://en.wikipedia.org/wiki/Border_Gateway_Protocol#Communities)

Currently, the MetalLB configmap implementation allows us
to create aliases for communities.
As part of the move to CRD based implementation, this commit
introduces the community CRD to reach feature parity.

The new community CRD has two fields:
Alias and Value, that are used for describing a community, i.e.:
```
kind: Community
metadata:
  name: no-advertise-community
  namespace: metallb-system
spec:
  alias: "NO_ADVERTISE"
  value: "65535:65282"
```
Also, added `// +kubebuilder:printcolumn` that will produce
a more verbose output when running "kubectl get community"
example:
```
NAMESPACE        NAME                  ALIAS          VALUE
metallb-system   no-adv-community      NO_ADVERTISE   65535:65282
metallb-system   no-export-community   NO_EXPORT      65535:65281
metallb-system   no-peer-community     NOPEER         65535:65284
```


liornoy:
(deleted for a typo in branch name)

PR #1253, opened 2022-03-09
Add Community CRD

liornoy:
In BGP, "community" is an optional attribute that can be passed to the
other BGP peers as an attached tag, that govern a certain behavior.
for more info, you can read [here](https://en.wikipedia.org/wiki/Border_Gateway_Protocol#Communities)

Currently, the MetalLB configmap implementation allows us
to create aliases for communities.
As part of the move to CRD based implementation, this commit
introduces the community CRD to reach feature parity.

The new community CRD can hold a collection of community definitions.
Example of the community YAML:
```
kind: Community
metadata:
  name: metallb-communities
  namespace: metallb-system
spec:
  communities:
  - name: NO_EXPORT
    value: 65535:65281
  - name: NO_ADVERTISE
    value: 65535:65282
```


manifests/metallb.yaml
```diff
@@ -1,5 +1,3 @@
-
```
sabinaaledort, line 1: did you manually delete these lines?

liornoy: yes

fedepaol:
@liornoy can you add an e2e tests that validates the CRD is working properly?

internal/k8s/controllers/config_controller.go
```diff
@@ -73,10 +73,17 @@ func (r *ConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		return ctrl.Result{}, err
 	}
 
+	var communities metallbv1beta1.CommunityList
+	if err := r.List(ctx, &communities, client.InNamespace(r.Namespace)); err != nil {
```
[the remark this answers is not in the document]
liornoy: done

liornoy:
> @liornoy can you add an e2e tests that validates the CRD is working properly?

Yes, sure. I think I prefer waiting for the CRD part 2 pr to merge and then rebase rather than doing it now and making you rebase. Plus your pr includes a new infra for the e2e with the new CRDs that I need.

msherif1234:
Hmm we have discussed this in the past and the decision was the avoid CRD overhead for this config and just use pass by value similar to what we have in the released code ? why we changing this now ?

fedepaol:
> Hmm we have discussed this in the past and the decision was the avoid CRD overhead for this config and just use pass by value similar to what we have in the released code ? why we changing this now ?

This is to have feature parity with what metallb offers today. We don't want to change the way we configure metallb and reduce the things a user can do.

msherif1234:
> > Hmm we have discussed this in the past and the decision was the avoid CRD overhead for this config and just use pass by value similar to what we have in the released code ? why we changing this now ?
> 
> This is to have feature parity with what metallb offers today. We don't want to change the way we configure metallb and reduce the things a user can do.

again that was what we discussed with Russel in the past and that was decision at the time he wasn't favor of spinning CR for this config just FYI

fedepaol:
> > > Hmm we have discussed this in the past and the decision was the avoid CRD overhead for this config and just use pass by value similar to what we have in the released code ? why we changing this now ?
> > 
> > 
> > This is to have feature parity with what metallb offers today. We don't want to change the way we configure metallb and reduce the things a user can do.
> 
> again that was what we discussed with Russel in the past and that was decision at the time he wasn't favor of spinning CR for this config just FYI

I understand, but that was when the CRDs were an alternative to the configmap. Now we are replacing the configmap, and I think we must offer the same degrees of configurability the users have today.

Also, from a user's perspective this is not adding any overhead, since they can still use the communities in the number format as we currently support in the operator, or add the aliases.

fedepaol:
@liornoy this needs to be rebased

e2etest/bgptests/bgp.go
```diff
@@ -55,7 +55,6 @@ const (
 	v4PoolAddresses      = "192.168.10.0/24"
 	v6PoolAddresses      = "fc00:f853:0ccd:e799::/124"
 	CommunityNoAdv       = "65535:65282" // 0xFFFFFF02: NO_ADVERTISE
-	IPLocalPref          = uint32(300)
```
fedepaol, line 58: Just "Remove unused const" in the commit subject, no need to tell it's a small fix

api/v1beta1/community_types.go
```diff
@@ -0,0 +1,64 @@
...
+limitations under the License.
+*/
+
+package v1beta1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type CommunityDefinition struct {
```
fedepaol, line 27: I'd call this "CommunityAlias"

internal/config/config.go
```diff
@@ -40,6 +40,7 @@ type ClusterResources struct {
 	BGPAdvs            []metallbv1beta1.BGPAdvertisement `json:"bgpadvertisements"`
 	L2Advs             []metallbv1beta1.L2Advertisement  `json:"l2advertisements"`
 	LegacyAddressPools []metallbv1beta1.AddressPool      `json:"legacyaddresspools"`
+	Communities        []metallbv1beta1.Community        `json:"communities"`
```
fedepaol, line 43: Nit about the commit message: try to be impersonal.

```
The new logic is tested under config_test.go,
Additional tests were added.
Pools were added to the tests involving BGPAdv, because otherwise it would pass for the wrong reason (such as lack of IPPool definition)
```

internal/config/config_test.go
```diff
@@ -788,41 +799,266 @@ func TestParse(t *testing.T) {
 			},
 		},
 		{
-			desc: "bad community literal (wrong format)",
+			desc: "bad community literal (wrong format) - in BGP adv",
 			crs: ClusterResources{
+				Pools: []v1beta1.IPPool{
```
fedepaol, line 804: How about having a variable with this pool and using it in all these tests?
The volume of the test will be smaller, and the content is not really relevant for the test.

api/v1beta1/bgpadvertisement_webhook.go
```diff
@@ -56,8 +56,13 @@ func (bgpAdv *BGPAdvertisement) ValidateCreate() error {
 		return err
 	}
 
+	communities, err := getExistingCommunities()
```
fedepaol, line 59: Same for the commit message, try to be impersonal.
Also, nit in existing

e2etest/bgptests/bgp.go
```diff
@@ -1005,6 +1005,19 @@ var _ = ginkgo.Describe("BGP", func() {
 
 				resources := metallbconfig.ClusterResources{
 					Peers: metallb.PeersForContainers(FRRContainers, ipFamily),
+					Communities: []metallbv1beta1.Community{
```
fedepaol, line 1007: By doing this, we are not testing metallb anymore with communities written in the advertisement.
I would change the table to accept:
- the adv community content (that can either be the alias or the community itself)
- the (optional) []metallbv1beta1.Community to add to resources

In this way, you'll be able to add a couple of entries that test the community crd AND the previous implementation (I'd avoid duplicating all the combinations).

e2etest/bgptests/infra_setup.go
```diff
@@ -125,6 +125,8 @@ func InfraSetup(ipv4Addresses, ipv6Addresses []string, cs *clientset.Clientset)
 			ASN:     externalASN,
 			BGPPort: 179,
 		},
+		HostIPv4: hostIPv4,
```
fedepaol, line 128: No need to say "small fix" in the commit subject.
Just say what we are doing here (and how, in the body).

Also, same impersonal comment:

`ebgpSingleHopContainerConfig` is missing the following:
		HostIPv4: hostIPv4,
		HostIPv6: hostIPv6,

so we add them because blablabla

api/v1beta1/community_types.go
liornoy: Done.

internal/config/config_test.go
liornoy: Done.

e2etest/bgptests/infra_setup.go
liornoy: Sound and clear. thanks.

e2etest/bgptests/bgp.go
liornoy: Sure, added for ipv4, ipv6, and ipv4 legacy.

api/v1beta1/bgpadvertisement_webhook.go
liornoy: what is the nit with getExistingCommunities?

api/v1beta1/bgpadvertisement_webhook.go
```diff
@@ -56,8 +56,13 @@ func (bgpAdv *BGPAdvertisement) ValidateCreate() error {
...
+	if err != nil {
+		return err
+	}
+
 	toValidate := bgpAdvListWithUpdate(existingBGPAdvList, bgpAdv)
-	err = Validator.Validate(toValidate, addressPools, ipPools)
+	err = Validator.Validate(toValidate, addressPools, ipPools, communities)
```
fedepaol, line 65: Since we have said we should not check for the presence of the community alias, I wonder if we should remove this whole commit

liornoy: I thought about it and I think we should.
The bgpAdv webhook should be able to handle the communities part
of the spec without the Community CRs.

liornoy: Probably the same for addressPool

fedepaol: Nope, addresspool is different because it makes sense to validate the consistency with what we know it's there. It's not about validating that something that should be there is not.

e2etest/bgptests/bgp.go
```diff
@@ -1065,6 +1078,7 @@ var _ = ginkgo.Describe("BGP", func() {
 					validateService(cs, svcNoAdvertisement, allNodes.Items, c)
 					Eventually(func() error {
 						for _, community := range advertisement.Spec.Communities {
+							community = tryTranslate(community, communitiesCRs)
```
fedepaol, line 1080: I'd rather be explicit and add the value we expect as an extra parameter.

e2etest/bgptests/bgp.go
```diff
@@ -1416,3 +1478,17 @@ func substringCount(substr string) interface{} {
 		return strings.Count(action, substr)
 	}
 }
+
+// tryTranslate checks if given Community string is alias of community value
+// in Community CRs collection.
+func tryTranslate(community string, cs []metallbv1beta1.Community) string {
```
fedepaol, line 1483: As a general note (I suggested  to remove this), we could have a better name for this.
Why try? What happens if it fails?
communityForAlias would be a better name, and I'd return an error in case we don't find a valid alias.

internal/config/config.go
```diff
@@ -693,7 +714,7 @@ func validateBGPAdvPerPool(adv *BGPAdvertisement, pool *Pool) error {
 func ParseCommunity(c string) (uint32, error) {
 	fs := strings.Split(c, ":")
 	if len(fs) != 2 {
-		return 0, fmt.Errorf("invalid community string %q", c)
+		return 0, TransientError{fmt.Sprintf("invalid community string %q", c)}
```
fedepaol, line 717: Why this is transient? If the parsing failed it will never pass.
The only case where it makes sense to have it transient is, when we have a string referencing a non existing community

internal/config/config.go
```diff
@@ -484,7 +498,10 @@ func addressPoolFromLegacyCR(p metallbv1beta1.AddressPool, bgpCommunities map[st
 		}
 		ret.L2Advertisements = []*L2Advertisement{{}}
 	case BGP:
-		ads, err := bgpAdvertisementsFromLegacyCR(p.Spec.BGPAdvertisements, cidrsPerAddresses, bgpCommunities)
+		ads, err := bgpAdvertisementsFromLegacyCR(p.Spec.BGPAdvertisements, cidrsPerAddresses, communities)
+		if errors.As(err, &TransientError{}) {
```
fedepaol, line 502: Same as above

internal/config/config.go
```diff
@@ -288,6 +282,9 @@ func For(resources ClusterResources, validate Validate) (*Config, error) {
 
 	for _, p := range resources.LegacyAddressPools {
 		pool, err := addressPoolFromLegacyCR(p, communities)
+		if errors.As(err, &TransientError{}) {
```
fedepaol, line 285: no need to do this. The Transient check is done in the validator.
If your concern is that the error is replaced by fmt.Errorf, use `errors.Wrapf`. See https://go.dev/blog/go1.13-errors 
"When operating on wrapped errors, however, these functions consider all the errors in a chain."

liornoy: fixed. thanks for the reference.

internal/config/config.go
liornoy: moved the TransientError out of the ParseComunnity function

e2etest/bgptests/bgp.go
liornoy: done.

internal/config/config.go
```diff
@@ -661,6 +659,27 @@ func bgpAdvertisementsFromLegacyCR(ads []metallbv1beta1.LegacyBgpAdvertisement,
 	return ret, nil
 }
 
+func (ad *BGPAdvertisement) communtiesForBGPAd(adCommunities []string, communities map[string]uint32) error {
```
fedepaol, line 662: typo: communties
Also, communitiesFor... implies that this is returning an object, which is not.
Does this have to be a method? It would be clearer as a function and having the community map assigned outside.

internal/config/config.go
```diff
@@ -289,7 +283,7 @@ func For(resources ClusterResources, validate Validate) (*Config, error) {
 	for _, p := range resources.LegacyAddressPools {
 		pool, err := addressPoolFromLegacyCR(p, communities)
 		if err != nil {
-			return nil, fmt.Errorf("parsing address pool %s: %s", p.Name, err)
+			return nil, fmt.Errorf("parsing address pool %s: %w", p.Name, err)
```
fedepaol, line 286: why this change?

internal/config/config.go
```diff
@@ -661,6 +659,27 @@ func bgpAdvertisementsFromLegacyCR(ads []metallbv1beta1.LegacyBgpAdvertisement,
 	return ret, nil
 }
 
+func (ad *BGPAdvertisement) communtiesForBGPAd(adCommunities []string, communities map[string]uint32) error {
+	for _, c := range adCommunities {
+		v, err := ParseCommunity(c)
+		if err != nil {
+			if strings.Contains(err.Error(), "invalid community string") {
```
fedepaol, line 666: Let's use an instance of error and errors.Is

internal/config/config.go
```diff
@@ -661,6 +659,27 @@ func bgpAdvertisementsFromLegacyCR(ads []metallbv1beta1.LegacyBgpAdvertisement,
...
+				// Check if the community exists.
+				if v, ok := communities[c]; ok {
```
fedepaol, line 668: Inverting will remove the need for else:
!ok {
   return the error
}

internal/config/config.go
```diff
@@ -661,6 +659,27 @@ func bgpAdvertisementsFromLegacyCR(ads []metallbv1beta1.LegacyBgpAdvertisement,
...
+					ad.Communities[v] = true
+					continue
+				} else {
+					return TransientError{fmt.Sprintf("No such community exists: %s", err)}
+				}
+			} else {
```
fedepaol, line 674: I'd move this on top and return early, will save another else.

internal/config/config.go
```diff
@@ -661,6 +659,27 @@ func bgpAdvertisementsFromLegacyCR(ads []metallbv1beta1.LegacyBgpAdvertisement,
 	return ret, nil
 }
 
+func (ad *BGPAdvertisement) communtiesForBGPAd(adCommunities []string, communities map[string]uint32) error {
+	for _, c := range adCommunities {
+		v, err := ParseCommunity(c)
```
fedepaol, line 664: This is hard to follow, let's try to untangle

internal/config/config.go
fedepaol: Ah, didn't know about it

internal/config/config.go
```diff
@@ -670,23 +677,33 @@ func bgpAdvertisementsFromLegacyCR(ads []metallbv1beta1.LegacyBgpAdvertisement,
...
 
+func getCommunityValue(community string, communities map[string]uint32) (uint32, error) {
+	if v, ok := communities[community]; ok {
+		return v, nil
+	}
+
+	v, err := ParseCommunity(community)
+	if errors.Is(err, ErrFormat) {
+		return 0, err
+	} else if errors.Is(err, ErrInvalid) {
```
fedepaol, line 700: nit: no need for the else, just another if

internal/config/config.go
```diff
@@ -670,23 +677,33 @@ func bgpAdvertisementsFromLegacyCR(ads []metallbv1beta1.LegacyBgpAdvertisement,
...
 
 	return ret, nil
 }
 
+func getCommunityValue(community string, communities map[string]uint32) (uint32, error) {
+	if v, ok := communities[community]; ok {
+		return v, nil
+	}
+
+	v, err := ParseCommunity(community)
```
fedepaol, line 697: looks much better!

internal/config/config.go
```diff
@@ -716,18 +733,21 @@ func validateBGPAdvPerPool(adv *BGPAdvertisement, pool *Pool) error {
 	return nil
 }
 
+var ErrInvalid = errors.New("invalid string")
```
fedepaol, line 736: nit: no need to make these public, also I'd name them more explicitly:
errInvalidCommunity (or just invalidCommunity / invalidCommunityFormat)

internal/config/config.go
liornoy: fixed

internal/config/config.go
liornoy: fixed

internal/config/config.go
liornoy: done

internal/config/config.go
liornoy: done

internal/config/config.go
liornoy: done

internal/config/config.go
liornoy: done

internal/config/config.go
liornoy: done

internal/k8s/controllers/pool_controller.go
```diff
@@ -57,9 +57,16 @@ func (r *PoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		return ctrl.Result{}, err
 	}
 
+	var communities metallbv1beta1.CommunityList
+	if err := r.List(ctx, &communities, client.InNamespace(r.Namespace)); err != nil {
```
fedepaol, line 61: You need to add the watch below (line 99)

fedepaol:
Just one last thing to change, looks great!

internal/k8s/controllers/pool_controller.go
liornoy: done

fedepaol:
LGTM

PR #1254, opened 2022-03-10
Manifests: reorganize with kustomize 

sabinaaledort:
This reorganizes the manifests so we can use kustomize for more flexible deployment and to support the next steps
of moving the CRDs from the operator (for example adding webhooks).


fedepaol:
I'd still keep the manifests, (at least the current flavours we have, metallb and metallb-frr) but generated by kustomize.
On top of that, we must ensure via CI that the manifests do not drift from kustomize.

tasks.py
```diff
@@ -252,6 +252,12 @@ def validate_kind_version():
     if delta < 0:
         raise Exit(message="kind version >= {} required".format(min_version))
 
+@task(help={
```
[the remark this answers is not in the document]
sabinaaledort: I forgot to add the help, thanks.

tasks.py
```diff
@@ -333,15 +333,12 @@ def dev_env(ctx, architecture="amd64", name="kind", protocol=None, frr_volume_di
     else:
         run("kubectl delete po -nmetallb-system --all", echo=True)
 
-        manifests_dir = os.getcwd() + "/manifests"
+        generate_manifests(ctx)
+        manifests_dir = "config/frr" if bgp_type == "frr" else "config/native"
```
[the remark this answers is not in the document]
sabinaaledort: It means having two more layers (one to kustomize native and one to kustomize frr) which I think is too much since we are also going to have more layers for the webhooks.

tasks.py
```diff
@@ -327,13 +352,12 @@ def dev_env(ctx, architecture="amd64", name="kind", protocol=None, frr_volume_di
     else:
         run("kubectl delete po -nmetallb-system --all", echo=True)
 
-        manifests_dir = os.getcwd() + "/manifests"
+        generate_manifests(ctx, bgp_type=bgp_type)
```
fedepaol, line 355: I am a bit afraid of this, because running dev-env will have a possible side effect of changing the manifests.
I think a cleaner way is to have generate_manifests to take the output path as an optional parameter, and let it generate directly to the temp path. If we do that, we won't even need to tweak the manifest name

tasks.py
```diff
@@ -252,6 +252,31 @@ def validate_kind_version():
...
+    "crd_options": "CRD Options."
+                    "Default: crd:crdVersions=v1.",
+    "kustomize_cli": "YAML files customization CLI."
+                    "Default: kustomize (version v4.4.0).",
+    "bgp_type": "Type of BGP implementation to use."
+                    "Supported: 'native' (default), 'frr'",
+    "namespace": "MetalLB deployment namespace."
+                    "Default: 'metallb-system'.",
+})
+def generate_manifests(ctx, controller_gen="controller-gen", crd_options="crd:crdVersions=v1",
```
fedepaol, line 267: This is more `generate_manifest` as it's getting one single type

website/content/installation/_index.md
```diff
@@ -59,16 +59,14 @@ kubectl apply -f - -n kube-system
...
+kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/main/config/manifests/metallb-native.yaml
 ```
 
 {{% notice note %}}
 If you want to deploy MetalLB using the [experimental FRR mode](https://metallb.universe.tf/configuration/#enabling-bfd-support-for-bgp-sessions), apply the manifests:
 
 ```bash
-kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/main/manifests/namespace.yaml
-kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/main/manifests/metallb-frr.yaml
+kubectl apply -f https://raw.githubusercontent.com/metallb/metallb/main/config/manifests/metallb-frr.yaml
```
fedepaol, line 69: You need to amend / extend the kustomize section below

fedepaol:
LGTM

PR #1245, opened 2022-03-03
Move the CRDs to MetalLB, step 2

fedepaol:
This is based on https://github.com/metallb/metallb/pull/1237 , and is the second part related to what's being proposed in https://github.com/metallb/metallb/pull/942

Here we move from the version of CRDs shipped with the operator to the final design.
We still lack a CRD for the communities and a label selector for selecting the addresspool from the L2 and the BGPAdvertisements, which are gonna be added as a next step.

Also, the e2e tests are updated to reflect this new structure.

PR #1264, opened 2022-03-15
Store BGP passwords in secrets

pperiyasamy:
provides a feature to store bgp peer passwords in secrets instead of plain text.


fedepaol:
@sabinaaledort mind reviewing this?

internal/config/config.go
```diff
@@ -319,9 +325,35 @@ func peerFromCR(p metallbv1beta1.BGPPeer) (*Peer, error) {
...
+			if srcPass, ok = secret.Data["password"]; !ok {
+				return nil, fmt.Errorf("password not specified in the secret %q", p.Spec.SecretRef)
+			}
+			dstPass, err := base64.StdEncoding.DecodeString(string(srcPass))
+			if err != nil {
+				return nil, err
+			}
+			password = string(dstPass)
+		} else {
+			return nil, fmt.Errorf("invalid secret specified in the secret %q", p.Spec.SecretRef)
```
sabinaaledort, line 355: can we give more info in the err msg? maybe mentioning the type mismatch.

api/v1beta1/bgppeer_types.go
```diff
@@ -81,6 +82,10 @@ type BGPPeerSpec struct {
 	// +optional
 	Password string `json:"password,omitempty" yaml:"password,omitempty"`
 
+	// secretRef is name of the authentication secret for BGP Peer
```
sabinaaledort, line 85: secretRef is the name or a reference? i'm still not sure how it's going to look like

sabinaaledort:
looks good. the only thing i'm not sure about is how this field is going to look like. maybe worth adding an example or a unit test? (or maybe it's just me that don't get it)

api/v1beta1/bgppeer_types.go
```diff
@@ -81,6 +82,10 @@ type BGPPeerSpec struct {
...
+	// +optional
+	SecretRef *v1.LocalObjectReference `json:"secretRef,omitempty" yaml:"secretRef,omitempty"`
```
fedepaol, line 87: I just found out that there's a corev1.SecretReference which is maybe better https://pkg.go.dev/k8s.io/api/core/v1#SecretReference

fedepaol, line 87: I'd rename this as PasswordSecret, meaning the secret that holds the password.

internal/config/config.go
```diff
@@ -169,7 +175,7 @@ func validateHoldTime(ht time.Duration) error {
 }
 
 // Parse loads and validates a Config from bs.
-func For(resources ClusterResources, validate Validate) (*Config, error) {
+func For(ctx context.Context, cli client.Client, resources ClusterResources, validate Validate) (*Config, error) {
```
fedepaol, line 178: This function is pure, meaning it has no side effects and the result does not depend on the system state.
It's important that stays this way, because otherwise we'd need to inject a mock of the client to test it.

So, we need to collect the secrets upfront and pass them here, in a way or another.

internal/config/config.go
```diff
@@ -319,9 +325,35 @@ func peerFromCR(p metallbv1beta1.BGPPeer) (*Peer, error) {
 		}
 	}
 
+	if p.Spec.Password != "" && p.Spec.SecretRef != nil {
+		return nil, fmt.Errorf("can not have both password and secret ref set in the peer config")
```
fedepaol, line 329: Return the peer name, it's gonna be easier for the user to understand which bgppeer is wrong

internal/config/config.go
```diff
@@ -319,9 +325,35 @@ func peerFromCR(p metallbv1beta1.BGPPeer) (*Peer, error) {
...
+		err := cli.Get(ctx, types.NamespacedName{Namespace: p.Namespace, Name: p.Spec.SecretRef.Name}, secret)
+		if err != nil {
+			return nil, err
+		}
+		if secret.Type == corev1.SecretTypeOpaque || secret.Type == corev1.SecretTypeBasicAuth {
+			var (
+				srcPass []byte
+				ok      bool
+			)
+			if srcPass, ok = secret.Data["password"]; !ok {
```
fedepaol, line 346: I'd wrap the parsing in a function.

api/v1beta1/bgppeer_types.go
fedepaol: Also, I'd add this in the v1beta2 version (which https://github.com/metallb/metallb/pull/1245 is adding)

e2etest/pkg/config/update.go
```diff
@@ -152,7 +163,26 @@ func (o beta1Updater) poolToCR(a AddressPool) *operatorv1beta1.AddressPool {
 }
 
 func (o beta1Updater) createPeer(p Peer) error {
-	peer, err := o.peerToCR(p)
+	secretName := fmt.Sprintf("bgppeer-%d-%d-secret", p.ASN, p.Port)
+	existing := &corev1.Secret{}
+	err := o.Client.Get(context.TODO(), types.NamespacedName{Name: secretName, Namespace: o.namespace}, existing)
```
fedepaol, line 168: We need to ensure both work, so I'd add a specific test that just fills the secret and checks the session is established (as opposed to changing all the tests)

internal/config/config.go
```diff
@@ -319,9 +325,35 @@ func peerFromCR(p metallbv1beta1.BGPPeer) (*Peer, error) {
 		}
 	}
 
+	if p.Spec.Password != "" && p.Spec.SecretRef != nil {
```
fedepaol, line 328: Can you add a unit test that verifies this fails?

api/v1beta1/bgppeer_types.go
pperiyasamy: changed it to use `corev1.SecretReference `and added few unit tests. hope that explains it clearly.

api/v1beta1/bgppeer_types.go
pperiyasamy: good find! done.

pperiyasamy: done

e2etest/pkg/config/update.go
pperiyasamy: yes, added specific e2e tests to cover this.

internal/config/config.go
pperiyasamy: sure, done.

internal/config/config.go
pperiyasamy: done, added relevant unit tests to cover this change.

internal/config/config.go
pperiyasamy: done

internal/config/config.go
```diff
@@ -319,9 +325,35 @@ func peerFromCR(p metallbv1beta1.BGPPeer) (*Peer, error) {
...
 	var password string
 	if p.Spec.Password != "" {
 		password = p.Spec.Password
+	} else if p.Spec.SecretRef != nil {
+		secret := &corev1.Secret{}
+		err := cli.Get(ctx, types.NamespacedName{Namespace: p.Namespace, Name: p.Spec.SecretRef.Name}, secret)
+		if err != nil {
+			return nil, err
+		}
+		if secret.Type == corev1.SecretTypeOpaque || secret.Type == corev1.SecretTypeBasicAuth {
```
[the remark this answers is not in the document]
pperiyasamy: done

internal/config/config.go
pperiyasamy: done

internal/config/config.go
pperiyasamy: sure, done.

manifests/metallb-frr.yaml
```diff
@@ -577,6 +577,7 @@ rules:
   - services
   - endpoints
   - nodes
+  - secrets
```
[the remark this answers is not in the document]
pperiyasamy: I tried with having secrets in Role, but below error occurred from reconciler while listing secrets from metallb-system namespace, so still holding it in cluster role. 
`E0317 23:21:27.475103     100 reflector.go:138] pkg/mod/k8s.io/client-go@v0.20.2/tools/cache/reflector.go:167: Failed to watch *v1.Secret: failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:metallb-system:speaker" cannot list resource "secrets" in API group "" at the cluster scope`
Any idea what can be done to avoid this ?

pperiyasamy (commented):
Thanks for great design inputs ! rebased it with new changes.

internal/config/config.go
```diff
@@ -370,9 +372,23 @@ func peerFromCR(p metallbv1beta2.BGPPeer) (*Peer, error) {
...
+			p.Name)
+	}
+
 	var password string
 	if p.Spec.Password != "" {
 		password = p.Spec.Password
+	} else if p.Spec.PasswordSecret.Name != "" {
+		var secret corev1.Secret
+		var ok bool
+		if secret, ok = passwordSecrets[p.Spec.PasswordSecret.Name]; !ok {
```
fedepaol, line 386: I'd rather have the := form and checking if !ok on the next line
secret, ok: = ...
if !ok {
}

internal/config/config.go
```diff
@@ -370,9 +372,23 @@ func peerFromCR(p metallbv1beta2.BGPPeer) (*Peer, error) {
 		nodeSels = []labels.Selector{labels.Everything()}
 	}
 
+	if p.Spec.Password != "" && p.Spec.PasswordSecret.Name != "" {
```
fedepaol, line 375: How about wrapping this all in a `passwordForPeer` function ?

internal/config/config.go
```diff
@@ -391,6 +407,21 @@ func peerFromCR(p metallbv1beta2.BGPPeer) (*Peer, error) {
...
+func parsePasswordFromSecret(secret *corev1.Secret) (string, error) {
+	if secret.Type != corev1.SecretTypeBasicAuth {
+		return "", fmt.Errorf("secret type mismatch on %q/%q, type %q is expected ", secret.Namespace,
+			secret.Name, corev1.SecretTypeBasicAuth)
+	}
+	var (
+		srcPass []byte
+		ok      bool
+	)
+	if srcPass, ok = secret.Data["password"]; !ok {
```
fedepaol, line 419: Same: 
pass, ok := ...
if !ok {
}

internal/k8s/controllers/config_controller.go
```diff
@@ -92,18 +94,28 @@ func (r *ConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 	}
 
-	metallbCRs := config.ClusterResources{
+	var passwordSecrets corev1.SecretList
+	if err := r.List(ctx, &passwordSecrets, client.InNamespace(r.Namespace)); err != nil {
+		level.Error(r.Logger).Log("controller", "ConfigReconciler", "error", "failed to get secrets", "error", err)
+		return ctrl.Result{}, err
+	}
+
+	metallbCRsAndSecrets := config.ClusterResources{
```
fedepaol, line 103: I'd just call this "resources"

internal/k8s/controllers/config_controller.go
```diff
@@ -92,18 +94,28 @@ func (r *ConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
 		Pools:              ipPools.Items,
 		Peers:              bgpPeers.Items,
 		BFDProfiles:        bfdProfiles.Items,
 		L2Advs:             l2Advertisements.Items,
 		BGPAdvs:            bgpAdvertisements.Items,
 		LegacyAddressPools: addressPools.Items,
 	}
+	metallbCRsAndSecrets.PasswordSecrets = make(map[string]corev1.Secret)
```
fedepaol, line 111: I'd wrap this (and the List above) in a func to put in the bottom of the file, and then use the result when you do the literal initializer above.

e2etest/pkg/metallb/config.go
```diff
@@ -74,3 +85,22 @@ func WithRouterID(peers []metallbv1beta2.BGPPeer, routerID string) []metallbv1be
 	}
 	return peers
 }
+
+func BGPPeerSecretRefernces(containers []*frrcontainer.FRR) map[string]corev1.Secret {
```
fedepaol, line 89: nit: BGPPeerSercretReferences

e2etest/pkg/metallb/config.go
```diff
@@ -27,6 +28,11 @@ func init() {
 
 // PeersForContainers returns the metallb config peers related to the given containers.
 func PeersForContainers(containers []*frrcontainer.FRR, ipFamily ipfamily.Family) []metallbv1beta2.BGPPeer {
+	return PeersForContainersWithSecret(containers, ipFamily, false)
+}
+
+// PeersForContainersWithSecret returns the metallb config peers related to the given containers.
+func PeersForContainersWithSecret(containers []*frrcontainer.FRR, ipFamily ipfamily.Family, useSecret bool) []metallbv1beta2.BGPPeer {
```
fedepaol, line 35: I would use a vararg with a modifier function, pretty much like we do in https://github.com/metallb/metallb/blob/10a02498f259870492a2f3992803f0a7e1a4995d/e2etest/pkg/service/create.go#L20

and expose a `WithSecret` func that does the modification.

fedepaol:
Re: using namespaced secrets, I filed https://github.com/metallb/metallb/pull/1287 today, which seems to make the trick.
I'd wait that pr to get in (that in turn depends on https://github.com/metallb/metallb/pull/1275), the changes will happen mostly at manifest level and a few lines in k8s.go.

internal/config/config.go
```diff
@@ -391,6 +393,32 @@ func peerFromCR(p metallbv1beta2.BGPPeer) (*Peer, error) {
...
 }
 
+func passwordForPeer(p metallbv1beta2.BGPPeer, passwordSecrets map[string]corev1.Secret) (string, error) {
+	if p.Spec.Password != "" && p.Spec.PasswordSecret.Name != "" {
+		return "", fmt.Errorf("can not have both password and secret ref set in peer config %q/%q", p.Namespace,
+			p.Name)
+	}
+	var password string
+	if p.Spec.Password != "" {
+		password = p.Spec.Password
```
fedepaol, line 403: you can just return here

e2etest/pkg/metallb/config.go
```diff
@@ -74,3 +78,22 @@ func WithRouterID(peers []metallbv1beta2.BGPPeer, routerID string) []metallbv1be
 	}
 	return peers
 }
+
+func BGPPeerSecretReferences(containers []*frrcontainer.FRR) map[string]corev1.Secret {
```
fedepaol, line 82: This can return a slice, I think?

pperiyasamy: I think this can be still be a map to align with [PasswordSecrets](https://github.com/metallb/metallb/blob/1e364f511ba9e29fb0ecbfe395392576376a82db/internal/config/config.go#L43) in ClusterResources, I chose map over slice as its mainly used for a secret lookup for secret ref in bgp peer.

fedepaol: You are right.

fedepaol:
@pperiyasamy https://github.com/metallb/metallb/pull/1287 was merged, you can now make the secrets namescoped too

pperiyasamy:
> @pperiyasamy #1287 was merged, you can now make the secrets namescoped too

Thanks @fedepaol for the PR https://github.com/metallb/metallb/pull/1287, now rebased it accordingly.

fedepaol:
LGTM, thanks @pperiyasamy !

fedepaol:
Hi @junkiebev , thanks for the feedback! You are right, the documentation is lacking details about this change, as it slipped under the other changes of the last release.
Would you mind filing an issue for this so we won't loose track of it? Thanks again.

PR #1275, opened 2022-03-21
Bump controller runtime / k8s api

fedepaol:
Following up the updates from dependabot, we bump up the versions of both the pinned k8s api and the version of the controller runtime.


.circleci/config.yml
```diff
@@ -10,23 +10,27 @@ commands:
...
 jobs:
-  test-1-16:
+  test-1-17:
     machine:
       image: ubuntu-2004:202104-01
     steps:
+      - run: |
+          curl -o /tmp/go1.17.1.linux-amd64.tar.gz https://dl.google.com/go/go1.17.1.linux-amd64.tar.gz
+          sudo rm -rf /usr/local/go
+          sudo tar -C /usr/local -xzf /tmp/go1.17.1.linux-amd64.tar.gz
```
liornoy, line 20: how come with go 1.16 we didn't have to curl?

fedepaol: the issue here is, we are using a vm (as opposed to the docker image for the other lanes) because frr tests are using docker. The version of go that came with it was old and not compatible with our codebase (some of the dependencies, actually).

oribon:
lgtm

Issue #1276, opened 2022-03-22
GH Actions publish job broken on main


johananl:
Following https://github.com/metallb/metallb/pull/1137 it seems the CI is broken on `main` right now:

https://github.com/metallb/metallb/runs/5634442521?check_suite_focus=true#step:7:10

>Error: Cannot perform an interactive login from a non TTY device

A Google search tells me this error can appear when the env vars for `docker login` aren't set. Looking at the repo's settings, it seems there are no secrets configured.

@fedepaol @tylerauerbeck

PR #1277, opened 2022-03-22
Revert "Add publish workflow (#1137)"

fedepaol:
This reverts commit 0c30dd32f09cb782dbcc85bcb94185a27ce7f559.

We found that secrets might be handled in a more effective way, and we
can use a github action to publish.
So, we temporary restore the old circleci publish workflow to restore
image publishing and prepare a better solution.

Fixes https://github.com/metallb/metallb/issues/1276


johananl (approved):
LGTM. Let's ensure CI is green before merging.

PR #1282, opened 2022-03-23
CI: enforce generated files

fedepaol:
There are currently two things we can check:
- go mod tidy, to ensure our go.mod is always clean
- the manifest generation

We add infrastructure to CI to validate those don't drift, and the committed code is always consistent.


fedepaol:
cc @sabinaaledort

oribon:
lgtm

PR #1285, opened 2022-03-24
Readme: Add a warning about consuming from main

fedepaol:
After removing the configmap, there were a few users complaining about
metallb not working anymore, mostly because they were consuming the main
branch. Here we add a note to warn about consuming main.


liornoy:
LGTM

PR #1286, opened 2022-03-24
Manifests: Fix: Add missing FRR config

sabinaaledort:
Part of the FRR config was missing in https://github.com/metallb/metallb/pull/1254


sabinaaledort:
fixed in https://github.com/metallb/metallb/pull/1282

PR #1293, opened 2022-03-30
E2E: Split L2 test to IPV4 and IPV6

sabinaaledort:
Since native bgp mode does not support ipv6 we get the following error in the log when
running the E2E tests in native mode ipv4 env:

"error":"pool \"l2-test\" has ipv6 CIDR fc00:f853:ccd:e793::7/128, native bgp mode does not support ipv6"

Splitting the test to ipv4 and ipv6 allows skipping the ipv6 tests in native mode ipv4 env.

### DIFF
--- a/e2etest/l2tests/l2.go
+++ b/e2etest/l2tests/l2.go
@@ -214,25 +214,36 @@ var _ = ginkgo.Describe("L2", func() {
 			err = wget.Do(address, executor.Host)
 			framework.ExpectNoError(err)
 		},
-			table.Entry("AddressPool defined by address range", func() []metallbv1beta1.IPPool {
+			table.Entry("IPV4 - AddressPool defined by address range", func() []metallbv1beta1.IPPool {
 				return []metallbv1beta1.IPPool{
 					{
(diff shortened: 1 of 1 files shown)


fedepaol:
L2 supports dual stack in both modes, and if I remember correctly we did not have failures for that. What am I missing?

sabinaaledort:
> L2 supports dual stack in both modes, and if I remember correctly we did not have failures for that. What am I missing?

maybe i'm missing something and it's time to remove it?
https://github.com/metallb/metallb/blob/13dbc189d59aaf100a6fc2e9770e16195f423afe/internal/config/validate.go#L40

sabinaaledort:
This PR can be ignored. The log appeared since we had a BGP test running before the L2 test which created BGPAdvertisements that weren't cleaned up yet since we are not waiting for the MetalLB resources to be deleted when doing the cleanup.
ConfigReconciler was called again after the error, this time without the BGPAdvertisements (which were already deleted by this time). The validation passed, the config was loaded successfully and the test passed.

PR #1294, opened 2022-03-30
E2E: reorder validation function call

liornoy:
In the e2e-test bgp.go, we create external docker FRR containers
and peer them to the kind cluster nodes.
We validate that the external FRR containers are configured well
by calling the function: validateFRRPeeredWithNodes.

This commit only re-positions the calls for this function
to after calling the pairing function, to make it more
clear that it's validating the previous step in the test.

### DIFF
--- a/e2etest/bgptests/bgp.go
+++ b/e2etest/bgptests/bgp.go
@@ -135,12 +135,12 @@ var _ = ginkgo.Describe("BGP", func() {
 			framework.ExpectNoError(err)
 		}
 
-		err := ConfigUpdater.Update(resources)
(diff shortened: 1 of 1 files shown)


fedepaol:
This won't work I guess? The pairing won't happen if we don't define the bgppeers

fedepaol:
Also, the order in "doing the setup -> calling Update" and asserting it works is fine as it is, I believe.

liornoy:
Thank you for the feedback. It appears that I misunderstood what kind of validation this function is doing.
I'll close this PR.

PR #1321, opened 2022-04-12
Add controllers unit tests

liornoy:
Adding unit tests to cover the controllers.
- [x] config_controller 
- [x] pool_controller
- [x] node_controller
- [x] service_controller

### DIFF
--- a/internal/k8s/controllers/config_controller_test.go
+++ b/internal/k8s/controllers/config_controller_test.go
@@ -1,0 +1,243 @@
+/*
+
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"context"
+	"testing"
+
+	"github.com/go-kit/kit/log"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
+	v1beta1 "go.universe.tf/metallb/api/v1beta1"
+	v1beta2 "go.universe.tf/metallb/api/v1beta2"
+	"go.universe.tf/metallb/internal/config"
+	metallbcfg "go.universe.tf/metallb/internal/config"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func TestConfigController(t *testing.T) {
+	tests := []struct {
+		desc                    string
+		handlerRes              SyncState
+		validResources          bool
+		expectReconcileFails    bool
+		expectForceReloadCalled bool
+	}{
+		{
+			desc:                    "handler returns SyncStateSuccess, valid resources",
+			handlerRes:              SyncStateSuccess,
+			validResources:          true,
+			expectReconcileFails:    false,
+			expectForceReloadCalled: false,
+		},
+		{
+			desc:                    "handler returns SyncStateError, valid resources",
+			handlerRes:              SyncStateError,
+			validResources:          true,
+			expectReconcileFails:    true,
+			expectForceReloadCalled: false,
+		},
+		{
+			desc:                    "handler returns SyncStateErrorNoRetry, valid resources",
+			handlerRes:              SyncStateErrorNoRetry,
+			validResources:          true,
+			expectReconcileFails:    false,
+			expectForceReloadCalled: false,
+		},
+		{
+			desc:                    "handler returns SyncStateReprocessAll, valid resources",
+			handlerRes:              SyncStateReprocessAll,
+			validResources:          true,
+			expectReconcileFails:    false,
+			expectForceReloadCalled: true,
+		},
+		{
+			desc:                    "handler returns SyncStateSuccess, invalid resources",
+			handlerRes:              SyncStateSuccess,
+			validResources:          false,
+			expectReconcileFails:    false,
+			expectForceReloadCalled: false,
+		},
+	}
+	for _, test := range tests {
+		var resources metallbcfg.ClusterResources
+		if test.validResources {
+			resources = configControllerValidResources
+		} else {
+			resources = configControllerInvalidResources
+		}
+
+		initObjects := objectsFromResources(resources)
+		fakeClient, err := newFakeClient(initObjects)
+		if err != nil {
+			t.Fatalf("test %s failed to create fake client: %v", test.desc, err)
+		}
+
+		expectedCfg, err := config.For(resources, config.DontValidate)
+		if err != nil && test.validResources {
+			t.Fatalf("test %s failed to create config, got unexpected error: %v", test.desc, err)
+		}
+
+		cmpOpt := cmpopts.IgnoreUnexported(metallbcfg.Pool{})
+
+		mockHandler := func(l log.Logger, cfg *config.Config) SyncState {
+			if !cmp.Equal(expectedCfg, cfg, cmpOpt) {
+				t.Errorf("test %s failed, handler called with unexpected config: %s", test.desc, cmp.Diff(expectedCfg, cfg, cmpOpt))
+			}
+			return test.handlerRes
+		}
+
+		calledForceReload := false
+		mockForceReload := func() { calledForceReload = true }
+
+		r := &ConfigReconciler{
+			Client:         fakeClient,
+			Logger:         log.NewNopLogger(),
+			Scheme:         scheme,
+			Namespace:      testNamespace,
+			ValidateConfig: config.DontValidate,
+			Handler:        mockHandler,
+			ForceReload:    mockForceReload,
+		}
+		req := reconcile.Request{
+			NamespacedName: types.NamespacedName{
+				Namespace: testNamespace,
+			},
+		}
+
+		_, err = r.Reconcile(context.TODO(), req)
+		failedReconcile := err != nil
+
+		if test.expectReconcileFails != failedReconcile {
+			t.Errorf("%s: fail reconcile expected: %v, got: %v. err: %v", test.desc, test.expectReconcileFails, failedReconcile, err)
+		}
+
+		if test.expectForceReloadCalled != calledForceReload {
+			t.Errorf("%s: call force reload expected: %v, got: %v", test.desc, test.expectForceReloadCalled, calledForceReload)
+		}
+	}
+}
+
+var (
+	testNamespace                  = "test-controller"
+	scheme                         = runtime.NewScheme()
+	configControllerValidResources = metallbcfg.ClusterResources{
+		Peers: []v1beta2.BGPPeer{
+			{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "peer1",
+					Namespace: testNamespace,
+				},
+				Spec: v1beta2.BGPPeerSpec{
+					MyASN:      42,
+					ASN:        142,
+					Address:    "1.2.3.4",
+					BFDProfile: "default",
+				},
+			},
+		},
+		BFDProfiles: []v1beta1.BFDProfile{
+			{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "default",
+					Namespace: testNamespace,
+				},
+			},
+		},
+		Pools: []v1beta1.IPAddressPool{
+			{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "pool1",
+					Namespace: testNamespace,
+				},
+				Spec: v1beta1.IPAddressPoolSpec{
+					Addresses: []string{
+						"10.20.0.0/16",
+					},
+				},
+			},
+		},
+		BGPAdvs: []v1beta1.BGPAdvertisement{
+			{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "adv1",
+					Namespace: testNamespace,
+				},
+				Spec: v1beta1.BGPAdvertisementSpec{
+					Communities: []string{"bar"},
+				},
+			},
+		},
+		L2Advs: []v1beta1.L2Advertisement{
+			{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "l2adv1",
+					Namespace: testNamespace,
+				},
+			},
+		},
+		PasswordSecrets: map[string]corev1.Secret{
+			"bgpsecret": {Type: corev1.SecretTypeBasicAuth, ObjectMeta: metav1.ObjectMeta{Name: "bgpsecret", Namespace: testNamespace},
+				Data: map[string][]byte{"password": []byte([]byte("nopass"))}},
+		},
+		LegacyAddressPools: []v1beta1.AddressPool{
+			{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "legacypool1",
+					Namespace: testNamespace,
+				},
+				Spec: v1beta1.AddressPoolSpec{
+					Addresses: []string{
+						"10.21.0.0/16",
+					},
+					Protocol: "bgp",
+				},
+			},
+		},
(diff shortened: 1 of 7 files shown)


internal/k8s/controllers/config_controller_test.go
```diff
@@ -0,0 +1,335 @@
+/*
```
fedepaol, line 1: Just so you know, no need to copy this manually. On new files, you can run inv bumplicense

internal/k8s/controllers/config_controller_test.go
```diff
@@ -0,0 +1,335 @@
...
+		expectedCfg, err := config.For(resources, config.DontValidate)
+		if err != nil {
+			t.Fatalf("test %s failed, got unexpected error: %v", test.desc, err)
+		}
+
+		initObjects := objectsFromResources(resources)
+		fakeClient := fake.NewClientBuilder().WithScheme(scheme).WithObjects(initObjects...).Build()
+
+		mockHandler := func(l log.Logger, cfg *config.Config) SyncState {
+			if diff := cmp.Diff(expectedCfg, cfg, cmpopts.IgnoreUnexported(metallbcfg.Pool{})); diff != "" {
```
fedepaol, line 267: use cmp.Cmp and print the diff in case of failure

internal/k8s/controllers/config_controller_test.go
```diff
@@ -0,0 +1,335 @@
...
+		req := reconcile.Request{
+			NamespacedName: types.NamespacedName{
+				Namespace: testNamespace,
+			},
+		}
+
+		_, err = r.Reconcile(context.TODO(), req)
+
+		if err != nil && test.retSyncState != SyncStateError {
+			t.Fatalf("%s: shouldn't return error: %v", test.desc, err)
```
fedepaol, line 290: Why the difference between fatal and errorf?

internal/k8s/controllers/config_controller_test.go
```diff
@@ -0,0 +1,335 @@
...
+		t.Fatalf("add to scheme failed: %s", err)
+	}
+
+	if err := corev1.AddToScheme(scheme); err != nil {
+		t.Fatalf("add to scheme failed: %s", err)
+	}
+
+	tests := []struct {
+		desc         string
+		retSyncState SyncState
```
fedepaol, line 237: handlerRes
I would also add here the expected outcome from reconcile, it will untangle all the if-s below there

internal/k8s/controllers/config_controller_test.go
```diff
@@ -0,0 +1,335 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func TestConfigController(t *testing.T) {
+	var (
+		testNamespace = "test-controller"
+		scheme        = runtime.NewScheme()
+		// Defining config using all features.
+		resources = metallbcfg.ClusterResources{
```
fedepaol, line 46: I don't think we need to have this so fully detailed.
What we want is to make sure that all the crds are passed to config.For , so it's sufficient to have the names.
If we need to have the full details because config.For would fail otherwise, let's move this as a global to the bottom of the file, so it won't distract from the test (also, give it a name so it's clear that we don't care about the content, like "validConfig". Also also, I'd test the scenario where config.For returns error, so it makes sense to have also an "invalidConfig". At that point, config can be a parameter of the test

internal/k8s/controllers/config_controller_test.go
```diff
@@ -0,0 +1,335 @@
...
+					},
+				},
+			},
+		}
+
+		calledForceReload = false
+		mockForceReload   = func() { calledForceReload = true }
+	)
+
+	if err := v1beta1.AddToScheme(scheme); err != nil {
```
fedepaol, line 223: This is is distracting. Wrap everything in a "newFakeClient" function that you put in the bottom of the file

internal/k8s/controllers/config_controller_test.go
liornoy: It's my bad, wanted it to be errorf here.
however on the err := corev1.AddToScheme(scheme) I return fatal because this error is not related to the test,
but to the environment.
does it sound right to you, or do we want to always return fatal?

fedepaol: Here we may want to use error, just because it's a table test. We don't want to stop at the first failure but let the other entries of the table to run.

liornoy: on fatal the test fails and stops running the TestConfigController function.
on error it fails but continues to other test cases.
so it made me believe to add this separation between test-logic fail (errorf) and env-related fail (fatal)

internal/k8s/controllers/config_controller_test.go
liornoy: done

internal/k8s/controllers/config_controller_test.go
liornoy: done

internal/k8s/controllers/config_controller_test.go
liornoy: this is a great suggestion thanks!

internal/k8s/controllers/config_controller_test.go
liornoy: done!

internal/k8s/controllers/config_controller_test.go
```diff
@@ -0,0 +1,265 @@
...
+		failReconcile   bool
+		callForceReload bool
+		validResources  bool
+	}{
+		{
+			desc:            "handler returns SyncStateSuccess, valid resources",
+			handlerRes:      SyncStateSuccess,
+			failReconcile:   false,
+			callForceReload: false,
+			validResources:  true,
```
fedepaol, line 52: nit: move validResources above, as it's an "input" for the tests, so looking at each test, it's visible what's the set of inputs and what's the expectation

fedepaol:
Couple of nits, looks much much better!

internal/k8s/controllers/config_controller_test.go
```diff
@@ -0,0 +1,265 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+)
+
+func TestConfigController(t *testing.T) {
+	tests := []struct {
+		desc            string
+		handlerRes      SyncState
+		failReconcile   bool
```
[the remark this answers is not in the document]
liornoy: fixed thanks

internal/k8s/controllers/config_controller_test.go
liornoy: moved.

liornoy:
> Couple of nits, looks much much better!

Thank you, I fixed the nits.
also rephrased the error string as follows:
t.Errorf("test %s failed, **handler called with unexpected config:** %s", test.desc, cmp.Diff(expectedCfg, cfg, cmpOpt))

internal/k8s/controllers/config_controller_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+	for _, l2Adv := range r.L2Advs {
+		objects = append(objects, l2Adv.DeepCopy())
+	}
+
+	for _, legacyPool := range r.LegacyAddressPools {
+		objects = append(objects, legacyPool.DeepCopy())
+	}
+
+	for _, community := range r.Communities {
```
sabinaaledort, line 265: same

internal/k8s/controllers/config_controller_test.go
```diff
@@ -0,0 +1,270 @@
...
+
+	for _, secret := range r.PasswordSecrets {
+		objects = append(objects, secret.DeepCopy())
+	}
+
+	for _, peer := range r.Peers {
+		objects = append(objects, peer.DeepCopy())
+	}
+
+	for _, bfdProfile := range r.BFDProfiles {
```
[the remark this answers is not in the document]
liornoy: No, it's supposed to be in this test also.
I trimmed too much.
thanks!

internal/k8s/controllers/pool_controller_test.go
```diff
@@ -0,0 +1,199 @@
...
+			t.Errorf("%s: fail reconcile expected: %v, got: %v", test.desc, test.expectReconcileFails, failedReconcile)
+		}
+
+		if test.expectForceReloadCalled != calledForceReload {
+			t.Errorf("%s: call force reload expected: %v, got: %v", test.desc, test.expectForceReloadCalled, calledForceReload)
+		}
+	}
+}
+
+var (
```
sabinaaledort, line 135: should we keep the vars at the top of the file like we usually do?

internal/k8s/controllers/node_controller_test.go
```diff
@@ -0,0 +1,97 @@
...
+				Namespace: testNamespace,
+				Name:      testNodeName,
+			},
+		}
+
+		_, err = r.Reconcile(context.TODO(), req)
+		failedReconcile := err != nil
+
+		if test.expectReconcileFails != failedReconcile {
+			t.Errorf("%s: fail reconcile expected: %v, got: %v", test.desc, test.expectReconcileFails, failedReconcile)
```
sabinaaledort, line 94: maybe worth printing err here? it's only printing expected/got true/false

internal/k8s/controllers/node_controller_test.go
```diff
@@ -0,0 +1,97 @@
...
+				Name:      testNodeName,
+				Namespace: testNamespace,
+			}})
+
+		mockHandler := func(log.Logger, *corev1.Node) SyncState {
+			return test.handlerRes
+		}
+
+		if err != nil {
+			t.Fatalf("test %s failed, got unexpected error: %v", test.desc, err)
```
sabinaaledort, line 72: maybe worth mentioning is failed creating the fakeClient

internal/k8s/controllers/config_controller_test.go
```diff
@@ -0,0 +1,279 @@
...
+					Namespace: testNamespace,
+				},
+			},
+		},
+	}
+)
+
+func newFakeClient(t *testing.T, resources metallbcfg.ClusterResources) client.WithWatch {
+	if err := v1beta1.AddToScheme(scheme); err != nil {
+		t.Fatalf("add to scheme failed: %s", err)
```
sabinaaledort, line 227: maybe mention here which AddToScheme failed? they all return the same output so it might be hard to know which line failed

internal/k8s/controllers/config_controller_test.go
```diff
@@ -0,0 +1,279 @@
...
+			resources = configControllerValidResources
+		} else {
+			resources = configControllerInvalidResources
+		}
+
+		fakeClient := newFakeClient(t, resources)
+
+		expectedCfg, err := config.For(resources, config.DontValidate)
+		if err != nil && test.validResources {
+			t.Fatalf("test %s failed, got unexpected error: %v", test.desc, err)
```
sabinaaledort, line 95: same here, maybe saying it failed to create the config?

internal/k8s/controllers/service_controller_test.go
```diff
@@ -0,0 +1,99 @@
...
+			NamespacedName: types.NamespacedName{
+				Namespace: testNamespace,
+			},
+		}
+
+		_, err := r.Reconcile(context.TODO(), req)
+		failedReconcile := err != nil
+
+		if test.expectReconcileFails != failedReconcile {
+			t.Errorf("%s: fail reconcile expected: %v, got: %v", test.desc, test.expectReconcileFails, failedReconcile)
```
sabinaaledort, line 92: same

internal/k8s/controllers/pool_controller_test.go
fedepaol: I suggested doing it this way. We refer to the variables by their name, and makes sense to understand what the test does when we open the file, so we can see at glance what the tests are.

fedepaol:
Few nits, the most notable comment is the one about the parameters to the handler.

internal/k8s/controllers/node_controller_test.go
liornoy: done

internal/k8s/controllers/config_controller_test.go
liornoy: done

internal/k8s/controllers/config_controller_test.go
liornoy: done

internal/k8s/controllers/service_controller_test.go
liornoy: done

internal/k8s/controllers/config_controller_test.go
```diff
@@ -0,0 +1,279 @@
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "peer1",
+					Namespace: testNamespace,
+				},
+			},
+		},
+	}
+)
+
+func newFakeClient(t *testing.T, resources metallbcfg.ClusterResources) client.WithWatch {
```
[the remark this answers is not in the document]
liornoy: done

internal/k8s/controllers/node_controller_test.go
```diff
@@ -0,0 +1,97 @@
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      testNodeName,
+				Namespace: testNamespace,
+			}})
+
+		mockHandler := func(log.Logger, *corev1.Node) SyncState {
+			return test.handlerRes
+		}
+
+		if err != nil {
```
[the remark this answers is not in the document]
liornoy: changed it so all the needed objects will be passed as init objects, with the fake client constructor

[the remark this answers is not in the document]
liornoy: so no need to fakeClient.create

internal/k8s/controllers/service_controller_test.go
```diff
@@ -0,0 +1,99 @@
...
+		},
+		{
+			desc:                    "handler returns SyncStateReprocessAll",
+			handlerRes:              SyncStateReprocessAll,
+			expectReconcileFails:    false,
+			expectForceReloadCalled: true,
+		},
+	}
+	for _, test := range tests {
+		mockHandler := func(log.Logger, string, *v1.Service, epslices.EpsOrSlices) SyncState {
```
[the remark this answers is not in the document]
liornoy: added

internal/k8s/controllers/servicereload_controller_test.go
```diff
@@ -0,0 +1,95 @@
...
+			}})
+		if err != nil {
+			t.Fatalf("test %s failed, got unexpected error: %v", test.desc, err)
+		}
+
+		mockHandler := func(log.Logger, string, *corev1.Service, epslices.EpsOrSlices) SyncState {
+			return test.handlerRes
+		}
+
+		r := &ServiceReloadReconciler{
```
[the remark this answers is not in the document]
liornoy: updated

internal/k8s/controllers/node_controller_test.go
liornoy: added thanks!

internal/k8s/controllers/fakeclient_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+	for _, l2Adv := range r.L2Advs {
+		objects = append(objects, l2Adv.DeepCopy())
+	}
+
+	for _, legacyPool := range r.LegacyAddressPools {
+		objects = append(objects, legacyPool.DeepCopy())
+	}
+
+	for _, community := range r.Communities {
```
sabinaaledort, line 82: there are still no Communities in the tested config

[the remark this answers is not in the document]
liornoy: fixed thanks!

internal/k8s/controllers/service_controller_test.go
```diff
@@ -0,0 +1,255 @@
...
+					Namespace: testNamespace,
+					Name:      testServiceName,
+				},
+			}
+		}
+
+		ctx, cancel := context.WithTimeout(context.Background(), contextTimeOutDuration)
+		defer cancel()
+
+		done := make(chan bool, 1)
```
fedepaol, line 229: why do we need a goroutine here?

Also, you can use a syn.WaitGroup if the purpouse is only to wait

liornoy: we need the goroutine for catching the signal in the mockReload channel

liornoy: nvm. you're right, it's possible by making the mockReload buffered. will fix. thanks!

fedepaol:
LGTM

PR #1323, opened 2022-04-18
Add unit tests conversion webhooks

sabinaaledort:


liornoy:
lgtm, only one small not critical comment I have is that maybe you could extract the 
definitions of the objects outside the test functions, and have something like:
convertBGPPeer := v1beta1BGPPeer
expectedBGPPeer := v1beta2BGPPeer
And this way you wouldn't need to have to repeat defining those objects twice in the
ConvertTo and ConverFrom functions.
Same for the addresspool.

api/v1alpha1/addresspool_conversion.go
```diff
@@ -13,11 +13,20 @@ func (src *AddressPool) ConvertTo(dstRaw conversion.Hub) error {
 	dst.Spec.Protocol = src.Spec.Protocol
 	dst.Spec.Addresses = make([]string, len(src.Spec.Addresses))
 	copy(dst.Spec.Addresses, src.Spec.Addresses)
-	*dst.Spec.AutoAssign = *src.Spec.AutoAssign
-	dst.Spec.BGPAdvertisements = make([]v1beta1.LegacyBgpAdvertisement, 0)
+	if src.Spec.AutoAssign != nil {
+		autoAssign := *src.Spec.AutoAssign
+		dst.Spec.AutoAssign = &autoAssign
```
fedepaol, line 18: How about utility copy functions? Another option is to use the function of the pointer package, that will allocate a new pointer based on the value you are passing in.

fedepaol:
LGTM

PR #1327, opened 2022-04-19
Fix SetPools function

### LINKED ISSUE #1330: Bug: controller assigns IP without existing CRs
To reproduce:
create an ipaddresspool.
delete the ipaddresspool.
create a service.
on the deletion, the pool_controller will invoke the handler: SetPools,
as there are no more pools, it will hit the if statement: if len(pools)==0, and
exit. so it doesn't reach the line c.pools=pools, meaning that
if we create an LB service next, it will allocate, even without any existing CRs.

This behavior exists also in v0.12 (pre-CRDs)

liornoy:
SetPools might be called with no pools,
e.g. when a user deletes the last ipaddresspool,

This commit fixes the if statement, as it's supposed  to
check if the pools parameter is nil, not if it's empty.

Added the check if pools is not empty before
calling c.ips.SetPools(pools)

And removed a test assertion that expects calling SetPools
with empty Pools to return SyncStateErrorNoRetry.

fix #1330

### DIFF
--- a/controller/main.go
+++ b/controller/main.go
@@ -102,7 +102,7 @@ func (c *controller) SetPools(l log.Logger, pools map[string]*config.Pool) contr
 	level.Debug(l).Log("event", "startUpdate", "msg", "start of config update")
 	defer level.Debug(l).Log("event", "endUpdate", "msg", "end of config update")
 
-	if len(pools) == 0 {
+	if pools == nil {
 		level.Error(l).Log("op", "setConfig", "error", "no MetalLB configuration in cluster", "msg", "configuration is missing, MetalLB will not function")
 		return controllers.SyncStateErrorNoRetry
 	}
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -854,7 +854,7 @@ func TestControllerConfig(t *testing.T) {
 	}
 
 	// Now that an IP is allocated, removing the IP pool is not allowed.
-	if c.SetPools(l, map[string]*config.Pool{}) == controllers.SyncStateError {
+	if c.SetPools(l, map[string]*config.Pool{}) != controllers.SyncStateError {
 		t.Fatalf("SetPools that deletes allocated IPs was accepted")
 	}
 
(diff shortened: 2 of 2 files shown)


fedepaol:
Thanks!

controller/main.go
```diff
@@ -102,14 +102,17 @@ func (c *controller) SetPools(l log.Logger, pools map[string]*config.Pool) contr
...
-	if len(pools) == 0 {
+	if pools == nil {
 		level.Error(l).Log("op", "setConfig", "error", "no MetalLB configuration in cluster", "msg", "configuration is missing, MetalLB will not function")
 		return controllers.SyncStateErrorNoRetry
 	}
 
-	if err := c.ips.SetPools(pools); err != nil {
-		level.Error(l).Log("op", "setConfig", "error", err, "msg", "applying new configuration failed")
-		return controllers.SyncStateError
+	if len(pools) != 0 {
```
fedepaol, line 110: Why do you need this check? This is changing the previous behaviour I guess

liornoy: yes. that's why I looked at v0.12 to see if it's happened there too.

liornoy: and without this check, the case of calling SetPools with empty pools would
return SyncStateError

fedepaol: Not sure about that, I think it would have returned SyncStateReprocessAll
See https://github.com/metallb/metallb/blob/f691c94622a5ab0aab29511df9911cf7bf1f94ac/internal/allocator/allocator.go#L66 , that will return nil

liornoy: I saw it and poolFor is returning "" for empty pools resulting of
returning an error

fedepaol: Right, but that makes sense. If the pool is allocated to some service we want to fail

liornoy: right. but it will fail anyway even without the service being allocated.

fedepaol: if a service is not allocated, a.allocated will be empty and poolFor won't be called? What am I missing?

liornoy: to resolve this conversation, you were right, shouldn't add this check.

PR #1328, opened 2022-04-19
Add label selector to l2 and bgp advertisement

pperiyasamy:
This adds label selector option for l2 and bgp advertisement to
select ip pools which would get advertised to peers via this
advertisement configuration settings.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -238,27 +238,9 @@ func For(resources ClusterResources, validate Validate) (*Config, error) {
 		cfg.Pools[p.Name] = pool
 	}
 
-	for _, l2Adv := range resources.L2Advs {
-		adv, err := l2AdvertisementFromCR(l2Adv, resources.Nodes)
-		if err != nil {
-			return nil, err
-		}
-		// No pool selector means select all pools
-		if len(l2Adv.Spec.IPAddressPools) == 0 {
-			for _, pool := range cfg.Pools {
-				if !containsAdvertisement(pool.L2Advertisements, adv) {
-					pool.L2Advertisements = append(pool.L2Advertisements, adv)
-				}
-			}
-			continue
-		}
-		for _, poolName := range l2Adv.Spec.IPAddressPools {
-			if pool, ok := cfg.Pools[poolName]; ok {
-				if !containsAdvertisement(pool.L2Advertisements, adv) {
-					pool.L2Advertisements = append(pool.L2Advertisements, adv)
-				}
-			}
-		}
+	err = setL2AdvertisementsToPools(resources.Pools, resources.L2Advs, resources.Nodes, cfg.Pools)
+	if err != nil {
+		return nil, err
 	}
 
 	err = validateDuplicateBGPAdvertisements(resources.BGPAdvs)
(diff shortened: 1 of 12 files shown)


internal/config/config.go
```diff
@@ -243,16 +243,20 @@ func For(resources ClusterResources, validate Validate) (*Config, error) {
 		if err != nil {
 			return nil, err
 		}
+		ipPools, err := selectPools(resources.Pools, l2Adv.Spec.IPAddressPoolSelectors)
+		if err != nil {
+			return nil, err
+		}
 		// No pool selector means select all pools
-		if len(l2Adv.Spec.IPAddressPools) == 0 {
+		if len(l2Adv.Spec.IPAddressPools) == 0 && len(ipPools) == 0 {
```
fedepaol, line 251: This must be empty label selector.
If the user puts a non matching selector, we don't want to associate the advertisement to all pools

internal/config/config.go
```diff
@@ -243,16 +243,20 @@ func For(resources ClusterResources, validate Validate) (*Config, error) {
 		if err != nil {
 			return nil, err
 		}
+		ipPools, err := selectPools(resources.Pools, l2Adv.Spec.IPAddressPoolSelectors)
```
fedepaol, line 246: This block is now big enough to deserve a function

internal/config/config.go
```diff
@@ -271,8 +275,12 @@ func For(resources ClusterResources, validate Validate) (*Config, error) {
 		if err != nil {
 			return nil, err
 		}
+		ipPools, err := selectPools(resources.Pools, bgpAdv.Spec.IPAddressPoolSelectors)
```
fedepaol, line 278: Same

internal/config/config.go
```diff
@@ -910,3 +918,28 @@ OUTER:
...
+OUTER:
+	for _, pool := range pools {
+		for _, s := range labelSelectors {
+			poolLabels := labels.Set(pool.Labels)
+			if s.Matches(poolLabels) {
+				ipPools = append(ipPools, pool.Name)
+				continue OUTER
+			}
+		}
+
```
fedepaol, line 942: nit: empty line

internal/config/config.go
```diff
@@ -910,3 +918,28 @@ OUTER:
 	return res, nil
 
 }
+
+func selectPools(pools []metallbv1beta1.IPAddressPool, selectors []metav1.LabelSelector) ([]string, error) {
```
fedepaol, line 922: nit: selectedPools

internal/config/config_test.go
```diff
@@ -1693,6 +1693,120 @@ func TestParse(t *testing.T) {
 				},
 			},
 		},
+		{
+			desc: "use ip pool selectors",
+			crs: ClusterResources{
```
fedepaol, line 1698: Can you add a test where we select a non existing label and we verify no advertisement is applied?

internal/config/config.go
pperiyasamy: valid point, changed it.

internal/config/config_test.go
pperiyasamy: done

fedepaol:
LGTM, thanks!

PR #1329, opened 2022-04-19
Squash the service controller and the service reload together

fedepaol:
Having one single controller doing both the full reload and handling service events prevents races, especially when handling delete events.


internal/k8s/controllers/service_controller.go
```diff
@@ -27,26 +27,36 @@ import (
...
 
 func (r *ServiceReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	if !isReloadReq(req) {
+		return r.reconcileService(ctx, req)
+	}
+	return r.reprocessAll(ctx, req)
+}
+
+func (r *ServiceReconciler) reconcileService(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+
```
sabinaaledort, line 59: empty line

internal/k8s/controllers/service_controller_reload.go
```diff
@@ -0,0 +1,103 @@
...
+	return false
+}
+
+func (r *ServiceReconciler) reprocessAll(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	level.Info(r.Logger).Log("controller", "ServiceReconciler - reprocessAll", "start reconcile", req.NamespacedName.String())
+	defer level.Info(r.Logger).Log("controller", "ServiceReconciler - reprocessAll", "end reconcile", req.NamespacedName.String())
+
+	var services v1.ServiceList
+	if err := r.List(ctx, &services); err != nil {
+		level.Error(r.Logger).Log("controller", "ServiceReconciler - reprocessAll", "error", "failed to list the services", "error", err)
```
sabinaaledort, line 66: "error" is used twice, i think we said it's only showing the second one this way

PR #1332, opened 2022-04-20
Node selector e2e

sabinaaledort:


e2etest/bgptests/node_selector.go
```diff
@@ -189,6 +189,81 @@ var _ = ginkgo.Describe("BGP Node Selector", func() {
...
+			ginkgo.By("Validating service IP not advertised")
+			checkServiceNotOnNodes(cs, svc, allNodes.Items, pairingIPFamily)
+
+			node := allNodes.Items[0]
+			node.Labels["bgp-node-selector-test"] = "true"
+			err = ConfigUpdater.Client().Update(context.Background(), &node)
+			framework.ExpectNoError(err)
+			ginkgo.By(fmt.Sprintf("Adding advertisement label to node %s", node.Name))
+
+			defer func() {
```
fedepaol, line 250: maybe it makes sense to add this in aftereach, so it will run even if we fail (same for the l2 test)

e2etest/bgptests/node_selector.go
```diff
@@ -189,6 +189,81 @@ var _ = ginkgo.Describe("BGP Node Selector", func() {
...
+			err = ConfigUpdater.Update(resources)
+			framework.ExpectNoError(err)
+
+			svc, _ := testservice.CreateWithBackend(cs, f.Namespace.Name, "external-local-lb", testservice.TrafficPolicyCluster)
+			defer testservice.Delete(cs, svc)
+
+			ginkgo.By("Validating service IP not advertised")
+			checkServiceNotOnNodes(cs, svc, allNodes.Items, pairingIPFamily)
+
+			node := allNodes.Items[0]
```
fedepaol, line 244: How about wrapping this labelling work in an util function under k8s ? Same for unlabeling I guess

e2etest/bgptests/node_selector.go
fedepaol: if we write it in a way that is idempotent, why not?

e2etest/bgptests/validate.go
```diff
@@ -193,6 +193,22 @@ func checkServiceOnlyOnNodes(cs clientset.Interface, svc *corev1.Service, expect
...
+		return
+	}
+	ip := svc.Status.LoadBalancer.Ingress[0].IP
+	nodeIps := k8s.NodeIPsForFamily(expectedNodes, ipFamily)
+
+	for _, c := range FRRContainers {
+		Eventually(func() bool {
+			routes, err := frr.RoutesForFamily(c, ipFamily)
+			framework.ExpectNoError(err)
+			return len(routes[ip].NextHops) == len(nodeIps)
```
fedepaol, line 207: sorry just noticed this. This is a bit fragile, as we are checking the lenght and a double bug may result in having them being different but with the svc being there.
Can we check that the nodes are really not there?

fedepaol:
LGTM

PR #1341, opened 2022-04-26
Specify the interface list that the speaker uses to announce Service IP

qingwusunny:
Use ConfigMap to specify the interface list that the speaker uses to announce Service IP for layer2.

ConfigMap data add properties 'bind-interfaces', it specifies the interface list to announce IP. If users don't config 'bind-interface', Speaker announces Service IP from all interfaces.

When ConfigMap change, update properties 'bindInterface' of struct 'Announce'. It will update ARP and NDP responders.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -37,6 +37,7 @@ type configFile struct {
 	BGPCommunities map[string]string `yaml:"bgp-communities"`
 	Pools          []addressPool     `yaml:"address-pools"`
 	BFDProfiles    []bfdProfile      `yaml:"bfd-profiles"`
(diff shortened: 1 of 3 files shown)


fedepaol:
Hi @qingwusunny , there's no configmap anymore, so the given configuration must be performed via CRDs.
Also, I'd appreciate discussing new substantial features over a design proposal before the implementation.

qingwusunny:
> Hi @qingwusunny , there's no configmap anymore, so the given configuration must be performed via CRDs. Also, I'd appreciate discussing new substantial features over a design proposal before the implementation.

Thanks Federico, I pushed this commit by accident and I'm going to close this PR.

PR #1345, opened 2022-04-27
API: remove yaml marshalling names

fedepaol:
The yaml names are a legacy of the initial implementation of the
operator, which was leveraging them to render the configmap. They are
not needed anymore.


liornoy:
I'm wondering if those yaml names will help with the CR generator from configmap.

fedepaol:
> I'm wondering if those yaml names will help with the CR generator from configmap.

I thought about that, but ended up thinking that it makes no sense to couple the CRs to the configmap and it's better to have the configmap to have it's own datamodel.

liornoy:
lgtm

PR #1347, opened 2022-04-28
Filter out no IP interfaces for L2 advertisement

pperiyasamy:
In case of Layer 2 mode, it doesn't make sense to advertise service
load balancer ip on the interfaces which don't have ip configured.
This commit filers out those interfaces so that it won't be chosen
for sending or responding arp/ndp  messages.


internal/layer2/announcer.go
```diff
@@ -86,20 +86,17 @@ func (a *Announce) updateInterfaces() {
...
-		}
 
 		for _, a := range addrs {
 			ipaddr, ok := a.(*net.IPNet)
 			if !ok {
 				continue
 			}
-			if ipaddr.IP.To4() != nil || !ipaddr.IP.IsLinkLocalUnicast() {
-				continue
+			if ipaddr.IP.To4() != nil && (ifi.Flags&net.FlagBroadcast) != 0 {
```
fedepaol, line 95: Would it just check len (addrs) be enough to skip the interface?

fedepaol:
Code aside, we need to understand if this is always the behaviour we want, or if we are cutting out corner case that are still working.

internal/layer2/announcer.go
```diff
@@ -86,20 +86,17 @@ func (a *Announce) updateInterfaces() {
...
+				keepARP[ifi.Index] = true
```
fedepaol, line 96: you can continue here, and no need for the else below, just the if

pperiyasamy: yes @fedepaol , done.

PR #1349, opened 2022-04-29
bump go-kit plus switch from deprecated module

fedepaol:
Follow up from https://github.com/metallb/metallb/pull/1279 , plus removing warnings for using a deprecated module


liornoy:
lgtm

PR #1352, opened 2022-05-03
Doc: generate the docs for the API

fedepaol:
This is the first part of the new version docs.
Leveraging https://github.com/ahmetb/gen-crd-api-reference-docs to generate an API reference docs section.
Also, cleaning up the CRDs so the comments look good when translated to doc.


liornoy (commented):
a couple of nits. besides that, the final md looks very nice

PR #1353, opened 2022-05-03
E2E: clean the crds after running the suite

fedepaol:
All the tests clean before running. This means, the last test leaves the
cluster dirty.

### DIFF
--- a/e2etest/e2etest_suite_test.go
+++ b/e2etest/e2etest_suite_test.go
@@ -49,6 +49,7 @@ var (
 	ipv6ForContainers string
 	useOperator       bool
 	reportPath        string
+	updater           testsconfig.Updater
 )
 
 // handleFlags sets up all flags and parses the command line.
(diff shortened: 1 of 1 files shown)


fedepaol:
cc @sabinaaledort

e2etest/e2etest_suite_test.go
```diff
@@ -49,6 +49,7 @@ var (
 	ipv6ForContainers string
 	useOperator       bool
 	reportPath        string
+	Updater           testsconfig.Updater
```
sabinaaledort, line 52: maybe i missing something here, but why it has to start with an uppercase letter if we only use it in this file?

fedepaol: That's correct, it can be private

fedepaol: fixed

PR #1360, opened 2022-05-08
Remove the dependency from cert-manager

sabinaaledort:


internal/k8s/k8s.go
```diff
@@ -125,7 +147,13 @@ func New(cfg *Config) (*Client, error) {
 				&corev1.Secret{}:                   namespaceSelector,
 			},
 		}),
+		//todo: check if needed
+		CertDir: cfg.CertDir,
+		MapperProvider: func(c *rest.Config) (meta.RESTMapper, error) {
```
fedepaol, line 152: This is not needed, as it's equal to the default:
https://github.com/kubernetes-sigs/controller-runtime/blob/master/pkg/cluster/cluster.go#L214

internal/k8s/k8s.go
```diff
@@ -149,48 +177,117 @@ func New(cfg *Config) (*Client, error) {
...
+		if err != nil {
+			level.Error(l).Log("op", "listenAndServe", "err", err, "msg", "cannot listen and serve", "host", cfg.MetricsHost, "port", cfg.MetricsPort)
+		}
+	}(c.logger)
+
+	return c, nil
+}
+
+func (c *Client) SetupControllers(cfg *Config) error {
+	if cfg.EnableWebhook {
```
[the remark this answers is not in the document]
fedepaol: Also, I would enable the rotator by default, and provide a mechanism to disable it when needed (at this point, the only remaining case is OCP).

controller/main.go
```diff
@@ -117,14 +117,17 @@ func (c *controller) SetPools(l log.Logger, pools map[string]*config.Pool) contr
...
-		enablePprof     = flag.Bool("enable-pprof", false, "Enable pprof profiling")
-		enableWebhook   = flag.Bool("enable-webhook", false, "Enable validation webhook")
+		port               = flag.Int("port", 7472, "HTTP listening port for Prometheus metrics")
+		namespace          = flag.String("namespace", os.Getenv("METALLB_NAMESPACE"), "config / memberlist secret namespace")
+		mlSecret           = flag.String("ml-secret-name", os.Getenv("METALLB_ML_SECRET_NAME"), "name of the memberlist secret to create")
+		deployName         = flag.String("deployment", os.Getenv("METALLB_DEPLOYMENT"), "name of the MetalLB controller Deployment")
+		logLevel           = flag.String("log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", logging.Levels.String()))
+		disableEpSlices    = flag.Bool("disable-epslices", false, "Disable the usage of EndpointSlices and default to Endpoints instead of relying on the autodiscovery mechanism")
+		enablePprof        = flag.Bool("enable-pprof", false, "Enable pprof profiling")
+		enableWebhook      = flag.Bool("enable-webhook", false, "Enable validation webhook")
```
fedepaol, line 127: How about having the webhooks always enabled? I think now that the cert-rotator is embedded it's safe to

controller/main.go
```diff
@@ -168,9 +171,15 @@ func main() {
...
 		},
-		ValidateConfig: validation,
-		EnableWebhook:  *enableWebhook,
-	})
+		ValidateConfig:     validation,
+		EnableWebhook:      *enableWebhook,
+		EnableCertRotation: *enableCertRotation,
+	}
+	if *enableWebhook && *enableCertRotation {
+		config.CertDir = *certDir
```
fedepaol, line 179: Just forward the parameters to the config without the if. If they are not set, they'll be empty.
If certRotation is not enabled, they'll not be used.

controller/main.go
```diff
@@ -117,14 +117,17 @@ func (c *controller) SetPools(l log.Logger, pools map[string]*config.Pool) contr
...
-		enableWebhook   = flag.Bool("enable-webhook", false, "Enable validation webhook")
+		port               = flag.Int("port", 7472, "HTTP listening port for Prometheus metrics")
+		namespace          = flag.String("namespace", os.Getenv("METALLB_NAMESPACE"), "config / memberlist secret namespace")
+		mlSecret           = flag.String("ml-secret-name", os.Getenv("METALLB_ML_SECRET_NAME"), "name of the memberlist secret to create")
+		deployName         = flag.String("deployment", os.Getenv("METALLB_DEPLOYMENT"), "name of the MetalLB controller Deployment")
+		logLevel           = flag.String("log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", logging.Levels.String()))
+		disableEpSlices    = flag.Bool("disable-epslices", false, "Disable the usage of EndpointSlices and default to Endpoints instead of relying on the autodiscovery mechanism")
+		enablePprof        = flag.Bool("enable-pprof", false, "Enable pprof profiling")
+		enableWebhook      = flag.Bool("enable-webhook", false, "Enable validation webhook")
+		enableCertRotation = flag.Bool("enable-cert-rotation", true, "enable automatic generation and rotation of webhook TLS certificates/keys")
```
fedepaol, line 128: I'd change to disableCertRotation, it's more intuitive having the default value to false.

internal/k8s/k8s.go
```diff
@@ -244,11 +262,52 @@ func New(cfg *Config) (*Client, error) {
...
+				return nil, err
+			}
+		} else {
+			close(setupFinished)
 		}
+
+		go func() {
+			// Block until the setup (certificate generation) finishes.
+			<-setupFinished
+			err := enableWebhook(c.mgr, cfg.ValidateConfig, cfg.Namespace, cfg.Logger)
```
fedepaol, line 306: I'd log here that we are enabling the webhooks, so we can see when it happens

fedepaol:
Looks great! I'd sort out the "do we want to have the webhook always enabled" discussion before moving to the helm / manifests, I suspect they'll be even slimmer if we do that.

controller/main.go
sabinaaledort: i was thinking to do that in another pr since it requires changes not just here but in the manifests+helm+ci+dev-env. do you think it's better to have it here?

fedepaol: Nope, I guess doing it in a separate PR is totally fine, I wasn't just aware of that. Let me move on the other bits then.

charts/metallb/templates/rbac.yaml
```diff
@@ -88,7 +96,7 @@ rules:
 {{- if .Values.speaker.memberlist.enabled }}
 - apiGroups: [""]
   resources: ["secrets"]
-  verbs: ["create", "get", "list", "watch"]
+  verbs: ["create", "delete", "get", "list", "patch", "update", "watch"]
```
fedepaol, line 99: This is under the memberlist if, so it could happen that it's not created

config/frr-with-webhook/clusterrole-permissions.yaml
```diff
@@ -0,0 +1,61 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app: metallb
+  name: metallb-system:controller
+rules:
+  - apiGroups:
```
fedepaol, line 8: Would it make sense to patch with the ones needed for the webhooks and relying on the base for the others?

config/native-with-webhook/clusterrole-permissions.yaml
```diff
@@ -0,0 +1,61 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    app: metallb
+  name: metallb-system:controller
+rules:
+  - apiGroups:
```
fedepaol, line 8: same

internal/k8s/k8s.go
sabinaaledort: there is such log in the enableWebhook function

fedepaol:
LGTM

fedepaol:
The operator lane is expected to fail as we need to mirror the changes.
Specifically, the operator controller tries to deploy the cert manager `Certificate` object which doesn't exist anymore.

fedepaol:
Merging as it's a chicken - egg issue we can't overcome.

PR #1361, opened 2022-05-10
Add BFDProfile validation webhook

liornoy:
This PR is adding a validation webhook to the BFDProfile CRD.
This webhook only validates a delete action, asserting that the deleted
profile is not being used by a BGPPeer. 
For "create" and "update" validation, the CRD handles it itself with limits values.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -188,7 +188,12 @@ jobs:
         run: |
           inv build
           inv dev-env -b frr --enable-webhooks
-
+      
+      # Patch the old e2etest to cleanup the resources in the correct order.
+      - name: Apply patch
+        run: |
+          patch metallb-v0.12.1/e2etest/pkg/config/update.go < e2etest/backwardcompatible/patchfile
+          
       - name: E2E
         run: |
           cat <<EOF | kubectl apply -f -
--- a/api/v1beta1/bfdprofile_webhook.go
+++ b/api/v1beta1/bfdprofile_webhook.go
@@ -1,0 +1,64 @@
+/*
+
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1beta1
+
+import (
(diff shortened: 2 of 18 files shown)


.github/workflows/ci.yaml
```diff
@@ -189,6 +189,10 @@ jobs:
           inv build
           inv dev-env -b frr --enable-webhooks
 
+      - name: Apply patch
```
fedepaol, line 192: Can you add a comment here on why we need this?

api/v1beta1/bfdprofile_webhook.go
```diff
@@ -0,0 +1,64 @@
...
+// ValidateDelete implements webhook.Validator so a webhook will be registered for BFDProfile.
+func (bfdProfile *BFDProfile) ValidateDelete() error {
+	level.Debug(Logger).Log("webhook", "bfdprofile", "action", "delete", "name", bfdProfile.Name, "namespace", bfdProfile.Namespace)
+
+	existingBGPPeers, err := v1beta2.GetExistingBGPPeers()
+	if err != nil {
+		return err
+	}
+
+	for _, peer := range existingBGPPeers.Items {
```
fedepaol, line 58: Would it make sense to use the same config.For trick we used for other webhooks?
It will be future proof, I guess.

e2etest/backwardcompatible/patchfile
```diff
@@ -0,0 +1,17 @@
+--- update.go	2022-05-18 11:49:32.937404800 +0300
```
fedepaol, line 1: I'd name the file according to what it does (not sure we can even drop a comment inside)

api/v1beta1/bfdprofile_webhook.go
liornoy: correct me if I'm wrong but the logic of checking if a given bfdprofile is present in one of the bgppeers is not existing in the current config.For function.
And I think that for this specific deletion validation, it's problematic to use the config.For

.github/workflows/ci.yaml
liornoy: sure

api/v1beta1/bfdprofile_webhook.go
fedepaol: What you can build is the list of bfdprofiles without the one being deleted, and feed it to config.For, which will complain about a bgpprofile not having a bfdprofile associated.
It's quite similar to the mechanism we use for updates.

e2etest/backwardcompatible/patchfile
liornoy: apparently we can write whatever above the @@

fedepaol:
The failure on the operator lane is expected as the operator doesn't deploy the webhook (yet).
Merging as it's a chicken - egg problem.

PR #1366, opened 2022-05-12
Remove the avoidBuggyIps field from the api

fedepaol:
When defining the first version of the CRDs, it was decided not to
expose this flag.
If .0 / .255 ips are problematic, users can define custom ranges to skip
them. Given this is the first version of the api, we are removing the
field in order to propose a slimmer api surface, with the possibility of
readding it later on.

Link to the discussion: https://github.com/metallb/metallb-operator/pull/17#discussion_r652824338


fedepaol:
> should we also remove it from the pool-configuration design-doc?

I don't like changing the design after it's merged. It's normal for things to take a different direction after implementing them, because reality hits.

oribon (approved):
lgtm

PR #1367, opened 2022-05-12
Disable directly connected check for IPv6 eBGP sessions by default

msherif1234:
There is an issue with how MHT works for IPv6 neighbors
specifically if DHCPv6 was used to assign interfaces IP
because of this issue eBGP directly connected sessions
won't get to establish state.

using disable directly connected check mitigate this
issue for the time being till frr has fix.

for more details about this issue pls refer to
https://github.com/FRRouting/frr/issues/11185

### DIFF
--- a/internal/bgp/frr/config.go
+++ b/internal/bgp/frr/config.go
@@ -102,6 +102,9 @@ router bgp {{$r.MyASN}}
 {{- if ne .BFDProfile ""}} 
   neighbor {{.Addr}} bfd profile {{.BFDProfile}}
 {{- end }}
+{{- if  mustDisableConnectedCheck .IPFamily $r.MyASN .ASN .EBGPMultiHop }}
+  neighbor {{.Addr}} disable-connected-check
+{{- end }}
 {{- end }}
 {{range $n := .Neighbors -}}
 {{/* no bgp default ipv4-unicast prevents peering if no address families are defined. We declare an ipv4 one for the peer to make the pairing happen */}}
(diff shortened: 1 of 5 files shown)


internal/bgp/frr/config.go
```diff
@@ -102,6 +102,9 @@ router bgp {{$r.MyASN}}
 {{- if ne .BFDProfile ""}} 
   neighbor {{.Addr}} bfd profile {{.BFDProfile}}
 {{- end }}
+{{- if eq .IPFamily.String "ipv6" }}
```
fedepaol, line 105: Should we specifically check for ebgp only too?

msherif1234: I was debating this with myself too , ebgp is the only type that does directly connection check since by default ebgp nbr needs to one one hop away unless user specify otherwise while for iBGP no such restriction so its kind of no-op 
on the other side its easy to add compare myasn with asn too so not sure WDYT ?

msherif1234: I limited it to eBGP only and add more test cases

internal/bgp/frr/config.go
```diff
@@ -102,6 +102,9 @@ router bgp {{$r.MyASN}}
 {{- if ne .BFDProfile ""}} 
   neighbor {{.Addr}} bfd profile {{.BFDProfile}}
 {{- end }}
+{{- if and (eq .IPFamily.String "ipv6") (ne $r.MyASN .ASN) }}
```
fedepaol, line 105: How about adding a custom "mustDisableConnectedCheck" with a comment of the reason on its definition instead of the and-ed if?

See https://github.com/metallb/metallb/blob/889a4bded1b46a19d5c99053f6440781439ff00e/internal/bgp/frr/config.go#L225

fedepaol:
LGTM

PR #1370, opened 2022-05-15
Change e2etests cleanup order v0.12

liornoy:
For backward_compatible CI lane in MetalLB's main branch, adding
BFDProfile requires us to change the cleanup order so the Peers
will be deleted before the BFDProfiles, so the webhook won't deny
the cleanup and fail the tests.

### DIFF
--- a/e2etest/pkg/config/update.go
+++ b/e2etest/pkg/config/update.go
@@ -131,12 +131,12 @@ func (o operatorUpdater) Clean() error {
 		return err
 	}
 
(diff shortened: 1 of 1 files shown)


fedepaol:
I am a bit torn about doing a 0.12 release for this (or pinning to a branch). How about we provide a (well commented) patch file to change the tests in a deterministic way?
As part of the CI we invert the order in the implementation, before running the tests.

liornoy:
using patch file instead.

PR #1374, opened 2022-05-16
Replace publish circle ci job with github action

### LINKED ISSUE #1278: Image publishing with github actions
https://github.com/metallb/metallb/pull/1137 was translating the current publishing mechanism in circleci to github actions.
However, while trying to add the secrets to make the action added there work, I realized that those would have been dumped as part of the lane run.

It probably makes sense to look for gh actions that do the publishing for us, instead of manually doing `docker login` as the old circleci publishing job was doing.

Once that is merged and we see it passing successfully, we can retire the circle ci job.

Lastly, we realized image publishing to dockerhub is still in place, although all the current manifests are consuming images from quay.
For this reason, it makes sense to have the new job to publish only to quay and announce the retirement of the images hosted on dockerhub.

pperiyasamy:
This commit replaces existing Circle CI job to build and publish
controller and speaker images with GitHub actions.

Fixes #1278


.github/workflows/publish.yaml
```diff
@@ -0,0 +1,75 @@
...
+    branches:
+      - 'main'
+
+jobs:
+  publish:
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        image: [controller, speaker, mirror-server]
```
fedepaol, line 20: I think mirror-server is not needed

pperiyasamy: ok, done.

jtcarnes:
You know best, but gh build-push action allows you to push to multiple registries (quay and dockerhub here) in one action, docs [here](https://github.com/docker/build-push-action/blob/master/docs/advanced/push-multi-registries.md). Might make the job simpler. Can also build multi-arch if that's something yall are interested in doing so.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,104 @@
...
+
+      - name: Extract docker metadata
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: quay.io/metallb/${{ matrix.image }}
+          tags: |
+            type=raw,value=${{ github.ref_name }}-${{ matrix.arch }}
+
+      - name: Build and push ${{ matrix.image }} image
```
fedepaol, line 98: This won't work.
We either choose to build the images with this, or with `inv build`.
If we want to go this way, we need to rethink the dockerfiles and have the build as part of the dockerfile so we are in control.
Given the comment about multiarch, we should consider what's faster, but the goal must be to have the images built locally (and by CI) with the same process done here.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,104 @@
...
+          inv checkchanges --action="run inv generateapidocs"
+
+  build-and-push:
+    needs: unit-tests
+    runs-on: ubuntu-20.04
+    strategy:
+      fail-fast: false
+      matrix:
+        image: [controller, speaker]
+        arch: [amd64]
```
fedepaol, line 58: amd64 is not enough. Have a look at what the circleci job was doing.
MetalLB users expect it to run on arm too for example.

fedepaol:
> You know best, but gh build-push action allows you to push to multiple registries (quay and dockerhub here) in one action, docs [here](https://github.com/docker/build-push-action/blob/master/docs/advanced/push-multi-registries.md). Might make the job simpler. Can also build multi-arch if that's something yall are interested in doing so.

Thanks for the suggestion! I left a comment about it. MetalLB currently  (and must continue to) supports multi arch images, but another thing that I want to preserve is to build released images with the same method as CI / dev-env . Given build-push takes a all-in-one dockerfile while our build process builds the binary from outside and then copies it in the image, we need to reconcile one way or another (which can be done in two steps too).

jtcarnes:
Its a little more idiomatic to build inside the container instead of copying in the finished build product (use multistage build), so that would be my suggestion but both ways work at the end of the day.

.github/workflows/publish.yaml
pperiyasamy: > You know best, but gh build-push action allows you to push to multiple registries (quay and dockerhub here) in one action, docs [here](https://github.com/docker/build-push-action/blob/master/docs/advanced/push-multi-registries.md). Might make the job simpler. Can also build multi-arch if that's something yall are interested in doing so.

Thanks for the info, currently this gh action is made to use `tasks.py `to build and publish metallb images, will revisit to this when we've Dockerfile itself as self contained one.

pperiyasamy: yes Federico, this would break, I didn't have a closer look at Dockerfile earlier, now changed it to use `inv push-multiarch `command for this purpose.

fedepaol: I agree, I filed https://github.com/metallb/metallb/issues/1408 for the follow up.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,86 @@
...
+          go-version: "1.17.1"
+
+      - name: Install Dependencies
+        run: |
+          sudo apt-get update
+          sudo apt-get install python3-pip arping ndisc6
+          sudo pip3 install invoke semver pyyaml
+          go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.7.0
+
+      - name: Unit Tests
```
fedepaol, line 35: Can we leverage the same set of tests we have here https://github.com/metallb/metallb/blob/0cf5b8209c2eb4128f7e64a28615d6a073d10065/.github/workflows/ci.yaml#L47 ?

pperiyasamy: Federico, as we discussed offline, raised an [issue](https://github.com/metallb/metallb/issues/1411) to address this concern as rework has to be done on the CI workflow as well.

PR #1375, opened 2022-05-16
FRR: restore the image from dockerhub

fedepaol:
Moving back to dockerhub since it's the official FRR image, plus it supports multi - architecture.


oribon (approved):
lgtm

PR #1381, opened 2022-05-19
CI improvements :earth_africa: 

oribon:
Some refactors that should save us CI resources and time:
* Creating dependencies between E2E and "basic" jobs
* Stop running everything again on merges


.github/workflows/ci.yaml
```diff
@@ -52,6 +56,22 @@ jobs:
...
           inv checkchanges --action="run inv generateapidocs"
 
+      - name: Build and save MetalLB images
+        run: |
+          inv build --binaries all
+          docker save quay.io/metallb/speaker:dev-amd64 > speaker.tar
+          docker save quay.io/metallb/controller:dev-amd64 > controller.tar
+          docker save quay.io/metallb/mirror-server:dev-amd64 > mirror-server.tar
+
+      - name: Upload images
```
fedepaol, line 66: Neat! Can we delete the artifacts as final step? I don't think we are gonna need those images for triaging failures.

fedepaol:
I like this! Just one comment about not leaving the images in the artifacts.

.github/workflows/ci.yaml
oribon: There isn't an official workflow for cleanup :confused: 
set the retention-days to 1 instead: https://github.com/actions/upload-artifact#retention-period

fedepaol: Makes sense, thanks

PR #1383, opened 2022-05-19
Bug 2063720: align with upstream 190522

fedepaol:
We align with upstream and we import the fixes for 2063720 and 2088431


fedepaol:
Wrong fork :-)

PR #1384, opened 2022-05-19
WIP: Add a CR generator from a configmap

liornoy:
This is an implementation of a feature suggested in the latest `MetalLB configuration options` [design doc](https://github.com/metallb/metallb/blob/main/design/pool-configuration.md).
The CR generator is a tool for converting MetalLB configmap to custom resources yaml.
This allows users to copy their configuration from the configmap-based MetalLB,
(which is not supported anymore) into the current CR-based version.
The generator is shipped as a container, building script included.

A readme file was added with further details + examples.


generator/README.md
```diff
@@ -0,0 +1,95 @@
+# Custom Resources Generator
+
+Up until release v0.12 metallb was configured via metallb's Configmap.
```
oribon, line 3: metallb -> MetalLB (anywhere this shows up)
Configmap -> configmap

generator/README.md
```diff
@@ -0,0 +1,95 @@
...
+```
+inv build-cr-generator
+```
+### Procedure
+
+Run the container and redirect the output to a file:
+```
+docker run -v <PATH_TO_CONFIG>:/var/input <IMAGE_ID> -origin <CONFIGMAP_FILE_NAME> > metallb_resources.yaml
+```
+
```
oribon, line 22: I think we would build this image and push it ourselves, the user won't have to build.
IMAGE_ID will be something like metallb/generator

generator/README.md
```diff
@@ -0,0 +1,95 @@
...
+        - 64512:1234
+```
+
+The generator will output the following yaml:
+```
+# This was autogenerated by MetalLB's custom resource generator.
+apiVersion: metallb.io/v1beta2
+kind: BGPPeer
+metadata:
+  creationTimestamp: null
```
oribon, line 56: no need to have this, can you remove it?

generator/README.md
```diff
@@ -0,0 +1,95 @@
...
+  name: peer1
+  namespace: metallb-system
+spec:
+  holdTime: 1m30s
+  keepaliveTime: 30s
+  myASN: 64512
+  passwordSecret: {}
+  peerASN: 64512
+  peerAddress: 10.96.0.100
+status: {}
```
oribon, line 66: same, no need to have empty status in the yamls, can you remove it?

generator/README.md
```diff
@@ -0,0 +1,95 @@
...
+kind: BGPPeer
+metadata:
+  creationTimestamp: null
+  name: peer1
+  namespace: metallb-system
+spec:
+  holdTime: 1m30s
+  keepaliveTime: 30s
+  myASN: 64512
+  passwordSecret: {}
```
oribon, line 63: same, I assume this shouldn't show up anywhere (it isn't supported in the configmap iirc)

generator/main.go
```diff
@@ -0,0 +1,485 @@
...
+	"go.universe.tf/metallb/internal/pointer"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/runtime/serializer"
+	"sigs.k8s.io/yaml"
+)
+
+const (
+	resourcesNameSpace = "metallb-system"
```
oribon, line 25: I think you should first try to read the configmap's namespace (using this as default)

generator/main.go
```diff
@@ -0,0 +1,485 @@
...
+	inputDirPath = "/var/input"
+)
+
+func main() {
+	var origin = flag.String("origin", "./config.yaml", "name of the configmap file to covnert")
+	flag.Parse()
+
+	resourcesYAML, err := generate(*origin)
+	if err != nil {
+		fmt.Println(err.Error())
```
oribon, line 40: does this mean when an error occurs we redirect it to the file :sweat_smile: ?
if so I rather we create the output file from here, not relying on the user to redirect from outside, maybe even adding some logs along the way - e.g commit id, "reading file.../ generated ...". wdyt?

tasks.py
```diff
@@ -821,3 +821,26 @@ def checkchanges(ctx, action="check uncommitted files"):
         print("{} must be committed".format(res))
         raise Exit(message="#### Uncommitted files found, you may need to {} ####\n".format(action))
 
+@task(help={
+          "tag": "Docker image tag to use. Default 'cr-generator'.",
+      })
+def build_cr_generator(ctx, tag="cr-generator"):
```
oribon, line 827: does this add something on top of the other `build` task?

oribon (commented):
nice work :)

tasks.py
liornoy: it is different from the other build task by not creating the image for every arch.

generator/main.go
liornoy: missed that, you're right. I'll create the output file and add logs.

generator/main.go
```diff
@@ -0,0 +1,485 @@
...
+	spacer             = "---\n"
+	autoGenComment     = "# This was autogenerated by MetalLB's custom resource generator.\n"
+)
+
+var (
+	inputDirPath = "/var/input"
+)
+
+func main() {
+	var origin = flag.String("origin", "./config.yaml", "name of the configmap file to covnert")
```
fedepaol, line 35: nit: convert

fedepaol, line 35: you can do origin := ...
Also, I'd rename it `source`

generator/main.go
fedepaol: you can use log.Fatal that prints the error and exits.

generator/main.go
```diff
@@ -0,0 +1,485 @@
...
+		return 0, fmt.Errorf("invalid keepalive time %q: %s", ka, err)
+	}
+	rounded := time.Duration(int(d.Seconds())) * time.Second
+	return rounded, nil
+}
+
+func ipAddressPoolsFor(c *configFile) []v1beta1.IPAddressPool {
+	ret := make([]v1beta1.IPAddressPool, len(c.Pools))
+	for i, addresspool := range c.Pools {
+		var ap v1beta1.IPAddressPool
```
fedepaol, line 288: this has an extra complication:
the configMap has avoidBuggyIPs while we don't. So we need to convert CIDR to ranges excluding the extremes if the flag is set, and to exclude extremes from ranges.

generator/main.go
```diff
@@ -0,0 +1,485 @@
...
+	var origin = flag.String("origin", "./config.yaml", "name of the configmap file to covnert")
+	flag.Parse()
+
+	resourcesYAML, err := generate(*origin)
+	if err != nil {
+		fmt.Println(err.Error())
+		os.Exit(1)
+	}
+
+	fmt.Println(resourcesYAML)
```
fedepaol, line 44: Can we write it to a file rather than printing it out?

generator/main.go
```diff
@@ -0,0 +1,485 @@
...
+}
+
+// generate gets a name of a metallb configmap file, converts it to
+// the matching metallb custom resources yamls, and returns it as a string.
+func generate(origin string) (string, error) {
```
fedepaol, line 49: You can pass a Writer to generate, so you can choose from outside whether write to a file or to stdoutput.

liornoy: how would we let the end-user choose?
maybe the default is stdout and optional adding --out=filename to write to a file?

fedepaol: I would not, but the writer comes handy in test scenarios.

generator/README.md
liornoy: will do.

generator/README.md
fedepaol: We can't avoid having this printed, unless we convert the field to be a pointer (and we don't want to make a pointer out of it for this).
See https://www.sohamkamani.com/golang/omitempty/#values-that-cannot-be-omitted

fedepaol: We can't avoid having this printed, unless we convert the field to be a pointer (and we don't want to make a pointer out of it for this).
See https://www.sohamkamani.com/golang/omitempty/#values-that-cannot-be-omitted

generator/README.md
fedepaol: Same

generator/README.md
fedepaol: And this comes directly from the go stdlib:
```
// UnmarshalJSON implements the json.Unmarshaler interface.
// The time is expected to be a quoted string in RFC 3339 format.
func (t *Time) UnmarshalJSON(data []byte) error {
	// Ignore null, like in the main JSON package.
	if string(data) == "null" {
		return nil
	}
	// Fractional seconds are handled implicitly by Parse.
	var err error
	*t, err = Parse(`"`+RFC3339+`"`, string(data))
	return err
}
```
(note that we use k8s deserializer, which converts from go to json and from json to yaml).

fedepaol: And this comes directly from the go stdlib:
```
// UnmarshalJSON implements the json.Unmarshaler interface.
// The time is expected to be a quoted string in RFC 3339 format.
func (t *Time) UnmarshalJSON(data []byte) error {
	// Ignore null, like in the main JSON package.
	if string(data) == "null" {
		return nil
	}
	// Fractional seconds are handled implicitly by Parse.
	var err error
	*t, err = Parse(`"`+RFC3339+`"`, string(data))
	return err
}
```
(note that we use k8s deserializer, which converts from go to json and from json to yaml).

PR #1389, opened 2022-05-25
GitHub helm docs

### LINKED ISSUE #1373: Automate Helm readme generation
We should generate the helm readme automatically, and we should enforce it via CI.

jtcarnes:
Change to CI to check if generated docs are different from what is currently in PR.

Ironic enough, because the current helm docs do not match, this will not pass the CI test. (which I suppose is the feature?)
Fixed #1373


.github/workflows/ci.yaml
```diff
@@ -92,6 +92,21 @@ jobs:
...
           args: ct lint
 
+      - name: Helm doc generate
+        uses: docker://jnorwood/helm-docs:v1.10.0
+
+      - name: Check if  docs are different
+        run: |
+          dir_dirty=$(git diff HEAD)
+          if [ -z "$dir_dirty" ]; then
+            echo 'Helm-docs generated readmes match chart readmes.'
```
fedepaol, line 102: Do we need this? I think you can just `if !` and let the step succeed without a message

Also, is there a way to have it part of tasks so we can instruct the user on what command to run to bump?
Also also, I'd change the error message with the command to run to do the bump.

jtcarnes: Sure, I'll change the if statement.

I can put something similar to it in `tasks.py` if that's desired. The original idea was to just have  it in CI but I can include it in there if desired.

jtcarnes: @fedepaol if you want me to add it to `tasks.py`, let me know.

fedepaol:
Looks good overall, thanks for the addition!

I see CI is failing with 
```
/home/runner/work/_temp/96083f90-93b2-4170-b039-787f8c6ad3a0.sh: line 2: helm-docs: command not found
```
Also, maybe we can use the invtask instead of the gh action, which will ensure it works.
I'd ask to refine the commits. I see the merge one contains changes to the CI commit. Also, please add a body to the commit with the reason. We are trying to enforce https://cbea.ms/git-commit/ as a guideline for commits.

jtcarnes:
@fedepaol See the issue. Fixing now. The CI will fail though intentionally as your helm docs are out of sink. Do you want me to update them as part of this PR (so it passes that check) or leave that for a different MR?

fedepaol:
> @fedepaol See the issue. Fixing now. The CI will fail though intentionally as your helm docs are out of sink. Do you want me to update them as part of this PR (so it passes that check) or leave that for a different MR?

I'd say let's fix the docs, it's fine as long as you do it in a separate commit

jtcarnes:
@fedepaol sent the update, but because the version set in the Charts.yaml is still 0, it needs to be addressed.

PR #1390, opened 2022-05-25
Bump K8s API to 0.24

fedepaol:
The apis are now stable, we can bump. Also, we'll leverage the changes
to the test framework to comply to pod security admission.


oribon (approved):
lgtm

PR #1394, opened 2022-05-25
Switch add to copy for dockerfiles

jtcarnes:
While docker add and copy are similar, docker recommends as a best practice to use COPY over ADD (reference [here](https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy)).

Additionally, docker allows you to chmod during the copy command. By changing to using chmod during copy, we can eliminate a layer.

Signed-off-by: Joshua Carnes <56089764+jtcarnes@users.noreply.github.com>


fedepaol:
LGTM

jtcarnes:
@fedepaol it seems like there are issues with e2e testing? this change should have no effect so I'm unsure why it's failing

fedepaol:
> @fedepaol it seems like there are issues with e2e testing? this change should have no effect so I'm unsure why it's failing

Yes we are in the process of moving away from cert-manager and the operator needs to be updated. It's a chicken-egg problem we couldn't avoid.

PR #1396, opened 2022-05-26
Exit with code 1 on startup errors

oribon:
If there's a problem with client.Run the pod should
be in "Error" instead of "Completed".


fedepaol:
LGTM

PR #1403, opened 2022-05-29
Enable webhooks by default

sabinaaledort:
Now that we don't have the cert-manager dependency for webhooks, we can enable the webhooks by default.


controller/main.go
```diff
@@ -124,7 +124,7 @@ func main() {
 		logLevel            = flag.String("log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", logging.Levels.String()))
 		disableEpSlices     = flag.Bool("disable-epslices", false, "Disable the usage of EndpointSlices and default to Endpoints instead of relying on the autodiscovery mechanism")
 		enablePprof         = flag.Bool("enable-pprof", false, "Enable pprof profiling")
-		enableWebhook       = flag.Bool("enable-webhook", false, "Enable validation webhook")
+		enableWebhook       = flag.Bool("enable-webhook", true, "Enable validation and conversion webhooks")
```
fedepaol, line 127: How about getting rid of the flag completely and always turn the webhooks on? Is there a reason why we wouldn't want that?
Even if we had to disable them because of some mis-behaviour, we could remove the webhook configuration.

sabinaaledort: ok, sounds good to me.

controller/main.go
```diff
@@ -172,7 +171,7 @@ func main() {
 			PoolChanged:    c.SetPools,
 		},
 		ValidateConfig:      validation,
-		EnableWebhook:       *enableWebhook,
+		EnableWebhook:       true,
```
fedepaol, line 174: I'd get rid of the field and all the consequent logic, since we can't set it to false

fedepaol: Discussed offline, it's needed as the speaker must not enable the webhooks.

sabinaaledort: if we don't set if both the controller and speaker enable webhooks.

fedepaol:
Tried to clone the repo, I still see a frr-with-webhook / native-with-webhook folder which I think is redundant now? Also, they try to set the webhook parameter to the speaker.

fedepaol:
> Tried to clone the repo, I still see a frr-with-webhook / native-with-webhook folder which I think is redundant now? Also, they try to set the webhook parameter to the speaker.

Ignore this, they are marked as non tracked. Must be leftovers of my local copy.

fedepaol:
LGTM

PR #1416, opened 2022-06-07
Add validation for BGP and L2 advertisement

pperiyasamy:
This commit validates duplicate values in ip address pool, ip address
pool selector, node selector and peers fields and throw an error
when such duplicates are found.


fedepaol:
LGTM, thanks!

PR #1419, opened 2022-06-08
BFDProfile Webhook: Fix delete message

sabinaaledort:
The name of the BGP peer that's using the BFDProfile should be specified.


cgoncalves:
LGTM

PR #1420, opened 2022-06-08
Helm: Move CRDs to a subchart

sabinaaledort:
The recommended method in the helm documentation is using the crds/ directory to manage the CRDs, but this way we can't template the CRDs, which is needed for the conversion webhooks. We decided to apply the second recommended method and manage the CRDs in a subchart that will be packaged with the main chart.

https://helm.sh/docs/chart_best_practices/custom_resource_definitions/#method-2-separate-charts

Changing the webhook service name is required because it can't be templated in the subchart since helm is adding the name of both charts to the service prefix.


charts/metallb/crds/crds.yaml
```diff
@@ -25,8 +22,8 @@ spec:
         # but we're going to set it later using the cert-controller.
         caBundle: Cg==
         service:
-          namespace: {{ .Release.Namespace }}
-          name: {{ template "metallb.fullname" . }}-webhook-service
+          namespace: default
```
gclawes, line 25: Should `namespace` be `default` here?  Will it get patched later?

sabinaaledort: as far as i understood from the docs if we use the crds/ we can't template the file

gclawes: Looks like this is addressed here: https://github.com/helm/helm/issues/8171

fedepaol: So if I get the suggestion right, the idea is to provide a separate chart (without crds/, just a regular chart) to deploy the crds, and that can be templated. Is that right @gclawes ?
We'll have to publish two separate charts and adjust our release toolery, but that won't be an issue I guess.

sabinaaledort: from what i saw so far, i think keeping the chart as it is would be the best. i don't like the fact the crds are not templated and i don't see the advantage of adding another chart just for crds, but maybe i'm missing something here.

gclawes: I think it may be possible to use a sub-chart to do this: https://helm.sh/docs/chart_template_guide/subcharts_and_globals/

This would probably solve our issues around templating the webhook config in the CRDs, though it does mean the CRDs get auto-updated whenever they are changes (unless operations want to use `crds.enabled` to toggle them, see below).

I know some projects like `kube-prometheus-stack` are very deliberate about updating breaking CRD changes outside of helm (see [here](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack#upgrading-chart)), but they also a) use the `crds/` directory and b) do not use upgrade webhooks.  That being said, I'm _fairly_ sure the subchart approach would work for us.

Assuming we'd name the CRD sub-chart chart `crds`, place it in `charts/metallb/charts/crds/`.  Template values for that could be put in `charts/metallb/values.yaml` under `crds:`

In `charts/metallb/Chart.yaml`, add this as a dependency to allow the CRD sub-chart to be toggled (otherwise it is just enabled by default):
```
dependencies:
- name: crds
  condition: "crdsenabled"
```

This way, the sub-chart will just be packaged with the main chart (see my example of `main-chart`/`sub-chart` below).  I think the only change to the release process would be to update `version:`/`appVersion:` in `charts/metallb/charts/crds/Chart.yaml`

charts/metallb/charts/crds/templates/crds.yaml
```diff
@@ -26,7 +26,7 @@ spec:
         caBundle: Cg==
         service:
           namespace: {{ .Release.Namespace }}
-          name: {{ template "metallb.fullname" . }}-webhook-service
```
gclawes, line 29: Now that these are in a sub-chart, they should bue fully templatable.

sabinaaledort: but the name will be different between the subchart and the main chart, it is adding both names here to the prefix - metallb-metallb-crds (i wrote that in my commit message)

sabinaaledort: and anyway, why do we need the webhook service name templated?

fedepaol:
@gclawes mind giving one last look?

charts/metallb/charts/crds/templates/crds.yaml
sabinaaledort: I think it's ok to un-template the service name. Why do you think it will be incorrect if the user is using another name for the release? if we un-template that and use the same service name all over the manifests I think we are good. Am I missing something here?

gclawes: If a user uses a non-standard release name, (say `foo`) I'm fairly certain the templated service name will come out as `foo-metallb-webhook-service` because of this block:
https://github.com/metallb/metallb/blob/10df3fa158dd99b0ba687b625d49129a6483e03b/charts/metallb/templates/_helpers.tpl#L19-L23

By my read the above means "if the release name contains the chart name, render the fullname as the release name, else render the fullname as releasename-chartname".

This will be different than the non-templated name `metallb-webhook-service` defined in the CRD for conversion, and will cause conversion to fail. 

@fedepaol would you mind double-checking my logic here? Relevant docs for the helm `contains` function in that code block [here](https://helm.sh/docs/chart_template_guide/function_list/#contains).

fedepaol: Spoke with @gclawes , I think we can move forward with the current status (which is, the service name is fixed).
Having multiple instance of metallb using different names, especially in the same namespace seems non realistic so it should not be a problem.

fedepaol:
Operator lane failed because of the asn limit change not in the operator yet

fedepaol:
LGTM, thanks!

PR #1424, opened 2022-06-09
Add controller-runtime logger

sabinaaledort:
We need to set a controller-runtime logger to print all the logs created by it. For exmaple:
https://github.com/metallb/metallb/blob/359e54d75df4c930125535cd9b4ce4dff1c45ee6/internal/k8s/k8s.go#L60

Log is the base logger used by controller-runtime. It delegates to another logr.Logger that must be set by SetLogger to get any actual logging.


fedepaol:
LGTM

PR #1427, opened 2022-06-13
Add webhook mode

fedepaol:
We want to be able to deploy the controller as standalone backend for
the webhook. Also, for such deployments we want the "real" controller
not to act as webhook, so we add a webhook mode parameter to enable,
disable or run only in webhook mode.


oribon:
lgtm

PR #1429, opened 2022-06-13
WIP: Migrate frr components to their own images

### LINKED ISSUE #1428: Migrate frr components to their own containers
Currently both frr metrics and reloader are being carried inside of speaker and copied over to other containers during
the init process. This breaks container packaging, requires additional startup time during spinning up and forces speaker
to have a shell, which it doesn't need.

Instead, two separate containers should be created to better package the frr components, frr-reloader and frr-metrics. In this way, no startup copying is required, scratch space for passing binaries isn't required, and security can be improved.

This requires:
 - [x] creating two new container images/dockerfiles
 - [x] modifying the helm chart to provide values and defaults for these images
 - [ ] modifying the build and publish process to build and publish the two additional containers
 - [ ] remove frr-metrics and frr reloader from the speaker image

I am happy to work on the issue.

jtcarnes:
Currently, FRR components for metrics and reloaded are packaged into speaker and are then copied out into
other containers via initContainers. This requires speaker to have a shell and to copy these components to
other containers. Instead, containers which have the requisite components can be created instead so that
they are properly packaged and secured.

Fixes #1428

### DIFF
--- a/charts/metallb/README.md
+++ b/charts/metallb/README.md
@@ -83,6 +83,13 @@ A network load-balancer implementation for Kubernetes using standard routing pro
 | speaker.frr.image.pullPolicy | string | `nil` |  |
 | speaker.frr.image.repository | string | `"frrouting/frr"` |  |
 | speaker.frr.image.tag | string | `"v7.5.1"` |  |
+| speaker.frr.metrics.enabled | bool | `true` |  |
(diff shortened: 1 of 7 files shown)


fedepaol:
Sorry if I left this behind.
I remember the original discussion, and the idea was that we wanted to have the freedom to update the frr image without having to ship our own images.
This is the reason why we added all this bulkiness.
Closing this PR

PR #1430, opened 2022-06-14
Validate node, ip pool selector match expressions

pperiyasamy:
This validates duplicate values specified in node and ip pool
selectors match expressions under bgp and l2 advertisements.


internal/config/config.go
```diff
@@ -1005,8 +1011,32 @@ func validateDuplicate(strSlice []string, sliceType string) error {
 	return nil
 }
 
+func validateLabelSelectorMatchExpressions(matchExpressions []metav1.LabelSelectorRequirement) error {
```
fedepaol, line 1014: Since this is used in the function below, can we put it after it?
Ideally, higher level functions should come first.

pperiyasamy: done.

PR #1414, opened 2022-06-01
Build in container and update CI

jtcarnes:
This MR changes the build process for metallb to build inside a docker container instead of building outside and copying in. It further refactors much of the CI to better build images inside containers and leverage github actions instead of custom scripting.

Fixes #1392 #1408 #1413

PR #1437, opened 2022-06-15
WIP: Cosign images during publish

jtcarnes:
For security, artifact integrity and attestation are important. While a container sha takes care of integrity,
it does not satisfy attestation. The Linux foundations open source Sigstore projects are becoming the
defacto standard for container attestation. Kubernetes latest release was signed with cosign (the
sigstore signing utility).

This MR adds cosign to the release process easily. We use the github actions oidc token to authenticate
to fulcio and receive a short lived key to sign the images after we publish them. This allows
anyone to trivially check where the images have come from and their integrity.

Signing containers in this way is also the standard template for new repos that use github
actions.


jtcarnes:
@fedepaol this MR relies on #1414 to work properly. Once this is done, this should be an easy rebase.

jtcarnes:
@fedepaol whenever you get the chance, lmk

.github/workflows/publish.yaml
```diff
@@ -98,3 +106,9 @@ jobs:
           build-args: |
             GIT_BRANCH: ${{ github.ref_name }}
             GIT_COMMIT: ${{ github.sha }}
+
+      - name: Sign the images with GitHub OIDC Token
+        run: cosign sign ${TAGS}
+        env:
+          TAGS: ${{ steps.meta.outputs.tags }}
```
mattmoor, line 113: It is better practice to sign the digest here, e.g. https://github.com/actions/starter-workflows/blob/3603e1f7d1629d3e9b99a7fd2e6ad72e42e8e84e/ci/docker-publish.yml#L93

mattmoor:
I came here to suggest y'all sign stuff with `cosign`, and found this PR!  FWIW, I use metallb a fair amount and would love to see this land.

.github/workflows/publish.yaml
jtcarnes: True and also possible. Only reason i hasn't is this seemed a smaller change but i can add trivially.

jtcarnes: Should be good to go now, but happy to hear your thoughts. If fine, mark as resolved.

fedepaol:
LGTM, I am gonna test the full release process against my fork tomorrow before merging

PR #1439, opened 2022-06-17
feat: support private registry deploy for helm chart

tgfree7:
Charts: add support for private registry deploy
Signed-off-by: tgfree <tgfree7@gmail.com>

### DIFF
--- a/charts/metallb/templates/_helpers.tpl
+++ b/charts/metallb/templates/_helpers.tpl
@@ -79,3 +79,10 @@ Create the name of the settings Secret to use.
 {{- define "metallb.secretName" -}}
     {{ default ( printf "%s-memberlist" (include "metallb.fullname" .)) .Values.speaker.secretName | trunc 63 | trimSuffix "-" }}
 {{- end -}}
+
+{{/*
+Create the json of the registry Secret to use.
+*/}}
+{{- define "metallb.registrySecret" }}
+{{- printf "{\"auths\": {\"%s\": {\"auth\": \"%s\"}}}" .Values.privateRegistry.registryUrl (printf "%s:%s" .Values.privateRegistry.registryUser .Values.privateRegistry.registryPasswd | b64enc) | b64enc }}
+{{- end }}
--- a/charts/metallb/templates/controller.yaml
+++ b/charts/metallb/templates/controller.yaml
@@ -30,10 +30,6 @@ spec:
       {{- with .Values.controller.runtimeClassName }}
       runtimeClassName: {{ . | quote }}
       {{- end }}
-      {{- with .Values.imagePullSecrets }}
-      imagePullSecrets:
-        {{- toYaml . | nindent 8 }}
-      {{- end }}
       serviceAccountName: {{ template "metallb.controller.serviceAccountName" . }}
       terminationGracePeriodSeconds: 0
 {{- if .Values.controller.securityContext }}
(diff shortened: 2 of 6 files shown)


gclawes:
I don't think this is necessary.

1. Cluster operators will likely have their own registry secret in the cluster that is independent of the metallb install (If they are installing metallb from a private registry, it is likely they are installing other containers from the same).  I don't believe it's necessary for the metallb chart to install this secret.  
2. `.Values.privateRegistry.registryUrl` is unnecessary, the `.Values.*.image.repository ` values can just be modified to use the private registry.  For example:
```
controller:
  image:
    repository: myprivateregistry.com/metallb/controller
```
3.  There may be value in moving `imagePullSecrets:` from the controller pod templates to service accounts, I'd be open to a discussion in a separate PR about the merits of this.

tgfree7:
> I don't think this is necessary.
> 
> 1. Cluster operators will likely have their own registry secret in the cluster that is independent of the metallb install (If they are installing metallb from a private registry, it is likely they are installing other containers from the same).  I don't believe it's necessary for the metallb chart to install this secret.
> 2. `.Values.privateRegistry.registryUrl` is unnecessary, the  values can just be modified to use the private registry.  For example:`.Values.*.image.repository `
> 
> ```
> controller:
>   image:
>     repository: myprivateregistry.com/metallb/controller
> ```
> 
> 3. There may be value in moving  from the controller pod templates to service accounts, I'd be open to a discussion in a separate PR about the merits of this.`imagePullSecrets:`

thanks for reply.
After few months of use, in our team, we will push the images to our private registry or repository. so the imagePullSecrets may public (setting in .value. imagePullSecrets) or private (setting in .value.privateRegistry) in the k8s cluster. this why i want support the private registry or repository.

gclawes:
I don't see why it's necessary to change the chart, you can already support that with the way the chart works today.  You just need to set the `.Values.*.image.repository` values as I showed above, and provide your own secret, which we already support referencing with `.Values.imagePullSecrets`.

There's no distinction between `.Values.imagePullSecrets` and `.Values.privateRegistry.registrySecret` when rendered here in the template:
https://github.com/metallb/metallb/blob/4e3690a3117c00786f7b044d248bbf0dc6330e0a/charts/metallb/templates/service-accounts.yaml#L14-L21
Nothing is "private" or "public" about the secrets referenced here. 

If you need to render a pull secret in a helm chart, there appear to be a chart to do this on ArtifactHub: https://artifacthub.io/packages/helm/snowplow-devops/dockerconfigjson

tgfree7:
> I don't see why it's necessary to change the chart, you can already support that with the way the chart works today. You just need to set the `.Values.*.image.repository` values as I showed above, and provide your own secret, which we already support referencing with `.Values.imagePullSecrets`.
> 
> There's no distinction between `.Values.imagePullSecrets` and `.Values.privateRegistry.registrySecret` when rendered here in the template:
> 
> https://github.com/metallb/metallb/blob/4e3690a3117c00786f7b044d248bbf0dc6330e0a/charts/metallb/templates/service-accounts.yaml#L14-L21
> 
> 
> Nothing is "private" or "public" about the secrets referenced here.
> If you need to render a pull secret in a helm chart, there appear to be a chart to do this on ArtifactHub: https://artifacthub.io/packages/helm/snowplow-devops/dockerconfigjson

> 
thanks for reply. sorry, not to make myself clear. 

such as `harbor`, we can create userA for private projectA, and we push metallb's images to private projectA.
so we should create our own imagePullSecrets by `.Values.privateRegistry` , that's `private secret`.  
if we push to a public projectB, and imagePullSecrets already existed `public secret`, just set 
`.Values.imagePullSecrets` is working.
image

About ` moving from the controller pod templates to service accounts`, i will open a separate PR first.

PR #1440, opened 2022-06-17
Fix docker metadata action

jtcarnes:
Fixes release CI's docker metadata action to properly generate tags
instead of failing.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -75,7 +75,7 @@ jobs:
             quay.io/metallb/${{ matrix.image }}
           # generate Docker tags based on the following events/attributes
           tags: |
-            type=ref,event=push
+            type=ref,event=branch
             type=semver,pattern={{version}}
           labels: |
             org.opencontainers.image.title=${{ matrix.image }}


jtcarnes:
@fedepaol this should fix the issue with publish from #1414

.github/workflows/publish.yaml
```diff
@@ -75,7 +75,7 @@ jobs:
             quay.io/metallb/${{ matrix.image }}
           # generate Docker tags based on the following events/attributes
           tags: |
-            type=ref,event=push
+            type=ref,event=branch
```
fedepaol, line 78: Will this work when we do a tag?

jtcarnes: Yes, the other tag type (type=Denver) will generate the server tag

fedepaol: sorry, which other tag type?

jtcarnes: The next line, line 79, `type=semver,pattern={{version}}`

jtcarnes: An example ci job front docker metadata: https://github.com/docker/metadata-action#semver

fedepaol:
LGTM, let's see if CI is happy now.
Thanks!

fedepaol:
Another issue, sorry 
https://github.com/metallb/metallb/runs/6935416539?check_suite_focus=true#step:6:122

fedepaol:
np, I'll have a look and fix myself

fedepaol:
@jtcarnes fixed it. What was missing was also the checkout action

PR #1441, opened 2022-06-17
Publish: checkout before publishing

fedepaol:
We need to checkout otherwise the dockerfiles won't be available.
Moreover, we set push to true to push the images.


fedepaol:
Double checked the differences with the examples and with the `build-test-images` job, seems checkout was missing.
Also, adding push: true to push the new images.

Issue #1402, opened 2022-05-28
MP BGP (ipv6) is broken in main with crds


nberlee:
Using 0.12 i was able to do MP-BGP with an ipv4 bgp peer (advertising also ipv6). However, when running latest main with crds, I am unable to get MP-BGP working.

The controller says in events it will advertise the ipv6 service on two nodes. But when doing in the corresponded speaker pod frr:
```
vtysh
show running-config
```
It is missing a `address-family ipv6 unicast` section and `ipv6 prefix-list 100-v6localpref-prefixes seq 5 permit` ... line(s)  compared to 0.12

In the remote bgp router i only see AFI `ip` sesions, not `ip ipv6` sessions.

Reverting back to 0.12 makes MP-BGP function again.

oribon:
Hi :smiley:, can you please share the speakers and controller logs (setting the loglevel parameter to debug before sending the logs if possible), CRs, the services spec when it is not working + the configmap that works when you're using this in 0.12?

nberlee:
I runned the speaker in debug and saw noting suspicious. But I already spotted the bug in the code:
https://github.com/metallb/metallb/commit/0f33257c2b1533d211e82746fd5189c292052c95#diff-799fb51a85bb84147c1487eb5d728663c098fffce7f326c412fac1f619860516R272

I am build now a version without the:
```
			if family != neighbor.IPFamily {
				continue
			}
```
To confirm that this is the problem. If this fixes it, I will create a pull request. If not, i will sanitize my logs (as they contain my public ipv6 address) and upload them here

Jean-Daniel:
> Could you also share the usecase - why you need to advertise v6 prefixes to a v4 peer? Does the peer have an IPv6 that you can peer with for now as a workaround?

Isn't it the main purpose of MP-BGP. Being able to advertise IPv6 prefixes on a v4 peers ? I'm using it too because having to create 2 sessions per peer would create too much sessions on my router. 

Moreover, I have devices where configuring BGP session using link-local addresses is more complex than it should be (and poorly supported),  so it would requires to create IPv6 networks only used to pass BGP advertisements. Actually,  my peers are advertising there link-local address as next-hop.

fedepaol:
@nberlee any news about this?

[...]

PR #1444, opened 2022-06-20
Restore MP-BGP

fedepaol:
Changing the way we render the config to accommodate MP-BGP.
We advertise the ips regardless of the ip family of the session established with the neighbor, plus we add e2e tests for that.

Fixes https://github.com/metallb/metallb/issues/1402


oribon (commented):
nice! few nits

oribon:
lgtm

PR #1446, opened 2022-06-22
Allow for 4-byte AS numbers

leppeK:
To accomodate for 4-byte AS numbers as described in rfc6793 we need to up the maximum value for the AS fields

Currently running the changes without issues, the implementation does already support these values.


fedepaol:
Hi @leppeK , thanks for the PR.

The native bgp implementation supports 4 bytes asn but the new frr one does not, so in order to have this working I'd also change the frr validation function (here https://github.com/metallb/metallb/blob/9fd6ec9ce038a4ade0e60a45d32cd63d8fbb29da/internal/config/validation.go#L66 ) so that the validating webhooks will refuse a 4 bytes asn if we are running using frr as the bgp backend.

Also, it may make sense to add an e2e test that validates that the webhook is working.

Thanks!

leppeK:
> Hi @leppeK , thanks for the PR.
> 
> The native bgp implementation supports 4 bytes asn but the new frr one does not, so in order to have this working I'd also change the frr validation function (here
> 
> https://github.com/metallb/metallb/blob/9fd6ec9ce038a4ade0e60a45d32cd63d8fbb29da/internal/config/validation.go#L66
> 
> ) so that the validating webhooks will refuse a 4 bytes asn if we are running using frr as the bgp backend.
> Also, it may make sense to add an e2e test that validates that the webhook is working.
> 
> Thanks!

Hi @fedepaol i checked the FRR implementation and 4 bytes seems to just work fine.

e2etest/bgptests/infra_setup.go
```diff
@@ -18,7 +18,7 @@ import (
 const (
 	multiHopNetwork = "multi-hop-net"
 	metalLBASN      = 64512
-	externalASN     = 64513
+	externalASN     = 4200000000
```
fedepaol, line 21: @oribon @gclawes wdyt about this change?
I think that using a 2 bytes for ibgp and 4 bytes for external is enough to demonstrate that we support it, but would like to hear your thought as this is gonna impact all our e2es.

oribon: seems good enough

fedepaol:
I did not notice the commit messages should have been fixed, I guess we'll live with that

PR #1452, opened 2022-06-23
Always use the main branch of the operator for CI

fedepaol:
We normally align the operator to metallb changes, and this will avoid
the extra jump of bumping back the operator on metallb ci due to a
change in metallb itself.


oribon:
lgtm

PR #1458, opened 2022-06-28
Change the helm releaser script

fedepaol:
The official helm release gh action doesn't work from tags (see https://github.com/helm/chart-releaser-action/issues/60)
Here we take inspiration from https://github.com/helm/chart-releaser-action/issues/60#issuecomment-983754215 in order to craft a custom workflow to release the charts.
Tested on my local repo.


fedepaol:
@gclawes I think this is a bit better than using a fork, ptal

gclawes:
Looks good to me.

PR #1460, opened 2022-06-28
Restore the k8s events metrics

fedepaol:
In migrating to CRDs, we omitted filling the metrics related to k8s
events. Here we re introduce them.


oribon (commented):
lgtm

Issue #1462, opened 2022-06-29
FRR interaction race condition


fedepaol:
As per https://kubernetes.slack.com/archives/C8FSTGU56/p1656390088462949 , there's a race when:
- the metallb configuration is already in place
- the speaker pod is not started yet
- the speaker container is faster than the reloader to come up (then, the pid file is not avaliable and signaling fails)

```
{"caller":"level.go:63","cause":"reload","config":{"Loglevel":"informational","Hostname":"node-2","Routers":{"64916@\u003cnil\u003e":{"MyASN":10000,"RouterId":"","Neighbors":{"20000@peer-1:179":{"ASN":20000,"Addr":"peer-1","SrcAddr":"","Port":179,"HoldTime":90,"KeepaliveTime":30,"Password":"supersecret","Advertisements":[{"IPFamily":"ipv4","Prefix":"8.8.8.8/32","Communities":[],"LocalPref":0}],"BFDProfile":"full1","EBGPMultiHop":false}}}},"BFDProfiles":[{"Name":"full1","ReceiveInterval":250,"TransmitInterval":250,"DetectMultiplier":null,"EchoInterval":50,"EchoMode":true,"PassiveMode":false,"MinimumTTL":254}],"PrefixesV4ForCommunity":[],"PrefixesV6ForCommunity":[],"PrefixesV4ForLocalPref":[],"PrefixesV6ForLocalPref":[]},"error":"open /etc/frr_reloader/reloader.pid: no such file or directory","level":"error","op":"reload","ts":"2022-06-28T01:36:33.496165268Z"}
{"caller":"level.go:63","level":"info","msg":"triggering discovery","op":"memberDiscovery","ts":"2022-06-28T01:38:55.551459352Z"}
```

PR #1463, opened 2022-06-29
FRR configuration retries

fedepaol:
Making the interaction with FRR more robust introducing retries after failures in both the call of the reloader (if the pid is not there yet) and if we find that the interaction between the reloader and the container fails.

Fixes https://github.com/metallb/metallb/issues/1462

### DIFF
--- a/internal/bgp/frr/config.go
+++ b/internal/bgp/frr/config.go
@@ -167,6 +167,11 @@ type frrConfig struct {
 	BFDProfiles []BFDProfile
 }
 
+type reloadEvent struct {
+	config *frrConfig
+	useOld bool
+}
+
 // TODO: having global prefix lists works only because we advertise all the addresses
 // to all the neighbors. Once this constraint is changed, we may need prefix-lists per neighbor.
 
@@ -334,7 +339,7 @@ func generateAndReloadConfigFile(config *frrConfig, l log.Logger) error {
 // the update requests are sent, and squashes any requests coming in a given timeframe
 // as a single request.
 func debouncer(body func(config *frrConfig) error,
-	reload <-chan *frrConfig,
+	reload <-chan reloadEvent,
 	reloadInterval time.Duration,
 	failureRetryInterval time.Duration) {
 	go func() {
@@ -347,7 +352,12 @@ func debouncer(body func(config *frrConfig) error,
 				if !ok { // the channel was closed
 					return
 				}
-				config = newCfg
+				if newCfg.useOld && config == nil {
+					continue // just ignore the event
+				}
+				if !newCfg.useOld {
+					config = newCfg.config
+				}
 				if !timerSet {
 					timeOut = time.After(reloadInterval)
 					timerSet = true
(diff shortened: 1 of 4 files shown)


internal/bgp/frr/config.go
```diff
@@ -345,13 +352,23 @@ func debouncer(body func(config *frrConfig),
 				if !ok { // the channel was closed
 					return
 				}
-				config = newCfg
+				if newCfg.useOld && config == nil {
+					continue // just ignore the event
+				}
```
oribon, line 357: not sure I understand this :/ why sending something to the chan in the first place only to do nothing with it?

internal/bgp/frr/debounce_test.go
```diff
@@ -44,3 +48,69 @@ func TestDebounce(t *testing.T) {
...
+	}
+	// reload to see if the debouncer uses the old config
+	reload <- reloadEvent{useOld: true}
+	time.Sleep(3 * timer)
+	if len(result) != 1 {
+		t.Fatal("received extra updates", len(result))
+	}
+	updated = <-result
+	if updated.Hostname != "1" {
+		t.Fatal("Config was not updated")
```
oribon, line 114: shouldn't this be "Config was updated"?

internal/bgp/frr/config.go
```diff
@@ -345,13 +352,23 @@ func debouncer(body func(config *frrConfig),
...
 				if !timerSet {
 					timeOut = time.After(reloadInterval)
 					timerSet = true
 				}
 			case <-timeOut:
-				body(config)
+				err := body(config)
+				if err != nil {
+					timeOut = time.After(failureRetryInterval)
+					timerSet = true
```
oribon, line 369: does this mean that after a fail the next config update won't happen until the failureretryinterval finishes? shouldn't we let new updates override this timeout?

internal/bgp/frr/config.go
fedepaol: This should never happen, because it means that we received a "reload old config" event but the old config is nil (meaning, we already tried a reload that failed).
Nonethless, I'd prefer not having the speaker crash but ignore the event in such corner case.

internal/bgp/frr/config.go
fedepaol: That is correct, and I thought that it makes sense to go slow after a failure happened (think about the case of the slow container startup, after a failure we want to give it time to recover). 5 seconds is still a bearable timeout for applying configuration changes.

internal/bgp/frr/debounce_test.go
fedepaol: Nope, because what writes to result is the action to update the config. So in this case, if we don't get the value, it means that the retry mechanism did not trigger and that the update function was not called

fedepaol:
Tested also manually by putting a sleep in the reloader script first, and in the frr entry point after, and the speaker recovered.

internal/bgp/frr/config.go
oribon: sorry, I misread it as `newCfg.config == nil` :sweat_smile:

oribon (commented):
lgtm

Issue #1457, opened 2022-06-28
talos v1.1/Kubernetes v.1.24.2/MetalLB gets all kinds of pod security issues, and speaker will NOT run.


lrosenman:
Kubernetes: v1.24.2
base OS: Talos V1.1.0
MetalLB: 0.12.1
Warning; I'm just starting with k8s.
When I deploy the manifests from the install doc, the speaker gets:
```❯ kubectl describe -n metallb-system ds speaker
Name:           speaker
Selector:       app=metallb,component=speaker
Node-Selector:  kubernetes.io/os=linux
Labels:         app=metallb
                component=speaker
Annotations:    deprecated.daemonset.template.generation: 1
Desired Number of Nodes Scheduled: 0
Current Number of Nodes Scheduled: 0
Number of Nodes Scheduled with Up-to-date Pods: 0
Number of Nodes Scheduled with Available Pods: 0
Number of Nodes Misscheduled: 0
Pods Status:  0 Running / 0 Waiting / 0 Succeeded / 0 Failed
Pod Template:
  Labels:           app=metallb
                    component=speaker
  Annotations:      prometheus.io/port: 7472
                    prometheus.io/scrape: true
  Service Account:  speaker
  Containers:
   speaker:
    Image:       quay.io/metallb/speaker:v0.12.1
    Ports:       7472/TCP, 7946/TCP, 7946/UDP
    Host Ports:  7472/TCP, 7946/TCP, 7946/UDP
    Args:
      --port=7472
      --config=config
      --log-level=info
    Liveness:   http-get http://:monitoring/metrics delay=10s timeout=1s period=10s #success=1 #failure=3
    Readiness:  http-get http://:monitoring/metrics delay=10s timeout=1s period=10s #success=1 #failure=3
    Environment:
      METALLB_NODE_NAME:       (v1:spec.nodeName)
      METALLB_HOST:            (v1:status.hostIP)
      METALLB_ML_BIND_ADDR:    (v1:status.podIP)
      METALLB_ML_LABELS:      app=metallb,component=speaker
      METALLB_ML_SECRET_KEY:  <set to the key 'secretkey' in secret 'memberlist'>  Optional: false
    Mounts:                   <none>
  Volumes:                    <none>
Events:
  Type     Reason        Age                  From                  Message
  ----     ------        ----                 ----                  -------
[...]

PR #1467, opened 2022-07-04
Pod security admission support

fedepaol:
From 1.23, pod security admission can require the namespace to be labelled in order to allow the speaker pod to access to the network stack.
Here we add the labels to kustomize / manifests, and we document what labels need to be added to the ns in case the user is using helm (which requires the ns to be created externally).


fedepaol:
Fixes https://github.com/metallb/metallb/issues/1457

PR #1470, opened 2022-07-06
Publish job: change the branch / tag selector

fedepaol:
During the latest release, the job did not trigger. Using the same as
the helm chart which was triggered.


cgoncalves:
LGTM

Issue #1481, opened 2022-07-06
webhook 500 errors


jr0dd:
## environment

MetalLB `v0.13.2`
Kubernetes `v1.24.2`
CNI `cilium v1.11.6`
kube-proxy `not deployed`

## The bug itself, as detailed as you can.

After the initial install it took multiple attempts to apply the CR's to the cluster. The validating webhook was returning error 500 on the attempts and then finally accepted them. This greatly affects clusters managed through GitOps as reconciliations are failing.

`k apply -f cluster/apps/networking/metallb/custom-resources/bgp-advertisement.yaml`
``` 
Warning: resource bgpadvertisements/bgp-lb is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by kubectl apply. kubectl apply should only be used on resources created declaratively by either kubectl create --save-config or kubectl apply. The missing annotation will be patched automatically.
Error from server (InternalError): error when applying patch:
{"metadata":{"annotations":{"kubectl.kubernetes.io/last-applied-configuration":"{\"apiVersion\":\"metallb.io/v1beta1\",\"kind\":\"BGPAdvertisement\",\"metadata\":{\"annotations\":{},\"name\":\"bgp-lb\",\"namespace\":\"networking\"},\"spec\":{\"ipAddressPools\":[\"bgp-pool\"],\"peers\":[\"mikrotik\"]}}\n"}}}
to:
Resource: "metallb.io/v1beta1, Resource=bgpadvertisements", GroupVersionKind: "metallb.io/v1beta1, Kind=BGPAdvertisement"
Name: "bgp-lb", Namespace: "networking"
for: "cluster/apps/networking/metallb/custom-resources/bgp-advertisement.yaml": Internal error occurred: failed calling webhook "bgpadvertisementvalidationwebhook.metallb.io": failed to call webhook: Post "https://metallb-webhook-service.networking.svc:443/validate-metallb-io-v1beta1-bgpadvertisement?timeout=10s": proxy error from 127.0.0.1:6443 while dialing 172.20.0.186:9443, code 500: 500 Internal Server Error
```

### controller errors with info logging
```
[...]

PR #1487, opened 2022-07-07
Charts: add the selectorlabels to the webhook service and give controller permission to watch communities

fedepaol:
The selector is only aiming at app.kubernetes.io/component, so if
there's another pod in the same namespace labeled controller, the
webhook service won't work (or it will work only if the service balances
the request to the real controller).

Also, give the controller the required permissions to watch communities.


fedepaol:
Potentially, a fix for https://github.com/metallb/metallb/issues/1481 given what was discussed in https://github.com/metallb/metallb/issues/1481#issuecomment-1177209775

oribon:
lgtm

Issue #1471, opened 2022-07-06
New Helm chart released before publishing container images


VeetiPoutsalo:
As of writing this issue, there has been a period of almost an hour where attempting to install the latest Helm release of MetalLB results in an error due to it referring to an unpublished version of the container images.

The v0.13.0 release updated the MetalLB Helm chart (https://artifacthub.io/packages/helm/metallb/metallb) before the v0.13.0 containers had been pushed to quay.io (https://quay.io/repository/metallb/speaker?tab=tags).

To fix this, any new release Helm charts should not be published before all prerequisite container images have been successfully published, to ensure minimal disruption to Helm installation.

fedepaol:
Apologies, we have some unexpected issues with the changes in our publishing process.
We are removing the release and try to fix it smoothly

fedepaol:
Update: now the release is reverted, as the changes to the chart (not sure if artifacthub will reflect that, but we have no control).
We'll make sure the pipeline is working and retry.

fedepaol:
The new 0.13.2 release is out, hopefully works as expected.
I am gonna keep this issue open to track the fact that we want to have dependencies between the image publishing job and teh helm release job.

Also, if you are automatically upgrading metallb, please be aware of the fact that the configmap won't work anymore: https://github.com/metallb/metallb/issues/1473

PR #1496, opened 2022-07-11
CI: publish the charts after the images

fedepaol:
The chart publishing job is faster than publishing the images, so when
we release there's a time window where the charts are ready but the
images are not, putting automatic upgrades at risk.
Here we add the chart job to the same lane of images, making it depend
on the images in order to run.

To be tested, fixes https://github.com/metallb/metallb/issues/1471


fedepaol:
Tested on my fork, https://github.com/fedepaol/metallb/actions/runs/2649286951

oribon:
lgtm

Issue #1494, opened 2022-07-10
change webhook name


jr0dd:
I just wanted to suggest changing the webhook name to something more descriptive.

```
mb-pro home-ops % k get validatingwebhookconfigurations.admissionregistration.k8s.io                                   
NAME                               WEBHOOKS   AGE
cert-manager-webhook               1          4d8h
ingress-nginx-admission            1          164d
kube-prometheus-stack-admission    1          7d22h
validating-webhook-configuration   7          3d20h.            <~~~ metallb-admission or something something
```

fedepaol:
Makes sense. As discussed previously, we did not think about a use case where metallb was mixed with other components in the namespace, so despite the name was generic it did not see a big issue.
We'll change it.

jr0dd:
> Makes sense. As discussed previously, we did not think about a use case where metallb was mixed with other components in the namespace, so despite the name was generic it did not see a big issue.
> We'll change it.

Awesome! Actually, I just realized that `validatingwebhookconfigurations` are not namespaced anyways.

fedepaol:
> > Makes sense. As discussed previously, we did not think about a use case where metallb was mixed with other components in the namespace, so despite the name was generic it did not see a big issue.
> > We'll change it.
> 
> Awesome! Actually, I just realized that `validatingwebhookconfigurations` are not namespaced anyways.

Right, so that's even worse :/
Will change that

nayrnet:
found this issue the hard way when I tried to install MetaLB into a cluster that had RabbitMQ, which also decided to use the name `validating-webhook-configuration`

looks like they recently changed that behavior too: https://github.com/rabbitmq/messaging-topology-operator/commit/f56f89b293e5453f7f5c7272a0b08fc3af723372

PR #1497, opened 2022-07-11
Change the validating webhook configuration to contain metallb

fedepaol:
The validating wh configuration is a cluster scoped object, so it makes
sense to have its name reference metallb instead of having a generic
name that makes it difficult to understand its relationship with
metallb.

Fixes https://github.com/metallb/metallb/issues/1494


charts/metallb/templates/webhooks.yaml
```diff
@@ -1,7 +1,7 @@
 apiVersion: admissionregistration.k8s.io/v1
 kind: ValidatingWebhookConfiguration
 metadata:
-  name: validating-webhook-configuration
+  name: metallb-webhook-configuration
```
gclawes, line 4: Should this be templated? `{{ template "metallb.fullname" . }}`. Or is it always hardcoded to `metallb-webhook-configuration` in the code?

fedepaol: It's hardcoded here https://github.com/metallb/metallb/pull/1497/files#diff-14c68fe5b25ee985e2af6d5dc8c7eaef1b4d91f0dac32019bf971d5681c803beR62 (that's needed for the cert-controller to patch the webhooks), so I guess it's fine as it is.
If we want to make it configurable, we should also change the controller to accept it as a parameter, but I am not sure it's worth the effort.

oribon (approved):
lgtm

Issue #1495, opened 2022-07-10
CR Converstion tool, failed to generate resources: invalid aggregation length 24


jonwilliams84:
Take this configmap:
```
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: metallb-system
  name: config
data:
  config: |
    peers:
  config: |
    peers:
    - peer-address: 10.32.100.1
      peer-asn: 64588
      my-asn: 64500
    address-pools:
    - name: default
      protocol: bgp
      addresses:
      - 10.0.8.0/24
      avoid-buggy-ips: true
      bgp-advertisements:
      - aggregation-length: 32
        localpref: 100
      - aggregation-length: 24
```

When running it through the CM to CR conversion tool during upgrade of MetalLB 12.1 > 13.3, I get this error:

failed to generate resources: invalid aggregation length 24: prefix 26 in this pool is more specific than the aggregation length for addresses 10.0.8.1-10.0.8.254

Commenting out the /24 aggregation-length line the CR is generated:
```
# This was autogenerated by MetalLB's custom resource generator.
apiVersion: metallb.io/v1beta2
kind: BGPPeer
metadata:
  creationTimestamp: null
  name: peer1
  namespace: metallb-system
spec:
  holdTime: 1m30s
  keepaliveTime: 30s
  myASN: 64500
  passwordSecret: {}
  peerASN: 64588
  peerAddress: 10.32.100.1
status: {}
---
apiVersion: metallb.io/v1beta1
kind: IPAddressPool
metadata:
  creationTimestamp: null
  name: default
  namespace: metallb-system
spec:
  addresses:
  - 10.0.8.1-10.0.8.254
status: {}
---
apiVersion: metallb.io/v1beta1
kind: BGPAdvertisement
metadata:
  creationTimestamp: null
  name: bgpadvertisement1
  namespace: metallb-system
spec:
  aggregationLength: 32
  ipAddressPools:
  - default
  localPref: 100
status: {}
---
```

Functionally, I realise this will work but is there any reason it doesn't like the /24 aggregation on a pool that is /24?

fedepaol:
> 10.0.8.1-10.0.8.254

I see what the problem is, still need to find the best way to address it.
Your `avoidBuggyIPs` is effectively translating the /24 pool to a pool that is `10.0.8.1-10.0.8.254`. 
[...]

PR #1503, opened 2022-07-14
Remove the cidr check with ip ranges

fedepaol:
Implementing a reliable cidr check without offering the remove buggy ips
flag is not possible.

A user that wants a full cidr without buggy ips will have to specify the
range without the extremes, and trying to reverse the logic of removing
the buggy ips is not reliable.

For this reason, we remove the consistency check between the cidr and
the aggregation length in case the cidr comes from a manual range.

Potential fix for the first issue described in https://github.com/metallb/metallb/issues/1495 (but also affecting regular users who want to avoid the buggyips).

Potential alternatives to this implementation:
- we competely remove the check, making the behaviour simmetric
- we restore the avoidbuggyips flag

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -753,8 +753,12 @@ func bgpAdvertisementsFromLegacyCR(ads []metallbv1beta1.LegacyBgpAdvertisement, 
 				maxLength = ad.AggregationLengthV6
 			}
 
-			// in case of range format, we may have a set of cidrs associated to a given address.
-			// We reject if none of the cidrs are compatible with the aggregation length.
+			// we don't check in case of range formats, because we don't know if the originating range
+			// is a full range minus buggy ips.
+			if strings.Contains(addr, "-") {
+				continue
+			}
+
 			lowest := lowestMask(cidrs)
 			if maxLength < lowest {
 				return nil, fmt.Errorf("invalid aggregation length %d: prefix %d in "+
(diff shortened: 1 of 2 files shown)


oribon (commented):
can you please explain why  `trying to reverse the logic of removing
the buggy ips is not reliable` / impossible?

nobiit:
I'm quite interested in this PR, hope I can help

fedepaol:
> can you please explain why `trying to reverse the logic of removing the buggy ips is not reliable` / impossible?

Because in general the behaviour is not reliable and we can't always reverse engineer the user's will.
The clear scenario is a /24 cidr that becomes:
`10.0.1.1 - 10.0.1.254`

In this case, we decrement the lowest, increment the highest, see if it results in a single CIDR translation and everything works.

But if we take a /26 like `10.0.10.0/26` for example translates to:
`10.0.1.1-10.0.1.63` (because then, only the .0 is problematic)

Plus, you never know if a user will ask for `10.0.1.1 - 10.0.1.254` because those 2 other ips are used somewhere else and advertising the /24 cidr will break something.

All in all, my feeling is that we have to deal with too many corner cases and the value added by the extra validation is not worth the complexity.
Also, now that I read it, I even realize the conversion logic in case of avoidbuggyips is wrong, because we increment / decrement anyway.

fedepaol:
Closing this in favor of bringing back the avoidbuggyips field.
The use case demonstrated it's not equivalent to specifying the range because of the aggrlength incompatibility.

PR #1506, opened 2022-07-15
L2: add ndp responders when the address is v6 only

fedepaol:
A regression introduced by
https://github.com/metallb/metallb/pull/1347, we were adding records to
the keepNDP map regardless of the fact that a v6 address was present on
the interface.
Here we enforce the if checking that the ip is not v4.

Fixes https://github.com/metallb/metallb/issues/1507

### DIFF
--- a/internal/layer2/announcer.go
+++ b/internal/layer2/announcer.go
@@ -95,7 +95,7 @@ func (a *Announce) updateInterfaces() {
 			if ipaddr.IP.To4() != nil && (ifi.Flags&net.FlagBroadcast) != 0 {
 				keepARP[ifi.Index] = true
 			}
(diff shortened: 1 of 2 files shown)


cgoncalves:
LGTM. It would be nice to have some test coverage to prevent regressions.

fedepaol:
> LGTM. It would be nice to have some test coverage to prevent regressions.

Yep, the problem is that docker gets the dual stack and this scenario slipped from CI. I'll try to investigate if there's a way to force v4 only nodes with kind.

fedepaol:
Filed https://github.com/metallb/metallb/issues/1508 to track the issue

pperiyasamy:
Thanks @fedepaol for the fix !

PR #1514, opened 2022-07-18
Manifests: fix the config webhook namespace

fedepaol:
More for consistency rather than real need, as the namespace gets
overridden when applying kustomize.


oribon (approved):
lgtm

PR #1515, opened 2022-07-18
Add avoidBuggyIPs to the IPAddressPool CIDR

fedepaol:
We originally thought that the avoidBuggyIPs flag was replaceable by picking the equivalent range of the cidr minus the extremes.
The issue is, the aggregation length validation won't work in that case because a 10.0.1.1-10.1.1.254 range doesn't correspond to a /24 cidr.
Because of that, we believe that bringing the flag back is safer than trying to understand if a range belongs to a cidr minus the buggy ips (and if the user's intention was to avoid them in first place).


oribon (approved):
lgtm

Issue #1517, opened 2022-07-18
Validating webhooks: make sure the crs are created in the metallb namespace


fedepaol:
Currently, there's nothing preventing the CRs to be deployed in a namespace different from the one metallb is deployed to.
We should validate that they are being created in the same ns, to prevent human errors.

muffl0n:
Is this fixed cause #1523 is merged?

oribon:
> Is this fixed cause #1523 is merged?

I think so

PR #1523, opened 2022-07-19
added namespace validation for custom resources

mgsh:
Fixed https://github.com/metallb/metallb/issues/1517


mgsh:
Hi @fedepaol , I've attempted a fix for namespace validation. Can you please review when you get a chance and suggest if any changes?

fedepaol:
> Hi @fedepaol , I've attempted a fix for namespace validation. Can you please review when you get a chance and suggest if any changes?

sure thing, and thanks for the contribution! :heart:

fedepaol:
Looks good! A couple of quick nits:

- Can you change the commit message according to https://chris.beams.io/posts/git-commit/
- Can you add an e2e test under https://github.com/metallb/metallb/blob/main/e2etest/webhookstests/webhooks.go ?
- I will ask you to add it to the release notes, but it will probably end up in a latter release so adding to 0.13.4 doesn't make sense now. Let's see if we are able to cut a release today

fedepaol:
> Sure @fedepaol. I'll try and resolve your changes by today.

no rush!

mgsh:
Just an update on status:
It took me some time to understand ginkgo setup and testing. I have created test cases and it's working as expected for all CRs except BFDProfiles which is not failing as expected when tested using the e2etests. I'm still debugging. Should be done by tomorrow.

fedepaol:
> Just an update on status: It took me some time to understand ginkgo setup and testing. I have created test cases and it's working as expected for all CRs except BFDProfiles which is not failing as expected when tested using the e2etests. I'm still debugging. Should be done by tomorrow.

Sure thing! Again, no rush :-)

mgsh:
@fedepaol, I've added e2tests and pushed another commit (BFDProfile test was failing earlier because its webhook config did NOT have a CREATE hook registered, I've added it now, hope that's fine).

As for clear commit message, I currently have 2 separate commits which I've pushed already. Would you recommend that I rebase/squash both commits into one single commit and force push in its current branch (I was reluctant to try it without checking first, because it's usually not advised)? OR perhaps a new PR with just the changes applied as a single commit with a clear commit message? Please advise.

As for the message content, I think your description on the issue was sufficiently verbose. Based on your advise on rebase/squash or squash/merge we can use the below message accordingly (with your changes if any).
```
Added namespace validation in the create webhook for all CRs

Currently, there's nothing preventing the CRs to be created in a
namespace different from the one metallb is deployed to. This change
adds namespace validation to prevent creation of CRs in a different
namespace.
```

e2etest/webhookstests/webhooks.go
```diff
@@ -381,3 +383,175 @@ var _ = ginkgo.Describe("Webhooks", func() {
...
+			k8s.DumpInfo(Reporter, ginkgo.CurrentGinkgoTestDescription().TestText)
+		}
+
+		// Clean previous configuration.
+		err := ConfigUpdaterOtherNS.Clean()
+		framework.ExpectNoError(err)
+	})
+
+	ginkgo.Context("For BFDProfile", func() {
+		ginkgo.It("Should reject creating BFDProfile in a different namespace", func() {
```
fedepaol, line 405: I think you could use a Table test here, providing resources as a table entry. This will avoid duplication, since we always expect the same behaviour.
See https://github.com/metallb/metallb/blob/65a54f3a97eb1a5d34a3671b187c3cd09fa65bae/e2etest/bgptests/bgp.go#L152 for example

e2etest/webhookstests/webhooks.go
```diff
@@ -381,3 +383,175 @@ var _ = ginkgo.Describe("Webhooks", func() {
 		})
 	})
 })
+
+var _ = ginkgo.Describe("Webhooks", func() {
+	ginkgo.BeforeEach(func() {
+		ginkgo.By("Clearing any previous configuration")
+		err := ConfigUpdaterOtherNS.Clean()
```
fedepaol, line 390: I'd expect there's no need for cleaning before or after, as all these tests won't add anything

e2etest/webhookstests/webhooks.go
```diff
@@ -381,3 +383,175 @@ var _ = ginkgo.Describe("Webhooks", func() {
 		})
 	})
 })
+
+var _ = ginkgo.Describe("Webhooks", func() {
```
fedepaol, line 387: I'd add it under the same describe, just as a different Context

config/crd/bases/metallb.io_addresspools.yaml
```diff
@@ -1,10 +1,9 @@
-
 ---
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.7.0
+    controller-gen.kubebuilder.io/version: v0.9.2
```
fedepaol, line 6: Can you use the same kubebuilder versoin? I know it's pretty old, but I'd be deliberate in changing the CRDs for that (I guess that's what is removing the status and adding `x-kubernetes-map-type: atomic` below)

fedepaol:
> @fedepaol, I've added e2tests and pushed another commit (BFDProfile test was failing earlier because its webhook config did NOT have a CREATE hook registered, I've added it now, hope that's fine).
> 
> As for clear commit message, I currently have 2 separate commits which I've pushed already. Would you recommend that I rebase/squash both commits into one single commit and force push in its current branch (I was reluctant to try it without checking first, because it's usually not advised)? OR perhaps a new PR with just the changes applied as a single commit with a clear commit message? Please advise.
> 
> As for the message content, I think your description on the issue was sufficiently verbose. Based on your advise on rebase/squash or squash/merge we can use the below message accordingly (with your changes if any).
> 
> ```
> Added namespace validation in the create webhook for all CRs
> 
> Currently, there's nothing preventing the CRs to be created in a
> namespace different from the one metallb is deployed to. This change
> adds namespace validation to prevent creation of CRs in a different
> namespace.
> ```

Commit message is fine. We normally split the changes with a commit for the changes and another one for the tests.
Rebasing is fine. 
I'd also add a section to the release notes (you can create v0.13.5) in a separate commit to mention this change.

e2etest/webhookstests/webhooks.go
mgsh: The latest version doesn't use setup code for namespace validation.

e2etest/webhookstests/webhooks.go
mgsh: The existing Describe has setup code which we don't need for the new specs. So I've retained it as a separate set of specs under a separate DescribeTable (since we don't need the setup code for it). Is that ok?

config/crd/bases/metallb.io_addresspools.yaml
mgsh: I've used the exact version of controller-gen and retested it in the latest changes.

mgsh:
Let me check the e2e failures and fix, certain scenarios are failing.

mgsh:
_My bad that I hadn't run all test possibilities (am exploring the use of `act` to run github actions locally before pushing changes, yet to get it fully working)._

The helm based e2etests were failing. I had to make 2 changes:
- Add `CREATE` op in the helm chart bfdprofile validation webhook config (and also fixed a possible typo) https://github.com/metallb/metallb/pull/1523/commits/163cf03ac1941ef844e3bd50faa8ea935019a756#diff-dbf01c8949b94f349b1a6e20f162b6cacb22d3348e400200384e29100b990669
- Updated the e2etest setup to use `kube-system` as the other namespace to expect failures in CR creation. I had used `default` earlier, but the helm based tests are using the `default` namespace to install metallb. Is this expected, shouldn't we be testing with `helm install --create-namespace --namespace metallb-system ...`?

mgsh:
The operator based e2etest required a change in the metallb-operator repo (to enable validation webhook for BFDProfile create op). Tracking that change under https://github.com/metallb/metallb-operator/pull/240 (from what I understand, it fetches configs from this repo and that's causing its manifest validation tests to fail, looking into it).

fedepaol:
>     * Updated the e2etest setup to use `kube-system` as the other namespace to expect failures in CR creation. I had used `default` earlier, but the helm based tests are using the `default` namespace to install metallb. Is this expected, shouldn't we be testing with `helm install --create-namespace --namespace metallb-system ...`?

Yep, this is something that we can definitely improve. Regarding the 'other namespace', why not creating one? Despite it's reasonable to expect `kube-system` , maybe it makes sense to be more in control with that (deleting the namespace when the suite exits)

mgsh:
I've made these two changes:
1. create a different namespace for the namespace validation tests.
2. use the fork version of metallb-operator (needs to be changed back to main once that fork is merged).

fedepaol:
>     2. use the fork version of metallb-operator (needs to be changed back to main once that fork is merged).

when we deal with this chicken-egg situations, it's fine to have the operator lane temporary failing. I'd go that way, to avoid the need of reverting again.

mgsh:
> when we deal with this chicken-egg situations, it's fine to have the operator lane temporary failing. I'd go that way, to avoid the need of reverting again.

Got it. In that case, let me remove that commit (https://github.com/metallb/metallb/pull/1523/commits/6c78ab3bbeb623855fc67c8dcb287036328b3923) and push it again. I've tested the above commit already for all tests passing in my forked repo (https://github.com/mgsh/metallb/pull/1), to cross-check that it was the only missing dependency. So once the metallb-operator changes are sorted out e2e-use-operator tests will be automatically fixed.

e2etest/e2etest_suite_test.go
```diff
@@ -134,11 +137,24 @@ var _ = ginkgo.BeforeSuite(func() {
 	updater, err = testsconfig.UpdaterForCRs(clientconfig, metallb.Namespace)
 	framework.ExpectNoError(err)
 
+	// for testing namespace validation, we need an existing namespace that's different from the
+	// metallb installation namespace
+	otherNamespace := fmt.Sprintf("%s-other", metallb.Namespace)
```
fedepaol, line 142: Can you delete this in aftersuite?
Also, ignore the err if it is AlreadyExists, so if we interrupt the tests before aftersuite the new run won't fail.

e2etest/webhookstests/webhooks.go
```diff
@@ -381,3 +384,102 @@ var _ = ginkgo.Describe("Webhooks", func() {
 		})
 	})
 })
+
+var _ = table.DescribeTable("Webhooks namespace validation",
+	func(resources *metallbconfig.ClusterResources) {
+		err := ConfigUpdaterOtherNS.Update(*resources)
+		framework.ExpectError(err)
```
fedepaol, line 391: I guess this is not necessary given the next row

fedepaol:
Couple of small nits, and I think this is good to land!

e2etest/webhookstests/webhooks.go
mgsh: do you mean that `framework.ExpectError` is redundant because of the following `Expect(err.Error()).To(...)`?

fedepaol: yes, if err is nil that row will complain

mgsh: @fedepaol ☝🏽 if that was your suggestion, won't `err.Error()` panic if it's `nil` and we remove the `framework.ExpectError` before it?

fedepaol: That's right. I misread the string checking line.

I suggest using
`Expect(err).To(MatchError(ContainSubstring(....))`

That will avoid adding the double check

fedepaol:
LGTM, thanks!

PR #1530, opened 2022-07-21
Bumprelease: change the kustomize reference to be more greedy

fedepaol:
The current replacement rules work only for main (and the frr one is
even bugged). Fixing to work also when moving from a minor version to
the next one.


oribon (approved):
lgtm

PR #1359, opened 2022-05-07
Docs: design for announcing IP from specified interfaces for layer2

qingwusunny:
This PR provide the design of new feature: announcing IP from specified interfaces for layer2.
Migrate from [this issue](https://github.com/metallb/metallb/issues/1358)

PR #1536, opened 2022-07-26
Support L2 mode specify interfaces for announcing LB IP

qingwusunny:
This PR implement the feature that MetalLB announce IP from specified interfaces for layer2. Related design is https://github.com/metallb/metallb/pull/1359.

I has add two basic e2etest about interface selector.


speaker/layer2_controller.go
```diff
@@ -133,9 +142,20 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce
 	return "notOwner"
 }
 
+func getInterfaces(localNode string, l2Advertisements []*config.L2Advertisement) []string {
+	ifs := sets.NewString()
```
fedepaol, line 146: This won't handle the case where you have one advertisement with no interface (and thus, collecting all of them) and one other advertisement with a selector.
The desired result is to advertise all of them, but we'll advertise to a subset only.

speaker/layer2_controller_test.go
```diff
@@ -2498,3 +2499,101 @@ func TestClusterPolicy(t *testing.T) {
 		t.Fatalf("All services assigned to speaker1")
 	}
 }
+
+func equalStringSlice(a []string, b []string) bool {
```
fedepaol, line 2503: put this on the bottom, as it's a util function

speaker/layer2_controller.go
```diff
@@ -133,9 +142,20 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce
...
+		}
+		ifs = ifs.Insert(l2.Interfaces...)
+	}
+	return ifs.List()
+}
+
 func (c *layer2Controller) SetBalancer(l log.Logger, name string, lbIPs []net.IP, pool *config.Pool) error {
 	for _, lbIP := range lbIPs {
-		c.announcer.SetBalancer(name, lbIP)
+		c.announcer.SetBalancer(name, lbIP, pool.Name)
```
fedepaol, line 158: Assuming you maintain the mapping between pool name and interface outside, here you can just pass the list of interfaces.

internal/layer2/announcer.go
```diff
@@ -140,33 +155,37 @@ func (a *Announce) updateInterfaces() {
 
 func (a *Announce) spamLoop() {
 	// Map IP to spam stop time.
-	m := map[string]time.Time{}
+	type spamInfo struct {
```
fedepaol, line 158: I'd call this `timedSpam` or `deferredSpam`.

fedepaol:
Overall looks good. I need to sleep a bit over the select loop and look at it carefully.
Also, I'd be more comfortable in having at least one or two basic e2e tests as part of this PR. This will give us confidence that we are merging something that works.

speaker/layer2_controller.go
qingwusunny: I has handle this case when announcer，if len(interfaces)==0，it won't filter interface.

speaker/layer2_controller.go
```diff
@@ -34,7 +35,15 @@ type layer2Controller struct {
 	sList     SpeakerList
 }
 
-func (c *layer2Controller) SetConfig(log.Logger, *config.Config) error {
+func (c *layer2Controller) SetConfig(l log.Logger, cfg *config.Config) error {
+	interfacesMap := make(map[string][]string)
```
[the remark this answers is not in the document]
qingwusunny: ok

speaker/layer2_controller_test.go
qingwusunny: ok

internal/layer2/announcer.go
```diff
@@ -12,8 +12,14 @@ import (
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
+type spam struct {
+	ip       net.IP
+	poolName string
```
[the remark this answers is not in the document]
qingwusunny: If the struct is (ip, interfaces), when L2Advertisement changed，I am afraid the announced LB IP can't update its related interfaces.
By the way, I am sorry that I found I hasn't modify func (a *Announce) shouldAnnounce(ip net.IP). And "poolName" is not unique identifier, also requires "namespace".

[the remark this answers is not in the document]
fedepaol: > If the struct is (ip, interfaces), when L2Advertisement changed，I am afraid the announced LB IP can't update its related interfaces. 

When a configuration changes, MetalLB will reprocess all the services, which in turn will call SetBalancer again, so this won't be an issue. 

See https://github.com/metallb/metallb/blob/9fce695d5b5b7ed1bc7e865f8122ffc7f5c5945d/speaker/main.go#L434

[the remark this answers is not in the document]
fedepaol: Regarding the namespace, MetalLB will filter objects not belonging to its namespace, so it's safe to assume namespace is not part of the key

speaker/layer2_controller.go
fedepaol: Yes, but the problem is, if you have two l2 advertisements, one with some interfaces and another one with no interfaces, the semantic there is : advertise to all (because you have one l2 advertisement with no interfaces) but your if won't trigger because len(interfaces) is not 0

fedepaol:
Couple of extra notes: I'd change the commit message (we are trying to adhere to https://cbea.ms/git-commit/ as guidelines)
For later, when we converge on the review:
I'd add also a commit changing the release notes (0.13.5 is not there yet, you can add it) to mention this feature.
I'd add a commit expanding the "advanced l2 configuration" section of the doc with this change

speaker/layer2_controller.go
qingwusunny: yes, I'm ill-considered.

internal/layer2/announcer.go
qingwusunny: ok

internal/layer2/announcer.go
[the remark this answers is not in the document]
qingwusunny: ok, I get it.

internal/layer2/announcer.go
```diff
@@ -219,9 +251,9 @@ func (a *Announce) shouldAnnounce(ip net.IP) dropReason {
 }
 
 // SetBalancer adds ip to the set of announced addresses.
-func (a *Announce) SetBalancer(name string, ip net.IP) {
+func (a *Announce) SetBalancer(name string, ip net.IP, poolName string) {
 	// Call doSpam at the end of the function without holding the lock
-	defer a.doSpam(ip)
+	defer a.doSpam(ip, poolName)
```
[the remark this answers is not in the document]
qingwusunny: It has been modified.

speaker/layer2_controller.go
qingwusunny: ok

internal/layer2/announcer.go
```diff
@@ -140,33 +147,37 @@ func (a *Announce) updateInterfaces() {
 
 func (a *Announce) spamLoop() {
 	// Map IP to spam stop time.
-	m := map[string]time.Time{}
+	type timedSpam struct {
+		t time.Time
```
fedepaol, line 151: I'd call this "until"

internal/layer2/announcer.go
```diff
@@ -176,14 +187,15 @@ func (a *Announce) spamLoop() {
...
 }
 
-func (a *Announce) doSpam(ip net.IP) {
-	a.spamCh <- ip
+func (a *Announce) doSpam(ip net.IP, interfaces sets.String) {
+	a.spamCh <- ipAdvertisement{ip, interfaces}
 }
 
-func (a *Announce) gratuitous(ip net.IP) {
+func (a *Announce) gratuitous(ipA ipAdvertisement) {
```
fedepaol, line 194: I'd call ipA -> adv

internal/layer2/announcer.go
```diff
@@ -192,51 +204,89 @@ func (a *Announce) gratuitous(ip net.IP) {
...
+				}
 				return dropReasonNone
 			}
 		}
 	}
 	return dropReasonAnnounceIP
 }
 
+// updateIPs update announce.ips depend on service info.
+func (a *Announce) updateIPs(name string, ip net.IP, interfaces sets.String) (hasUpdate bool) {
```
fedepaol, line 253: This is used by SetBalancer, please put it below it, reasonably after the exported functions

fedepaol, line 253: I'd avoid using named returns, just return the value

internal/layer2/announcer.go
```diff
@@ -192,51 +204,89 @@ func (a *Announce) gratuitous(ip net.IP) {
...
+	hasFound := false
+	if ips, ok := a.ips[name]; ok {
+		for i, _ := range ips {
+			if ip.Equal(ips[i].ip) {
+				if ips[i].interfaces.Equal(interfaces) {
+					hasUpdate = false
+					return
+				}
+				ips[i].interfaces = interfaces
+				hasFound = true
```
fedepaol, line 263: we can just return here I think? If so we can remove hasFound

internal/layer2/announcer.go
```diff
@@ -192,51 +204,89 @@ func (a *Announce) gratuitous(ip net.IP) {
...
 
 	// Kubernetes may inform us that we should advertise this address multiple
 	// times, so just no-op any subsequent requests.
-	if ips, ok := a.ips[name]; ok {
-		for i, ip := range ips {
-			if ip.Equal(a.ips[name][i]) {
-				continue
-			}
-		}
+	hasUpdate := a.updateIPs(name, ip, interfaces)
```
fedepaol, line 285: nit: hasUpdate -> updated

internal/layer2/announcer.go
```diff
@@ -192,51 +204,89 @@ func (a *Announce) gratuitous(ip net.IP) {
...
+					return dropReasonNotMatchInterface
+				}
 				return dropReasonNone
 			}
 		}
 	}
 	return dropReasonAnnounceIP
 }
 
+// updateIPs update announce.ips depend on service info.
```
fedepaol, line 252: "and returns true if the announcement has changed"

internal/layer2/announcer.go
fedepaol, line 253: nit: updateBalancerIPs

internal/layer2/announcer.go
```diff
@@ -192,51 +204,89 @@ func (a *Announce) gratuitous(ip net.IP) {
...
+			}
 			if err := client.Gratuitous(ip); err != nil {
 				level.Error(a.logger).Log("op", "gratuitousAnnounce", "error", err, "ip", ip, "msg", "failed to make gratuitous NDP announcement")
 			}
 		}
 	}
 }
 
-func (a *Announce) shouldAnnounce(ip net.IP) dropReason {
+func isMatchIntf(intf string, ifs sets.String) bool {
```
fedepaol, line 228: nit: `advertiseOnInterface`

speaker/layer2_controller.go
```diff
@@ -133,9 +134,24 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce
 	return "notOwner"
 }
 
+func getInterfaces(localNode string, l2Advertisements []*config.L2Advertisement) sets.String {
+	ifs := sets.NewString()
+	for _, l2 := range l2Advertisements {
+		if matchNode, ok := l2.Nodes[localNode]; !ok || !matchNode {
+			continue
+		}
+		if len(l2.Interfaces) == 0 {
```
fedepaol, line 143: Can you add a comment here explaining the semantic? I am also tempted to wrap the set under a 
struct {
   allInterfaces bool
   interfaces Set
}

WDYT? I am not sure I like propagating the len==0 -> all interfaces down there. A flag might be more explicit and understandable.
Ideally, that flag could be even a property of the "internal" l2advertisement, so the only place where we consider  len==0 -> all interfaces is the CRD.

speaker/layer2_controller.go
```diff
@@ -133,9 +134,24 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce
 	return "notOwner"
 }
 
+func getInterfaces(localNode string, l2Advertisements []*config.L2Advertisement) sets.String {
```
fedepaol, line 137: Same, put below where it's used. See https://kentcdodds.com/blog/newspaper-code-structure for reference

fedepaol:
Another round, mostly nits. It's converging :-)

internal/layer2/announcer.go
qingwusunny: done

internal/layer2/announcer.go
qingwusunny: done

internal/layer2/announcer.go
qingwusunny: This file's all export functions are in the last. Maybe the function's position can not change for consistency?

qingwusunny: done

internal/layer2/announcer.go
qingwusunny: done

internal/layer2/announcer.go
qingwusunny: Done

internal/layer2/announcer.go
qingwusunny: Done

speaker/layer2_controller.go
qingwusunny: ok

speaker/layer2_controller.go
qingwusunny: done

internal/layer2/announcer.go
qingwusunny: done

e2etest/l2tests/specify_interfaces.go
```diff
@@ -0,0 +1,132 @@
...
+						},
+					},
+				},
+				L2Advs: []metallbv1beta1.L2Advertisement{
+					{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: "with-interfaces",
+						},
+						Spec: metallbv1beta1.L2AdvertisementSpec{
+							Interfaces: []string{"eth0"},
```
fedepaol, line 69: not sure it makes sense, but eth0 and eth1 are totally arbitrary and related to the kind deployment tests normally run against. 
We always tried to have the tests runnable on multiple clusters, would it make sense to pick the default gateway for the nodes here (running ip inside the speaker containers), and passing a nonsensical interface in the next test?

e2etest/l2tests/specify_interfaces.go
```diff
@@ -0,0 +1,132 @@
...
+
+			svc, _ := service.CreateWithBackend(cs, f.Namespace.Name, "lb-service")
+			defer func() {
+				err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
+				framework.ExpectNoError(err)
+			}()
+
+			ginkgo.By("checking connectivity to its external VIP")
+
+			gomega.Eventually(func() error {
```
fedepaol, line 127: When we run negative tests, we use Consistently (with a lower timeout)
Also, you can use `Should(HaveOccurred())`

e2etest/l2tests/specify_interfaces.go
```diff
@@ -0,0 +1,132 @@
...
+				return service.ValidateL2(svc)
+			}, 2*time.Minute, 1*time.Second).Should(gomega.BeNil())
```
fedepaol, line 87: Should(Not(HaveOccurred()))

internal/config/config.go
```diff
@@ -639,9 +643,15 @@ func l2AdvertisementFromCR(crdAd metallbv1beta1.L2Advertisement, nodes []corev1.
 	if err != nil {
 		return nil, errors.Wrapf(err, "Failed to parse node selector for %s", crdAd.Name)
 	}
-	return &L2Advertisement{
+	l2 := &L2Advertisement{
 		Nodes: selected,
-	}, nil
+	}
+	if len(crdAd.Spec.Interfaces) == 0 {
+		l2.AllInterfaces = true
```
fedepaol, line 650: just return here and omit the else

internal/layer2/announcer.go
```diff
@@ -192,61 +197,89 @@ func (a *Announce) gratuitous(ip net.IP) {
...
+					return dropReasonNotMatchInterface
+				}
 				return dropReasonNone
 			}
 		}
 	}
 	return dropReasonAnnounceIP
 }
 
+// updateBalancerIPs update announce.ips depend on service info, and returns true if the announcement has changed.
```
fedepaol, line 237:
```suggestion
// updateBalancerIPs updates announce.ips depend on service info, and returns true if the announcement has changed.
```

internal/layer2/arp.go
```diff
@@ -97,7 +97,10 @@ func (a *arpResponder) processRequest() dropReason {
 	}
 
 	// Ignore ARP requests that the announcer tells us to ignore.
-	if reason := a.announce(pkt.TargetIP); reason != dropReasonNone {
+	if reason := a.announce(pkt.TargetIP, a.intf); reason != dropReasonNone {
```
fedepaol, line 100: let's avoid nested ifs:

```suggestion
	reason := a.announce(pkt.TargetIP, a.intf)
	if reason == dropReasonNotMatchInterface {
	log
	}
	if reason != dropReasonNone {
	   return reason
	}
```

speaker/layer2_controller.go
```diff
@@ -153,6 +155,20 @@ func (c *layer2Controller) SetNode(log.Logger, *v1.Node) error {
 	return nil
 }
 
+func assembleIPAdvertisement(ip net.IP, localNode string, l2Advertisements []*config.L2Advertisement) layer2.IPAdvertisement {
```
fedepaol, line 158: ipAdvertisementFor(

tasks.py
```diff
@@ -121,6 +122,16 @@ def _get_subnets_allocated_ips():
 
     return sorted(v4_ips), sorted(v6_ips)
 
+def _add_nic(cluster_name):
```
fedepaol, line 125: Is this required now?

fedepaol:
Just nits, I like this a lot!
One extra thought: if we are able to maintain a "allInterfacesForAllIPs" sort of map (maybe with a better name) we could even avoid creating the responders here, providing a better optimization.
That list is updated every 10 seconds.

fedepaol:
Also, I'd like to get an extra :eyes: from @oribon, to see if I missed something

fedepaol:
> Just nits, I like this a lot! One extra thought: if we are able to maintain a "allInterfacesForAllIPs" sort of map (maybe with a better name) we could even avoid creating the responders here, providing a better optimization. That list is updated every 10 seconds.

Just to be clear: please note that we can have this in without the optimization I wrote about ^^^^
The behaviour will be similar to what we have today, and I'd rather measure the benefit before even adding the extra logic.

qingwusunny:
@fedepaol I have modified the func SetBalancer in announcer.go，review again?

qingwusunny:
> Couple of extra notes: I'd change the commit message (we are trying to adhere to https://cbea.ms/git-commit/ as guidelines) For later, when we converge on the review: I'd add also a commit changing the release notes (0.13.5 is not there yet, you can add it) to mention this feature. I'd add a commit expanding the "advanced l2 configuration" section of the doc with this change

Add two commit for release notes and doc in this PR?

qingwusunny:
> @fedepaol I have modified the func SetBalancer in announcer.go，review again?

And @oribon

speaker/layer2_controller.go
```diff
@@ -153,6 +155,20 @@ func (c *layer2Controller) SetNode(log.Logger, *v1.Node) error {
 	return nil
 }
 
+func ipAdvertisementFor(ip net.IP, localNode string, l2Advertisements []*config.L2Advertisement) layer2.IPAdvertisement {
+	ifs := sets.NewString()
+	for _, l2 := range l2Advertisements {
+		if matchNode, ok := l2.Nodes[localNode]; !ok || !matchNode {
+			continue
```
fedepaol, line 162: I think we will never end up in that situation (because ShouldAnnounce would prevent it), but would it make sense to throw an error and log if there are no matching advertisements?

fedepaol:
> > Couple of extra notes: I'd change the commit message (we are trying to adhere to https://cbea.ms/git-commit/ as guidelines) For later, when we converge on the review: I'd add also a commit changing the release notes (0.13.5 is not there yet, you can add it) to mention this feature. I'd add a commit expanding the "advanced l2 configuration" section of the doc with this change
> 
> Add two commit for release notes and doc in this PR?

Yes exactly, so when we cut a new release we won't have to add a new pr for that.

fedepaol:
> @fedepaol I have modified the func SetBalancer in announcer.go，review again?

Sorry for the delay, I was off. Reviewed again!

internal/layer2/announcer.go
```diff
@@ -192,61 +197,97 @@ func (a *Announce) gratuitous(ip net.IP) {
...
-		}
-	}
-
-	a.ips[name] = append(a.ips[name], ip)
-
-	a.ipRefcnt[ip.String()]++
-	if a.ipRefcnt[ip.String()] > 1 {
-		// Multiple services are using this IP, so there's nothing
-		// else to do right now.
+	old, updated := a.updateBalancerIPs(name, adv)
```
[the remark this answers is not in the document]
fedepaol: I also filed https://github.com/metallb/metallb/pull/1570 (which is likely to land before this) to fix the issue and to ease backports.

internal/layer2/announcer.go
```diff
@@ -192,61 +197,97 @@ func (a *Announce) gratuitous(ip net.IP) {
...
+					return old, false
+				}
+				ips[i] = adv
+				return old, true
+			}
+		}
+	}
+
+	a.ips[name] = append(a.ips[name], adv)
+	return old, true
```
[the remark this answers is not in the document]
fedepaol: this function is gonna be removed I think.

fedepaol:
> > > Couple of extra notes: I'd change the commit message (we are trying to adhere to https://cbea.ms/git-commit/ as guidelines) For later, when we converge on the review: I'd add also a commit changing the release notes (0.13.5 is not there yet, you can add it) to mention this feature. I'd add a commit expanding the "advanced l2 configuration" section of the doc with this change
> > 
> > 
> > Add two commit for release notes and doc in this PR?
> 
> Yes exactly, so when we cut a new release we won't have to add a new pr for that.

As a second thought, let's leave them out for now. We'll add documentation and declare the feature when we have enough coverage to assure us the feature is working (I am assuming we'll add new tests in another PR)

e2etest/l2tests/interface_selector.go
```diff
@@ -0,0 +1,148 @@
...
+				err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
+				framework.ExpectNoError(err)
+			}()
+			ingressIP := e2eservice.GetIngressPoint(&svc.Status.LoadBalancer.Ingress[0])
+			out, err := executor.Host.Exec("ip", "neigh", "flush", "to", ingressIP)
+			framework.ExpectNoError(err, out)
+
+			ginkgo.By("checking connectivity to its external VIP")
+			gomega.Eventually(func() error {
+				err := mac.RequestAddressResolution(ingressIP, executor.Host)
```
fedepaol, line 95: I think @oribon 's comment was mis-interpreted here. What we should check is that the interface we expect to reply replies, by matching it's mac address using arping (we have a few utility options around).
So, find the node that is announcing the IP (and it's going to be one only), get the mac address of eth0 / eth1, do the arping and validate it matches with what we selected.

e2etest/l2tests/interface_selector.go
```diff
@@ -0,0 +1,148 @@
...
+				if err != nil {
+					return err
+				}
+				return service.ValidateL2(svc)
+			}, 1*time.Minute, 1*time.Second).Should(gomega.Not(gomega.HaveOccurred()))
+		})
+	})
+
+	ginkgo.Context("Specify interfaces which disconnect client", func() {
+		ginkgo.It("Shouldn't work", func() {
```
fedepaol, line 105: Something I didn't notice before. Why using the second interface shouldn't work?
Both interfaces should be accessible from the host via their respective bridges I think.

I think what is making it not working is the fact that the ipv4service range is related to the default bridge range. I did a quick test hardcoding the ip to the one of network one and it passed, so arping looks for the subnet in order to use the interface to request the arp from.

In short, we should validate when requesting arping via both bridges (using the -I option) work only from the selected ones.
IF we want to validate wget via the secondary network, we need to know its subnet and set the iprange of the addresspool accordingly.

Does it make sense? Also cc @oribon

fedepaol:
A couple of nits around the code, which is getting to look really good in my opinion, plus a comment about the e2e tests which I think needs to be sorted out.

internal/layer2/announcer.go
```diff
@@ -192,60 +197,76 @@ func (a *Announce) gratuitous(ip net.IP) {
...
 	a.RLock()
 	defer a.RUnlock()
-	for _, ips := range a.ips {
-		for _, i := range ips {
-			if i.Equal(ip) {
-				return dropReasonNone
+	matchIP := false
+	for _, ipAdvertisements := range a.ips {
+		for _, i := range ipAdvertisements {
+			if i.ip.Equal(ip) {
```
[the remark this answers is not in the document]
fedepaol: Can you check this comment?

speaker/layer2_controller.go
fedepaol: Can you check this comment?

internal/layer2/announcer.go
```diff
@@ -255,26 +276,26 @@ func (a *Announce) DeleteBalancer(name string) {
...
-		a.ipRefcnt[ip.String()]--
-		if a.ipRefcnt[ip.String()] > 0 {
+
+	for _, cur := range advs {
+		a.ipRefcnt[cur.ip.String()]--
+		if a.ipRefcnt[cur.ip.String()] > 0 {
 			// Another service is still using this IP, don't touch any
 			// more things.
-			return
+			continue
```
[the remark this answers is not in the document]
fedepaol: Can you check this comment?

oribon (commented):
apart from the few unaddressed comments, lgtm

internal/layer2/announcer.go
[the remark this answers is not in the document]
fedepaol: discussed on slack, it's fine as it is

e2etest/l2tests/interface_selector.go
qingwusunny: I has update the e2etest. One is about mac check, another is about arp check.

internal/layer2/announcer.go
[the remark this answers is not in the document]
fedepaol: So this is a bug we had also before, that you fixed. Can you add it in a separate commit so it will be easier to backport to 0.12x if required?

e2etest/l2tests/interface_selector.go
```diff
@@ -0,0 +1,138 @@
...
+	"go.universe.tf/metallb/e2etest/pkg/service"
+	internalconfig "go.universe.tf/metallb/internal/config"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/kubernetes/test/e2e/framework"
+	e2eservice "k8s.io/kubernetes/test/e2e/framework/service"
+	admissionapi "k8s.io/pod-security-admission/api"
+)
+
+var _ = ginkgo.Describe("L2-interface selector", func() {
```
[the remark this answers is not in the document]
qingwusunny: The e2etest "modify L2 interface" is check arp

e2etest/e2etest_suite_test.go
```diff
@@ -78,6 +78,8 @@ func handleFlags() {
 	flag.StringVar(&ipv6ForContainers, "ips-for-containers-v6", "0", "a comma separated list of IPv6 addresses available for containers")
 	flag.StringVar(&l2tests.IPV4ServiceRange, "ipv4-service-range", "0", "a range of IPv4 addresses for MetalLB to use when running in layer2 mode")
 	flag.StringVar(&l2tests.IPV6ServiceRange, "ipv6-service-range", "0", "a range of IPv6 addresses for MetalLB to use when running in layer2 mode")
+	flag.StringVar(&l2tests.DefaultInterface, "default-interface", "", "default gateway device use when running in interface selector")
+	flag.StringVar(&l2tests.ExtraInterface, "extra-interface", "", "extra interface use when running in interface selector")
```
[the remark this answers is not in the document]
qingwusunny: I has delete the param

fedepaol:
Thanks for taking the time to iterate over this! The result is really neat.
LGTM

fedepaol:
@qingwusunny it'd be great if you can extend the e2e tests in a separate PR as we discussed in the past.

PR #1542, opened 2022-07-28
retry when applying bgp config when running local dev env

### LINKED ISSUE #1535: Dev-env -p bgp / l2 fails the webhook validation
When running dev-env with a prefixed configuration (-p bgp for example) we don't wait for the controller to be in place, so the creation of the configuration fails because the backend of the webhook is not running yet.

karampok:
Fixed #1535


karampok:
The `-p l2` did not fail so I only added in the bgp setup.

PR #1545, opened 2022-07-29
Add an option to expose prometheus metrics securely via kube-rbac-proxy

fedepaol:
Here we adjust the manifests and the helm chart to optionally set a kube-rbac-proxy sidecar to interact with prometheus, in order to expose the metrics securely.

At the same time, we change CI to deploy the prometheus k8s stack and validate the metrics against the real prometheus instance (as opposed to validating that the pods are exposing the metrics). This will give us more confidence in the fact that the integration with prometheus is really working.


e2etest/pkg/metrics/metrics.go
```diff
@@ -11,26 +11,39 @@ import (
...
 		for _, p := range c.Ports {
-			if p.Name == "monitoring" {
+			if p.Name == "monitoring" ||
+				p.Name == "frr-monitoring" {
 				ports = append(ports, int(p.ContainerPort))
 			}
 		}
 	}
 
-	podExecutor := executor.ForPod(namespace, source.Name, source.Spec.Containers[0].Name)
```
oribon, line 33: this means the source pod has to be the controller pod? shouldn't we take the container's name as a parameter instead?

e2etest/pkg/metrics/metrics.go
```diff
@@ -47,6 +60,13 @@ func ForPod(source, target *corev1.Pod, namespace string) ([]map[string]*dto.Met
 	return allMetrics, nil
 }
 
+type CheckType int
+
+const (
+	NotThere CheckType = iota
+	There
+)
```
oribon, line 68: could there be other values here in the future? if not, wouldn't it be a bit clearer just using a boolean?

e2etest/pkg/metrics/prometheus.go
```diff
@@ -0,0 +1,61 @@
...
+	"net/url"
+
+	"go.universe.tf/metallb/e2etest/pkg/executor"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+// ValidateOnPrometheus checks the existance of the given metric directly on prometheus pods.
+func ValidateOnPrometheus(exec executor.Executor, query string, expected CheckType) error {
+	fmt.Println("Checking ", query, expected)
```
oribon, line 18: does an `fmt.Print..` belong in these pkgs?

e2etest/pkg/metrics/prometheus.go
```diff
@@ -0,0 +1,61 @@
...
+	url := fmt.Sprintf("localhost:9090/api/v1/query?%s", (url.Values{"query": []string{query}}).Encode())
+	stdout, err := exec.Exec("wget", "-qO-", url)
```
oribon, line 20: since this heavily relies on the exec being a prometheus pod, would it make sense having a harder restriction rather than accepting an exec.Executor?

oribon, line 20: we can use wget.Do here

e2etest/pkg/metrics/prometheus.go
```diff
@@ -0,0 +1,61 @@
...
+	if err := json.Unmarshal([]byte(stdout), &result); err != nil {
+		return fmt.Errorf("unable to parse query response for %s: %v", query, err)
+	}
+	metrics := result.Data.Result
+	if result.Status != "success" {
+		data, _ := json.MarshalIndent(metrics, "", "  ")
+		return fmt.Errorf("promQL query: %s had reported incorrect status:\n%s", query, data)
+	}
+
+	if expected == NotThere && len(metrics) > 0 {
```
oribon, line 35: is it possible to extend the other metrics validation functions to use a similar semantic?
the current negative tests expect an error for the other validation functions but noerror for this one (when you're setting metrics.NotThere), I think it would be nice if they all behave the same

e2etest/l2tests/l2.go
```diff
@@ -521,11 +564,16 @@ var _ = ginkgo.Describe("L2", func() {
...
+				if err != nil {
+					return err
+				}
 
 				return nil
-			}, 1*time.Minute, 1*time.Second).Should(gomega.BeNil())
+			}, time.Minute, 1*time.Second).Should(gomega.BeNil())
 		},
-			table.Entry("IPV4 - Checking service", "ipv4"),
+			table.FEntry("IPV4 - Checking service", "ipv4"),
```
oribon, line 576: FEntry

config/frr/speaker-patch.yaml
```diff
@@ -95,7 +95,7 @@ spec:
             - --metrics-port=7473
           ports:
             - containerPort: 7473
-              name: monitoring
```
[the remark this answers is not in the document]
fedepaol: Ah, TIL. Will double check. That will conflict with the fact that we must have different named ports for the frr container.

config/manifests/metallb-frr.yaml
```diff
@@ -1665,9 +1665,6 @@ spec:
       component: controller
   template:
     metadata:
-      annotations:
-        prometheus.io/port: "7472"
-        prometheus.io/scrape: "true"
```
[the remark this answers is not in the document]
fedepaol: as above, I wasn't aware of this way of configure prometheus

config/manifests/metallb-frr-prometheus.yaml
```diff
@@ -0,0 +1,2313 @@
...
+    metadata:
+      annotations:
+        prometheus.io/scrape: "true"
+      labels:
+        app: metallb
+        component: speaker
+    spec:
+      containers:
+      - args:
+        - --secure-listen-address=0.0.0.0:8443
```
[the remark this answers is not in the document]
fedepaol: Indeed

e2etest/pkg/metrics/metrics.go
fedepaol: I'd just rename source -> controller, because the reality is, we always use the controller.
It's better than bloating the code with a parameter that is always the same

e2etest/pkg/metrics/metrics.go
fedepaol: makes sense

e2etest/pkg/metrics/prometheus.go
fedepaol: nope, it's a debug leftover

e2etest/pkg/metrics/prometheus.go
fedepaol: I'll change to accept a pod and will craft the executor out of the pod

e2etest/pkg/metrics/prometheus.go
fedepaol: Tried to give a look at them, it would basically mean adding a "thisWouldReturnAnError" parameter, not sure it'd make the codebase better or worse.

e2etest/l2tests/l2.go
fedepaol: Will remove

e2etest/pkg/metrics/prometheus.go
fedepaol: we can't, unless we extend it to produce the output too

oribon (approved):
LGTM

PR #1546, opened 2022-08-03
Helm: setting the command for the controller and the speaker

fedepaol:
Another change required to align the operator deployment:

- setting the command for the two pods (in case the image used doesn't have a default entry point)
- setting the webhook mode (in the operator the webhook is deployed together with the operator, and in the controller the webhook is disabled)


pperiyasamy:
LGTM

PR #1548, opened 2022-08-04
configmaptocrs: allow it to run directly against a cluster

### LINKED ISSUE #1527: configmaptocrs: allow it to run in k8s and work only with the data part of the configMap
The described way to run the configmaptocrs image isn't ideal, as it requires to have Docker installed and the exported configMap available locally.

I thought about an approach like
```
kubectl -n metallb-system run configmaptocrs --restart=Never -it --rm --image overriden --overrides '
{
  "spec": {
    "containers": [
      {
        "name": "configmaptocrs",
        "image": "quay.io/metallb/configmaptocrs",
        "command": [
          "/configmaptocrs",
          "-source",
          "config"
        ],
        "volumeMounts": [
          {
            "name": "config",
            "mountPath": "/var/input"
          }
        ]
      }
    ],
    "volumes": [
      {
        "name": "config",
        "configMap": {
          "name": "config"
        }
      }
    ]
  }
}'
```
as that would make it much easier. But that brings a few issues:
- the tool forces the output file to be in /var/input as well (which would be R/O, if the configMap is mounted there)
- the tool expects the entire configMap defintion (which wouldn't be present, as only the configMap data is presented)

Suggestion:
- Make it possible to pass an output file, or fallback to /dev/stdout, if the resources.yaml can't be created due to R/O filesystem
- Make the entire configMap spec optional and allow the tool to work just with the exported configFile itself

liornoy:
This PR allows users to run configmaptocrs while mounting
an existing ConfigMap.

Added `-only-data` flag, set to true if the input file is only the configMap data
And: `-stdout` - set to true to write crs to stdout

README was updated accordingly.

Added Unit tests to cover the new feature.

fix https://github.com/metallb/metallb/issues/1527


configmaptocrs/README.md
```diff
@@ -92,3 +92,43 @@ status: {}
 ---
 
 ```
+
+## Run with ConfigMap Volume
```
fedepaol, line 96: I'd rename this "Running directly against a cluster" and remove the mention of the issue (or add to the bottom).

configmaptocrs/README.md
```diff
@@ -92,3 +92,43 @@ status: {}
...
+        "name": "config",
+        "configMap": {
+          "name": "config"
+        }
+      }
+    ]
+  }
+}'
+```
+This will mount the MetalLB's config into the container's /var/input path, and print the result into stdout.
```
fedepaol, line 132: I'd either document the full behaviour (from a user perspective) and / or the single flags meaning.
The full behaviour is "this will print the CRs corresponding to the configmap deployed on the cluster to the standard output".

configmaptocrs/main.go
```diff
@@ -34,6 +34,7 @@ const (
 var (
 	resourcesNameSpace = "metallb-system"
 	inputDirPath       = "/var/input"
+	dataOnly           = flag.Bool("data-only", false, "set this to true if the source file includes only the configmap's data")
```
fedepaol, line 37: "config as yaml" -> set this to true if the config is provided directly as yaml and not as the content of a configmap

configmaptocrs/main.go
```diff
@@ -44,7 +45,8 @@ func main() {
 
 	f, err := os.Create(filepath.Join(inputDirPath, outPutFileName))
 	if err != nil {
-		log.Fatalf("failed to create output file: %s", err)
+		log.Println("failed to create output file, writing to stdout")
```
fedepaol, line 48: I'd be explicit about our intentions, and provide a parameter to drive the behaviour, otherwise this will hide errors (i.e. the user thinks he set a file)

configmaptocrs/conversion_test.go
```diff
@@ -17,9 +17,21 @@ import (
...
 
+const (
+	testConfigMapSource = false
+	testDataOnlySource  = true
+	configMapTestDir    = "./testdata/configmap"
+	dataOnlyTestDir     = "./testdata/data-only"
+)
+
 func TestGenerateResources(t *testing.T) {
+	testGenerate(t, configMapTestDir, testConfigMapSource)
```
fedepaol, line 28: Can these be two different tests? It will be easier to understand what failed, to focus, etc.

fedepaol:
Just a few nits. Thanks!

configmaptocrs/README.md
liornoy: done

configmaptocrs/README.md
liornoy: done

configmaptocrs/main.go
liornoy: done

configmaptocrs/conversion_test.go
liornoy: done

configmaptocrs/main.go
liornoy: done. 
--output flag added

liornoy:
@fernferret let me know if you have any comments regarding the update to the README.

fernferret:
@liornoy this looks great. After seeing your PR I'm actually excited to try out the in-cluster mode! I already migrated all of my lab clusters but I'll have to spin up a test one once this is merged.

I think being able to point it to the cluster (for me) makes a lot of sense. This would open up the ability to let a Helm `hook` perform the upgrade in the chart. That might make sense, might not.

configmaptocrs/main.go
```diff
@@ -28,12 +28,13 @@ import (
...
 	separator      = "---\n"
 	autoGenComment = "# This was autogenerated by MetalLB's custom resource generator.\n"
-	outPutFileName = "resources.yaml"
 )
 
 var (
 	resourcesNameSpace = "metallb-system"
 	inputDirPath       = "/var/input"
+	configAsYAML       = flag.Bool("config-as-yaml", false, " set this to true if the config is provided directly as yaml and not as the content of a configmap")
+	outPutFilePath     = flag.String("output", "/var/input/resources.yaml", "set this to the desired output file path")
```
fedepaol, line 37: nit: outputFilePath

configmaptocrs/main.go
```diff
@@ -42,57 +43,59 @@ func main() {
 	log.Printf("MetalLB generator starting. commit: %s branch: %s goversion: %s",
 		version.CommitHash(), version.Branch(), version.GoString())
 
-	f, err := os.Create(filepath.Join(inputDirPath, outPutFileName))
+	res, err := generate(*source)
```
fedepaol, line 46: I think I remember we said to add a parameter to output the crds to the std output. If that's not the case anymore, the previous version was enough I think, and we only need to add the new mode of parsing of the yaml.

Likewise, if we want to bring the parameter back, passing `os.Stdout` as the writer will make the trick (I think)

configmaptocrs/main.go
```diff
@@ -42,57 +43,59 @@ func main() {
...
+	log.Println("Generator finished successfully!")
+
+	f, err := os.OpenFile(*outPutFilePath, os.O_CREATE|os.O_WRONLY, 0666)
 	if err != nil {
 		log.Fatalf("failed to create output file: %s", err)
 	}
 	defer f.Close()
 
-	err = generate(f, *source)
+	_, err = f.Write(res)
```
fedepaol, line 58: Again, if we stick to always writing to a file, let's revert the change and leave the writer there.

configmaptocrs/README.md
```diff
@@ -92,3 +93,52 @@ status: {}
...
+kubectl run configmaptocrs -n metallb-system --restart=Never -it --rm --image overriden --overrides '
+{
+  "spec": {
+    "containers": [
+      {
+        "name": "configmaptocrs",
+        "image": "quay.io/metallb/configmaptocrs",
+        "command": [
+          "/configmaptocrs",
+          "-source",
```
fedepaol, line 112: should we add config-as-yaml too here?

liornoy: This naming is confusing and not distinguishing IMO. both approaches use YAML,
we have the ConfigMap YAML or the ConfigMap's data YAML.

configmaptocrs/main.go
liornoy: done

configmaptocrs/main.go
liornoy: done

configmaptocrs/main.go
liornoy: done

liornoy:
@fedepaol following your comments, I revert to writing to a file, with a --stdout flag,
and renamed --config-as-yaml to only-data, as I think it's clearer. but of course, we 
can think of some other name.

configmaptocrs/main.go
```diff
@@ -28,31 +28,38 @@ import (
...
 	autoGenComment = "# This was autogenerated by MetalLB's custom resource generator.\n"
-	outPutFileName = "resources.yaml"
+	outputFileName = "resources.yaml"
 )
 
 var (
 	resourcesNameSpace = "metallb-system"
 	inputDirPath       = "/var/input"
+	source             = flag.String("source", "./config.yaml", "name of the configmap file to convert")
+	onlyData           = flag.Bool("only-data", false, "set this to true if the input file is only the configMap data")
```
oribon, line 38: nit:
```suggestion
	onlyData           = flag.Bool("only-data", false, "set this to true if the input file contains only the ConfigMap's data field")
```

configmaptocrs/main.go
```diff
@@ -76,7 +83,7 @@ func generate(w io.Writer, origin string) error {
 		return err
 	}
 
-	log.Println("Validating the custom resources")
+	log.Println("Checking the resources are parsed correctly")
```
oribon, line 86: does this / other log.Println show up in stdout when using -stdout?

oribon (commented):
small nits, overall looks good

configmaptocrs/main.go
liornoy: Thanks for the review! Yes, it shows. And I made sure the logs don't mix up with the result YAML.
Plus I'll note that I made this change to the log text because the
following line is: _, err = config.For(resources, config.**DontValidate**)
and the new log text will prevent confusion.
Here is an example of the output on stdout:
```
2022/08/30 06:44:51 Reading configmap
2022/08/30 06:44:51 Decoding configmap
2022/08/30 06:44:51 Creating custom resources
2022/08/30 06:44:51 Checking the resources are parsed correctly
2022/08/30 06:44:51 Creating the output YAML
# This was autogenerated by MetalLB's custom resource generator.
kind: IPAddressPool
metadata:
  creationTimestamp: null
  name: default
  namespace: metallb-system
spec:
  addresses:
  - 192.168.1.240-192.168.1.250
status: {}
---
kind: L2Advertisement
metadata:
  creationTimestamp: null
  name: l2advertisement1
  namespace: metallb-system
spec:
  ipAddressPools:
  - default
status: {}
---
pod "configmaptocrs" deleted
```

configmaptocrs/main.go
liornoy: done thanks!

fedepaol:
LGTM
Waiting for CI to merge

PR #1551, opened 2022-08-09
Support optional creation of rbac resources for prometheus

cyclinder:
In general, Prometheus is able to watch the various resources needed such as serviceMonitor, So we don't need to explicitly create prometheus rbac resources in the matallb. Optional, You can also pass "--set prometheus.rbacPrometheus.enabled=true --set prometheus.rbacPrometheus.serviceAccount= --set prometheus.rbacPrometheus.namespace=<prometheus-operator-namespace" to create rbac resources for prometheus.

Signed-off-by:  cyclinder <qifeng.guo@daocloud.io>

### DIFF
--- a/charts/metallb/templates/podmonitor.yaml
+++ b/charts/metallb/templates/podmonitor.yaml
@@ -75,6 +75,7 @@ spec:
 {{- toYaml .Values.prometheus.podMonitor.relabelings | nindent 4 }}
 {{- end }}
 ---
+{{- if .Values.prometheus.rbacPrometheus }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
@@ -101,4 +102,5 @@ subjects:
   - kind: ServiceAccount
     name: {{ required ".Values.prometheus.serviceAccount must be defined when .Values.prometheus.podMonitor.enabled == true" .Values.prometheus.serviceAccount }}
     namespace: {{ required ".Values.prometheus.namespace must be defined when .Values.prometheus.podMonitor.enabled == true" .Values.prometheus.namespace }}
+{{- end }}
 {{- end }}
--- a/charts/metallb/values.yaml
+++ b/charts/metallb/values.yaml
@@ -42,12 +42,15 @@ prometheus:
   # certificate to be used.
   controllerMetricsTLSSecret: ""
 
+  # prometheus doens't have the permission to scrape all namespaces so we give it permission to scrape metallb's one
+  rbacPrometheus: true
+
   # the service account used by prometheus
-  # required when .Values.prometheus.podMonitor.enabled == true
+  # required when " .Values.prometheus.rbacPrometheus == true " and " .Values.prometheus.podMonitor.enabled=true or prometheus.serviceMonitor.enabled=true "
   serviceAccount: ""
 
   # the namespace where prometheus is deployed
-  # required when .Values.prometheus.podMonitor.enabled == true
+  # required when " .Values.prometheus.rbacPrometheus == true " and " .Values.prometheus.podMonitor.enabled=true or prometheus.serviceMonitor.enabled=true "
   namespace: ""
 
   # the image to be used for the kuberbacproxy container
--- a/charts/metallb/README.md
+++ b/charts/metallb/README.md
@@ -85,6 +85,7 @@ A network load-balancer implementation for Kubernetes using standard routing pro
 | prometheus.prometheusRule.extraAlerts | list | `[]` |  |
 | prometheus.prometheusRule.staleConfig.enabled | bool | `true` |  |
 | prometheus.prometheusRule.staleConfig.labels.severity | string | `"warning"` |  |
+| prometheus.rbacPrometheus | bool | `true` |  |
 | prometheus.rbacProxy.repository | string | `"gcr.io/kubebuilder/kube-rbac-proxy"` |  |
 | prometheus.rbacProxy.tag | string | `"v0.12.0"` |  |
 | prometheus.scrapeAnnotations | bool | `false` |  |
--- a/charts/metallb/templates/servicemonitor.yaml
+++ b/charts/metallb/templates/servicemonitor.yaml
@@ -155,6 +155,7 @@ spec:
   sessionAffinity: None
   type: ClusterIP
 ---
+{{- if .Values.prometheus.rbacPrometheus }}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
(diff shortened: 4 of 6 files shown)


fedepaol:
@cyclinder couple of comments:
- I am not sure the rbacs are optional (with helm, I had to add this https://github.com/metallb/metallb/commit/8b28ef52ec3383cfc6d844ac94a5565a261faa0d when we started deploying metallb to a namespace different than default)
- can you check the ci failures / rebase?
- can you change the commit body to explain WHY we don't need those rbacs by default?

cyclinder:
Thanks @fedepaol , Please review the latest changes for this PR. I think we should create rbac resources for it in the prometheus manifest. 

I think we should create rbac resources for it in the prometheus manifest. In the [officially](https://github.com/prometheus-operator/kube-prometheus/blob/main/manifests/prometheusOperator-clusterRole.yaml), it has already  rbac permissions, so I don't even think we need to create rbac permissions for prometheus in the metallb.

cyclinder:
I found a CI Bug: https://github.com/metallb/metallb/runs/8206972097?check_suite_focus=true,  and I raise a [PR](https://github.com/metallb/metallb/pull/1594) for it. The PR should be merged firstly.

And refer to [officialy docs](https://prometheus-operator.dev/docs/operator/rbac/#prometheus-rbac), The Prometheus server itself accesses the Kubernetes API to discover targets and Alertmanagers. So we should just add the `list`, `watch`, and `get` permissions for the desired resources in [there](https://github.com/metallb/metallb/blob/8377501adec634ecfa7a215560daa2c81c31c025/dev-env/kube-prometheus/manifests/prometheus-clusterRole.yaml#L12), without even creating special rbac permissions for it in the metallb.

WDYT? @fedepaol

fedepaol:
> I found a CI Bug: https://github.com/metallb/metallb/runs/8206972097?check_suite_focus=true, and I raise a [PR](https://github.com/metallb/metallb/pull/1594) for it. The PR should be merged firstly.
> 
> And refer to [officialy docs](https://prometheus-operator.dev/docs/operator/rbac/#prometheus-rbac), The Prometheus server itself accesses the Kubernetes API to discover targets and Alertmanagers. So we should just add the `list`, `watch`, and `get` permissions for the desired resources in [there](https://github.com/metallb/metallb/blob/8377501adec634ecfa7a215560daa2c81c31c025/dev-env/kube-prometheus/manifests/prometheus-clusterRole.yaml#L12), without even creating special rbac permissions for it in the metallb.
> 
> WDYT? @fedepaol

I kind of agree, with the difference that I'd try to change our kube-prometheus configuration to consume the file that was added with https://github.com/prometheus-operator/kube-prometheus/pull/380 , which should patch that manifest (instead of patching it manually).
See also https://github.com/prometheus-operator/kube-prometheus/issues/350 for reference.

charts/metallb/values.yaml
```diff
@@ -42,13 +42,16 @@ prometheus:
   # certificate to be used.
   controllerMetricsTLSSecret: ""
 
-  # the service account used by prometheus
-  # required when .Values.prometheus.podMonitor.enabled == true
-  serviceAccount: ""
+  # Create rbac resources for prometheus
+  rbacPrometheus:
+    enabled: false
```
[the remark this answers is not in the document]
fedepaol: Also, document why we would require it, which is "prometheus doens't have the permission to scrape all namespaces so we give it permission to scrape metallb's one"

charts/metallb/values.yaml
```diff
@@ -42,13 +42,16 @@ prometheus:
...
+    # the service account used by prometheus
+    # required when .Values.prometheus.rbacPrometheus.enabled == true
+    serviceAccount: ""
```
fedepaol, line 50: moving this inside is a breaking change for users who might already have values

charts/metallb/templates/rbac.yaml
```diff
@@ -188,4 +188,35 @@ roleRef:
 subjects:
 - kind: ServiceAccount
   name: {{ include "metallb.controller.serviceAccountName" . }}
-{{- end -}}
+{{- end }}
+---
+{{- if .Values.prometheus.rbacPrometheus.enabled }}
```
fedepaol, line 193: if we move this, it must be conditioned also by having either podmonitor OR servicemonitor enabled

charts/metallb/values.yaml
cyclinder: Agree, I made these changes for it. I think it keep backward compatibility.

charts/metallb/templates/rbac.yaml
cyclinder: Done

charts/metallb/values.yaml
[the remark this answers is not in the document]
cyclinder: Agree, Done.

cyclinder:
Hi @fedepaol Could you please take a look?  I think the current changes have addressed your concerns.

charts/metallb/templates/podmonitor.yaml
```diff
@@ -102,3 +105,4 @@ subjects:
     name: {{ required ".Values.prometheus.serviceAccount must be defined when .Values.prometheus.podMonitor.enabled == true" .Values.prometheus.serviceAccount }}
     namespace: {{ required ".Values.prometheus.namespace must be defined when .Values.prometheus.podMonitor.enabled == true" .Values.prometheus.namespace }}
 {{- end }}
+{{- end }}
```
fedepaol, line 108: nit: newline

charts/metallb/templates/podmonitor.yaml
```diff
@@ -84,6 +85,8 @@ rules:
       - ""
     resources:
       - pods
+      - services
```
fedepaol, line 88: I don't think we do need services / endpoints in case of podmonitor, pods are enough

charts/metallb/templates/podmonitor.yaml
```diff
@@ -75,6 +75,7 @@ spec:
 {{- toYaml .Values.prometheus.podMonitor.relabelings | nindent 4 }}
 {{- end }}
 ---
+{{- if and .Values.prometheus.rbacPrometheus .Values.prometheus.podMonitor.enabled }}
```
fedepaol, line 78: the whole file is under `.Values.prometheus.podMonitor.enabled`, so I think this one is redundant

fedepaol, line 78: Don't we need something similar under servicemonitor too?

fedepaol:
> Hi @fedepaol Could you please take a look? I think the current changes have addressed your concerns.

left a few comments, sorry for the delay!

charts/metallb/templates/podmonitor.yaml
cyclinder: 👍 thanks, I didn't notice that..

cyclinder: yes!

cyclinder:
> 
> left a few comments, sorry for the delay!

No matter！

charts/metallb/templates/podmonitor.yaml
```diff
@@ -84,6 +85,7 @@ rules:
       - ""
     resources:
       - pods
+      - endpoints
```
fedepaol, line 88: I don't think we need to read endpoints if we are using podmonitors

cyclinder: Thanks!

fedepaol:
LGTM, thanks!

PR #1564, opened 2022-08-24
fix invalid Go dependencies

xrstf:
Our product imports `go.universe.tf/metallb v0.13.3` via our `go.mod` file. Since doing so, we cannot list our dependencies anymore in our project.

```bash
$ go list -u -m -json all
go: k8s.io/kubernetes@v1.21.23: invalid version: unknown revision v1.21.23
go: k8s.io/pod-security-admission@v0.0.0: invalid version: unknown revision v0.0.0
```

It seems those dependencies come from metallb, which never noticed due to the `replace` directives. The issue occured both with Go 1.18 and 1.19.

This PR adjusts to `go.mod` to reflect the actually used version, hopefully fixing the issue for downstream projects. It's not an urgent fix, as downstream projects can easily include the two additional replace() directives in their own `go.mod` files to work around this.


xrstf:
@fedepaol PTAL :)

xrstf:
We've only used it in a single e2e test, but have since moved to unstructured.Unstructured to get rid of the dependency (which incidentally also solved all dependency issues we faced): https://github.com/kubermatic/kubermatic/commit/bef31e35c5ac529e317f25a5635202901a26ecbb#diff-dd50ff5e07afd76292b6c7496e75c685ef2e566b157c42b14ca63cebedb8f2b6L334

fedepaol:
> We've only used it in a single e2e test, but have since moved to unstructured.Unstructured to get rid of the dependency (which incidentally also solved all dependency issues we faced): [kubermatic/kubermatic@bef31e3#diff-dd50ff5e07afd76292b6c7496e75c685ef2e566b157c42b14ca63cebedb8f2b6L334](https://github.com/kubermatic/kubermatic/commit/bef31e35c5ac529e317f25a5635202901a26ecbb#diff-dd50ff5e07afd76292b6c7496e75c685ef2e566b157c42b14ca63cebedb8f2b6L334)

Ah, I see. I was bitten by this multiple times :-)

PR #1569, opened 2022-08-29
Remove the pod security policy creation

fedepaol:
Pod security policy is now not supported anymore, so we remove it.
Note: it's currently not possible to bump the CI up to validate against 1.25 as a 1.25 is not available on kind yet.
I tested it running minikube and checking if the pods get up properly.


oribon (approved):
lgtm

PR #1570, opened 2022-08-30
L2 SetBalancer: return instead of continue if we find the ip

fedepaol:
In transitioning to dual stack we introduced an issue: if the ip was found, instead of returning the new code continued, grewing the a.ips[name] list on every service update.


oribon (approved):
lgtm

Issue #1578, opened 2022-09-02
Detection of availability of EndpointSlices may fail on k8s 1.25


ssams:
I'm running MetalLB on OpenShift / OKD and since the update to the latest version, which is based on k8s 1.24, I'm receiving warnings from OpenShift that an API which is deprecated in the next release (i.e. 1.25) is in use.

MetalLB apparently accesses the API _endpointslices.v1beta1.discovery.k8s.io_, as far as I can see this is used at least in the detection to check whether EndpointSlices are available in https://github.com/metallb/metallb/blob/ea5396dc5572b89e910edfde0f35781b5829dfa4/internal/k8s/k8s.go#L489-L492

Kubernetes 1.25 stops serving the v1beta1 version (https://kubernetes.io/docs/reference/using-api/deprecation-guide/#v1-25). If I see this correctly, this check may be problematic on 1.25 and could fail due to the beta removal, even though the v1 API would be available?

### Versions
OpenShift / OKD: 4.11.0-0.okd-2022-08-20-022919
k8s: v1.24.0+4f0dd4d
MetalLB: v0.13.5 (installed via official Helm chart)

fedepaol:
Thanks for raising the issue! Yes, we probably need to update. It won't make metallb stop working because it will fallback on endpoints, but it's better to update it.

PR #1579, opened 2022-09-02
Epslices detection: use discoveryv1 instead of v1beta1

fedepaol:
v1beta1 is not being served anymore in 1.25, which will make metallb
unable to detect the availability of epslices.
Fixes https://github.com/metallb/metallb/issues/1578


oribon (approved):
lgtm

Issue #1581, opened 2022-09-02
Endless loop and crash


fedepaol:
Opening an upstream issue to track slack discussion.

Hello! I'm having an issue w/ MetalLB where a speaker pod on one of my nodes appears to be leaking memory until the node crashes (~1 GiB per hour).
While running, it seems to be working as intended. I'm able to access the services on the IP assigned by MetalLB.
I'm running  v0.13.4 of MetalLB using the operator on Kubernetes v1.22.7 (Rancher K3s). I've got MetalLB configured to perform BGP routing peering w/ my OPNsense router.
I'm not really sure how to start debugging this, so if anyone has any tips that'd be great. In the reply to this message I'll include the logs for the affected speaker pod, though I don't see much there.

Speaker logs:
https://gist.github.com/glmdev/31d67284a2d7f7f06c0d3a506477b842
https://gist.github.com/glmdev/aa7da8b2cabd1bdf2c8c7acf1926cdd3

fedepaol:
The `nginx-ingress-ingress-nginx-controller` is bouncing between `172.21.0.11` and `172.20.80.1`.

```
{"caller":"service_controller.go:83","controller":"ServiceReconciler","level":"debug","processing service":"{\"kind\":\"Service\",\"apiVersion\":\"v1\",\"metadata\":{\"name\":\"nginx-ingress-ingress-nginx-controller\",\"namespace\":              ster\"},\"status\":{\"loadBalancer\":{\"ingress\":[{\"ip\":\"172.20.80.1\"}]}}}","ts":"2022-08-24T15:48:01Z"}
{"caller":"service_controller.go:83","controller":"ServiceReconciler","level":"debug","processing service":"{\"kind\":\"Service\",\"apiVersion\":\"v1\",\"metadata\":{\"name\":\"nginx-ingress-ingress-nginx-controller\",\"namespace\":              \"status\":{\"loadBalancer\":{\"ingress\":[{\"ip\":\"172.21.0.11\"}]}}}","ts":"2022-08-24T15:48:01Z"}
[...]

PR #1583, opened 2022-09-02
Slower debounce time in FRR reloading

fedepaol:
In corner cases where we have a bouncing service, the number of updates requested to the reloader is higher than how fast the reloader is able to consume those requests. The requests are mutexed, so what happens is that we have a long (and growing) list of reloader child processes spun up by the sighup that we are not able to estinguish.
This consumes memory and may end up in having the speaker killed.

To fix this, we make the debounce timeout longer

Fixes https://github.com/metallb/metallb/issues/1581


oribon (approved):
lgtm

PR #1590, opened 2022-09-06
Fixed #1589

### LINKED ISSUE #1589: feature-request: fallback to SingleStack when PreferDualStack
metallb-version 0.13.4
k8s-version 1.23.8

https://github.com/metallb/metallb/blob/91b34a0a7b8d6fe3735a204c6b2c205133bb9d6a/internal/allocator/allocator.go#L262

is should be better if we could fallback to singleStack when preferDualStack.

woodgear:


fedepaol:
I won't discuss the code yet but comment the behaviour.

Falling back to using one ip only might be fine, but we should give priority to pools with two ips as they'll offer a better service. With this implementation, if we find a single stack pool that matches, we assign it even though we'd have another one available.
This also poses another question: what to do in case we have two pools, one with v4 only and another one with v6 only?
Which one would the user expect to be used? Do we want to be deliberate about that (ie favouring the "primary" ip)?

And how about "RequireDualStack" services? Should we give them one single ip too in case two are not available?

PR #1600, opened 2022-09-08
Layer2: fix bug when delete LB service

qingwusunny:
L2 DeleteBalancer: continue instead of return if one of the service's IP has been used by another service.


fedepaol:
Good catch! LGTM

PR #1607, opened 2022-09-13
Skip unnecessary node events in config reconciler

pperiyasamy:
The config controller watches node event to populate l2 and bgp advertisement objects with selected nodes based on node selector labels.

But currently a change in the node results in reprocessing the configuration, which results in reprocessing the services. This results in an unnecessary cpu load.

Hence this change skips processing unnecessary node events other than node create, node update (if its with new labels) and node delete events.


internal/k8s/controllers/config_controller.go
```diff
@@ -114,6 +119,10 @@ func (r *ConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 		Nodes:              nodes.Items,
 	}
 
+	for _, node := range nodes.Items {
+		r.Nodes[node.Name] = node
```
fedepaol, line 123: I think every round must be deleted and recreated otherwise we won't be able to deal with deletions (or, you must iterate over it and remove nodes not in items, but recreating seems enough)

internal/k8s/controllers/config_controller.go
```diff
@@ -149,6 +158,28 @@ func (r *ConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 }
 
 func (r *ConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	p := predicate.NewPredicateFuncs(
+		func(obj client.Object) bool {
+			newNodeObj, ok := obj.(*corev1.Node)
+			if !ok {
+				return true
+			}
+			err := r.Get(context.Background(), types.NamespacedName{Namespace: newNodeObj.Namespace,
```
fedepaol, line 167: I'd use the predicate version with the action instead of doing the get.

internal/k8s/controllers/config_controller.go
```diff
@@ -149,6 +158,28 @@ func (r *ConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+				Name: newNodeObj.Name}, &corev1.Node{})
+			if err != nil {
+				if apierrors.IsNotFound(err) {
+					return true
+				}
+				level.Error(r.Logger).Log("controller", "ConfigReconciler", "error", "failed to get node", "error", err)
+				return false
+			}
+			if existingNodeObj, ok := r.Nodes[newNodeObj.Name]; ok &&
+				labels.Set(existingNodeObj.Labels).String() == labels.Set(newNodeObj.Labels).String() {
```
fedepaol, line 177: will these be always in the right order?

fedepaol:
Can you add a unit test for this scenario?

internal/k8s/controllers/config_controller.go
pperiyasamy: correct, but anyway this local cache is no longer needed with new predicate version which overrides `UpdateFunc`.

internal/k8s/controllers/config_controller.go
pperiyasamy: done.

internal/k8s/controllers/config_controller.go
pperiyasamy: yes, changed to use `labels.Equals` method.

tasks.py
```diff
@@ -666,8 +666,13 @@ def _replace(pattern):
 @task
 def test(ctx):
     """Run unit tests."""
-    run("go test -short ./...")
-    run("go test -short -race ./...")
+    envtest_asset_dir = os.getcwd() + "/testbin"
+    run('sudo rm -rf "%s"' % envtest_asset_dir)
+    os.mkdir(envtest_asset_dir)
```
fedepaol, line 671: what is this for?

fedepaol:
LGTM, just a question about the test setup

tasks.py
pperiyasamy: this is for setting up envtest cluster as it is needed for running `TestNodeEvent` test.

pperiyasamy: yes, done.

tasks.py
```diff
@@ -666,9 +666,10 @@ def _replace(pattern):
 @task
 def test(ctx):
     """Run unit tests."""
-    run("go test -short ./...")
-    run("go test -short -race ./...")
-
+    envtest_asset_dir = os.getcwd() + "/dev-env/unittest"
+    run("source {}/setup-envtest.sh; fetch_envtest_tools {}; setup_envtest_env {}; go test -short ./...".format(envtest_asset_dir, envtest_asset_dir, envtest_asset_dir), echo=True)
+    run("source {}/setup-envtest.sh; fetch_envtest_tools {}; setup_envtest_env {}; go test -short -race ./...".format(envtest_asset_dir, envtest_asset_dir, envtest_asset_dir), echo=True)
+    run("sudo rm -rf {}/bin".format(envtest_asset_dir), echo=True)
```
fedepaol, line 672: I'd just add it to gitignore and not remove it, otherwise every time we run unit tests (locally) we'll download everything again.

tasks.py
```diff
@@ -666,9 +666,10 @@ def _replace(pattern):
 @task
 def test(ctx):
     """Run unit tests."""
-    run("go test -short ./...")
-    run("go test -short -race ./...")
-
+    envtest_asset_dir = os.getcwd() + "/dev-env/unittest"
+    run("source {}/setup-envtest.sh; fetch_envtest_tools {}; setup_envtest_env {}; go test -short ./...".format(envtest_asset_dir, envtest_asset_dir, envtest_asset_dir), echo=True)
```
fedepaol, line 670: We can run fetch once I guess, maybe even setup, before running the two tests?

fedepaol:
LGTM

Issue #1610, opened 2022-09-14
[Enhancement] helm should wait critical condition of web-hook service ready


panpan0000:
### [Root Cause]

The metallb controller should be treated as `ready` until webhook-service ready.
Currently , the readiness use `/metrics` which will be ready sooner when pod starts, but the `webhook-service` will be slower.
And `helm --wait` should wait critical condition of web-hook service ready.

### [Reproduce steps]

for example: In below script , the second line of `kubectl apply` tends to fail
```
    helm --install metallb  metallb/metallb  --wait

    cat << EOF | kubectl apply -f -
apiVersion: metallb.io/v1beta1
kind: IPAddressPool
metadata:
  name: "${poolName}"
  namespace: metallb-system
spec:
  addresses:
  - ${vipRange}
---
EOF
```

The **error message** will be 
```
 Error from server (InternalError): error when creating "STDIN": Internal error occurred: 
  failed calling webhook "ipaddresspoolvalidationwebhook.metallb.io": 
  failed to call webhook: Post "https://metallb-webhook-service.metallb-system.svc:443/validate-metallb-io-v1beta1-ipaddresspool?timeout=10s": 
  dial tcp 100.67.81.84:443: connect: connection refused
```

The webhook will be slower, so the controller pod will not be treat as ready until the webhook URL serviceable.

I will create a PR to enhance this.

panpan0000:
close with the PR merged

PR #1611, opened 2022-09-14
Enhancement: adjust the readinessProbe to wait webhook-server to be ready

panpan0000:
To address issue https://github.com/metallb/metallb/issues/1610

### DIFF
--- a/charts/metallb/templates/controller.yaml
+++ b/charts/metallb/templates/controller.yaml
@@ -100,8 +100,9 @@ spec:
         {{- if .Values.controller.readinessProbe.enabled }}
         readinessProbe:
           httpGet:
-            path: /metrics
-            port: monitoring
+            path: /
+            port: webhook-server
+            scheme: HTTPS
           initialDelaySeconds: {{ .Values.controller.readinessProbe.initialDelaySeconds }}
           periodSeconds: {{ .Values.controller.readinessProbe.periodSeconds }}
           timeoutSeconds: {{ .Values.controller.readinessProbe.timeoutSeconds }}
--- a/charts/metallb/templates/controller.yaml
+++ b/charts/metallb/templates/controller.yaml
@@ -100,7 +100,7 @@ spec:
         {{- if .Values.controller.readinessProbe.enabled }}
         readinessProbe:
           httpGet:
-            path: /
+            path: /validate-metallb-io-v1beta1-addresspool
             port: webhook-server
             scheme: HTTPS
           initialDelaySeconds: {{ .Values.controller.readinessProbe.initialDelaySeconds }}
--- a/config/controllers/controller.yaml
+++ b/config/controllers/controller.yaml
@@ -48,7 +48,7 @@ spec:
           failureThreshold: 3
         readinessProbe:
           httpGet:
-            path: /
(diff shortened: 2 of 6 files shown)


gclawes:
Chart/manifest diff looks fine to me.  I'm less familiar with the webook logic, if someone else can verify this is OK then I think it's good to go.

fedepaol:
@panpan0000 thanks for the PR!
Couple of notes:
- the commit subject is cut, please fix it and add a body with the reason why we need this (we are trying to adhere to https://www.sangfor.com/)
- I think you'll have to run inv generatemanifests to generate the all-in-one manifests out of the config

panpan0000:
Thanks @fedepaol . I've update the commit message as you suggested.

But I'm newbie for invoke, I met problem with `inv generatemanifests`, Can somebody please give some hints ?

```
#inv generatemanifests
Error: rawResources failed to read Resources: Load from path ../native failed: '../native' must be a file (got d='/root/metallb/config/native')
```

I guess maybe tool version compatible issue, but no luck when digging into contribution guide doc of metallb. 
```
# controller-gen --version
Version: (unknown)  --------> I complied from src code https://github.com/kubernetes-sigs/controller-tools.git
# kustomize version
Version:kustomize/v4.5.7 
# go version
go version go1.19.1 linux/amd64
```

panpan0000:
I think I've addressed the second controller-gen problem.
The controller-gen should be exactly v0.7.0, otherwise, many additional change will show up ( I previously used v0.9.2 ). 
 
I  have to **re-compile** `controller-gen` binary to ensure "controller-gen --version" shows "v0.7.0".

```
#controller-gen --version
Version: v0.7.0

#controller-gen crd:crdVersions=v1 rbac:roleName=manager-role webhook paths=\"./api/...\" output:crd:artifacts:config=config/crd/bases^C

#git diff
# nothing more shows up,  ** I guess my PR is all set now ?**

```

This takes the newbie contribution a bit hard. maybe we should note it somewhere like contributor guide docs.

panpan0000:
github action still fails the `inv generatemanifests` with error message:
```
Error: rawResources failed to read Resources: Load from path ../native failed: '../native' must be a file (got d='/root/metallb/config/native')
```

After digging into `tasks.py`, I found out the error comes from command `kubectl kustomize config/frr > config/manifests/metallb-frr.yaml `
although my kustomize version is up to date, but my `kubectl` is very old , v1.20.2.

after I upgrade my `kubectl` to `v1.23.7`, issue gone!

panpan0000:
way is not easy...
Directly call webhook root path "IP:9443:/" will get HTTP code 404,  and kubernetes treats it as not ready.that's why previous e2e test fails.
while API like "IP:9443:/validate-metallb-io-v1beta1-addresspool" without data shows:
```
{"response":{"uid":"","allowed":false,"status":{"metadata":{},"message":"contentType=,expected application/json","code":400}}}
```
but at least it responses 200 HTTP code.

So I added second commit : b351e473ddd642a1144a17354a97b17c565fc35f

I know it's not elegant, because when the readinessProbe path may change if v1beta1 bumps to v1. 

But seems it's the only solution out ?

fedepaol:
Thanks!

panpan0000:
Thanks @fedepaol

PR #1619, opened 2022-09-19
update mirror-server golang version

my-git9:
Thanks for sending a pull request! A few things before we get started:

1. If this is your first time, please read the [contributing guide](https://metallb.universe.tf/community/#contributing)
2. For non-trivial pull requests, please [file an
   issue](https://github.com/metallb/metallb/issues/new) first, and get
   agreement that the change is a good idea, and a general guideline
   for how it should be implemented, before sending code. Large PRs
   that weren't first discussed and agreed upon in an issue won't be
   accepted.
3. If the PR fixes a particular bug, please include the words "Fixed
   #<issue number>" in the PR text, so that the bug auto-closes when
   the PR is merged.

PR #1621, opened 2022-09-24
remove mirror-server

my-git9:
remove the mirror-server as it's no longer used.
Associated discussion: https://github.com/metallb/metallb/pull/1619#issuecomment-1255936967


liornoy:
Looks good. The mirror-server was removed completely as far as I can see.

Issue #1459, opened 2022-06-28
Extend crds with additionalPrinterColumns


oribon:
With the new crds it would be useful to have a bit more info when using `kubectl get`, for example:
```
$ kubectl get bgppeers
NAME    IP            MYASN    ASN          ...
peer1   1.2.3.4       64512    64512
peer1   5.6.7.8       64512    64513

$ kubectl get ipaddresspools
NAME    ADDRESSES              ...
pool1   1.2.3.1-1.2.3.254
```
Should be a matter of adding (and thinking about) the relevant `+kubebuilder:printcolumn` markers and re-generating the crds.
https://book.kubebuilder.io/reference/generating-crd.html#additional-printer-columns
https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definitions/#additional-printer-columns

attiss:
@oribon @fedepaol I'd be happy to contribute this.

fedepaol:
Thanks! Assigning this to you

PR #1632, opened 2022-10-07
CRDs: additionalPrinterColumns configuration

attiss:
This commit adds `+kubebuilder:printcolumn` markers to CRD structs so the generated CRDs are extended with `additionalPrinterColumns` properties.

---

Closes: https://github.com/metallb/metallb/issues/1459

### DIFF
--- a/api/v1beta1/bgpadvertisement_types.go
+++ b/api/v1beta1/bgpadvertisement_types.go
@@ -73,6 +73,10 @@ type BGPAdvertisementStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="IPAddressPools",type=string,JSONPath=`.spec.ipAddressPools`
+//+kubebuilder:printcolumn:name="IPAddressPool Selectors",type=string,JSONPath=`.spec.ipAddressPoolSelectors`
+//+kubebuilder:printcolumn:name="Peers",type=string,JSONPath=`.spec.peers`
+//+kubebuilder:printcolumn:name="Node Selectors",type=string,JSONPath=`.spec.nodeSelectors`,priority=10
 
 // BGPAdvertisement allows to advertise the IPs coming
 // from the selected IPAddressPools via BGP, setting the parameters of the
--- a/api/v1beta1/bfdprofile_types.go
+++ b/api/v1beta1/bfdprofile_types.go
@@ -74,6 +74,10 @@ type BFDProfileStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="Passive Mode",type=boolean,JSONPath=`.spec.passiveMode`
+//+kubebuilder:printcolumn:name="Transmit Interval",type=integer,JSONPath=`.spec.transmitInterval`
+//+kubebuilder:printcolumn:name="Receive Interval",type=integer,JSONPath=`.spec.receiveInterval`
+//+kubebuilder:printcolumn:name="Multiplier",type=integer,JSONPath=`.spec.detectMultiplier`
 
 // BFDProfile represents the settings of the bfd session that can be
 // optionally associated with a BGP session.
--- a/api/v1beta1/bgppeer_types.go
+++ b/api/v1beta1/bgppeer_types.go
@@ -96,6 +96,10 @@ type BGPPeerStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="Address",type=string,JSONPath=`.spec.peerAddress`
+//+kubebuilder:printcolumn:name="ASN",type=string,JSONPath=`.spec.peerASN`
+//+kubebuilder:printcolumn:name="BFD Profile",type=string,JSONPath=`.spec.bfdProfile`
+//+kubebuilder:printcolumn:name="Multi Hops",type=string,JSONPath=`.spec.ebgpMultiHop`
 
 // BGPPeer is the Schema for the peers API.
 type BGPPeer struct {
--- a/api/v1beta1/ipaddresspool_types.go
+++ b/api/v1beta1/ipaddresspool_types.go
@@ -50,6 +50,9 @@ type IPAddressPoolStatus struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:subresource:status
 // +kubebuilder:storageversion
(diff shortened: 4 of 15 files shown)


attiss:
# BFDProfile

```
$ kubectl get bfdprofiles -A
NAMESPACE        NAME             PASSIVE MODE   TRANSMIT INTERVAL   RECEIVE INTERVAL   MULTIPLIER
metallb-system   passive-bfd      true           270                 380                5
metallb-system   testbfdprofile                  270                 380
```

---

# BGPPeer

```
$ kubectl get bgppeers -A
NAMESPACE        NAME         ADDRESS      ASN     BFD PROFILE      MULTI HOPS
metallb-system   peer         172.18.0.5   64514                    true
metallb-system   peersample   172.30.0.3   64512   testbfdprofile
```

---

# IPAddressPool

```
$ kubectl get ipaddresspools -A
NAMESPACE        NAME          AUTO ASSIGN   AVOID BUGGY IPS   ADDRESSES
metallb-system   dev-env-bgp   true          false             ["192.168.10.0/24","fc00:f853:0ccd:e799::/124"]
metallb-system   second-pool   true          false             ["192.168.20.0/24","192.168.9.6-192.168.9.10"]
metallb-system   third-pool    true          false             ["10.0.0.0/8"]
```

---

# Community

```
$ kubectl get communities -A
NAMESPACE        NAME          AGE
metallb-system   communities   157m
```

I wanted to add a single column that provides a list of alias names that the individual resources define using the following marker:

```
+kubebuilder:printcolumn:name="Communities",type=string,JSONPath=`.spec.communities[*].name`
```

However, it doesn't work as expected, only the first item from the `.spec.communities` array is added to the output.
(Related kubectl issue: https://github.com/kubernetes/kubectl/issues/517)

---

# L2Advertisement ❗ 

```
$ kubectl get l2advertisements -A
NAMESPACE        NAME      IPADDRESSPOOLS   IPADDRESSPOOL SELECTORS   NODE SELECTORS                                                                                            INTERFACES
metallb-system   example   ["first-pool"]                             [{"matchLabels":{"kubernetes.io/hostname":"NodeA"}},{"matchLabels":{"kubernetes.io/hostname":"NodeB"}}]   ["eth0"]
```

I did not find any better way to print these selectors, I'm open for suggestions.

---

# BGPAdvertisement ❗ 

```
$ kubectl get bgpadvertisements -A
NAMESPACE        NAME      IPADDRESSPOOLS                  IPADDRESSPOOL SELECTORS                PEERS
metallb-system   empty     ["dev-env-bgp"]
metallb-system   example   ["dev-env-bgp","second-pool"]                                          ["peersample"]
metallb-system   private                                   [{"matchLabels":{"type":"private"}}]   ["peersample"]
```

I did not find any better way to print these selectors, I'm open for suggestions.

api/v1beta1/bgpadvertisement_types.go
```diff
@@ -73,6 +73,9 @@ type BGPAdvertisementStatus struct {
 
 //+kubebuilder:object:root=true
 //+kubebuilder:subresource:status
+//+kubebuilder:printcolumn:name="IPAddressPools",type=string,JSONPath=`.spec.ipAddressPools`
+//+kubebuilder:printcolumn:name="IPAddressPool Selectors",type=string,JSONPath=`.spec.ipAddressPoolSelectors`
+//+kubebuilder:printcolumn:name="Peers",type=string,JSONPath=`.spec.peers`
```
fedepaol, line 78: Any reason why we don't have the node selector here?

fedepaol:
I am torn about the selectors, but I think the only alternative is to leave them out (which might be even more confusing).
One mitigation can be leave them last.
In case of both ipaddresspool selector / node selector I'd give this order:
.... IPAddressPools - IPAddressPool selectors - Node Selectors

So at least they are on the right when a user sees them.

api/v1beta1/bgpadvertisement_types.go
attiss: I added those fields which seemed to be the most "important" for the resources. (So the output doesn't get too wide.)
Or in other words those configurations which I'd most often check.

In case of `BGPAdvertisement` for me it seemed these were the most important things:
- what addresses are we advertising,
- who are we advertising to.

I'm happy to add the node selector if that's also something users would usually set and check.

attiss: (Btw. we can even add "hidden" fields that are only appear with `-o wide`.)

fedepaol: Yep, I was curious about the asymmetry between l2 and bgp advertisement, where we had the node selector only on the l2 one.
Regarding hidden fields, TIL. I think it's reasonable to keep the node selector hidden in both.

fedepaol:
> > I am torn about the selectors, but I think the only alternative is to leave them out (which might be even more confusing).
> > One mitigation can be leave them last.
> 
> That'd definitely a more difficult change, but I think an alternative solution would be to add (and always keep up to date) an `IPAddressPool` / `Node` array to the Status of these resources which are matching either the "static" reference or the selector criteria.
> 
> Then we could print the name of all the matching `IPAddressPool`s or `Node`s in a single column.
> 

Sounds legit. Let me add that to the list of things it may make sense to expose as status. I am working on a proposal (that will turn in a lot of issues :P to be worked) to collect all the status we may want to expose.

I'd apply the change suggested above (hide the node selectors) and go out with this first version. It's not as binding as an api in any case, and we can change it later on.

> For example:
> 
> ```yaml
> apiVersion: metallb.io/v1beta1
> kind: BGPAdvertisement
> metadata:
>   name: example
>   namespace: metallb-system
> spec:
>   ipAddressPools:
>   - first-pool
>   - second-pool
>   ipAddressPoolSelectors:
>   - matchLabels:
>       type: example
> status:
>   ipAddressPools:
>   - first-pool
>   - second-pool
>   - example-1-pool
>   - example-2-pool
> ```
> 
> arrow_down
> 
> ```
> $ kubectl get bgpadvertisements -A
> NAMESPACE        NAME      IPADDRESSPOOLS                                                     PEERS
> metallb-system   example   ["first-pool", "second-pool", "example-1-pool", "example-2-pool"]  ["peersample"]                                    
> ```

attiss:
>I am working on a proposal (that will turn in a lot of issues :P to be worked) to collect all the status we may want to expose.

Sounds good! 🥳 

>I'd apply the change suggested above (hide the node selectors) and go out with this first version.

Changes applied. I also updated the output example in the comment: https://github.com/metallb/metallb/pull/1632#issuecomment-1272088432

fedepaol:
LGTM!
Thanks! Let's wait for the CI

PR #1633, opened 2022-10-07
Documentation: added missing namespaces

attiss:
This PR adds namespace configuration for the example resources where it was missing.

### DIFF
--- a/website/content/configuration/_advanced_bgp_configuration.md
+++ b/website/content/configuration/_advanced_bgp_configuration.md
@@ -44,6 +44,7 @@ apiVersion: metallb.io/v1beta1
 kind: BGPAdvertisement
 metadata:
   name: local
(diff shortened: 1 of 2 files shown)


design/layer2-bind-interfaces.md
```diff
@@ -136,8 +136,8 @@ in the form of veth pair. For example, if each node in the cluster has 3 physica
 apiVersion: metallb.io/v1beta1
 kind: L2Advertisement
 metadata:
-name: example-advertisement5
-namespace: metallb-system
+  name: example-advertisement5
```
fedepaol, line 139: this is a design proposal. As much as it's wrong, I would avoid changing it.

attiss: Ah, fair point! Removed the change from the commit.

Issue #1532, opened 2022-07-22
Remove the "experimental" from FRR mode


fedepaol:
MetalLB with FRR support was released with v0.12.0, adding this issue to track the removal of the experimental warning.
To add more information, the Openshift version of metallb is using FRR by default and was GA-ed around the same time.

fedepaol:
Just to add more color to the "experimental" notice, that was always meant to be "less battle tested" than the native BGP implementation.
We have running CI covering both native and FRR mode, and we have all the interest in fixing bugs related to the FRR mode as soon or later it is going to be the only BGP mode (as per https://github.com/metallb/metallb/blob/main/design/0001-frr.md )

I understand that the wording might scare some users away, so maybe it's now time to remove it, adding a note of being less mature only.

emaante:
Hello,
we've been using MetalLB in production over here at Deutsche Telekom TDI for awhile. Having BFD support (provided by FRR mode) would be nice, but due to telco requirements, "experimental" naming does not fly very well with customers, even if tested (upstream and locally).
Thanks for clarifications @fedepaol , removing this would be very helpful.

PR #1636, opened 2022-10-11
Docs: remove the experimental wording around FRR

fedepaol:
FRR is now be used by multiple users (and in OpenShift). Given that in long term we want it to be the BGP backend, it makes sense to remove the "experimental" wording that might intimidate some users.


fedepaol:
Fixes https://github.com/metallb/metallb/issues/1532

Issue #1448, opened 2022-06-22
Annotate a service with the pool used to provide its IP


fedepaol:
Especially now that we have multiple configurations tied to the ipaddresspool, it'd be useful to see which pool was used to assign the ip to a given service. It can be done contextually to the ip assignement.

fedepaol:
@cyclinder that'd be great!
Assigning it to you

cyclinder:
Yeah! But maybe I don't have extra time to work on this lately, maybe I'll look into it during the weekend.

fedepaol:
No rush! I was just going through the issues and found this

PR #1637, opened 2022-10-12
annotate a service with the pool used to provide its IP

cyclinder:
annotate a service with the pool used to provide its IP

Signed-off-by: cyclinder <qifeng.guo@daocloud.io>

Fixed https://github.com/metallb/metallb/issues/1448

### DIFF
--- a/controller/main.go
+++ b/controller/main.go
@@ -86,16 +86,24 @@ func (c *controller) SetBalancer(l log.Logger, name string, svcRo *v1.Service, _
 		return successRes
 	}
 
+	toWrite := svcRo.DeepCopy()
 	if !reflect.DeepEqual(svcRo.Status, svc.Status) {
-		var st v1.ServiceStatus
-		st, svc = svc.Status, svcRo.DeepCopy()
-		svc.Status = st
+		toWrite.Status = svc.Status
+	}
+
+	if !reflect.DeepEqual(svcRo.Annotations, svc.Annotations) {
+		toWrite.Annotations = svc.Annotations
+	}
+
+	if !reflect.DeepEqual(toWrite, svcRo) {
 		if err := c.client.UpdateStatus(svc); err != nil {
-			level.Error(l).Log("op", "updateServiceStatus", "error", err, "msg", "failed to update service status")
+			level.Error(l).Log("op", "updateServiceStatus", "error", err, "msg", "failed to update service")
 			return controllers.SyncStateError
 		}
+		level.Info(l).Log("event", "serviceUpdated", "msg", "updated service object")
 	}
-	level.Info(l).Log("event", "serviceUpdated", "msg", "updated service object")
+
+	level.Info(l).Log("event", "serviceUpdated", "msg", "service is not updated")
 	return successRes
 }
 
--- a/controller/service.go
+++ b/controller/service.go
@@ -30,8 +30,9 @@ import (
 )
 
 const (
-	annotationAddressPool     = "metallb.universe.tf/address-pool"
-	annotationLoadBalancerIPs = "metallb.universe.tf/loadBalancerIPs"
+	annotationAddressPool        = "metallb.universe.tf/address-pool"
+	annotationLoadBalancerIPs    = "metallb.universe.tf/loadBalancerIPs"
+	annotationIPAllocateFromPool = "metallb.universe.tf/ip-allocated-from-pool"
 )
 
 func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service) bool {
@@ -161,6 +162,10 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 		lbIngressIPs = append(lbIngressIPs, v1.LoadBalancerIngress{IP: lbIP.String()})
 	}
 	svc.Status.LoadBalancer.Ingress = lbIngressIPs
+	if svc.Annotations == nil {
+		svc.Annotations = make(map[string]string)
+	}
+	svc.Annotations[annotationIPAllocateFromPool] = pool
 	return true
 }
 
@@ -168,6 +173,7 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 // this controller.
 func (c *controller) clearServiceState(key string, svc *v1.Service) {
 	c.ips.Unassign(key)
+	delete(svc.Annotations, annotationIPAllocateFromPool)
 	svc.Status.LoadBalancer = v1.LoadBalancerStatus{}
 }
 
--- a/e2etest/l2tests/l2.go
+++ b/e2etest/l2tests/l2.go
@@ -676,10 +676,14 @@ var _ = ginkgo.Describe("L2", func() {
 			err = config.ValidateIPInRange([]metallbv1beta1.IPAddressPool{pool}, ingressIP)
 			framework.ExpectNoError(err)
 
+			ginkgo.By("validate annotating a service with the pool used to provide its IP")
+			framework.ExpectEqual(svc.Annotations["metallb.universe.tf/ip-allocated-from-pool"], pool.Name)
+
 			services = append(services, svc)
 			servicesIngressIP = append(servicesIngressIP, ingressIP)
 
 			for j := 0; j <= i; j++ {
+
 				ginkgo.By(fmt.Sprintf("validate service %d IP didn't change", j+1))
 				ip := e2eservice.GetIngressPoint(&services[j].Status.LoadBalancer.Ingress[0])
 				framework.ExpectEqual(ip, servicesIngressIP[j])


controller/service.go
```diff
@@ -32,8 +32,9 @@ import (
 )
 
 const (
-	annotationAddressPool     = "metallb.universe.tf/address-pool"
-	annotationLoadBalancerIPs = "metallb.universe.tf/loadBalancerIPs"
+	annotationAddressPool        = "metallb.universe.tf/address-pool"
+	annotationLoadBalancerIPs    = "metallb.universe.tf/loadBalancerIPs"
+	annotationIPAllocateFromPool = "metallb.universe.tf/allocate-pool-from"
```
fedepaol, line 37: Can we call this "ip-allocated-from-pool"?

fedepaol:
Neat! Can you modify one of the e2e tests to validate that the annotation is set properly? No need to create a new one I think.

controller/service.go
cyclinder: Sure!

e2etest/pkg/config/validate.go
```diff
@@ -14,6 +14,23 @@ import (
 	"k8s.io/kubernetes/test/e2e/framework"
 )
 
+func ValidateAnnotateIPAllocatedFromPool(addressPools []metallbv1beta1.IPAddressPool, annotations map[string]string) error {
```
fedepaol, line 17: If you call it with a single pool, this can be an item instead of a slice

e2etest/pkg/config/validate.go
```diff
@@ -14,6 +14,23 @@ import (
...
+	found := false
+	if poolName, ok := annotations["metallb.universe.tf/ip-allocated-from-pool"]; ok {
```
fedepaol, line 19: I'd return error early and avoid the extra nesting
poolName, ok := ... 
if !ok {
    return error
}

At that point you can return instead of break-ing and return the error as the last return statement (and get rid of found)

cyclinder: Thanks!

e2etest/pkg/config/validate.go
```diff
@@ -14,6 +14,13 @@ import (
 	"k8s.io/kubernetes/test/e2e/framework"
 )
 
+func ValidateAnnotateIPAllocatedFromPool(addressPool *metallbv1beta1.IPAddressPool, annotations map[string]string) error {
+	if poolName, ok := annotations["metallb.universe.tf/ip-allocated-from-pool"]; ok && poolName == addressPool.Name {
+		return nil
+	}
+	return errors.New("annotation of service with the pool used to provide its no found")
```
fedepaol, line 21: Much better :-)

fedepaol: But since a miss in a map returns the zero-value of the element (in this case the string), you can even do
`if annotations["metallb.universe.tf/ip-allocated-from-pool"] == addressPool.Name`
At this point I'd argue we need a function :-)

fedepaol:
Also, can you change the commit message to have the subject starting with a capital letter and provide a body on why the feature is useful? Thanks!

e2etest/pkg/config/validate.go
cyclinder: Do you mean it need another function here?  I don'k think so ..

`if poolName, ok := annotations["metallb.universe.tf/ip-allocated-from-pool"]; ok && poolName == addressPool.Name` it validate two things:

- Do map-key(`metallb.universe.tf/ip-allocated-from-pool`) is exist?
- If exist, checking the values of `addressPool.Name` and map-value are equal.

fedepaol: Yes, but since when you check the map and the item is not there, the map will give you "", there's no need for the "ok" part

fedepaol: It's not like java where you'd get a NPE, for example

cyclinder:
Yes, I saw it. I will test it in my local, and then update this PR.

controller/main.go
```diff
@@ -88,7 +88,9 @@ func (c *controller) SetBalancer(l log.Logger, name string, svcRo *v1.Service, _
 
 	if !reflect.DeepEqual(svcRo.Status, svc.Status) {
 		var st v1.ServiceStatus
-		st, svc = svc.Status, svcRo.DeepCopy()
+		var annotations map[string]string
+		st, annotations, svc = svc.Status, svc.Annotations, svcRo.DeepCopy()
+		svc.Annotations = annotations
```
fedepaol, line 93: I think we should reflect.DeepCopy annotations here, as what you are doing is copying the reference of the map (which is owned by svcRo)

fedepaol: Also, given this is becoming a bit beefy, consider moving it in a function

e2etest/pkg/config/validate.go
```diff
@@ -14,6 +14,13 @@ import (
 	"k8s.io/kubernetes/test/e2e/framework"
 )
 
+func ValidateAnnotateIPAllocatedFromPool(addressPool *metallbv1beta1.IPAddressPool, annotations map[string]string) error {
+	if annotations["metallb.universe.tf/ip-allocated-from-pool"] == addressPool.Name {
```
fedepaol, line 18: Given how much we simplified this, I think we can use the if directly and get rid of the function

cyclinder: Thanks!

controller/main.go
cyclinder: Thanks, Done.

cyclinder: Please review the latest changes, I appreciate it if you give me some advices

cyclinder:
Hi @fedepaol  Can you take a look?

fedepaol:
yep sorry, I've been stuck with other stuff. I will do a round today / tomrorrow.
Sorry about that :/

cyclinder:
@fedepaol  oh No sorry and  No rush. Take your time!

controller/service.go
```diff
@@ -161,16 +162,44 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
...
 	c.ips.Unassign(key)
+	delete(svc.Annotations, annotationIPAllocateFromPool)
 	svc.Status.LoadBalancer = v1.LoadBalancerStatus{}
 }
 
+// updateService return true if svc's status or annotations changed
+// and update svc. Or nothing to do and return false.
+func (c *controller) updateService(svcRo, svc *v1.Service) bool {
+	isUpdate := false
+	if !reflect.DeepEqual(svcRo.Status, svc.Status) {
```
fedepaol, line 184: What I don't like about this function is, it says "updateService" but doesn't tell how.
So I'd split it like this

We do the reflect.DeepEqual checks outside, and we split this function in two:
- updateSvcStatus
- updateSvcAnnotations

Also, this function is added as a method of the controller but it's not really using anything from it, I'd convert the two functions to "non methods"

controller/service.go
```diff
@@ -161,16 +162,44 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
...
+		var st v1.ServiceStatus
+		st, svc = svc.Status, svcRo.DeepCopy()
+		svc.Status = st
+		isUpdate = true
+	}
+
+	if !reflect.DeepEqual(svcRo.Annotations, svc.Annotations) {
+		var annotations map[string]string
+		annotations = svc.Annotations
+		if reflect.DeepEqual(svc, svcRo) {
```
fedepaol, line 194: is this necessary? We entered here because the annotations are different, so this will always be false

cyclinder: Here is a nit, ignore..

controller/service.go
cyclinder: Totally agree with you and Thanks. 

- Call `c.client.UpdateStatus` in `updateSvcStatus` and `updateSvcAnnotations` respectively.
- Or call `c.client.UpdateStatus` after calling `updateSvcStatus` and `updateSvcAnnotations`? 

Which is better?  I prefer option 2 , because this only call the k8s api once.

fedepaol: Yep, 2 is better. The "real" call can still happen outside

cyclinder: Thanks, Done

controller/main.go
```diff
@@ -87,14 +87,18 @@ func (c *controller) SetBalancer(l log.Logger, name string, svcRo *v1.Service, _
...
-			return controllers.SyncStateError
-		}
+		updateSvcStatus(svc, svcRo)
+	}
+
+	if !reflect.DeepEqual(svcRo.Annotations, svc.Annotations) {
+		updateSvcAnnotations(svc, svcRo)
 	}
+
+	if err := c.client.UpdateStatus(svc); err != nil {
```
fedepaol, line 97: You need to call this only if one of the deepequals above changed. I suggest using a flag.

controller/service.go
```diff
@@ -260,3 +266,18 @@ func isEqualIPs(ipsA, ipsB []net.IP) bool {
 	})
 	return reflect.DeepEqual(ipsA, ipsB)
 }
+
+// updateSvcStatus updating svc's status and svcRo's status
+func updateSvcStatus(svc, svcRo *v1.Service) {
+	var st v1.ServiceStatus
+	st, svc = svc.Status, svcRo.DeepCopy()
+	// we need be update svcRo's status
```
fedepaol, line 274: Why do we need this? svcRo is supposed to be read only

controller/main.go
cyclinder: sorry I don't notice this

fedepaol: I realize only now that this might not work because metallb has writing permissions only on the status subresource https://github.com/metallb/metallb/blob/main/config/rbac/role.yaml#L19 
So _it might be_ that this requires additional permissions, and if that is the case I am not sure it's worth it. If it's needed, we can pivot to expose this information via one of the status crds we are working to expose as part of https://github.com/metallb/metallb/pull/1685

controller/service.go
cyclinder: We should update status and annotations an same time , and then call `UpdateService`, Update svcRo's status here, I think it's OK.

cyclinder: Otherwise, when `updateSvcAnnotations ` is called, the parameter `svc` has been reassigned and annotations are lost

controller/main.go
cyclinder: Nope, I made an image with the current branch and did a test for it, and the annotaitons of the service were successfully updated.

controller/service.go
fedepaol: uhm, never easy... let me think about it, not sure I like this versin because we'll need to remember if we ever add another function that changes something else

fedepaol: So (and sorry if I stress too much on this), I think we can keep the "defensive" spirit of the original code by doing this:

```go
	if reflect.DeepEqual(svcRo, svc) {
		level.Debug(l).Log("event", "noChange", "msg", "service converged, no change")
		return successRes
	}

	toWrite := svcRo.DeepCopy()
	if !reflect.DeepEqual(toWrite.Status, svc.Status) {
		toWrite.Status = svc.Status
	}

	if !reflect.DeepEqual(toWrite.Annotations, svc.Annotations) {
		toWrite.Annotatins = svc.Annotations
	}

	if err := c.client.UpdateStatus(svc); err != nil {
		level.Error(l).Log("op", "updateServiceStatus", "error", err, "msg", "failed to update service")
	}
```

In theory, the only two things convergeBalancer is supposed to change are the annotations and the status.
We can be even more cautious and call updateStatus when at least one of the two changes

cyclinder: No sorry, Your suggestions make sense.

Agree with you, and do we also need a flag to decide `UpdateStatus`? When annotations or status changed.

fedepaol:
This looks great, super clean!
LGTM!

Issue #1591, opened 2022-09-06
Reassignment Duplicate IP after deleting IP pool


SihengCui:
Hello! I have find a problem caused by IP reassignment after deleting IP pool. 
k8s version: 1.25.0 metallb: 0.13.5
At first, I have two autoAssign ip pools.
`kubectl get ipaddresspools -n metallb-system -oyaml`
``` yaml
apiVersion: v1
items:
- apiVersion: metallb.io/v1beta1
  kind: IPAddressPool
  metadata:
    creationTimestamp: "2022-09-06T08:01:03Z"
    generation: 1
    name: cheap
    namespace: metallb-system
    resourceVersion: "1739387"
    uid: 64c320ee-0570-4685-a619-e533afa53b65
  spec:
    addresses:
    - 192.168.1.0/24
    autoAssign: true
    avoidBuggyIPs: true
- apiVersion: metallb.io/v1beta1
  kind: IPAddressPool
  metadata:
    creationTimestamp: "2022-09-06T08:05:47Z"
    generation: 1
    name: default
    namespace: metallb-system
    resourceVersion: "1740838"
    uid: 436518ec-f63c-4921-9510-c4d2e8a4941f
  spec:
    addresses:
    - 192.168.10.0/24
    autoAssign: true
    avoidBuggyIPs: true
kind: List
metadata:
  resourceVersion: ""
```
The services may be divided into ips in two different ip pools.
`kubectl get svc -A | grep Load`
```
          test         sanxianyimian-ng                     LoadBalancer   10.240.181.10    192.168.10.1    80:26491/TCP                                 3d22h
istio-system           istio-ingressgateway                 LoadBalancer   10.240.160.224   192.168.1.1   15021:37434/TCP,80:39462/TCP,443:33942/TCP   4d22h
monitor                kube-state-metrics                   LoadBalancer   10.240.92.35     192.168.1.2   8080:21949/TCP,8081:22126/TCP                4d1h
```
Then I deleted the default1 ippool, restart controller, in my opinion, the reassigned ip for `istio-ingressgateway` and `kube-state-metrics` should not use 192.168.10.1.
`kubectl delete ipaddresspools cheap -n metallb-system`
But
```
test                   sanxianyimian-ng                     LoadBalancer   10.240.181.10    192.168.10.1   80:26491/TCP                                 3d23h
[...]

PR #1647, opened 2022-10-14
Revert "When services share IP, modifying any service should fail

fedepaol:
This reverts commit aea1881ce3478b72d5f907655cb9a9eb32d13f72.

The meaning of the annotation is that MetalLB _may_ colocate the two services sharing the same IP, but it's a best effort indication. If the two services are incompatible (i.e. they have different external traffic policies), they won't get the same IP and they should not. This also fixes a bug where:
- the controller is stopped
- a service without an external ip is created
- a service with an external ip is already present
- the controller is started

If the controller processes the service without IP first, it may get the first available IP (because metallb doesn't know the allocation yet), but when the second service is processed, instead of getting a new IP because it is not compatible with the first processed service, the controller silently fail and the service holds the same IP as the other, even if no request of sharing through the annotation were made.

Fixes https://github.com/metallb/metallb/issues/1591


oribon (approved):
lgtm

PR #1649, opened 2022-10-14
CI: wait after deploying the operator

fedepaol:
We must ensure the operator is running and can serve the webhook before applying the metallb resource.


fedepaol:
Yo-loing this as it's blocking all the other PRs to pass

PR #1666, opened 2022-10-26
Fixed #1639 Apply filter to NodeReconciler similarly as PR #1607

### LINKED ISSUE #1639: NodeReconciler being invoked too frequently, may trigger errant reloads
In issues #1562 and #1606 , there were frequent ConfigReconciler calls caused by frequent Node changes. That was fixed by #1607 by adding a filters inside the ConfigReconciler that ignored some Node changes. On the bright side, I can confirm that fix worked for ConfigReconciler!

But I think we need to apply a similar filter to the NodeReconciler: https://github.com/metallb/metallb/blob/a826be5f29547e418aaa52d0cf29d2219008675f/internal/k8s/controllers/node_controller.go#L67-L76 

so that this doesn't flood the logs:

```
{"caller":"node_controller.go:42","controller":"NodeReconciler","level":"info","start reconcile":"/n4-mac","ts":"2022-10-12T14:59:50Z"}
{"caller":"speakerlist.go:271","level":"info","msg":"triggering discovery","op":"memberDiscovery","ts":"2022-10-12T14:59:50Z"}
{"caller":"node_controller.go:64","controller":"NodeReconciler","end reconcile":"/n4-mac","level":"info","ts":"2022-10-12T14:59:50Z"}
{"caller":"node_controller.go:42","controller":"NodeReconciler","level":"info","start reconcile":"/n4-mac","ts":"2022-10-12T15:00:00Z"}
{"caller":"speakerlist.go:271","level":"info","msg":"triggering discovery","op":"memberDiscovery","ts":"2022-10-12T15:00:00Z"}
{"caller":"node_controller.go:64","controller":"NodeReconciler","end reconcile":"/n4-mac","level":"info","ts":"2022-10-12T15:00:00Z"}
```

I don't have hard proof, but I have seen very rare cases where metallb mistakenly triggers a failover of the VIP due to one of the node change events, even though none of the nodes appeared to have had any issues.

I think what we should learn from #1606 is that ALL node watching code needs some kind of filter, because nodes change very frequently under normal conditions (maybe the `conditions` or the `images` sections?).

What kind of node update events does the NodeReconciler really care about? Nodes being added or deleted? Or also nodes being NotReady/Cordoned? I think we should put those checks inside the filter [here](https://github.com/metallb/metallb/blob/a826be5f29547e418aaa52d0cf29d2219008675f/internal/k8s/controllers/node_controller.go#L75) to save CPU, reduce log spam, and avoid risks of errant failovers.

giulianopz:
Fix as suggested by @mac-chaffee in issue #1639.


fedepaol:
Hi @giulianopz , thanks for the PR!

I have a couple of asks:
- can you clear the commit message? We try to adhere to https://cbea.ms/git-commit/ style
- can you unit test the controller as we did in https://github.com/metallb/metallb/pull/1607 ?

giulianopz:
Hi @fedepaol,
I added a unit test just as the one in `config_controller_test.go`, but the test fails when I run `invoke test`:
```
W1104 00:20:23.634880   69068 warnings.go:70] metallb.io v1beta1 AddressPool is deprecated, consider using IPAddressPool
W1104 00:20:23.635792   69068 warnings.go:70] metallb.io v1beta1 AddressPool is deprecated, consider using IPAddressPool
--- FAIL: TestNodeReconciler_SetupWithManager (6.90s)
    node_controller_test.go:138: 
        Expected
            <*fmt.wrapError | 0xc0002937a0>: {
                msg: "error listening on :8080: listen tcp :8080: bind: address already in use",
                err: <*net.OpError | 0xc000533ae0>{
                    Op: "listen",
                    Net: "tcp",
                    Source: nil,
                    Addr: <*net.TCPAddr | 0xc000c3e630>{IP: nil, Port: 8080, Zone: ""},
                    Err: <*os.SyscallError | 0xc000293780>{
                        Syscall: "bind",
                        Err: <syscall.Errno>0x62,
                    },
                },
            }
        to be nil
FAIL
FAIL    go.universe.tf/metallb/internal/k8s/controllers 19.415s
```
It seems a conflict between the two unit tests, but I cannot understand where in the code the port 8080 is selected.
Can you help me? Did I miss some configuration to run the unit tests locally?

internal/k8s/controllers/node_controller.go
```diff
@@ -65,16 +68,23 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
 func (r *NodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	p := predicate.NewPredicateFuncs(
-		func(obj client.Object) bool {
-			node, ok := obj.(*v1.Node)
+	p := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			newNodeObj, ok := e.ObjectNew.(*corev1.Node)
 			if !ok {
-				level.Error(r.Logger).Log("controller", "NodeReconciler", "error", "object is not node", "name", obj.GetName())
+				return true
```
fedepaol, line 75: Can we log here? Something bad happens if this cast fails

internal/k8s/controllers/node_controller.go
```diff
@@ -65,16 +68,23 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 }
 
 func (r *NodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	p := predicate.NewPredicateFuncs(
-		func(obj client.Object) bool {
-			node, ok := obj.(*v1.Node)
+	p := predicate.Funcs{
```
fedepaol, line 71: nit: can you expand the commit body with the reason why we want this change?

internal/k8s/controllers/node_controller.go
```diff
@@ -65,16 +68,23 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+			}
+			oldNodeObj, ok := e.ObjectOld.(*corev1.Node)
+			if !ok {
+				return true
+			}
+			// If there is no changes in node labels, ignore event.
+			if labels.Equals(labels.Set(oldNodeObj.Labels), labels.Set(newNodeObj.Labels)) {
 				return false
 			}
-			return node.Name == r.NodeName
```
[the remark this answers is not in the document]
giulianopz: Sorry, didn't realize.

internal/k8s/controllers/node_controller.go
giulianopz: I squashed all the commits into a single one with a body provinding some context for this change.

giulianopz: I squashed all the commits into a single one with a body provinding some context for this change.

internal/k8s/controllers/node_controller.go
```diff
@@ -65,16 +68,25 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 }
 
 func (r *NodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
-	p := predicate.NewPredicateFuncs(
-		func(obj client.Object) bool {
-			node, ok := obj.(*v1.Node)
+	p := predicate.Funcs{
+		UpdateFunc: func(e event.UpdateEvent) bool {
```
fedepaol, line 72: By doing this, the function will be triggered only during an update because here we are passing the functions explicitly (as opposed to using the predicate.NewPredicateFuncs function).

You need to implement also the other functions (like the add one, the delete one, etc).

internal/k8s/controllers/node_controller.go
```diff
@@ -65,16 +68,25 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+				return true
+			}
+			// If there is no changes in node labels, ignore event.
+			if labels.Equals(labels.Set(oldNodeObj.Labels), labels.Set(newNodeObj.Labels)) {
 				return false
 			}
-			return node.Name == r.NodeName
-		})
-
+			return oldNodeObj.Name == newNodeObj.Name && newNodeObj.Name == newNodeObj.Name
```
fedepaol, line 87: I'd keep only the new name equal, trusting the framework that it passed an old and a new instance of the same object.

internal/k8s/controllers/node_controller.go
giulianopz: For the other events I've just restored the previous function which compares nodes' names: I don't know if other checks are needed.

internal/k8s/controllers/node_controller.go
```diff
@@ -65,18 +68,45 @@ func (r *NodeReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+				return true
+			}
+			// If there is no changes in node labels, ignore event.
+			if labels.Equals(labels.Set(oldNodeObj.Labels), labels.Set(newNodeObj.Labels)) {
 				return false
 			}
-			return node.Name == r.NodeName
-		})
-
+			return newNodeObj.Name == newNodeObj.Name
```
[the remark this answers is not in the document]
giulianopz: Sorry, totally missed it. Ok, I will reuse the new function.

fedepaol:
LGTM, thanks! Let's wait for CI

PR #1674, opened 2022-10-28
Add a design proposal for announcing a service via VRFs

fedepaol:
Adding the design proposal of a feature we are requested by a few telco users: making it possible to announce a service via an interface wrapped in a VRF.

Note: this covers only the announcement part and not how the CNI handles the traffic coming to / from a VRF.

### DIFF
--- a/design/vrf-announcement.md
+++ b/design/vrf-announcement.md
@@ -1,0 +1,206 @@
+# Announce LoadBalancer services from VRFs
+
+## Summary
+
+The purpose of this enhancement is allow MetalLB to peer with external BGP peers
+using interfaces belonging to [Linux VRFs](https://www.kernel.org/doc/html/latest/networking/vrf.html),
+and to announce traffic from within VRFs.
+
+The VRF support will be available only via the [FRR based implementation](0001-frr.md).
+
+## Motivation
+
+Having interfaces wrapped in VRFs allow to set a different default gateway than
+the regular one. This helps to avoid asymmetrical routing in scenarios where the
+nodes have multiple interfaces, providing that either the CNI is able to receive
+the traffic inside the VRF or the host networking is set up in a way that the VRF
+is able to communicate with the default one for the traffic related to the
+service's IP.
+
+This is the default behaviour for a client coming from a router connected via an
+interface different from the default one.
+
+```none
+             ▲
+             │
+             │
+        ┌────┴────┐
+        │ eth0    │
+        │         │
+┌───────┴─────────┴───────────┐
+│                             │               ┌──────────────┐
+│                             │               │              │
+│                             ├────────┐      │              │
+│                             │        │      │              │
+│                             │ eth1   │◄─────┤     DCGW     │◄───────  Client
+│                             │        │      │              │
+│                             ├────────┘      │              │
+│                             │               │              │
+│                      Node   │               └──────────────┘
+│                             │
+└─────────────────────────────┘
+```
+
+And this is how it would look like with VRF:
+
+```none
+
+
+
+               ┌─────────┐
+               │ eth0    │
+               │         │
+       ┌───────┴─────────┴───────────┐
+       │                             │               ┌──────────────┐
+       │                      ┌──────┤               │              │
+       │                      │      ├────────┐      │              │
+       │                      │      │        │      │              │
+       │                      │ VRF  │ eth1   │◄─────┤     DCGW     │◄───────  Client
+       │                      │      │        │      │              │
+       │                  ────┼────► ├────────┘      │              │
+       │                      └──────┤               │              │
+       │                      Node   │               └──────────────┘
+       │                             │
+       └─────────────────────────────┘
+
+```
+
+Other advantages of having VRFs on the node are:
+
+* being able to have more complex routing tables (i.e. multiple DCGWs for HA)
+without impacting the node networking
+* ease the troubleshooting of the traffic directed to the service
+* making it possible to have multiple clients (coming from different interfaces)
+sharing the same IP
+
+VRF for BGP is supported by FRR [out of the box](https://docs.frrouting.org/en/latest/bgp.html#clicmd-router-bgp-ASN-vrf-VRFNAME).
+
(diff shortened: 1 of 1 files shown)


design/vrf-announcement.md
```diff
@@ -0,0 +1,198 @@
...
+without impacting the node networking
+* ease the troubleshooting of the traffic directed to the service
+* making it possible to have multiple clients (coming from different interfaces)
+sharing the same IP
+
+VRF for BGP is supported by FRR [out of the box](https://docs.frrouting.org/en/latest/bgp.html#clicmd-router-bgp-ASN-vrf-VRFNAME).
+
+### Goals
+
+* Extablish a BGP session with a BGP peer using an interface with VRF as master
```
liornoy, line 80: nit: Establish

design/vrf-announcement.md
```diff
@@ -0,0 +1,198 @@
...
+**Note 2**: There might be checks that we want to relax, like having
+BGPPeers with the same IP might be allowed in case the peers are reached
+from different VRFs.
+
+### How to configure host networking
+
+The right combination of host routing through ip rules and iptables rules is
+strongly dependant on the CNI. Because of this reason, we don't think is safe
+to have a one size-fit all solution. What we will provide with this change is
+a sample script (and possibly a controller) that works for kindnet (in order
```
liornoy, line 174: nit: works for the Kind network

design/vrf-announcement.md
```diff
@@ -0,0 +1,198 @@
...
+to enable CI) and calico, with the idea of extending / validating it with the
+other major CNIs.
+
+### Test Plan
+
+**e2e tests**: the dev-env will be extended with an option to have additional
+interfaces inside a VRF, and to configure the host network in such a way that
+the service is accessible from inside a VRF.
+**unit tests**: Unit tests will be added (mostly to the frr package) to ensure
```
liornoy, line 183: Looking at this with the "rich diff" - can you please make another newline above `unit test`?

design/vrf-announcement.md
```diff
@@ -0,0 +1,198 @@
...
+
+The alternative is to embed all the logic inside MetalLB, but this means that
+we might end up with per-cni logic and feels a bit less modular.
+
+## Development Phases
+
+There are strictly no phases. We want to:
+
+- validate that FRR is able to implement what we think we require
+- extend dev-env to allow traffic directed to a LB to flow from / to a VRF
```
liornoy, line 196: Will this make us double the number of external containers up to 8?
4 - default VRF + 4 custom VRF?

design/vrf-announcement.md
```diff
@@ -0,0 +1,198 @@
...
+- implement the feature
```
liornoy, line 197: For the `provide a guideline..` goal - should we add an action item for adding the required docs?

design/vrf-announcement.md
```diff
@@ -0,0 +1,198 @@
...
+VRFs, we can have different BGP peers with the same IP.
+
+**Note 2**: There might be checks that we want to relax, like having
+BGPPeers with the same IP might be allowed in case the peers are reached
+from different VRFs.
+
+### How to configure host networking
+
+The right combination of host routing through ip rules and iptables rules is
+strongly dependant on the CNI. Because of this reason, we don't think is safe
```
liornoy, line 172: nit: think it is safe

design/vrf-announcement.md
```diff
@@ -0,0 +1,198 @@
...
+
+* Extablish a BGP session with a BGP peer using an interface with VRF as master
+* Announce a LoadBalancer service to a BGP peer peered via an interface belonging
+to a VRF
+* Eventually, provide a guideline and an example on how to setup the VRF and the
+host routing, knowing that the routing logic might be CNI dependant
+
+### Non-Goals
+
+* Creating the VRF
```
oribon, line 88: what about reacting to updates to a vrf (renames/deletions/?)? I assume we do not care about this, but think it is worth mentioning here so everything is clear (what MetalLB cares about regarding VRFs)
also, if a speaker's host does not have the vrf configured what's the behavior?

design/vrf-announcement.md
```diff
@@ -0,0 +1,198 @@
...
+  peerASN: 64512
+  peerAddress: 172.30.0.3
+  vrf: red
+```
+
+Any announcement meant to be performed towards the vrf-ed BGPPeer will be added
+in the related router's section.
+
+```none
+router bgp 100 vrf red
```
oribon, line 139: can you change this example to reflect the peer above (172.30.0.3/64512/..)?

design/vrf-announcement.md
```diff
@@ -0,0 +1,198 @@
...
+  exit-address-family
+```
+
+**Note**: We must be careful in handling the FRR configuration rendering,
+because all the route maps are handled in terms of the peer ip. With
+VRFs, we can have different BGP peers with the same IP.
+
+**Note 2**: There might be checks that we want to relax, like having
+BGPPeers with the same IP might be allowed in case the peers are reached
+from different VRFs.
```
oribon, line 167: can you extend a bit about how we might handle this? adding something to the "none-vrfs" route-maps/changing the webhook/...? imo if we do/don't want to take care of this at the initial implementation we should give a brief explanation (somewhere) here

design/vrf-announcement.md
fedepaol: https://www.tkng.io/cni/kindnet/ 
Kindnet is the CNI used by kind

design/vrf-announcement.md
fedepaol: yep let me add it

design/vrf-announcement.md
fedepaol: I was thinking mostly about the validating webhooks here, let me update it.

liornoy:
LGTM

oribon (approved):
lgtm

fedepaol:
Ci is not running because we currently dont perform validation against the proposals oath.
Merging

Issue #1577, opened 2022-09-01
Webhook improvements


fedepaol:
Follow up of https://github.com/metallb/metallb/issues/1538

- Creating a bgpadvertisement first, and an incompatible (from aggregatino lenght point of view) addresspool will pass the validation, while the inversed order fails
- Whenever incompatibility between a peer and a pool is found, the error doesn't tell which are the two elements involved, making it a bit difficult to see what are the issues at a glance

kevydotvinu:
Can I work on this?

liornoy:
Hi @kevydotvinu are you still intending to work on this item?

kevydotvinu:
@liornoy, I would love to.

liornoy:
@kevydotvinu Awesome thanks!

PR #1685, opened 2022-11-03
Design: add a proposal to expose MetalLB's status via CRDs

fedepaol:
Exposing part of the internal state was not possible until the introduction of CRD based configuration, and now that we have it it makes sense to expose part of the state in order to ease debugging and troubleshooting.

### DIFF
--- a/design/crd-status.md
+++ b/design/crd-status.md
@@ -1,0 +1,346 @@
+# Exposing the status with CRDs
+
+## Summary
+
+The purpose of this enhancement is to expose useful information to perform troubleshooting
+via CRDs.
+
+## Motivation
+
+Exposing part of the internal state was not possible until the introduction of CRD
+based configuration, and despite part of the status is exposed via Prometheus
+metrics, troubleshooting MetalLB often requires inspecting the logs of the various
+controllers, and it's not always easy to understand why a service is not working.
+
+For example, inspecting logs is required when:
+
+* an invalid configuration was applied
+* the session with a BGP peer is not established
+* a service was not announced to a BGP peer
+
+### Goals
+
+* Expose useful information to ease the debug
+
+A high level list of informations we want to retrieve is:
+
+* configuration status (valid/invalid and reason)
+* number of used / available IPs per IPAddressPool
+* session state for both BGP and BFD, per node - peer
+* service announcement status, per peer - node (BGP)
+* service announcement node (L2)
+
+### Non-Goals
+
+* Providing BGP status for the legacy implementation
+* Providing a Status subsection for each resource
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+As a cluster administrator, I want to see from which nodes my service is
+announced, and to which `BGPpeer`s.
+
+#### Story 2
+
+As a cluster administrator, I want to know if the BGP / BFD session with a given
+peer is established or not, for each node.
+
+#### Story 3
+
+As a cluster administrator, I want to know if the configuration applied
+is valid or not, and why it failed.
+
+## Design Details
+
+### Making the solution scalable
+
+The biggest challenge is the fact that all the concepts related to MetalLB
+are cluster scoped, but a lot of the information we care about is node scoped.
+
+A clear example is the state of the BGP session established with a given
+`BGPpeer`, where the `BGPpeer` is defined as a cluster concept, but sessions
+are established from different nodes.
+
+If we would add a `Status` field to the `BGPPeer` resource, it would add an
+unwanted load to the APIServer on clusters with an high number of nodes,
+especially when dealing with faulty networks where the connectivity is intermittent.
+
+This consideration is driving the proposed design.
+
+#### IPAddressPool Status
+
+This should be easy enough, as it will require extending the current `IPAddressPool`
+CRD with a `Status` section:
+
+```yaml
+apiVersion: metallb.io/v1beta1
+kind: IPAddressPool
+metadata:
+  name: example
+  namespace: metallb-system
+spec:
+  addresses:
+  - 192.168.10.0/24
+  - 192.168.9.1-192.168.9.5
+  - fc00:f853:0ccd:e799::/124
+status:
+  availableIPV4: 45
+  availableIPV6: 145
+  assignedIPV4: 5
+  assignedIPV6: 52
+```
+
+```go
+    type IPAddressPoolStatus struct {
+        assignedIPV4  int
+        assignedIPV6  int
+        availableIPV4 int
+        availableIPV6 int
+}
+```
+
+#### Configuration Status
+
+Given that the configuration is composed by multiple CRs, there are few cases
+where a given configuration is invalid because of a single CR (i.e. invalid IP formatting).
+
+The majority of the scenarios involve multiple CRs which are not compatible together.
+For this reason, we think a global `ConfigurationStatus` indicator is better and
+easier to understand, compared to a "per resource" status that tells if the resource
+is valid or not.
+
+```yaml
+apiVersion: metallb.io/v1beta1
+kind: ConfigurationStatus
+metadata:
+  name: config-status
+  namespace: metallb-system
+status:
+    validConfig: false
+    error: "peer 1.2.3.4 has myAsn different from 1.2.3.5, in FRR mode all myAsn must be equal"
+```
+
+```go
+    type MetalLBConfigurationStatus struct {
+        validConfig bool
+        lastError   string
+    }
+
+    type ConfigurationStatus struct {
+        metav1.TypeMeta   `json:",inline"`
+        metav1.ObjectMeta `json:"metadata,omitempty"`
+        Status            MetalLBConfigurationStatus `json:"status,omitempty"`
+    }
+```
+
+**Note:** given the fact that the configuration is parsed both by the speakers and
+by the controller, we might want to expose the status of each component. However,
+there is currently no logic in the configuration parsing that depends on the component.
+For this reason, we can have an initial status that is produced only by the controller
+(which runs in single instance) to validate the status.
+
+As an alternative, we might consider having a state per component, named after the
+single component that produces the status, but in general having a single place to
+check seems more straightforward.
+
+If we go with the _per component_ scenario, we might add a `loadedConfiguration`
+field that exposes the latest loaded configuration by that component. This can't
+be done if we let the controller to expose the single configuration because
+what's loaded might depend on the order the CRs are received with.
+
+#### BGP / BFD Session state
+
+Because of the scalability concerns expressed above, the idea is to produce an instance
+of the resource _per peer / node_, which exposes the state of the session between
+the speaker running on a given node and a given peer.
+
+The name of a given instance will be like `nodename-peer`, and each instance will
+be labeled with the name of the node and the peer it refers to, to make it easier
+to list the status of all the sessions related to a given `BGPPeer` and a given
+node.
+
+```go
+    type MetalLBBGPStatus struct {
+        bgpStatus string
+        bfdStatus string
+    }
+
+    type BGPSessionStatus struct {
+        metav1.TypeMeta   `json:",inline"`
+        metav1.ObjectMeta `json:"metadata,omitempty"`
+        Status            MetalLBBGPStatus `json:"status,omitempty"`
+    }
+```
+
+```yaml
+apiVersion: metallb.io/v1beta1
+kind: BGPSessionStatus
+metadata:
+  name: worker0-peer1
+  namespace: metallb-system
+  labels:
+    metallb.io/node: worker0
+    metallb.io/peer: peer1
+status:
+    bgpStatus: Established
+    bfdStatus: Up
+```
+
+The string exposed is taken directly from the output of FRR.
+
+If BFD is not configured for a given `BGPPeer`, the exposed bfdStatus will be "N/A".
+
+**A note about the implementation:** without entering to much into details,
+we will need to implement some sort of polling of the FRR status. Given the
+fact that this CR has no relation with the existing ones, a valid approach is
+to follow what was done for the [metrics exporter](https://github.com/metallb/metallb/blob/main/frr-metrics/exporter.go)
+and have a different component (or even the exporter itself) polling FRR and filling
+the session status. The polling interval must be configurable and large enough to
+avoid impacts both on FRR and on the API.
+The speaker should continue not to have direct interactions with FRR.
+
+#### Service Announcement - BGP
+
+Given a service and a node, we want to expose the `BGPpeer`s the service is configured
+to be advertised to.
+
+```yaml
+apiVersion: metallb.io/v1beta1
+kind: ServiceBGPStatus
+metadata:
+  name: service1-worker0
+  namespace: servicenamespace
+  labels:
+    metallb.io/node: worker0
+    metallb.io/service: service1
+status:
+    bgpPeers:
+        - peerA
+        - peerB
+```
+
+```go
+    type MetalLBServiceBGPStatus struct {
+        Peers []string
+    }
+
+    type ServiceBGPStatus struct {
+        metav1.TypeMeta   `json:",inline"`
+        metav1.ObjectMeta `json:"metadata,omitempty"`
+        Status            MetalLBServiceBGPStatus `json:"status,omitempty"`
+    }
+```
+
+The labels will allow to easily discover which services are advertised from
+a given node, or all the peers a given service is advertised to.
+
+**Note:** this status won't take into consideration the status of the
+session with the given BGP peer, but only the will to advertise to that peer.
+This, to overcome the considerations about scalability written in the preface.
+
+#### Service Announcement - L2
+
+The useful information related to L2 are related to the node that is exposing
+the service, and via which interfaces.
+
+```yaml
+apiVersion: metallb.io/v1beta1
+kind: ServiceL2Status
+metadata:
+  name: service1
+  namespace: servicenamespace
+  labels:
+    metallb.io/node: worker0
+    metallb.io/service: service1
+status:
+    node: worker0
+    interfaces:
+        - eth0
+        - eth1
+```
+
+```go
+    type MetalLBServiceL2Status struct {
+        Node       string
+        Interfaces []InterfaceInfo
+    }
+
+    type ServiceL2Status struct {
+        metav1.TypeMeta   `json:",inline"`
+        metav1.ObjectMeta `json:"metadata,omitempty"`
+        Status            MetalLBServiceL2Status `json:"status,omitempty"`
+    }
+
+    type InterfaceInfo struct {
+        Name string `json:"name,omitempty"`
+    }
+```
+
+The absence of interfaces means all interfaces are selected.
+
+## Alternatives
+
+We might consider exposing all the information (or part of it) as labels to be applied
+to the services. However, the current proposal is easier to query and to navigate,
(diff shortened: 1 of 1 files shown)


oribon (commented):
went over a part of this briefly, need to go over this again :upside_down_face:

design/crd-status.md
```diff
@@ -0,0 +1,237 @@
...
+The biggest challenge is the fact that all the concepts related to MetalLB
+are cluster scoped, but a lot of the information we care about is node scoped.
+
+A clear example is the state of the BGP session established with a given
+`BGPpeer`, where the `BGPpeer` is defined as a cluster concept, but sessions
+are established from different nodes.
+
+If we would add a `Status` field to the `BGPPeer` resource, it would add an
+unwanted load to the APIServer on clusters with an high number of nodes,
+especially when dealing with faulty networks where the connectivity is intermittent.
```
liornoy, line 70: Can you please clarify how having a BGPSessionStatus CRD for each BGP peer (which can be thousands and more) 
is less overhead than adding this status to the already existing BGPPeer?

design/crd-status.md
```diff
@@ -0,0 +1,237 @@
...
+After this proposal converges and is accepted, separate issues will be filed in order
+to ease the development and allow the development to move in parallel.
+
+### Test Plan
+
+**e2e tests**: E2E tests will be expanded to cover bfd, using external container(s)
+running FRR. The test will need to cover the cases where a node is dropped, verifying
+that the broken route is detected by BFD.
+We **must** ensure through tests that unnecessary updates are not generated if the
+exposed status does not change.
```
liornoy, line 233: I guess you mean having a BGP session backed up with BFD.
Plus, we should also mention covering the L2 Advertisement.

liornoy (commented):
Great and anticipated design proposal!
I left one question regarding solution choice and 2 other small comments.

design/crd-status.md
```diff
@@ -0,0 +1,237 @@
...
+
+```yaml
+apiVersion: metallb.io/v1beta1
+kind: ConfigurationStatus
+metadata:
+  name: config-status
+  namespace: metallb-system
+status:
+    validConfig: false
+    error: "error message here"
```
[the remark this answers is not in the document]
fedepaol: The webhooks work _eventually_, looking at the state contained in the cache.
This means that there's a non null chance that two different and incompatible CRs creation happen in parallel and are served by two different goroutines), and we can't do anything about it.
 
On top of that, another concrete example is https://github.com/metallb/metallb/issues/1577 where the wrong order causes an invalid configuration.

design/crd-status.md
```diff
@@ -0,0 +1,237 @@
...
+kind: BGPSessionStatus
+metadata:
+  name: worker0-peer1
+  namespace: metallb-system
+  labels:
+    metallb.io/node: worker0
+    metallb.io/peer: peer1
+status:
+    bgpStatus: Established
+    bfdStatus: Up
```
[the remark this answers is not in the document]
fedepaol: I think that matters and reflect the BGP status as per the RFC.
See https://www.noction.com/blog/debug-bgp-states , where `Idle` has a different meaning from `Active` for example.
In the metrics we had to be backward compatible with the behaviour of the legacy implementation IIRC.

design/crd-status.md
```diff
@@ -0,0 +1,237 @@
...
+
+A clear example is the state of the BGP session established with a given
+`BGPpeer`, where the `BGPpeer` is defined as a cluster concept, but sessions
+are established from different nodes.
+
+If we would add a `Status` field to the `BGPPeer` resource, it would add an
+unwanted load to the APIServer on clusters with an high number of nodes,
+especially when dealing with faulty networks where the connectivity is intermittent.
+
+This consideration is driving the proposed design.
```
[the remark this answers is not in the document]
fedepaol: The stale point is very valid. If the node selector of the speaker gets reduced, an instance might disappear but all the information handled by that speaker will stay there (i.e. a bgpannouncementstatus).
Need to think about it and see how the other controllers we know of handle this scenario. An option could be, as you say, to have those status resources owned by the pod that maintains them.

design/crd-status.md
fedepaol: First, thousands of bgppeers is not realistic (as opposed to thousands of nodes).
And the difference here is the size of each resource and the number of updates.

This is both about the size of the payload each session update will produce, the conflicts and the number of upgrades.
Think about 1000 nodes and two flickering sessions from two nodes (to the same peer).
Every flick will require to move a CR that is way bigger, the updates will be the combination of the two sessions (down down, down up, up up), not considerning the number of retries given by the fact that they may try to update their version at the same time. And this is with two sessions only.

Having one CR per session limits this, while not affecting the UX (too much).

design/crd-status.md
```diff
@@ -0,0 +1,237 @@
...
+apiVersion: metallb.io/v1beta1
+kind: L2AdvertisementStatus
+metadata:
+  name: service1-worker0
+  namespace: metallb-system
+  labels:
+    metallb.io/node: worker0
+    metallb.io/service: service1
+status:
+    node: worker0
```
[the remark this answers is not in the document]
fedepaol: mostly because without interfaces selector the resource would be empty, and because I thought in L2 the focus is more on the node (we have only one leader here).
The labels are both for discovery (find all the services announced via L2 on this node) and for exposing the status.
This makes me think that we can even get rid of the node in the name.

design/crd-status.md
fedepaol: ah damn. This is a leftover, I copied it from the bfd proposal :sweat_smile:

design/crd-status.md
```diff
@@ -0,0 +1,312 @@
...
+        - eth1
+```
+
+```go
+    type MetalLBServiceL2Status struct {
+        Node       string
+        Interfaces []string
+    }
+
+    type ServiceL2Status struct {
```
liornoy, line 272: @fedepaol nit: the "ServiceL2Status" struct name should match the "kind: L2AdvertisementStatus" name ?

design/crd-status.md
```diff
@@ -0,0 +1,312 @@
...
+  assignedIPV4: 5
+  assignedIPV6: 52
+```
+
+```go
+    type IPAddressPoolStatus struct {
+        assignedIPV4  int
+        assignedIPV6  int
+        availableIPV4 int
+        availableIPV6 int
```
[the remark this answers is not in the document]
fedepaol: this is the assigned part, or I did not get the question right

design/crd-status.md
```diff
@@ -0,0 +1,312 @@
...
+kind: L2AdvertisementStatus
+metadata:
+  name: service1
+  namespace: metallb-system
+  labels:
+    metallb.io/node: worker0
+    metallb.io/service: service1
+status:
+    node: worker0
+    interfaces:
```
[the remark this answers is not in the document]
fedepaol: let's start low, this is an information that should be easy to add if needed

design/crd-status.md
```diff
@@ -0,0 +1,312 @@
...
+        metav1.ObjectMeta `json:"metadata,omitempty"`
+        Status            MetalLBServiceL2Status `json:"status,omitempty"`
+    }
+```
+
+The absence of interfaces means all interfaces are selected.
+
+## Alternatives
+
+We might consider exposing all the information (or part of it) as labels to be applied
```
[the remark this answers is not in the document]
fedepaol: we already have it, see https://github.com/metallb/metallb/pull/1637

design/crd-status.md
```diff
@@ -0,0 +1,312 @@
...
+    node: worker0
+    interfaces:
+        - eth0
+        - eth1
+```
+
+```go
+    type MetalLBServiceL2Status struct {
+        Node       string
+        Interfaces []string
```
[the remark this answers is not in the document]
fedepaol: this might be useful, but I am afraid the information has to traverse too many layers. 
What I will do is, change the field type from []string to []InterfaceInfo or something like that so it's gonna be easier to extend

design/crd-status.md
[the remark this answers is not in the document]
pperiyasamy: yes, but assigned part may not give actual numbers in case of dual stack services.

design/crd-status.md
```diff
@@ -0,0 +1,312 @@
...
+* a service was not announced to a BGP peer
+
+### Goals
+
+* Expose useful information to ease the debug
+
+A high level list of informations we want to retrieve is:
+
+* configuration status (valid/invalid and reason)
+* number of used / available IPs per IPAddressPool
```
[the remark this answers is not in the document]
fedepaol: that (especially the unassigned part) can be pretty big in case of v6, so I'd rather avoid.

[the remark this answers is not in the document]
cyclinder: So here are a number of IP statistics, Do not consider showing the specific IP list that has been used or unused?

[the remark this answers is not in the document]
fedepaol: not sure I understood the question. I thought you were talking about adding the full list of ips, one by one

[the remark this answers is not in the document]
cyclinder: Maybe like this: 

```yaml
spec:
  addresses:
  - 10.6.168.140-10.6.168.150
  autoAssign: true
  avoidBuggyIPs: true
status:
  allocatedIPs:
    10.6.168.140:
       name: svc1
       namespace: default
    10.6.168.141:
       name: svc2
       namespace: default
```

Or :

```yaml
spec:
  addresses:
  - 10.6.168.140-10.6.168.150
  autoAssign: true
  avoidBuggyIPs: true
status:
  unallocatedIPs:
  - 10.6.168.140
  - 10.6.168.141
```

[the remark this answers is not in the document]
fedepaol: yeah the problem is, you might get a big list of unallocated in case of v6, plus I am not really sure it's useful. If you allow a service to pick an ip from a pool, it means you don't care which ip it is going to take.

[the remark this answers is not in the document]
cyclinder: This feature is really useful when we assign IPs to Services via annotations: `metallb.universe.tf/loadBalancerIPs`.  When creating the service we can know which IPs are available for assignment which cannot.

design/crd-status.md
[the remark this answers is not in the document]
oribon: cool, I guess we can sort these kind of stuff when we go over the implementation part (in a different pr)

design/crd-status.md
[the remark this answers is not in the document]
oribon: +1 for ServiceL2Status (or changing the bgp one to BGPAdvertisementStatus)

design/crd-status.md
```diff
@@ -0,0 +1,316 @@
...
+* service announcement node (L2)
+
+### Non-Goals
+
+* Providing BGP status for the legacy implementation
+* Providing a Status subsection for each resource
+
+## Proposal
+
+### User Stories
```
oribon, line 40: wdyt about detailing a bit here or in a next section how the user would achieve these with the proposal?
I think it would add some value here and form some sort of (non-guaranteed) base for the future docs

design/crd-status.md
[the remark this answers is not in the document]
mlguerrero12: I think it's useful to show the allocated IPs, then the user can infer which ones are still available

design/crd-status.md
```diff
@@ -0,0 +1,316 @@
...
+be done if we let the controller to expose the single configuration because
+what's loaded might depend on the order the CRs are received with.
+
+#### BGP / BFD Session state
+
+Because of the scalability concerns expressed above, the idea is to produce an instance
+of the resource _per peer / node_, which exposes the state of the session between
+the speaker running on a given node and a given peer.
+
+The name of a given instance will be like `nodename-peer`, and each instance will
```
mlguerrero12, line 161: I understand the scalability concerns. However, I believe that an instance per node and peer pair is too much. Instead, we can have once instance per node. Taking as an example the number of nodes you mentioned above and 2 peers, there will be 1000 instances instead of 2000.

design/crd-status.md
```diff
@@ -0,0 +1,316 @@
...
+fact that this CR has no relation with the existing ones, a valid approach is
+to follow what was done for the [metrics exporter](https://github.com/metallb/metallb/blob/main/frr-metrics/exporter.go)
+and have a different component (or even the exporter itself) polling FRR and filling
+the session status. The polling interval must be configurable and large enough to
+avoid impacts both on FRR and on the API.
+The speaker should continue not to have direct interactions with FRR.
+
+#### Service Announcement - BGP
+
+Given a service and a node, we want to expose the `BGPpeer`s the service is configured
```
mlguerrero12, line 208: I would prefer to have status in the bgp advertisement crd. For instance, you can list the nodes and the peers that the bgp adv is linked to, as well as, the services. In my opinion, this is the most important crd and what users expect to see.

Note that we can modify the controller to update the status. Not need to do it in each speaker.

design/crd-status.md
```diff
@@ -0,0 +1,316 @@
...
+        assignedIPV4  int
+        assignedIPV6  int
+        availableIPV4 int
+        availableIPV6 int
+}
+```
+
+#### Configuration Status
+
+Given that the configuration is composed by multiple CRs, there are few cases
```
mlguerrero12, line 108: I'm not sure about this new resource. I would add status in bgppeer and bgp adv only for errors. Note that we can modify the controller to update the status instead of doing in each speaker. So, not scalability concerns. For instance, the error you mention below can be catched by the controller and simply update the status on the affected bgp peer resource. Same for errors in bgp adv. We can even watch for node changes in the controller and run the same config that it is run in the speaker to detect issues with local preference for example.

fedepaol: The problem is, there's no straight way to say "this crd has an error". 
The configuration as a whole might be invalid, because of conflicting selectors for example.
Or because of the local pref thing that you added. And the "wrong one" depends on the order in which our controllers receive the various pieces, which could not be the same for each one.
Because of this, it's hard to say "the status of the peer is invalid" while it makes sense to say "the config as a whole is invalid because of this reason"

design/crd-status.md
fedepaol: What users will care about is: why my service is not working?
And, the configuration associated to a given service might come from the union of different bgp advertisements (i.e. multiple peers / multiple nodes / etc).
That's the reason why I think the service must be the key for this status.

Adding the status for all the services associated to a bgp advertisement will give a partial picture and will still make the troubleshooting hard to diagnose.

design/crd-status.md
fedepaol: yeah, that might work. My guess is we'll have a high number of nodes and a manageable number of peers.
This means that when a session (or two) flickers, we'll move the status related to the node, but the size should be manageable.

Moreover, we'll control how often we publish this, so it should be ok. @oribon thoughts?

design/crd-status.md
mlguerrero12: does it matter if the controllers in the speaker have different order in the cache? We only care about the "main" controller and also, it will be the same cache used by the webhook. We can also not return at the first error and simply report all conflict crds. At the end, all of them are faulty.

design/crd-status.md
mlguerrero12: but also here is incomplete. It is possible to see the nodes and peers but the user won't know where bgp advs are causing that configuration

design/crd-status.md
fedepaol: it matters because we can't say "this is right, this is wrong". You need to see this in a scenario where the webhooks are not present.
In a ideal world where the webhooks work fine, there would no be requirement for this feature. But you can throw all the manifests at a cluster, and they get processed by different threads. Or two crs together are invalid, but you apply them so fast that the webhook cache is not updated yet.
So there will always be scenarios where a faulty config goes through the webhooks.

And that's (all of them are faulty) is exactly why we can't say "this cr is invalid" but "the configuration is invalid".

design/crd-status.md
fedepaol: Then we could add the list of advs that contribute to a given config. But this version is still preferrable imo

mlguerrero12: what about having the 2 of them?

design/crd-status.md
mlguerrero12: I understand your point and also the fact that you don't want to say that a particular cr is invalid, but note that with the message we will also indirectly point to a crd which might not necessarily be the one causing the issue.

What I do agree, however, is that having all failures in a single place improves user experience.

design/crd-status.md
oribon: Not sure I follow, do you suggest here having a resource per node, and there specifying all of its peers + connection status?
If so wouldn't it (over)complicate querying how a given peer behaves with the nodes of the cluster compared to what suggested here (is it something we'd like to be simple?)
I'm not against it, but I feel that the current way "talks" better with the rest of the proposal

If you think the number of objects here play an important role then it sounds reasonable to condense them in a one-per-node object

design/crd-status.md
fedepaol: let's stick with the current proposal as it can be a starting point that we extend.
We can expand, but when we push an api it's really hard to get back.

design/crd-status.md
fedepaol: if a message is not clear, then the problem is on how we report the message. 
They all should be like "IPaddresspool A is not valid because the range overlaps with IPAddresspool B". If we missed some we'll fix them while they come.

design/crd-status.md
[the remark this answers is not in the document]
fedepaol: I am still not entirely sold on this. I'll add a section with "open items" in the bottom of the doc, which we can use as a starting point for the discussion.

design/crd-status.md
fedepaol: > Wouldn't it (over)complicate querying how a given peer behaves with the nodes of the cluster compared to what suggested here (is it something we'd like to be simple?) I'm not against it, but I feel that the current way "talks" better with the rest of the proposal
> 

Right, I did not remember this part :sweat_smile: 

> If you think the number of objects here play an important role then it sounds reasonable to condense them in a one-per-node object

I think both versions would be fine performance wise especially considering that we would sample the status anyway and that we would not update anything if the session keeps being stable.

This is probably (until now) the remaining pending discussion. Given both are more or less the same, but the original one would favor navigability, I am inclined to tend for that. Basically what we'd loose with the "all the peers for one node" is a way to say "give me the status of all the sessions for the peer xy".

@mlguerrero12 do you have reasons to say "an instance per node per peer" is too much? Is the concern about the number of objects?

mlguerrero12: yes, my concern is the number of objects and also from the user perspective, I think it is nicer to see in one place all sessions of a speaker instead of having to get each one of them separately

fedepaol: from the user's perspective it won't change. Because of the labels, they will still have a way to look at all the session from a given speaker selecting the right label.

The difference is, in the all-in-one it won't be possible to catch all the session for a given peer, and the amount of data moved when a session is flickering will be slightly more.
All in all, I feel that having one item representing a single session is more natural. The considerations made on endpointslices vs endpoints ([here](https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#motivation)) seem to tend to the fact that it's better to have smaller objects rather than a big one, so the number should not be a technical issue.

mlguerrero12: ok

oribon (approved):

oribon:
lgtm

PR #1692, opened 2022-11-08
Use file volume mount for memberlist secret key

liornoy:
This PR changes the way we insert the memberlist secret key to the speaker.
Switching from env var to a file in volume mount.

### DIFF
--- a/charts/metallb/policy/speaker.rego
+++ b/charts/metallb/policy/speaker.rego
@@ -11,15 +11,8 @@ deny[msg] {
 # validate METALLB_ML_SECRET_KEY (memberlist)
 deny[msg] {
 	input.kind == "DaemonSet"
-	not input.spec.template.spec.containers[0].env[5].name == "METALLB_ML_SECRET_KEY"
-	msg = "speaker env does not contain METALLB_ML_SECRET_KEY at env[5]"
-}
-
-deny[msg] {
-	input.kind == "DaemonSet"
-	not input.spec.template.spec.containers[0].env[5].valueFrom.secretKeyRef.name == "RELEASE-NAME-metallb-memberlist"
-	not input.spec.template.spec.containers[0].env[5].valueFrom.secretKeyRef.key == "secretkey"
-	msg = "speaker env METALLB_ML_SECRET_KEY secretKeyRef does not equal expected value"
+	not input.spec.template.spec.containers[0].env[5].name == "METALLB_ML_SECRET_KEY_PATH"
+	msg = "speaker env does not contain METALLB_ML_SECRET_KEY_PATH at env[5]"
 }
 
 # validate node selector includes builtin when custom ones are provided
--- a/charts/metallb/templates/speaker.yaml
+++ b/charts/metallb/templates/speaker.yaml
@@ -147,6 +147,12 @@ spec:
       terminationGracePeriodSeconds: 0
       hostNetwork: true
       volumes:
+      {{- if .Values.speaker.memberlist.enabled }}
+        - name: memberlist
+          secret:
+            secretName: {{ include "metallb.secretName" . }}
+            defaultMode: 420
+      {{- end }}
       {{- if .Values.speaker.frr.enabled }}
         - name: frr-sockets
           emptyDir: {}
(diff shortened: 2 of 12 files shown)


liornoy:
e2e-use-operator fails because the operator deploys metallb using YAMLs that inject env var instead of mounting a file.
so it's a chicken-egg situation, where we'll have to merge this and then right after sending a fix to the operator.

charts/metallb/templates/speaker.yaml
```diff
@@ -288,11 +292,15 @@ spec:
             - ALL
             add:
             - NET_RAW
-        {{- if .Values.speaker.frr.enabled }}
         volumeMounts:
+          {{- if .Values.speaker.memberlist.enabled }}
```
fedepaol, line 296: what if neither memberlist or frr are enabled?

charts/metallb/templates/speaker.yaml
```diff
@@ -147,6 +147,15 @@ spec:
...
       hostNetwork: true
       volumes:
+      {{- if .Values.speaker.memberlist.enabled }}
+        - name: memberlist
+          secret:
+            secretName: {{ include "metallb.secretName" . }}
+            defaultMode: 420
+            items:
+            - key: secretkey
+              path: metallb/speaker
```
fedepaol, line 157: just "key" as the rest is given by the full path

speaker/main.go
```diff
@@ -117,7 +116,13 @@ func main() {
 	}()
 	defer level.Info(logger).Log("op", "shutdown", "msg", "done")
 
-	sList, err := speakerlist.New(logger, *myNode, *mlBindAddr, *mlBindPort, *mlSecret, *namespace, *mlLabels, stopCh)
+	mlSecret, err := os.ReadFile("/etc/ml_secret_key/metallb/speaker")
```
fedepaol, line 119: Let's pass the path as a parameter

speaker/main.go
```diff
@@ -117,7 +116,13 @@ func main() {
...
+	if err != nil {
```
fedepaol, line 120: The secret is optional, so it's fine not to have it. By doing this we are making it mandatory

charts/metallb/templates/speaker.yaml
liornoy: done. added an OR statement above.

charts/metallb/templates/speaker.yaml
liornoy: done

speaker/main.go
liornoy: done

speaker/main.go
liornoy: done

liornoy:
(Updated speaker main.go - to include the const mlSecretKeyName = "secretkey"

charts/metallb/policy/speaker.rego
```diff
@@ -11,15 +11,14 @@ deny[msg] {
...
+	not input.spec.template.spec.containers[0].env[5].name == "METALLB_ML_SECRET_KEY_PATH"
+	msg = "speaker env does not contain METALLB_ML_SECRET_KEY_PATH at env[5]"
 }
 
 deny[msg] {
 	input.kind == "DaemonSet"
-	not input.spec.template.spec.containers[0].env[5].valueFrom.secretKeyRef.name == "RELEASE-NAME-metallb-memberlist"
-	not input.spec.template.spec.containers[0].env[5].valueFrom.secretKeyRef.key == "secretkey"
-	msg = "speaker env METALLB_ML_SECRET_KEY secretKeyRef does not equal expected value"
+	not input.spec.template.spec.containers[0].env[5].value == "/etc/ml_secret_key"
```
fedepaol, line 20: The value can be overridden, this is static, so if the user changes it this check will always fail.

speaker/main.go
```diff
@@ -61,13 +62,15 @@ type service interface {
 func main() {
 	prometheus.MustRegister(announcing)
 
+	const mlSecretKeyName = "secretkey"
```
fedepaol, line 65: Having this as a constant while it's used only in one place is not adding value. On the other hand, I think it makes sense to have it as a global constant in k8s.go and use it also there (where we create the secret).

charts/metallb/policy/speaker.rego
liornoy: you're right, removed.

speaker/main.go
liornoy: done! thanks

fedepaol:
LGTM, thanks!

PR #1693, opened 2022-11-08
Implement ip pool selector logic for service

pperiyasamy:
This changes ip allocation logic when ip pools are configured with namespace, service/namespace label selectors so that ip are allocated from desired pools based on its priority before it tries to allocate from other ip pools.

### DIFF
--- a/controller/main.go
+++ b/controller/main.go
@@ -42,7 +42,7 @@ type service interface {
 
 type controller struct {
 	client service
-	pools  map[string]*config.Pool
+	pools  *config.Pools
 	ips    *allocator.Allocator
 }
 
@@ -58,7 +58,7 @@ func (c *controller) SetBalancer(l log.Logger, name string, svcRo *v1.Service, _
 		return controllers.SyncStateReprocessAll
 	}
 
-	if c.pools == nil {
+	if c.pools == nil || c.pools.ByName == nil {
 		// Config hasn't been read, nothing we can do just yet.
 		level.Debug(l).Log("event", "noConfig", "msg", "not processing, still waiting for config")
 		return controllers.SyncStateSuccess
@@ -112,11 +112,11 @@ func (c *controller) deleteBalancer(l log.Logger, name string) {
 	}
 }
 
-func (c *controller) SetPools(l log.Logger, pools map[string]*config.Pool) controllers.SyncState {
+func (c *controller) SetPools(l log.Logger, pools *config.Pools) controllers.SyncState {
 	level.Debug(l).Log("event", "startUpdate", "msg", "start of config update")
 	defer level.Debug(l).Log("event", "endUpdate", "msg", "end of config update")
 
-	if pools == nil {
+	if pools == nil || pools.ByName == nil {
 		level.Error(l).Log("op", "setConfig", "error", "no MetalLB configuration in cluster", "msg", "configuration is missing, MetalLB will not function")
 		return controllers.SyncStateErrorNoRetry
 	}
--- a/controller/service.go
+++ b/controller/service.go
@@ -148,7 +148,7 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 	}
 
 	pool := c.ips.Pool(key)
-	if pool == "" || c.pools[pool] == nil {
+	if pool == "" || c.pools == nil || c.pools.IsEmpty(pool) {
 		level.Error(l).Log("bug", "true", "ip", lbIPs, "msg", "internal error: allocated IP has no matching address pool")
 		c.client.Errorf(svc, "InternalError", "allocated an IP that has no pool")
 		c.clearServiceState(key, svc)
@@ -214,7 +214,7 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error) 
 	}
 
 	// Okay, in that case just bruteforce across all pools.
-	return c.ips.Allocate(key, serviceIPFamily, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc))
+	return c.ips.Allocate(key, svc, serviceIPFamily, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc))
 }
 
 func (c *controller) isServiceAllocated(key string) bool {
--- a/internal/allocator/allocator.go
+++ b/internal/allocator/allocator.go
@@ -7,17 +7,20 @@ import (
 	"fmt"
 	"math"
 	"net"
+	"sort"
 	"strings"
 
 	"go.universe.tf/metallb/internal/config"
 	"go.universe.tf/metallb/internal/ipfamily"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/labels"
 
 	"github.com/mikioh/ipaddr"
 )
 
 // An Allocator tracks IP address pools and allocates addresses from them.
 type Allocator struct {
-	pools map[string]*config.Pool
+	pools *config.Pools
 
 	allocated       map[string]*alloc          // svc -> alloc
 	sharingKeyForIP map[string]*key            // ip.String() -> assigned sharing key
@@ -52,7 +55,7 @@ type alloc struct {
 // New returns an Allocator managing no pools.
 func New() *Allocator {
 	return &Allocator{
-		pools: map[string]*config.Pool{},
+		pools: &config.Pools{ByName: map[string]*config.Pool{}},
 
 		allocated:       map[string]*alloc{},
 		sharingKeyForIP: map[string]*key{},
@@ -63,19 +66,19 @@ func New() *Allocator {
 }
 
 // SetPools updates the set of address pools that the allocator owns.
-func (a *Allocator) SetPools(pools map[string]*config.Pool) error {
+func (a *Allocator) SetPools(pools *config.Pools) error {
 	// All the fancy sharing stuff only influences how new allocations
 	// can be created. For changing the underlying configuration, the
 	// only question we have to answer is: can we fit all allocated
 	// IPs into address pools under the new configuration?
 	for svc, alloc := range a.allocated {
-		if poolFor(pools, alloc.ips) == "" {
+		if poolFor(pools.ByName, alloc.ips) == "" {
 			return fmt.Errorf("new config not compatible with assigned IPs: service %q cannot own %q under new config", svc, alloc.ips)
 		}
 	}
 
-	for n := range a.pools {
-		if pools[n] == nil {
+	for n := range a.pools.ByName {
+		if pools.ByName[n] == nil {
 			stats.poolCapacity.DeleteLabelValues(n)
 			stats.poolActive.DeleteLabelValues(n)
 			stats.poolAllocated.DeleteLabelValues(n)
@@ -86,7 +89,7 @@ func (a *Allocator) SetPools(pools map[string]*config.Pool) error {
 
 	// Need to rearrange existing pool mappings and counts
 	for svc, alloc := range a.allocated {
-		pool := poolFor(a.pools, alloc.ips)
+		pool := poolFor(a.pools.ByName, alloc.ips)
 		if pool != alloc.pool {
 			a.Unassign(svc)
 			alloc.pool = pool
@@ -97,7 +100,7 @@ func (a *Allocator) SetPools(pools map[string]*config.Pool) error {
 	}
 
 	// Refresh or initiate stats
-	for n, p := range a.pools {
+	for n, p := range a.pools.ByName {
 		stats.poolCapacity.WithLabelValues(n).Set(float64(poolCount(p)))
 		stats.poolActive.WithLabelValues(n).Set(float64(len(a.poolIPsInUse[n])))
 	}
@@ -127,14 +130,14 @@ func (a *Allocator) assign(svc string, alloc *alloc) {
 		}
 		a.poolIPsInUse[alloc.pool][ip.String()]++
 	}
-	stats.poolCapacity.WithLabelValues(alloc.pool).Set(float64(poolCount(a.pools[alloc.pool])))
+	stats.poolCapacity.WithLabelValues(alloc.pool).Set(float64(poolCount(a.pools.ByName[alloc.pool])))
 	stats.poolActive.WithLabelValues(alloc.pool).Set(float64(len(a.poolIPsInUse[alloc.pool])))
 }
 
 // Assign assigns the requested ip to svc, if the assignment is
 // permissible by sharingKey and backendKey.
 func (a *Allocator) Assign(svc string, ips []net.IP, ports []Port, sharingKey, backendKey string) error {
-	pool := poolFor(a.pools, ips)
+	pool := poolFor(a.pools.ByName, ips)
 	if pool == "" {
 		return fmt.Errorf("%q is not allowed in config", ips)
 	}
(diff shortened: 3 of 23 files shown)


controller/service.go
```diff
@@ -213,10 +215,59 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error)
 		return ips, nil
 	}
 
+	pinnedPriorityPools, pinnedNoPriorityPools := c.getPinnedPools(svc)
```
fedepaol, line 218: I'd move all this logic to allocator's `Allocate()`

internal/config/config.go
```diff
@@ -470,6 +487,40 @@ func addressPoolFromCR(p metallbv1beta1.IPAddressPool) (*Pool, error) {
 		ret.cidrsPerAddresses[cidr] = nets
 	}
 
+	if p.Spec.AllocateTo != nil &&
```
fedepaol, line 490: Let's wrap in a function that returns the ServiceAllocation.
Also, return early if allocateTo is nil

internal/k8s/controllers/pool_controller.go
```diff
@@ -109,5 +117,6 @@ func (r *PoolReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		For(&metallbv1beta1.IPAddressPool{}).
 		Watches(&source.Kind{Type: &metallbv1beta1.AddressPool{}}, &handler.EnqueueRequestForObject{}).
 		Watches(&source.Kind{Type: &metallbv1beta1.Community{}}, &handler.EnqueueRequestForObject{}).
+		Watches(&source.Kind{Type: &corev1.Namespace{}}, &handler.EnqueueRequestForObject{}).
```
fedepaol, line 120: We need a predicate that discard events on the namespace that do not involve changing labels.

internal/k8s/controllers/config_controller.go
```diff
@@ -179,6 +186,7 @@ func (r *ConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
 		Watches(&source.Kind{Type: &metallbv1beta1.AddressPool{}}, &handler.EnqueueRequestForObject{}).
 		Watches(&source.Kind{Type: &metallbv1beta1.Community{}}, &handler.EnqueueRequestForObject{}).
 		Watches(&source.Kind{Type: &corev1.Secret{}}, &handler.EnqueueRequestForObject{}).
+		Watches(&source.Kind{Type: &corev1.Namespace{}}, &handler.EnqueueRequestForObject{}).
```
fedepaol, line 189: same, add a predicate here to avoid unnecessary events

controller/main.go
```diff
@@ -42,9 +42,11 @@ type service interface {
 }
 
 type controller struct {
-	client service
-	pools  map[string]*config.Pool
-	ips    *allocator.Allocator
+	client                 service
+	pools                  map[string]*config.Pool
```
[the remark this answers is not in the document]
pperiyasamy: sure, moved this higher level struct into config and consuming it in every places.

controller/service.go
pperiyasamy: yes, this is a cleaner solution. moved the logic into `Allocate()`.

internal/allocator/allocator.go
```diff
@@ -269,31 +272,84 @@ func (a *Allocator) AllocateFromPool(svc string, serviceIPFamily ipfamily.Family
...
-		if ips, err := a.AllocateFromPool(svc, serviceIPFamily, poolName, ports, sharingKey, backendKey); err == nil {
+		if ips, err := a.AllocateFromPool(key, serviceIPFamily, poolName, ports, sharingKey, backendKey); err == nil {
 			return ips, nil
 		}
 	}
 
 	return nil, errors.New("no available IPs")
 }
 
+func (a *Allocator) getPinnedPools(svc *v1.Service) ([]*config.Pool, []*config.Pool) {
```
fedepaol, line 312: Can we have this return a single sorted list with the ones with priority on top and the ones without priority after? That will make the caller easier I think

internal/allocator/allocator.go
```diff
@@ -269,31 +272,84 @@ func (a *Allocator) AllocateFromPool(svc string, serviceIPFamily ipfamily.Family
 }
 
 // Allocate assigns any available and assignable IP to service.
-func (a *Allocator) Allocate(svc string, serviceIPFamily ipfamily.Family, ports []Port, sharingKey, backendKey string) ([]net.IP, error) {
-	if alloc := a.allocated[svc]; alloc != nil {
-		if err := a.Assign(svc, alloc.ips, ports, sharingKey, backendKey); err != nil {
+func (a *Allocator) Allocate(key string, svc *v1.Service, serviceIPFamily ipfamily.Family, ports []Port, sharingKey, backendKey string) ([]net.IP, error) {
```
fedepaol, line 275: why we renamed svc -> key?

internal/allocator/allocator.go
```diff
@@ -269,31 +272,84 @@ func (a *Allocator) AllocateFromPool(svc string, serviceIPFamily ipfamily.Family
...
+		if err := a.Assign(key, alloc.ips, ports, sharingKey, backendKey); err != nil {
 			return nil, err
 		}
 		return alloc.ips, nil
 	}
 
-	for poolName := range a.pools {
-		if !a.pools[poolName].AutoAssign {
+	for poolName := range a.pools.PoolsByName {
+		if !a.pools.PoolsByName[poolName].AutoAssign || a.pools.PoolsByName[poolName].ServiceAllocations != nil {
```
fedepaol, line 301: Why do we need the double check of service allocations?

internal/config/config.go
```diff
@@ -471,9 +499,81 @@ func addressPoolFromCR(p metallbv1beta1.IPAddressPool) (*Pool, error) {
...
+		(len(p.Spec.AllocateTo.Namespaces) > 0 && len(p.Spec.AllocateTo.NamespaceSelectors) > 0 &&
+			len(p.Spec.AllocateTo.ServiceSelectors) > 0) {
+		return nil, nil
+	}
+	serviceAllocations := &ServiceAllocation{}
+	serviceAllocations.Priority = p.Spec.AllocateTo.Priority
+	if len(p.Spec.AllocateTo.Namespaces) > 0 || len(p.Spec.AllocateTo.NamespaceSelectors) > 0 {
+		serviceAllocations.Namespaces = sets.NewString()
+	}
+	for _, namespace := range p.Spec.AllocateTo.Namespaces {
```
fedepaol, line 522: Thinking out loudly as it may impact the implementation, but would it be more efficient if we propagate the namespace selector and pass the whole namespace object (or just its labels) to the allocator?
Here we iterate on all the namespaces every time there's a config change. Maybe it's not that much.

yuvalk:
hint for static-security-analysis,
see my fixes for older code here:
https://github.com/metallb/metallb/commit/b3e8e242da9cbb018e9b16cbd5392b866c146d52
and a nice explanation here:
https://medium.com/swlh/use-pointer-of-for-range-loop-variable-in-go-3d3481f7ffc9

pperiyasamy:
> hint for static-security-analysis,

Can  you also make this available from inv command via [tasks.py](https://github.com/metallb/metallb/blob/main/tasks.py) ?

controller/service.go
```diff
@@ -148,7 +148,7 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 	}
 
 	pool := c.ips.Pool(key)
-	if pool == "" || c.pools[pool] == nil {
+	if pool == "" || c.pools == nil || c.pools.ByName == nil {
```
fedepaol, line 151: can we add an `IsEmpty` method to `*config.Pools` and use it?

internal/allocator/allocator.go
```diff
@@ -269,31 +272,80 @@ func (a *Allocator) AllocateFromPool(svc string, serviceIPFamily ipfamily.Family
...
+	if alloc := a.allocated[svcKey]; alloc != nil {
+		if err := a.Assign(svcKey, alloc.ips, ports, sharingKey, backendKey); err != nil {
 			return nil, err
 		}
 		return alloc.ips, nil
 	}
 
-	for poolName := range a.pools {
-		if !a.pools[poolName].AutoAssign {
+	for poolName := range a.pools.ByName {
```
fedepaol, line 289: can we iterate over pools (instead that over the keys), so we won't  need to dereference the map again?

internal/allocator/allocator.go
```diff
@@ -269,31 +272,80 @@ func (a *Allocator) AllocateFromPool(svc string, serviceIPFamily ipfamily.Family
...
 		}
 	}
 
 	return nil, errors.New("no available IPs")
 }
 
+// This method returns sorted ip pools which are allocatable for given service.
+// When ip pool is not set with priority, then just append it after sorted
+// priority ip pools.
+func (a *Allocator) getSortedPinnedPools(svc *v1.Service) []*config.Pool {
```
fedepaol, line 304: I'd call this `sortedPoolsForService`, no need for the `get`

internal/k8s/controllers/config_controller.go
```diff
@@ -178,10 +173,42 @@ func (r *ConfigReconciler) SetupWithManager(mgr ctrl.Manager) error {
...
+		return true
+	}
+	if labels.Equals(labels.Set(oldNodeObj.Labels), labels.Set(newNodeObj.Labels)) {
+		return false
+	}
+	return true
+}
+
+func filterNamespaceEvent(e event.UpdateEvent) bool {
+	newNodeObj, ok := e.ObjectNew.(*corev1.Namespace)
```
fedepaol, line 197: nit: newNamespaceObj (and old)

e2etest/bgptests/bgp.go
```diff
@@ -1718,6 +1718,290 @@ var _ = ginkgo.Describe("BGP", func() {
 		table.Entry("IPV6", "fc00:f853:0ccd:e799::/116", "fc00:f853:ccd:e800::1/128", ipfamily.IPv6),
 	)
 
+	ginkgo.Context("IPV4 - Validate service allocation in address pools", func() {
```
fedepaol, line 1721: I think the right place for this is under l2/assignement.go

e2etest/bgptests/bgp.go
```diff
@@ -1718,6 +1718,290 @@ var _ = ginkgo.Describe("BGP", func() {
...
+
+			for _, c := range FRRContainers {
+				validateService(cs, svc1, allNodes.Items, c)
+				validateService(cs, svc2, allNodes.Items, c)
+				validateService(cs, svc3, allNodes.Items, c)
+			}
+		})
+	})
+
+	ginkgo.Context("IPV4 - Validate service allocation in address pools", func() {
```
fedepaol, line 1810: we can leverage the wrapping context I think, no need to add a new one

e2etest/bgptests/bgp.go
```diff
@@ -1718,6 +1718,290 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			svc2, _ := testservice.CreateWithBackend(cs, f.Namespace.Name, "external-local-lb-2")
+			svc3, _ := testservice.CreateWithBackend(cs, f.Namespace.Name, "external-local-lb-3")
+			defer func() {
+				testservice.Delete(cs, svc1)
+				testservice.Delete(cs, svc2)
+				testservice.Delete(cs, svc3)
+			}()
+
+			framework.ExpectEqual(f.Namespace.Name, f.Namespace.Name)
+			for _, i := range svc1.Status.LoadBalancer.Ingress {
```
fedepaol, line 1783: can you add a comment about the fact that this is deterministic because createWithBackend always wait for the service to get an ip?

e2etest/bgptests/bgp.go
```diff
@@ -1718,6 +1718,290 @@ var _ = ginkgo.Describe("BGP", func() {
...
+					{
+						ObjectMeta: metav1.ObjectMeta{Name: "bgp-test-3"},
+						Spec: metallbv1beta1.IPAddressPoolSpec{
+							Addresses: []string{
+								"192.168.20.0/32",
+							},
+						},
+					},
+				},
+				Peers:   metallb.PeersForContainers(FRRContainers, ipfamily.IPv4),
```
fedepaol, line 1757: I would avoid the peers / advertisement as we care only about the allocation

e2etest/bgptests/bgp.go
```diff
@@ -1718,6 +1718,290 @@ var _ = ginkgo.Describe("BGP", func() {
...
+				BGPAdvs: []metallbv1beta1.BGPAdvertisement{emptyBGPAdvertisement},
+			}
+
+			for _, c := range FRRContainers {
+				err := frrcontainer.PairWithNodes(cs, c, ipfamily.IPv4)
```
fedepaol, line 1762: same, no need to configure frr nor to check the connectivity

e2etest/bgptests/bgp.go
```diff
@@ -1718,6 +1718,290 @@ var _ = ginkgo.Describe("BGP", func() {
...
+				ginkgo.By("validate LoadBalancer IP is allocated from 1st higher priority address pool")
+				ingressIP := e2eservice.GetIngressPoint(&i)
+				framework.ExpectEqual(ingressIP, "192.168.10.0")
+			}
+			for _, i := range svc2.Status.LoadBalancer.Ingress {
+				ginkgo.By("validate LoadBalancer IP is allocated from 2nd higher priority address pool")
+				ingressIP := e2eservice.GetIngressPoint(&i)
+				framework.ExpectEqual(ingressIP, "192.168.5.0")
+			}
+			for _, i := range svc3.Status.LoadBalancer.Ingress {
```
fedepaol, line 1793: since we are doing this extensively, I'd wrap it in a "serviceIPIsFromPool` function

e2etest/bgptests/bgp.go
```diff
@@ -1718,6 +1718,290 @@ var _ = ginkgo.Describe("BGP", func() {
...
+		ginkgo.AfterEach(func() {
+			// Clean previous configuration.
+			err := ConfigUpdater.Clean()
+			framework.ExpectNoError(err)
+		})
+		ginkgo.It("with namespace and service label", func() {
+			resources := metallbconfig.ClusterResources{
+				Pools: []metallbv1beta1.IPAddressPool{
+					{
+						ObjectMeta: metav1.ObjectMeta{Name: "bgp-test-1"},
```
fedepaol, line 1918: This is making me think: what happens if we have a pool with a namespace selector and a label selector? They are or-ed right?

internal/allocator/allocator.go
```diff
@@ -269,31 +272,80 @@ func (a *Allocator) AllocateFromPool(svc string, serviceIPFamily ipfamily.Family
...
+	if alloc := a.allocated[svcKey]; alloc != nil {
+		if err := a.Assign(svcKey, alloc.ips, ports, sharingKey, backendKey); err != nil {
 			return nil, err
 		}
 		return alloc.ips, nil
 	}
 
-	for poolName := range a.pools {
-		if !a.pools[poolName].AutoAssign {
+	for poolName, pool := range a.pools.ByName {
```
fedepaol, line 289: at this point I'd get rid of poolName and use pool.Name below

controller/service.go
```diff
@@ -148,7 +148,7 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 	}
 
 	pool := c.ips.Pool(key)
-	if pool == "" || c.pools[pool] == nil {
+	if c.isEmpty(pool) {
```
fedepaol, line 151: I meant a method of pools, so we can call 
`if pool == "" || c.pools.IsEmpty()`

internal/config/config.go
```diff
@@ -170,6 +196,10 @@ type BFDProfile struct {
 	MinimumTTL       *uint32
 }
 
+func (p *Pools) IsEmpty() bool {
+	return p.ByName == nil
```
fedepaol, line 200: we should check also for p being nil here (it will work)

fedepaol:
LGTM!

PR #1700, opened 2022-11-11
WIP: E2E tests: add VRF support

fedepaol:
Here we change the e2e tests infrastructure to:
- add a new docker network to the nodes
- wrap the interface corresponding to that docker network in a vrf named "red"
- create a new container reacheable only from that network

Then, we refactor the infra_setup logic in order to split it in pieces that return the related containers instead of having a single entry point.

This does not extend the tests or add VRF support to metallb, but lays the ground to test and validate it.

Note: this PR is marked as WIP only because MetalLB is not tested e2e

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -167,6 +167,11 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v3
 
+      - name: Install kernel modules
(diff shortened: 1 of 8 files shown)


fedepaol:
Closing in favor of https://github.com/metallb/metallb/pull/1719

PR #1702, opened 2022-11-11
Added docs note about BGP advertisement

SlothCroissant:
This threw me for a loop while migrating from L2 to BGP in my setup, so want to give others a heads up to not make the same mistakes I did :)

   accepted.
   the PR is merged.

### DIFF
--- a/website/content/configuration/_index.md
+++ b/website/content/configuration/_index.md
@@ -187,6 +187,11 @@ spec:
   - first-pool
 ```
 
+{{% notice note %}}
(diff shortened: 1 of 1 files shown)


fedepaol:
Hi! Sorry for the delay, and thanks for the contribution. What was the expectation? Or, why were you expecting metallb to advertise routes with no services? MetalLB's purpouse is to advertise services, so I'd naturally expect no routes to be advertised if there are no services.

I am closing the PR because I don't feel the clarificatoin is needed, but feel free to expand with your considerations.

PR #1709, opened 2022-11-14
CI: move the api doc consistency check to the website lane

fedepaol:
Changes in the automatically generated api docs do not trigger the verification of consistency, because they live under `website` while the api live under `api`.

Because of this, we move the api check under the website lane and we trigger it also when a change under `api` occurs.


website/content/apis/_index.md
```diff
@@ -969,3 +969,4 @@ BGPPeerStatus
 Generated with <code>gen-crd-api-reference-docs</code>
 .
 </em></p>
+dsadasdsa
```
liornoy, line 972: Looks like the site build lane caught it, plus the CI lane works as intended.
LGTM!

PR #1713, opened 2022-11-16
Introduce gosec as Static Code Analysis tool

### LINKED ISSUE #1712: Add gosec to the ci pipeline
Adding another static analyzer should be easy and will save us from audits

yuvalk:
I've added gosec to the CI job
then went over the results and fixed all the high/medium issues.

for now, I've left LOW out of scanner result as this seems to include too many false positives and the benefit from investing in these is .. low..

this fixes #1712

### DIFF
--- a/configmaptocrs/main.go
+++ b/configmaptocrs/main.go
@@ -56,7 +56,11 @@ func main() {
 			log.Fatalf("failed to create output file: %s", err)
 		}
 	}
-	defer f.Close()
+	defer func() {
+		if tmpErr := f.Close(); tmpErr != nil {
+			err = tmpErr
+		}
+	}()
 
 	err = generate(f, *source)
 	if err != nil {
@@ -114,7 +118,11 @@ func readConfig(origin string) ([]byte, error) {
 	if err != nil {
 		return nil, fmt.Errorf("failed to open file %s: %v", fp, err)
 	}
-	defer f.Close()
+	defer func() {
+		if tmpErr := f.Close(); tmpErr != nil {
+			err = tmpErr
+		}
+	}()
 
 	raw, err := io.ReadAll(f)
 	if err != nil {
--- a/configmaptocrs/main.go
+++ b/configmaptocrs/main.go
@@ -8,8 +8,10 @@ import (
 	"io"
 	"log"
 	"os"
+	"path"
 	"path/filepath"
 	"sort"
+	"strings"
 	"time"
 
 	"go.universe.tf/metallb/api/v1beta1"
@@ -104,7 +106,11 @@ func generate(w io.Writer, origin string) error {
 
 func readConfig(origin string) ([]byte, error) {
 	fp := filepath.Join(inputDirPath, origin)
-	f, err := os.Open(fp)
+	fp = filepath.Clean(fp)
+	if !strings.HasPrefix(fp, path.Clean(inputDirPath)) {
+		return nil, fmt.Errorf("Unsafe path %s", origin)
+	}
+	f, err := os.Open(filepath.Clean(fp)) // Clean have to happen here to avoid https://github.com/securego/gosec/issues/893
 	if err != nil {
 		return nil, fmt.Errorf("failed to open file %s: %v", fp, err)
 	}
--- a/frr-metrics/exporter.go
+++ b/frr-metrics/exporter.go
@@ -8,6 +8,7 @@ import (
 	stdlog "log"
 	"net/http"
 	"os"
+	"time"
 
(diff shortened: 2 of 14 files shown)


configmaptocrs/main.go
```diff
@@ -105,7 +107,11 @@ func generate(w io.Writer, origin string) error {
 
 func readConfig(origin string) ([]byte, error) {
 	fp := filepath.Join(inputDirPath, origin)
-	f, err := os.Open(fp)
+	fp = filepath.Clean(fp)
+	if !strings.HasPrefix(fp, path.Clean(inputDirPath)) {
```
fedepaol, line 115: we do we need this? fp is the join of inputdir and origin, so it will have inputDirPath as prefix

configmaptocrs/main.go
```diff
@@ -9,8 +9,10 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"path"
```
fedepaol, line 12: What is this commit fixing? Or, what is G304 ?

frr-metrics/exporter.go
```diff
@@ -8,6 +8,7 @@ import (
 	stdlog "log"
 	"net/http"
 	"os"
+	"time"
```
fedepaol, line 11: Why is this required  / raised by gosec?

internal/config/config.go
```diff
@@ -384,6 +384,7 @@ func peerFromCR(p metallbv1beta2.BGPPeer, passwordSecrets map[string]corev1.Secr
 
 	var nodeSels []labels.Selector
 	for _, s := range p.Spec.NodeSelectors {
+		s := s // so we can use &s
```
fedepaol, line 387: Neat. This is not causing bugs as we are using &s in pure functions and not carrying it around, but might bite us in the future.

fedepaol:
Thanks for the contribution! Left a few requests for clarification.

frr-metrics/exporter.go
yuvalk: see below, without a ReadTimeout, the defulat is infinite,
which make it easy for denial-of-service (DoS) attacks

configmaptocrs/main.go
yuvalk: I can add a better commit msg
https://securego.io/docs/rules/g304.html

configmaptocrs/main.go
yuvalk: actually no
I had to do it to workaround https://github.com/securego/gosec/issues/893
but still, IMHO, is a better option than masking it (with a //#nosec comment)

yuvalk: again, I can add a better comment/commit msg so it'll be properly documented

configmaptocrs/main.go
```diff
@@ -105,7 +107,11 @@ func generate(w io.Writer, origin string) error {
 
 func readConfig(origin string) ([]byte, error) {
 	fp := filepath.Join(inputDirPath, origin)
-	f, err := os.Open(fp)
+	fp = filepath.Clean(fp)
+	if !strings.HasPrefix(fp, path.Clean(inputDirPath)) {
+		return nil, fmt.Errorf("Unsafe path %s", origin)
+	}
+	f, err := os.Open(filepath.Clean(fp))
```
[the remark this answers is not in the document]
yuvalk: actually no
I had to do it to workaround https://github.com/securego/gosec/issues/893
but still, IMHO, is a better option than masking it (with a //#nosec comment)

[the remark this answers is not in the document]
yuvalk: again, I can add a better comment/commit msg so it'll be properly documented

configmaptocrs/main.go
yuvalk: no, consider the case
inputDirPath = "/var/input"
and 
origin = "../../etc/passwd"

the join would be "/var/input/../../etc/passwd"
and the clean is "/etc/passwd" 
which doesnt have the inputDirPath prefix.

for (hopefully) obvious reasons, we want to block such cases.

fedepaol:
Would you mind fixing the conflicts?

internal/config/config.go
yuvalk: yeah, reinforcing the code :-)
could argue the other way too (adding bullshit, making unnecessary copy, etc etc)
but it's the easiest way for the code scanner to see there are no issues here

yuvalk:
> Would you mind fixing the conflicts?

I think this is all set now

fedepaol:
LGTM, thanks!

PR #1714, opened 2022-11-16
Refactor the frr configuration template

fedepaol:
Removing duplication and isolating separate parts in separate subtemplates to make it a bit easier to navigate it.

We also change the way the prefixes are being added, because the previous version was misleading in how the configuration was rendered, making the reader think that the prefixes are advertised per neighbor while they are per router (and we have per neighbor specific filters to choose what prefixes should go to a given neighbor).

Fixes https://github.com/metallb/metallb/issues/1715


fedepaol:
Fixes https://github.com/metallb/metallb/issues/1715

e2etest/bgptests/bgp.go
```diff
@@ -344,16 +344,6 @@ var _ = ginkgo.Describe("BGP", func() {
...
-						err = metrics.ValidateCounterValue(metrics.GreaterThan(0), "metallb_bgp_notifications_sent", map[string]string{"peer": addr}, speakerMetrics)
-						if err != nil {
-							return err
-						}
-
-						err = metrics.ValidateOnPrometheus(promPod, fmt.Sprintf(`metallb_bgp_notifications_sent{peer="%s"} >= 1`, addr), metrics.There)
-						if err != nil {
-							return err
-						}
-
```
liornoy, line 356: How is it non-deterministic?
When we configure the bgp peers and advertise the ip don't we always get bgp notification sent?
I don't remember seeing flakes here.

fedepaol: Before we always had notifications because of this https://github.com/metallb/metallb/issues/1715
With this new layout, we don't have that change anymore, so the notification is not sent in that scenario (while before, it was ALWAYS been sent), but it still sent _sometimes_.

internal/bgp/frr/config.go
```diff
@@ -264,14 +133,27 @@ func templateConfig(data interface{}) (string, error) {
 				}
 				return false
 			},
-		}).Parse(configTemplate)
+			"dict": func(values ...interface{}) (map[string]interface{}, error) {
+				if len(values)%2 != 0 {
+					return nil, errors.New("invalid dict call")
```
oribon, line 138: nit: wdyt about having the err be a little bit more detailed?  something like "dict expects even amount of values, got %d"?

internal/bgp/frr/config.go
```diff
@@ -264,14 +133,27 @@ func templateConfig(data interface{}) (string, error) {
...
+				}
+				dict := make(map[string]interface{}, len(values)/2)
+				for i := 0; i < len(values); i += 2 {
+					key, ok := values[i].(string)
+					if !ok {
+						return nil, errors.New("dict keys must be strings")
```
oribon, line 144: nit: same here, something like "got a non-string key: %v %T"

internal/bgp/frr/config.go
```diff
@@ -264,14 +133,27 @@ func templateConfig(data interface{}) (string, error) {
...
+					}
+					dict[key] = values[i+1]
+				}
+				return dict, nil
+			},
+		}).ParseFS(templates, "templates/*")
```
oribon, line 150: neat

oribon (commented):
minor nits, other than that looks good :)

internal/bgp/frr/config.go
fedepaol: Won't hurt (even though this kind of mistake will be easily caught by CI), will add

internal/bgp/frr/config.go
fedepaol: Done

internal/bgp/frr/testdata/TestSingleAdvertisementChange.golden
```diff
@@ -30,6 +32,14 @@ router bgp 100
...
     neighbor 10.2.2.254 activate
     neighbor 10.2.2.254 route-map 10.2.2.254-in in
-    network 172.16.1.11/24
     neighbor 10.2.2.254 route-map 10.2.2.254-out out
   exit-address-family
+  address-family ipv6 unicast
+    neighbor 10.2.2.254 activate
+    neighbor 10.2.2.254 route-map 10.2.2.254-in in
+    neighbor 10.2.2.254 route-map 10.2.2.254-out out
+  exit-address-family
```
liornoy, line 41: From looking at the `TestSingleAdvertisementChange` test It shouldn't have an ipv6 family here I think

internal/bgp/frr/testdata/TestSingleAdvertisement.golden
```diff
@@ -45,6 +47,14 @@ router bgp 100
...
     neighbor 10.2.2.254 activate
     neighbor 10.2.2.254 route-map 10.2.2.254-in in
-    network 172.16.1.10/24
     neighbor 10.2.2.254 route-map 10.2.2.254-out out
   exit-address-family
+  address-family ipv6 unicast
+    neighbor 10.2.2.254 activate
+    neighbor 10.2.2.254 route-map 10.2.2.254-in in
+    neighbor 10.2.2.254 route-map 10.2.2.254-out out
+  exit-address-family
```
liornoy, line 56: From looking at the `TestSingleAdvertisement` test, I think we shouldn't have this here.

internal/bgp/frr/testdata/TestSingleAdvertisementNoRouterID.golden
```diff
@@ -29,6 +31,14 @@ router bgp 100
...
     neighbor 10.2.2.254 activate
     neighbor 10.2.2.254 route-map 10.2.2.254-in in
-    network 172.16.1.10/24
     neighbor 10.2.2.254 route-map 10.2.2.254-out out
   exit-address-family
+  address-family ipv6 unicast
+    neighbor 10.2.2.254 activate
+    neighbor 10.2.2.254 route-map 10.2.2.254-in in
+    neighbor 10.2.2.254 route-map 10.2.2.254-out out
+  exit-address-family
```
liornoy, line 40: Same as above.

internal/bgp/frr/testdata/TestSingleAdvertisementWithPeerSelector.golden
```diff
@@ -45,6 +47,14 @@ router bgp 100
...
     neighbor 10.2.2.254 activate
     neighbor 10.2.2.254 route-map 10.2.2.254-in in
-    network 172.16.1.10/24
     neighbor 10.2.2.254 route-map 10.2.2.254-out out
   exit-address-family
+  address-family ipv6 unicast
+    neighbor 10.2.2.254 activate
+    neighbor 10.2.2.254 route-map 10.2.2.254-in in
+    neighbor 10.2.2.254 route-map 10.2.2.254-out out
+  exit-address-family
```
liornoy, line 56: Same as above.

oribon (approved):
lgtm

internal/bgp/frr/testdata/TestSingleAdvertisement.golden
liornoy: **Call the family:= assignment**
https://github.com/fedepaol/metallb/blob/b7ed3c93898d17165af9a3439ece45eb0c103673/internal/bgp/frr/frr.go#L272

**switch family:**
https://github.com/fedepaol/metallb/blob/b7ed3c93898d17165af9a3439ece45eb0c103673/internal/bgp/frr/frr.go#L292

liornoy (commented):
Left some comments after taking a closer look at the .golden files

internal/bgp/frr/testdata/TestTwoSessionsDuplicateRouter.golden
```diff
@@ -42,21 +44,21 @@ router bgp 100
...
   address-family ipv4 unicast
     neighbor 10.4.4.255 activate
     neighbor 10.4.4.255 route-map 10.4.4.255-in in
-	neighbor 10.4.4.255 route-map 10.4.4.255-out out
+    neighbor 10.4.4.255 route-map 10.4.4.255-out out
   exit-address-family
   address-family ipv6 unicast
     neighbor 10.4.4.255 activate
     neighbor 10.4.4.255 route-map 10.4.4.255-in in
-	neighbor 10.4.4.255 route-map 10.4.4.255-out out
```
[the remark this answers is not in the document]
fedepaol: That was a mistake that has been fixed here. It was working but the file was rendered poorly

internal/bgp/frr/testdata/TestSingleAdvertisementChange.golden
fedepaol: This the only effective change introduced by this PR. We always add the peers to both IPv4 and IPv6 families instead of adding to both if there are no advertisements, and adding them only to the family where we have advertisements in case at least one advertisement is present (which causes the route flapping), as explained in https://github.com/metallb/metallb/issues/1715

internal/bgp/frr/testdata/TestSingleAdvertisement.golden
fedepaol: same

internal/bgp/frr/testdata/TestSingleAdvertisementNoRouterID.golden
fedepaol: same

internal/bgp/frr/testdata/TestSingleAdvertisementWithPeerSelector.golden
fedepaol: same

internal/bgp/frr/testdata/TestTwoAdvertisementsTwoSessions.golden
```diff
@@ -76,25 +86,28 @@ router bgp 100
...
     neighbor 10.2.2.255 activate
     neighbor 10.2.2.255 route-map 10.2.2.255-in in
-    network 172.16.1.11/24
     neighbor 10.2.2.255 route-map 10.2.2.255-out out
   exit-address-family
+  address-family ipv4 unicast
+    network 172.16.1.10/24
+    network 172.16.1.10/24
+    network 172.16.1.11/24
+    network 172.16.1.11/24
```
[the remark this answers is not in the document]
fedepaol: nope, I can improve it

internal/bgp/frr/frr.go
```diff
@@ -277,15 +278,30 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
...
 				LocalPref:   adv.LocalPref,
 			}
 
 			neighbor.Advertisements = append(neighbor.Advertisements, &advConfig)
+			if _, ok := routerAnnouncements[prefix]; !ok {
+				switch family {
+				case ipfamily.IPv4:
+					router.IPV4Prefixes = append(router.IPV4Prefixes, prefix)
+				case ipfamily.IPv6:
+					router.IPV6Prefixes = append(router.IPV6Prefixes, prefix)
```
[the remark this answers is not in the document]
fedepaol: There's the if that was supposed to prevent dupes. Need to check why it didn't work

internal/bgp/frr/testdata/TestSingleAdvertisementChange.golden
liornoy: ok, now I get it. thanks,

fedepaol:
So, this turned out to be the occasion also to refactor the fact that we have maps we iterate on and so the rendering is not always consistent. I still wonder why we never hit this in CI....

internal/bgp/frr/frr.go
```diff
@@ -288,22 +299,26 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
...
-					router.IPV4Prefixes = append(router.IPV4Prefixes, prefix)
-				case ipfamily.IPv6:
-					router.IPV6Prefixes = append(router.IPV6Prefixes, prefix)
-				}
-				routerAnnouncements[prefix] = struct{}{}
+			switch family {
+			case ipfamily.IPv4:
+				rout.ipV4Prefixes[prefix] = prefix
+			case ipfamily.IPv6:
+				rout.ipV6Prefixes[prefix] = prefix
```
liornoy, line 306: Can't we make `rout.ipV4Prefixes` and `rout.ipV6Prefixes` to be slices?
I don't think maps are required here.

liornoy: On top of that, I think that we should refactor the whole createConfig() function in the sense
of separating into smaller functions.

fedepaol: The maps are to avoid duplicates

fedepaol: > On top of that, I think that we should refactor the whole createConfig() function in the sense of separating into smaller functions.

I am not sure it would be better. I kind of like the fact the rendering mechanism is clear. The lenght is mostly because of literals that take multiple lines.

liornoy: Extracting to separate, smaller, and more atomic functions can only improve clarity IMO.
But anyways, I see you addressed all the comments I wrote above and it's LGTM from me. :)

fedepaol: I am generally in favour of hiding complexity behind funcitons with meaningful names, but long function = not clear is not always necessarily true. If you need to jump between the loop and the subfunctions to understand the behaviour, the clarity is less.

This function is easy to follow, it's long but as I wrote, the biggest reason for the length is the literal initializiatoins that consume multiple rows.

PR #1721, opened 2022-11-25
Hide the bgp password from logs

fedepaol:
We expose the password in the logs in two places:
- when the speaker container dumps the configuration, because an error happened or because the log level is raised to debug
- in the reloader container, every time an frr configuration gets generated

Here, in both places we replace the password before logging.

### DIFF
--- a/frr-reloader/frr-reloader.sh
+++ b/frr-reloader/frr-reloader.sh
@@ -1,4 +1,7 @@
 #!/bin/bash
+
+set -o pipefail
+
 cleanup() {
   echo "Caught an exit signal.."
   clean_files
@@ -14,14 +17,14 @@ reload_frr() {
   kill_sleep
 
   echo "Checking the configuration file syntax"
-  if ! python3 /usr/lib/frr/frr-reload.py --test --stdout "$FILE_TO_RELOAD" ; then
+  if ! python3 /usr/lib/frr/frr-reload.py --test --stdout "$FILE_TO_RELOAD" 2>&1 | sed 's/password.*/password <retracted>/g'; then
     echo "Syntax error spotted: aborting.. $SECONDS seconds"
     echo -n "$(date +%s) failure"  > "$STATUSFILE"
     return
   fi
 
   echo "Applying the configuration file"
-  if ! python3 /usr/lib/frr/frr-reload.py --reload --overwrite --stdout "$FILE_TO_RELOAD" ; then
+  if ! python3 /usr/lib/frr/frr-reload.py --reload --overwrite --stdout "$FILE_TO_RELOAD" 2>&1 | sed 's/password.*/password <retracted>/g'; then
     echo "Failed to fully apply configuration file $SECONDS seconds"
     echo -n "$(date +%s) failure"  > "$STATUSFILE"
     return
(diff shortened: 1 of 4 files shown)


cgoncalves:
I'm of the opinion that logging the password when debug is on is useful when troubleshooting (only reason why one would enable debug)

frr-reloader/frr-reloader.sh
```diff
@@ -14,14 +14,14 @@ reload_frr() {
   kill_sleep
 
   echo "Checking the configuration file syntax"
-  if ! python3 /usr/lib/frr/frr-reload.py --test --stdout "$FILE_TO_RELOAD" ; then
+  if ! python3 /usr/lib/frr/frr-reload.py --test --stdout "$FILE_TO_RELOAD" 2>&1 | sed 's/password.*/password xxxxx/g'; then
```
yuvalk, line 17: `password <retracted>`
is more descriptive of what happend here IMHO

yuvalk:
> I'm of the opinion that logging the password when debug is on is useful when troubleshooting (only reason why one would enable debug)

you can read the secret if what you need is (only) the password.
if you need the configuration, u can read it from debug log, and if you want to try it, you need to add the password in.

but getting the password from debug log is wrong.

consider the fact that logs can be stored on unencrypted disks and even in backups as well as (most common!) shipped to aggregator tools such as logstash/ splunk/ etc. so the opportunities of unauthorized access and data leak are growing.
best practices for secure coding is to treat passwords as secrets and as such mask them in logs. (or more generally - touch them as less as possible, dont carry them around, dont store them, etc.)

fedepaol:
> > I'm of the opinion that logging the password when debug is on is useful when troubleshooting (only reason why one would enable debug)
> 
> you can read the secret if what you need is (only) the password. if you need the configuration, u can read it from debug log, and if you want to try it, you need to add the password in.
> 
> but getting the password from debug log is wrong.
> 
> consider the fact that logs can be stored on unencrypted disks and even in backups as well as (most common!) shipped to aggregator tools such as logstash/ splunk/ etc. so the opportunities of unauthorized access and data leak are growing. best practices for secure coding is to treat passwords as secrets and as such mask them in logs. (or more generally - touch them as less as possible, dont carry them around, dont store them, etc.)

This, plus the fact that we can still rely on `vtysh show running-config` if we suspect the passwords are wrong / there's a drift between the configured password and the expected one make me tend to keep the current version

oribon (approved):
lgtm

frr-reloader/frr-reloader.sh
```diff
@@ -14,14 +14,14 @@ reload_frr() {
   kill_sleep
 
   echo "Checking the configuration file syntax"
-  if ! python3 /usr/lib/frr/frr-reload.py --test --stdout "$FILE_TO_RELOAD" ; then
+  if ! python3 /usr/lib/frr/frr-reload.py --test --stdout "$FILE_TO_RELOAD" 2>&1 | sed 's/password.*/password <retracted>/g'; then
```
freedge, line 17: I believe the if might take into account the return code of the sed and not the PIPESTATUS of the python3 command (I could be wrong!)

fedepaol: this is a valid point I think. Thanks for the comment, let me dig more!

fedepaol: added set -o pipefail , thanks again for the catch!

PR #1726, opened 2022-11-29
L2: avoid having one failure on responder disable l2

fedepaol:
If a responder fails to be set (i.e. ndp on a ipv4 only node), then the list of interfaces is not being built.
That results in l2 mode entirely not working. Here we make the mechanism a bit more robust by continuing on errors.

Fixes https://github.com/metallb/metallb/issues/1727


oribon (approved):
lgtm

PR #1730, opened 2022-12-01
Allow to have different ASNs per VRF on metallb side

fedepaol:
When using VRFs it's allowed to have different asns per vrf.


oribon (approved):
lgtm

PR #1732, opened 2022-12-02
Add a liveness probe to the frr container

fedepaol:
In order to make sure that frr is up and running, we add a liveness probe to check that the daemons are available.


config/manifests/metallb-frr.yaml
```diff
@@ -1927,6 +1927,12 @@ spec:
         - name: TINI_SUBREAPER
           value: "true"
         image: frrouting/frr:v7.5.1
+        livenessProbe:
+          exec:
+            command:
+            - /etc/frr_liveness/liveness.sh
+          initialDelaySeconds: 5
```
oribon, line 1934: this seems a bit short ( as a guess ), is there a reason you went with 5 seconds?
also a startupprobe might be more fitting, wdyt?

fedepaol: yeah, it makes sense. Will change

fedepaol: done

config/frr/speaker-patch.yaml
```diff
@@ -78,6 +89,16 @@ spec:
...
               tail -f /etc/frr/frr.log
+          livenessProbe:
+            exec:
+              command: ["/etc/frr_liveness/liveness.sh"]
+            periodSeconds: 5
+            failureThreshold: 3
+          startupProbe:
+            exec:
+              command: ["/etc/frr_liveness/liveness.sh"]
+            failureThreshold: 30
```
liornoy, line 100: is it 30 on purpose here? or should it be 3?

fedepaol: it's on purpouse

charts/metallb/templates/speaker.yaml
```diff
@@ -341,6 +352,16 @@ spec:
         resources:
           {{- toYaml . | nindent 12 }}
         {{- end }}
+        livenessProbe:
+          exec:
+            command: ["/etc/frr_liveness/liveness.sh"]
+          initialDelaySeconds: 5
+          periodSeconds: 5
```
liornoy, line 359: Shouldn't this be: 
            periodSeconds: 5
            failureThreshold: 3
like in different places in the code?

fedepaol: yeah that's the old version, good catch

fedepaol: fixed

liornoy:
LGTM, also tested on my Kind cluster and works great.

PR #1745, opened 2022-12-15
Disable manual ip address pool selection

pperiyasamy:
In case we want to have the cluster administrator enforce the multi tenancy configurability of metallb, we need to disable the option for users to pin to a specific pool.
Hence this enhances ip address pool crd to disable manual ip address pool selection when its ip addresses are requested through service metallb.universe.tf/loadBalancerIPs or metallb.universe.tf/address-pool annotation.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -261,6 +261,7 @@ jobs:
           EOF
           rm -rf e2etest # we want to make sure we are not running current e2e by mistake
           cd metallb-v0.12.1
+          sed -i -e 's/quay.io\/frrouting\/frr:stable_7.5/frrouting\/frr:v7.5.1/g' e2etest/pkg/frr/container/container.go # replace with frr image from dockerhub since it has official image.
           FOCUS="L2.*should work for ExternalTrafficPolicy=Cluster|BGP.*A service of protocol load balancer should work with.*IPV4 - ExternalTrafficPolicyCluster$|validate FRR running configuration"
           sudo -E env "PATH=$PATH" inv e2etest --skip-docker --use-operator --focus "$FOCUS" -e /tmp/kind_logs
 
--- a/controller/service.go
+++ b/controller/service.go
@@ -94,7 +94,7 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 	if len(lbIPs) != 0 {
 		// This assign is idempotent if the config is consistent,
 		// otherwise it'll fail and tell us why.
-		if err = c.ips.Assign(key, lbIPs, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc)); err != nil {
+		if err = c.ips.Assign(key, svc, lbIPs, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc)); err != nil {
 			level.Info(l).Log("event", "clearAssignment", "error", err, "msg", "current IP not allowed by config, clearing")
 			c.clearServiceState(key, svc)
 			lbIPs = []net.IP{}
@@ -198,7 +198,7 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error) 
 		if serviceIPFamily != desiredLbIPFamily {
 			return nil, fmt.Errorf("requested loadBalancer IP(s) %q does not match the ipFamily of the service", desiredLbIPs)
 		}
-		if err := c.ips.Assign(key, desiredLbIPs, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc)); err != nil {
+		if err := c.ips.Assign(key, svc, desiredLbIPs, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc)); err != nil {
 			return nil, err
 		}
 		return desiredLbIPs, nil
@@ -206,7 +206,7 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error) 
 	// Otherwise, did the user ask for a specific pool?
 	desiredPool := svc.Annotations[annotationAddressPool]
 	if desiredPool != "" {
-		ips, err := c.ips.AllocateFromPool(key, serviceIPFamily, desiredPool, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc))
+		ips, err := c.ips.AllocateFromPool(key, svc, serviceIPFamily, desiredPool, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc))
 		if err != nil {
 			return nil, err
 		}
--- a/internal/allocator/allocator.go
+++ b/internal/allocator/allocator.go
@@ -72,7 +72,8 @@ func (a *Allocator) SetPools(pools *config.Pools) error {
 	// only question we have to answer is: can we fit all allocated
 	// IPs into address pools under the new configuration?
 	for svc, alloc := range a.allocated {
-		if poolFor(pools.ByName, alloc.ips) == "" {
+		pool := poolFor(pools.ByName, alloc.ips)
+		if pool == nil {
 			return fmt.Errorf("new config not compatible with assigned IPs: service %q cannot own %q under new config", svc, alloc.ips)
 		}
 	}
@@ -90,9 +91,12 @@ func (a *Allocator) SetPools(pools *config.Pools) error {
 	// Need to rearrange existing pool mappings and counts
 	for svc, alloc := range a.allocated {
 		pool := poolFor(a.pools.ByName, alloc.ips)
-		if pool != alloc.pool {
+		if pool == nil {
+			return fmt.Errorf("can't retrieve new pool for assigned IPs: service %q cannot own %q under new config", svc, alloc.ips)
+		}
+		if pool.Name != alloc.pool {
 			a.Unassign(svc)
-			alloc.pool = pool
+			alloc.pool = pool.Name
 			// Use the internal assign, we know for a fact the IP is
 			// still usable.
 			a.assign(svc, alloc)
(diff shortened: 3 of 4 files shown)


api/v1beta1/ipaddresspool_types.go
```diff
@@ -49,6 +49,13 @@ type IPAddressPoolSpec struct {
 	// multiple IPAddressPools have the same priority, choice will be random.
 	// +optional
 	AllocateTo *ServiceAllocation `json:"serviceAllocation,omitempty"`
+
+	// DisablePoolSelection disables ip pool selection when its ip addresses
+	// are requested via service's metallb.universe.tf/loadBalancerIPs or
+	// metallb.universe.tf/address-pool annotation.
```
fedepaol, line 55: I would mention spec.loadbalancerip even though it's gonna disappear soon

controller/service.go
```diff
@@ -118,10 +127,19 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 			c.client.Errorf(svc, "LoadBalancerFailed", "invalid requested loadbalancer IPs: %s", err)
 			return true
 		}
-		if len(desiredLbIPs) > 0 && !isEqualIPs(lbIPs, desiredLbIPs) {
-			level.Info(l).Log("event", "clearAssignment", "reason", "differentIPRequested", "msg", "user requested a different IP than the one currently assigned")
-			c.clearServiceState(key, svc)
-			lbIPs = []net.IP{}
+		if len(desiredLbIPs) > 0 {
```
fedepaol, line 130: Can we split this check in two, with one being the old one (the user requested a different set of ips). I think it will be eeasier to follow as the logic will be isolated.
So, two checks:
- user requested different ip
- user requested ips from a pool he has no access to

controller/service.go
```diff
@@ -118,10 +127,19 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
...
+			desiredPool, err := c.ips.GetPoolOwner(desiredLbIPs)
+			if err != nil {
+				level.Error(l).Log("event", "loadbalancerIP", "error", err, "msg", "invalid ip pool for requested loadbalancer IPs")
+				c.client.Errorf(svc, "LoadBalancerFailed", "invalid ip pool for requested loadbalancer IPs: %s", err)
+				c.clearServiceState(key, svc)
+				return true
+			}
+			if len(lbIPs) != len(desiredLbIPs) || (!isEqualIPs(lbIPs, desiredLbIPs) && !desiredPool.DisablePoolSelection) {
```
fedepaol, line 138: I  think checking the disable selection is not enough here. We need to chceck if the pool is not suitable for the service we are looking at (checking the selectors)

internal/allocator/allocator.go
```diff
@@ -108,6 +108,26 @@ func (a *Allocator) SetPools(pools *config.Pools) error {
 	return nil
 }
 
+// GetPoolOwner retrieves ip pool owner for the given list of ip addresses.
+func (a *Allocator) GetPoolOwner(ips []net.IP) (*config.Pool, error) {
```
fedepaol, line 112: I'd rather change poolFor returning the whole record (instead of just the name) and use it (and change all the calling places to do p.Name), rather than having a wrapper that do the lookup again

internal/allocator/allocator.go
```diff
@@ -108,6 +108,26 @@ func (a *Allocator) SetPools(pools *config.Pools) error {
...
+	}
+	if pool, ok := a.pools.ByName[poolStr]; ok {
+		return pool, nil
+	}
+	return nil, nil
+}
+
+// GetPool retrieves pool object for the given pool name.
+func (a *Allocator) GetPool(name string) (*config.Pool, error) {
+	if pool, ok := a.pools.ByName[name]; ok {
```
fedepaol, line 125: I'd avoid this wrapper, it's a map access

controller/service.go
```diff
@@ -103,11 +103,20 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 		// The user might also have changed the pool annotation, and
 		// requested a different pool than the one that is currently
 		// allocated.
-		desiredPool := svc.Annotations[annotationAddressPool]
-		if len(lbIPs) != 0 && desiredPool != "" && c.ips.Pool(key) != desiredPool {
-			level.Info(l).Log("event", "clearAssignment", "reason", "differentPoolRequested", "msg", "user requested a different pool than the one currently assigned")
-			c.clearServiceState(key, svc)
-			lbIPs = []net.IP{}
+		desiredPoolStr := svc.Annotations[annotationAddressPool]
+		if len(lbIPs) != 0 && desiredPoolStr != "" {
```
fedepaol, line 107: Let's split this in two too. One check is, the user requested a different pool, so we clean.
The other is, we check if the pool is compatible and if not we exit

controller/service.go
```diff
@@ -195,22 +213,34 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error)
 
 	// If the user asked for a specific IPs, try that.
 	if len(desiredLbIPs) > 0 {
-		if serviceIPFamily != desiredLbIPFamily {
-			return nil, fmt.Errorf("requested loadBalancer IP(s) %q does not match the ipFamily of the service", desiredLbIPs)
-		}
-		if err := c.ips.Assign(key, desiredLbIPs, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc)); err != nil {
+		desiredPool, err := c.ips.GetPoolOwner(desiredLbIPs)
```
fedepaol, line 216: same here. I'd keep the existing logic and eventually extend. If we got this far, we already did the pool compatibility checks?

controller/service.go
```diff
@@ -195,22 +213,34 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error)
...
+		if err != nil {
 			return nil, err
 		}
-		return desiredLbIPs, nil
+		if desiredPool != nil && !desiredPool.DisablePoolSelection {
```
fedepaol, line 220: same considerations about the logic being done in the caller of this

controller/service.go
```diff
@@ -123,6 +136,25 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
...
+				c.client.Errorf(svc, "LoadBalancerFailed", "invalid ip pool for requested loadbalancer IPs: %s", err)
+				c.clearServiceState(key, svc)
+				return true
+			}
+		}
+
+		// Check if desired ip pool is compatible for assigning ip for the service.
+		// If it is not, exit immediately with an error.
+		if desiredPool != nil {
+			if desiredPool.DisablePoolSelection || !isPoolCompatibleWithService(desiredPool, svc) {
```
fedepaol, line 152: just a thought: with this logic, there's no need for the flag right?
Because what is going to happen is, a service is able to pin to a pool only if the pool selects it.

internal/allocator/allocator.go
fedepaol: Ok, so how about delegating the compatibility check to allocator?
Outside we fetch the pool name, and we ask hte allocator if it's compatible with our service.

controller/service.go
```diff
@@ -123,6 +136,25 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 			c.clearServiceState(key, svc)
 			lbIPs = []net.IP{}
 		}
+		if len(desiredLbIPs) > 0 {
+			desiredPool, err = c.ips.GetPoolOwner(desiredLbIPs)
```
fedepaol, line 140: as I wrote below, I feel this logic belongs more to the allocator. Basically what we need to ask is, given a pool name, is it compatible with my service?

api/v1beta1/ipaddresspool_types.go
```diff
@@ -49,6 +49,14 @@ type IPAddressPoolSpec struct {
...
 	// +optional
 	AllocateTo *ServiceAllocation `json:"serviceAllocation,omitempty"`
+
+	// DisablePoolSelection disables ip pool selection when its ip addresses
+	// are requested via service's metallb.universe.tf/loadBalancerIPs or
+	// metallb.universe.tf/address-pool annotations or Spec.LoadBalancerIP
+	// field.
+	// +optional
+	// +kubebuilder:default:=false
+	DisablePoolSelection bool `json:"disablePoolSelection,omitempty"`
```
fedepaol, line 59: I start to think we don't need this anymore (see comment below) because we check if a service is asking for an ip (or a pool) it's allowed to use

controller/service.go
pperiyasamy: yes, as discussed offline, removed flag now from ip pool crd and doing only ip pool compatibility check for a given service during ip assignment.

.github/workflows/ci.yaml
```diff
@@ -261,6 +261,7 @@ jobs:
           EOF
           rm -rf e2etest # we want to make sure we are not running current e2e by mistake
           cd metallb-v0.12.1
+          sed -i -e 's/quay.io\/frrouting\/frr:stable_7.5/frrouting\/frr:v7.5.1/g' e2etest/pkg/frr/container/container.go
```
fedepaol, line 264: mind adding a comment here?

internal/allocator/allocator.go
```diff
@@ -136,20 +137,23 @@ func (a *Allocator) assign(svc string, alloc *alloc) {
...
 	pool := poolFor(a.pools.ByName, ips)
-	if pool == "" {
+	if pool == nil {
 		return fmt.Errorf("%q is not allowed in config", ips)
 	}
 	sk := &key{
 		sharing: sharingKey,
 		backend: backendKey,
 	}
+	if svc != nil && !a.isPoolCompatibleWithService(pool, svc) {
```
fedepaol, line 149: why svc shound be nil?

internal/allocator/allocator.go
```diff
@@ -340,11 +344,34 @@ func (a *Allocator) sortedPoolsForService(svc *v1.Service) []*config.Pool {
 	return append(priorityPools, noPriorityPools...)
 }
 
+func (a *Allocator) isPoolCompatibleWithService(p *config.Pool, svc *v1.Service) bool {
+	if p == nil || svc == nil {
+		return true
+	}
+	if p.ServiceAllocations != nil && p.ServiceAllocations.Namespaces.Len() > 0 &&
```
fedepaol, line 351: is the len check necessary?

internal/allocator/allocator.go
```diff
@@ -340,11 +344,34 @@ func (a *Allocator) sortedPoolsForService(svc *v1.Service) []*config.Pool {
...
+		!p.ServiceAllocations.Namespaces.Has(svc.Namespace) {
+		return false
+	}
+	svcLabels := labels.Set(svc.Labels)
+	if p.ServiceAllocations != nil && len(p.ServiceAllocations.ServiceSelectors) > 0 {
```
fedepaol, line 356: same for the len check, we can remove it

internal/allocator/allocator.go
```diff
@@ -340,11 +344,34 @@ func (a *Allocator) sortedPoolsForService(svc *v1.Service) []*config.Pool {
...
+	return true
+}
+
 // Pool returns the pool from which service's IP was allocated. If
 // service has no IP allocated, "" is returned.
 func (a *Allocator) Pool(svc string) string {
 	if alloc := a.allocated[svc]; alloc != nil {
-		return poolFor(a.pools.ByName, alloc.ips)
+		pool := poolFor(a.pools.ByName, alloc.ips)
+		if pool != nil {
```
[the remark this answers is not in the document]
pperiyasamy: This public method just intends to return empty string when pool is not found for given service. 
so i guess there is no need of returning error as well here.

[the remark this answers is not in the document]
fedepaol: makes sense

fedepaol:
lgtm

PR #1747, opened 2022-12-16
VRF: extend the tests and fix the broken things revealed by the tests

fedepaol:
Here we extend the tests to cover all the configuration options with vrf, and we also hammer our way through the failures.


frr-tools/metrics/collector/bfd_test.go
```diff
@@ -197,11 +201,19 @@ func TestBFDCollect(t *testing.T) {
 
 			l := log.NewNopLogger()
 			collector := NewBFD(l)
+			cmdOutput := map[string]string{
+				"show bgp vrf all json":                    vrfVtysh,
+				"show bfd vrf default peers json":          test.vtyshPeersOutput,
+				"show bfd vrf red peers json":              "[]",
```
oribon, line 207: wdyt about adding a test that has peers in another vrf?

frr-tools/metrics/collector/bgp_test.go
```diff
@@ -15,46 +15,47 @@ var (
 	metricsTmpl = `
 	# HELP metallb_bgp_announced_prefixes_total Number of prefixes currently being advertised on the BGP session
 	# TYPE metallb_bgp_announced_prefixes_total gauge
-	metallb_bgp_announced_prefixes_total{peer="{{ .NeighborIP }}"} {{ .AnnouncedPrefixes }}
+	metallb_bgp_announced_prefixes_total{peer="{{ .NeighborIP }}", vrf="{{ .NeighborVRF }}"} {{ .AnnouncedPrefixes }}
```
oribon, line 18: (reminder to document this vrf label)

e2etest/bgptests/metrics.go
```diff
@@ -0,0 +1,760 @@
...
+		address := c.Ipv4
+		if ipFamily == ipfamily.IPv6 {
+			address = c.Ipv6
+		}
+		peerAddr := address + fmt.Sprintf(":%d", c.RouterConfig.BGPPort)
+
+		labelsBGP := map[string]string{"peer": peerAddr}
+		labelsForQueryBGP := fmt.Sprintf(`peer="%s"`, peerAddr)
+		labelsBFD := map[string]string{"peer": address}
+		labelsForQueryBFD := fmt.Sprintf(`peer="%s"`, address)
```
oribon, line 739: I think we should have a `"vrf": default`, `vrf="default"` here, is this omitted because of native?

e2etest/bgptests/infra_setup.go
```diff
@@ -542,6 +548,16 @@ func addContainerToNetwork(containerName, network string) error {
 	return nil
 }
 
+func enableL3masterDomains(container string) error {
+	nodeExec := executor.ForContainer(container)
+	_, err := nodeExec.Exec("sysctl", "net.ipv4.udp_l3mdev_accept=1")
```
oribon, line 523: (reminder to document this)

oribon (commented):
overall looks good, the reminders can be ignored for now :)

e2etest/bgptests/metrics.go
fedepaol: I wanted to validate that if a user is not using vrfs, the old label selection would still work, so for the default vrf we select without labels. Let me add a comment for that so our future selves will thank.

frr-tools/metrics/collector/bgp_test.go
fedepaol: Filed https://github.com/metallb/metallb/issues/1751

frr-tools/metrics/collector/bfd_test.go
fedepaol: Adding a test with a different vrf won't change how the metrics look like (because "default" is a value like the others). We could add a test with two peers, but would overcomplicate the test and we still have e2e to cover that.

oribon (approved):
lgtm

PR #1749, opened 2022-12-19
FRR liveness probe: move the execution inside the container

fedepaol:
Using an exec probe means not binding the resource consumption of the
execution of the probe to the container, but to the container runtime.

By moving the execution of the vtysh behind the http endpoint, we ensure
that all the cpu quotas are respected because the actual vtysh execution
happens in the context of the container.


oribon (approved):
lgtm
(side note: we might want to use this `Run` in the collectors as well)

fedepaol:
> lgtm (side note: we might want to use this `Run` in the collectors as well)

Right, I forgot :-)
I will use the vtysh package on metrics too

frr-tools/metrics/liveness/liveness.go
```diff
@@ -0,0 +1,39 @@
...
+			"watchfrr": {},
+			"zebra":    {},
+		}
+
+		runningDaemons := strings.Split(strings.TrimSuffix(res, "\n"), " ")
+		for _, d := range runningDaemons {
+			delete(expected, d)
+		}
+		if len(expected) > 0 {
+			logger.Log("daemons not running ", expected, res)
```
liornoy, line 35: Maybe it could be more informative to have here something like: 
```suggestion
			logger.Log("daemons not running. got: ", res, "missing: ", expected)
```

fedepaol: Done

liornoy:
lgtm

PR #1752, opened 2022-12-20
E2E VRF: validate multi hop via vrf

fedepaol:
Extending the set of containers with another two reacheabile via multihop through the interface in the vrf.

The multihop setup function is made generic so that can be applied to the containers in vrf too.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -323,8 +323,18 @@ jobs:
           sed -i 's/quay.io\/metallb\/controller:main/quay.io\/metallb\/controller:dev-amd64/g' bin/metallb-operator.yaml
           sed -i 's/native/frr/g' bin/metallb-operator.yaml
           kubectl apply -f bin/metallb-operator.yaml
-          sleep 30s
 
+      - name: Ensure MetalLB operator is ready
+        run: |
+          COUNT=0
+          while [ "$(kubectl get pods -n metallb-system -l control-plane='controller-manager' -o jsonpath='{.items[*].status.containerStatuses[0].ready}')" != "true" ]; do
+            sleep 5
+            COUNT=$((COUNT+1))
+            if [[ $COUNT -gt 15 ]] ; then
+              exit 1;
+            fi
+            echo "Waiting for operator pod to be ready."
+          done
 
       - name: MetalLB E2E Tests with Operator Deployment
         run: |
--- a/e2etest/bgptests/infra_setup.go
+++ b/e2etest/bgptests/infra_setup.go
@@ -18,22 +18,39 @@ import (
 )
 
 const (
-	multiHopNetwork      = "multi-hop-net"
-	kindNetwork          = "kind"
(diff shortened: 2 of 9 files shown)


.github/workflows/ci.yaml
```diff
@@ -323,8 +323,14 @@ jobs:
           sed -i 's/quay.io\/metallb\/controller:main/quay.io\/metallb\/controller:dev-amd64/g' bin/metallb-operator.yaml
           sed -i 's/native/frr/g' bin/metallb-operator.yaml
           kubectl apply -f bin/metallb-operator.yaml
-          sleep 30s
 
+      - name: Ensure MetalLB operator is ready
+        run: |
+          sleep 5
+          while [ "$(kubectl get pods -n metallb-system -l control-plane='controller-manager' -o jsonpath='{.items[*].status.containerStatuses[0].ready}')" != "true" ]; do
```
oribon, line 330: can you add some retry counter? probably won't trigger but better be safe

e2etest/bgptests/infra_setup.go
```diff
@@ -177,26 +228,26 @@ func InfraTearDown(cs *clientset.Clientset, containers []*frrcontainer.FRR) erro
 
 // multiHopSetUp connects the ebgp-single-hop container to the multi-hop-net network,
```
oribon, line 229: can you change this comment + update the docs (can be in a different pr I guess)?

e2etest/bgptests/infra_setup.go
```diff
@@ -147,27 +164,61 @@ func VRFContainersSetup(cs *clientset.Clientset) ([]*frrcontainer.FRR, error) {
...
+
+func infraTearDown(cs *clientset.Clientset, containers []*frrcontainer.FRR, nextHop nextHopSettings, filter func(*frrcontainer.FRR) bool) error {
+	multiHopRoutes, err := container.Networks(nextHop.nextHopContainerName)
+	if err != nil {
+		return err
+	}
+
+	filtered := make([]*frrcontainer.FRR, 0)
+	for _, c := range containers {
+		if !filter(c) {
```
oribon, line 212: I think it's more clear doing this the other way (if it matches the filter it is considered)

e2etest/bgptests/infra_setup.go
```diff
@@ -18,22 +18,39 @@ import (
...
+	kindNetwork         = "kind"
+	vrfNetwork          = "vrf-net"
+	vrfName             = "red"
+	defaultRoutingTable = ""
+	metalLBASN          = 64512
+	metalLBASNVRF       = 64513
+	externalASN         = 4200000000
 )
 
+type nextHopSettings struct {
```
oribon, line 30: neat

e2etest/bgptests/infra_setup.go
fedepaol: fair enough, but then I'll change the name because filter doesn't sound right.

oribon (approved):
lgtm

Issue #1684, opened 2022-11-03
CIDR and usable addresses?


withinboredom:
With IPv4, given a CIDR, what addresses will metallb consider usable?

Generally, there are the traditional network/broadcast addresses at the beginning/end of the range. Will metallb allocate those addresses, and if not, is there a way a tell metallb to do so? With something as small as a /29, this is the difference between 6 or 8 usable addresses, and relatively substantial.

I'm sorry if this is in the documentation or a currently existing issue somewhere; I couldn't find it.

fedepaol:
If you don't specify anything, metallb will consider the .0 and the .255 as usable.
If you don't want to use them, you can use the AvoidBuggyIPs flag in the IPAddressPool CRD.

Leaving this open to restore the docs (at some point we thought it was fine to remove the flag, but now it's there)

fedepaol:
Doc to be changed: https://metallb.org/configuration/_advanced_ipaddresspool_config/#handling-buggy-networks

withinboredom:
> If you don't specify anything, metallb will consider the .0 and the .255 as usable.

This is what I wanted to hear! but what about 65.108.66.126/29 where 65.108.66.120 is the "network IP" and 65.108.66.127 is the "broadcast IP"? 

The current implementation is probably ideal (if .0 and .255 are hardcoded) for 99% of use cases. Still, if you're using metallb on an internal network, you probably only want to use CIDR's less than 25 with this setting, or there is a slight possibility of a bad day if it were to assign the first/last IP.

fedepaol:
The 255 are hardcoded for now: https://github.com/metallb/metallb/blob/f8007e04d3380496ef59ea2efe688e3aacdabd71/e2etest/pkg/config/validate.go#L105

PR #1756, opened 2022-12-23
Website: restore the avoidbuggyips doc

fedepaol:
The flag was removed by the doc when we removed it from the api. Then we restored the api but not the flag.

Fixes https://github.com/metallb/metallb/issues/1684


liornoy:
lgtm

PR #1759, opened 2022-12-25
Respect network unavailable condition

### LINKED ISSUE #1710: MetalLB ignores NodeNetworkUnavailable condition
Hi,

we're using MetalLB ~~3.0.10~~ 0.12.1 with Rancher v1.23.12-rancher1-1 (which equals Kubernetes 1.23.12) and the Canal network plugin.

We believe MetalLB executes a failback of an IP to a node that was just restarted (during a node reboot scenario), when the nodes networking stack isn't ready yet.

This bug seems to appear when a Kubernetes node containing both a MetalLB speaker and MetalLB controller Pod (which I'm calling $WORKER1 here) is restarted.
MetalLB always assigns our single configured IP address to the node containing the controller Pod as soon as possible, i.e. even when the speaker Pod has just started after a node reboot.

The following timeline might clarify the problem:

| Timestamp  | Event |
| ------------- | ------------- |
| 15:30:00 | $WORKER1 holds one IP address configured in MetalLB in layer2 mode |
| 15:31:07 | executing `systemctl reboot` on $WORKER1 |
| 15:31:17 | the MetalLB speaker Pod on $WORKER1 is terminated and the IP address switches to $WORKER2 |
| 15:31:58 | reboot of $WORKER1 |
| 15:33:43 | $WORKER1 has been rebooted and services start again |
| 15:33:44 | the MetalLB speaker Pod starts again on $WORKER1 |
| 15:33:44 | IP address switches back to $WORKER1 from $WORKER2 |
| 15:33:44 | Services in K8s cluster become unreachable from the outside as traffic is sent to the network interface of $WORKER1 |
| 15:33:53 | Container calico-node in Pod $CANAL_POD on $WORKER1 logs "Calico node started successfully"
| 15:33:59 | Container kube-flannel in Pod $CANAL_POD on $WORKER1 logs "Setting NodeNetworkUnavailable" (which sets that condition to false, i.e. the network is now ready)
| 15:33:59 | Container kube-flannel in Pod $CANAL_POD on $WORKER1 logs "Some iptables rules are missing; deleting and recreating rules" and creates and deletes some iptables rules
| 15:34:01 | Services in K8s cluster become reachable again

we are not using ipvs mode for kube-proxy but the standard iptables mode.

I believe this is a bug in MetalLB, as the condition NodeNetworkUnavailable isn't set to "false" until 15:33:59. MetalLB routes traffic to the node 15 seconds before this point, which is obviously not going to work.

liornoy:
This PR fixes the bug of MetalLB ignoring the node NetworkUnavailable condition and adds/updates
relevant unit tests and e2e tests.

fixes #1710

### DIFF
--- a/e2etest/l2tests/l2.go
+++ b/e2etest/l2tests/l2.go
@@ -232,6 +232,54 @@ var _ = ginkgo.Describe("L2", func() {
 			err = wget.Do(address, executor.Host)
 			framework.ExpectNoError(err)
 		})
+
+		ginkgo.It("should not be announced from a node with a NetworkUnavailable condition", func() {
+			svc, _ := service.CreateWithBackend(cs, f.Namespace.Name, "external-local-lb", service.TrafficPolicyCluster)
+			defer func() {
+				err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
+				framework.ExpectNoError(err)
+			}()
+
+			allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+			framework.ExpectNoError(err)
+
+			ginkgo.By("getting the advertising node")
+			ingressIP := e2eservice.GetIngressPoint(&svc.Status.LoadBalancer.Ingress[0])
+			n, err := advertisingNodeFromMAC(allNodes.Items, ingressIP, executor.Host)
+			framework.ExpectNoError(err)
+			nodeToSet := n.Name
+
+			err = k8s.SetNodeCondition(cs, nodeToSet, corev1.NodeNetworkUnavailable, corev1.ConditionTrue)
+			framework.ExpectNoError(err)
+			defer func() {
+				err = k8s.SetNodeCondition(cs, nodeToSet, corev1.NodeNetworkUnavailable, corev1.ConditionFalse)
+				framework.ExpectNoError(err)
+			}()
+
+			ginkgo.By("validating the service is announced from a different node")
+			gomega.Eventually(func() string {
+				advNode, err := advertisingNodeFromMAC(allNodes.Items, ingressIP, executor.Host)
+				if err != nil {
+					return err.Error()
+				}
+
+				return advNode.Name
+			}, time.Minute, time.Second).ShouldNot(gomega.Equal(nodeToSet))
+
+			ginkgo.By("setting the NetworkUnavailable condition back to false")
+			err = k8s.SetNodeCondition(cs, nodeToSet, corev1.NodeNetworkUnavailable, corev1.ConditionFalse)
+			framework.ExpectNoError(err)
+
+			ginkgo.By("validating the service is announced back again from the previous node")
+			gomega.Eventually(func() string {
+				advNode, err := advertisingNodeFromMAC(allNodes.Items, ingressIP, executor.Host)
+				if err != nil {
+					return err.Error()
+				}
+
+				return advNode.Name
+			}, time.Minute, time.Second).Should(gomega.Equal(nodeToSet))
+		})
 	})
 
 	ginkgo.Context("validate different AddressPools for type=Loadbalancer", func() {
--- a/e2etest/pkg/k8s/nodes.go
+++ b/e2etest/pkg/k8s/nodes.go
@@ -4,11 +4,17 @@ package k8s
 
 import (
 	"context"
+	"encoding/json"
+	"fmt"
 	"net"
+	"time"
 
+	"github.com/onsi/ginkgo/v2"
+	"github.com/onsi/gomega"
 	"go.universe.tf/metallb/e2etest/pkg/executor"
 	"go.universe.tf/metallb/e2etest/pkg/netdev"
 	"go.universe.tf/metallb/internal/ipfamily"
+	"go.universe.tf/metallb/internal/k8s/nodes"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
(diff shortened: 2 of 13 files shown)


e2etest/l2tests/node_selector.go
```diff
@@ -273,5 +274,83 @@ var _ = ginkgo.Describe("L2", func() {
 				return advNode.Name
 			}, 2*time.Minute, time.Second).Should(gomega.Equal(nodeToLabel.Name))
 		})
+
+		ginkgo.It("should respect network-unavailable condition", func() {
```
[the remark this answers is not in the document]
fedepaol: Also, another relevent test would be not using the node selector, doing a l2 advertisement, look at the node, make the node unavailable and see if the announcement moves (and gets back once we remove it)

liornoy:
@fedepaol I'm still working on this - and I took your suggestion to have a nodes collection
in the speaker's controller, and by this having the ShouldAnnounce fail if my node has a NetworkUnavailable condition.
However, we update this node collection via the Node listener that invokes the SetNodes method.
In the BGP controller, we're invoking syncPeers(l) and then the update of the announcements should occur.
On the other hand, in the L2 controller, the SetNodes method just invokes c.sList.Rejoin() which does not make the
ARP responders update.
I'm checking how can we modify the layer2_controller to have the SetNode update the responders.

fedepaol:
> @fedepaol I'm still working on this - and I took your suggestion to have a nodes collection in the speaker's controller, and by this having the ShouldAnnounce fail if my node has a NetworkUnavailable condition. However, we update this node collection via the Node listener that invokes the SetNodes method. In the BGP controller, we're invoking syncPeers(l) and then the update of the announcements should occur. On the other hand, in the L2 controller, the SetNodes method just invokes c.sList.Rejoin() which does not make the ARP responders update. I'm checking how can we modify the layer2_controller to have the SetNode update the responders.
> 
> I pushed what I have so far as a draft.

You need  to retrigger the processing of all the services. This can be done in two ways:
You extend the config we pass to the controller with the status of all the nodes
OR
You change the node listener to be able to trigger the reprocessall mechanism.

Probably the latter is better.

The clean way is to pass through reprocessAll in any case, because any service might be affected by this change.
Also note that bgp is affected only by the status of the node the speaker runs on, whereas l2 might be affected also by changes of the status of other nodes, because the elected leader might change.

liornoy:
@fedepaol The new implementation is ready for review. Took all of the comments into account and there I resolve them. thanks.

speaker/bgp_controller_test.go
```diff
@@ -211,7 +213,7 @@ func TestBGPSpeaker(t *testing.T) {
 	}
 	newBGP = b.NewSessionManager
 	c, err := newController(controllerConfig{
-		MyNode:        "pandora",
+		MyNode:        nodeName,
```
fedepaol, line 216: can we restore the previous version? I think having the explicit name is more clear

speaker/bgp_controller_test.go
```diff
@@ -1084,7 +1086,7 @@ func TestBGPSpeakerEPSlices(t *testing.T) {
...
```
fedepaol, line 1089: same

speaker/layer2_controller.go
```diff
@@ -94,7 +95,7 @@ func usableNodes(eps epslices.EpsOrSlices, speakers map[string]bool) []string {
 	return ret
 }
 
-func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce []net.IP, pool *config.Pool, svc *v1.Service, eps epslices.EpsOrSlices) string {
+func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce []net.IP, pool *config.Pool, svc *v1.Service, eps epslices.EpsOrSlices, allNodes map[string]*v1.Node) string {
```
fedepaol, line 98: nit: just "nodes" as without any classifier it's clear it refers to all the nodes of the cluster

internal/k8s/nodes/nodes.go
```diff
@@ -0,0 +1,22 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package nodes
+
+import (
+	corev1 "k8s.io/api/core/v1"
+)
+
+// ConditionStatus returns the status of the condition for a given node.
+func ConditionStatus(n *corev1.Node, ct corev1.NodeConditionType) corev1.ConditionStatus {
```
fedepaol, line 10: I'd add also a "IsNetworkUnavailable" which we can use

speaker/main.go
```diff
@@ -280,6 +284,10 @@ func (c *controller) SetBalancer(l log.Logger, name string, svc *v1.Service, eps
 		return c.deleteBalancer(l, name, "notLoadBalancer")
 	}
 
+	if nodes.ConditionStatus(c.nodes[c.myNode], v1.NodeNetworkUnavailable) == v1.ConditionTrue {
```
fedepaol, line 287: Is this necessary? If everything is correct we'll get here from shouldAnnounce

speaker/main.go
```diff
@@ -508,19 +516,21 @@ func (c *controller) SetConfig(l log.Logger, cfg *config.Config) controllers.Syn
...
+	c.nodes[node.Name] = node
+
 	for proto, handler := range c.protocolHandlers {
 		if err := handler.SetNode(l, node); err != nil {
 			level.Error(l).Log("op", "setNode", "error", err, "protocol", proto, "msg", "failed to propagate node info to protocol handler")
 			return controllers.SyncStateError
 		}
 	}
-	return controllers.SyncStateSuccess
+	return controllers.SyncStateReprocessAll
```
fedepaol, line 527: I know we already enforce this in the predicate, but let's be on the safe side and trigger reload only if the network status changed

e2etest/l2tests/l2.go
```diff
@@ -232,6 +232,54 @@ var _ = ginkgo.Describe("L2", func() {
 			err = wget.Do(address, executor.Host)
 			framework.ExpectNoError(err)
 		})
+
+		ginkgo.It("should not be announced from a node with a NetworkUnavailable condition", func() {
```
fedepaol, line 236: since the choosen node is random, I'd leverage the node selector feature to pin the announcement to one node, then check the service is not advertised, and then restore the condition back.
Here we won't know if the test passes because we are lucky or because it works.

liornoy: If I understand correctly, you are concerned that in this L2 test, the announcer node could change due to a random election. i.e. the first node which we mark as "nodeToSet" wouldn't get elected again after changing back the condition to false.
I don't think the election is random in the sense of we could have a different leader in both elections where all the nodes are available.
The elections are based on the hash of the nodes name and IP and those do not change during this test.
correct me if I'm wrong.

e2etest/pkg/k8s/nodes.go
```diff
@@ -91,3 +97,42 @@ func RemoveLabelFromNode(nodeName, key string, cs clientset.Interface) {
 	_, err = cs.CoreV1().Nodes().Update(context.Background(), nodeObject, metav1.UpdateOptions{})
 	framework.ExpectNoError(err)
 }
+
+// SetCondition sets the node's condition to the desired status and validates that the change is applied.
+func SetCondition(cs clientset.Interface, nodeName string, conditionType v1.NodeConditionType, status v1.ConditionStatus) error {
```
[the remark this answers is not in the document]
liornoy: done

speaker/main.go
liornoy: done

speaker/layer2_controller.go
liornoy: done

speaker/bgp_controller_test.go
liornoy: done

e2etest/l2tests/l2.go
fedepaol: yep, I mis-read the behaviour.

speaker/main.go
```diff
@@ -411,8 +415,8 @@ func (c *controller) deleteBalancerProtocol(l log.Logger, protocol config.Proto,
 			return controllers.SyncStateSuccess
 		}
 	}
-	delete(c.svcIPs, name)
 	level.Info(l).Log("event", "serviceWithdrawn", "ip", c.svcIPs[name], "reason", reason, "msg", "withdrawing service announcement")
+	delete(c.svcIPs, name)
```
fedepaol, line 419: nit: this should not be moved

liornoy: after the deletion c.svcIPs[name] will be nil?

fedepaol: right, good catch!
Can you move this to a separate commit? I think it deserves being tracked, plus it wasn't clear looking at this changeset.

PR #1773, opened 2023-01-11
Support specifying node name to indicate the preferred leader for lay2 LB IP

dongjieh:
Support specifying node name to indicate the preferred leader for layer 2 LB IP
  By default, the Layer2 leader node is elected automatically.  This modification support indicating a specific node as the leader during the election process. If the user prefer one node to serve the Load Balancer IP, add the `metallb.universe.tf/preferredNode` annotation to your service, with the name of the node as the annotation value.

  This can help user select the preferred node to serve the LB IP instead of the default electing hashing algorithm.
  It may also be used to perform manual switchover of LB IP.
  In case this preferred node is not available, the election process can fall back to the default hashing one.

Signed-off-by: dongjie.huang <dongjie.huang@nokia-sbell.com>


fedepaol:
Hi! Thanks for the PR.
I'd prefer discussing over a design proposal, in particular the use case and eventual api changes.
I am not sure having a single node is the right approach to follow, nor is having this as part of an annotation on the service.

dongjieh:
Hi @fedepaol,

Thanks very much for your reply! Let me provide more about my idea for disscussion.

This idea came from Router Priority and Preemption configuration of VRRP.
We know MetalLB’s layer2 mode is simliar to Keepalived which using VRRP, so I think we can introduce VRRP features to enhance metallb's L2 mode.
For VRRP, the priority configuration can determine which backup VRRP router is assigned as master if the current master fails; and if preemptive is enabled, when a VRRP router is configured with higher priority than the current master is up, it replaces the current master.

This PR provides simplified function of Priority , because it only set one node with high-priority. Maybe we should set priority of multiple nodes like VRRP.
And the manual switchover (like the scenario of VRRP Preemption enabled) function is also provided, I think it make sense when the leader node requires maintenance and the user wants the traffic go to one designated node.
 
In order not to change APIs, I put the configuration in an annotation of the service and made it service-oriented.

fedepaol:
Can you file a design proposal like those we have here https://github.com/metallb/metallb/tree/main/design? It doesn't have to be fancy but I'd rather discuss over it.
There are a couple of things I'd sort out, one is the choice of having the logic choosen by the service while we are resorting to l2 advertisements to put the responsibility on the admin.
The other one is the choice of a single node.

The last one is the rationale behind it. I understand keepalived is doing it, but I'd like to understand why as a user I would want to set priority among nodes and if the extra introduced complexity is worth it.

dongjieh:
OK, I will document the design, we can discuss after that.

design/layer2-prefer-node.md
```diff
@@ -0,0 +1,103 @@
+# Configuring Preferred Node To Be Layer2 Leader
+
+## Summary
```
fedepaol, line 3: sorry to bother, would you mind filing this as a separate pr, independent from the implementation?

dongjieh: Of course, the separate PR for design is created: https://github.com/metallb/metallb/pull/1804

liornoy:
I think we might want to add test for unavailable prefered node, asserting it is electing a different leader.

Issue #1431, opened 2022-06-14
Existing services with more than two external IPs prevent MetalLB from completing setup


penguineer:
Following initial situation:
* k3s (1.24.1+k3s1) cluster with three worker nodes
* ingress-nginx running, which got an external IP for each of the worker nodes

I decided to add MetalLB as load balancer, mostly to have a fixed IP for firewall and DNS handling. Setup via Helm chart went well so far, but I got this weird error message:

reason: "nolbIPsIPFamily"
message: "Failed to retrieve lbIPs family"

Searching did not help much but turn up this part of code: https://github.com/metallb/metallb/blob/889a4bded1b46a19d5c99053f6440781439ff00e/controller/service.go#L74

After reading some more code I ended up here: https://github.com/metallb/metallb/blob/b7a8f7b38f716e1f28d7a3ad4ece52b2c865e486/internal/ipfamily/ipfamily.go#L27

My understanding: MetalLB tries to determine IPv4/Ipv6/Dual Stack behavior based on the Service's IP addresses. In this regard, the service is expected to have either one IPv4 or one IPv6 or an IPv4 and IPv6 address.
My service, at this point, had three IPv4 addresses, which has been discarded as invalid in a Dual Stack context.

Unfortunately, because of this, MetalLB also did not go along and assigned the new address to my Ingress, leaving it with the three addresses that would let MetalLB run into the same issue over and over.

After two hours of code reading and thinking that I am too stupid to set up the lb based on its documentation, I decided to reinstall the nginx ingress controller → this solved my problem and it now got the correct IP address.

So, what happened: MetalLB stumbled over existing services with external IPs not matching its expectations
What I expected: MetalLB would re-configure these services

To solve this issue I needed to delete these services and deploy them again. 
(Luckily this wasn't too hard with Helm, but it means that MetalLB cannot be introduced into a larger existing setup.)

I am not sure how well the problem can be solved, but I'm hoping that, even if it cannot or will not be fixed, with this issue thre is another point to end an search other than the line of code that generates this error.

Thank you and keep up the good work!

protoz:
I am having a similar issue but, I am not trying to use multiple external IPs, but something keeps assigning the host ips as well as the specified load balancer port.  I'm at a loss as to what is going on as I can hit those host addresses and get to my resources but that defeats the purpose of a load balancer.

penguineer:
Has this bug been fixed?

protoz:
[...]

PR #1778, opened 2023-01-13
Wipe out the loadbalancers assigned to a service if they are not valid

fedepaol:
Currently, when metallb receives a service with malformed lb ips and it's not able to determine the ip family, it errors out and reprocesses the service.

Here if the lb part of the status is invalid we just wipe it and let the assignement logic add a new ip. This prevents malformed services to cause infinite loops and helps in recovering the service.

Fixes https://github.com/metallb/metallb/issues/1431


fedepaol:
Fixes https://github.com/metallb/metallb/issues/1431

auxworker:
I hope this gets merged soon, I havent tested it however

auxworker:
I tested it and this fix works in my environment, thanks @fedepaol

PR #1782, opened 2023-01-16
Assignement tests: make the generic pool broader

fedepaol:
If there is a service of type loadbalancer in the cluster we are testing against, it will steal the single ip we expect to be assigned to our services. For this reason, we broaden the number of ips available for the default svc in order to have free ips available for our test svc.


pperiyasamy:
LGTM

fedepaol:
Failure not related to the change.

PR #1788, opened 2023-01-19
E2E: Reduce default ip pool scope to namespace

pperiyasamy:
This changes default ip pool scope only to test namespace so that it avoids ip assignment to a service outside test namespace which is already waiting for lb ip address then it leads bigger churn of ip pool reconcile attempts while running the tests.

### DIFF
--- a/e2etest/l2tests/assignment.go
+++ b/e2etest/l2tests/assignment.go
@@ -148,16 +148,17 @@ var _ = ginkgo.Describe("IP Assignment", func() {
 					AllocateTo: &metallbv1beta1.ServiceAllocation{Priority: 10, Namespaces: []string{f.Namespace.Name}},
 				},
 			}
-			defaultPool := metallbv1beta1.IPAddressPool{
-				ObjectMeta: metav1.ObjectMeta{Name: "test-ns-pool-3"},
+			namespacePoolNoPriority := metallbv1beta1.IPAddressPool{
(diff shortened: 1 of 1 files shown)


e2etest/l2tests/assignment.go
```diff
@@ -148,16 +148,17 @@ var _ = ginkgo.Describe("IP Assignment", func() {
 					AllocateTo: &metallbv1beta1.ServiceAllocation{Priority: 10, Namespaces: []string{f.Namespace.Name}},
 				},
 			}
-			defaultPool := metallbv1beta1.IPAddressPool{
+			defaultNsPool := metallbv1beta1.IPAddressPool{
```
fedepaol, line 151: this should be "withNamespaceNoPriority", according to the other tests. Also, if we pin here we can revert the pool to be /32

e2etest/l2tests/assignment.go
```diff
@@ -208,17 +209,18 @@ var _ = ginkgo.Describe("IP Assignment", func() {
 					AllocateTo: &metallbv1beta1.ServiceAllocation{Priority: 10, NamespaceSelectors: []metav1.LabelSelector{{MatchLabels: f.Namespace.Labels}}},
 				},
 			}
-			defaultPool := metallbv1beta1.IPAddressPool{
+			defaultNsPool := metallbv1beta1.IPAddressPool{
```
fedepaol, line 212: same, "namespacePoolWithNoPriority", also same for the ones below. I'd try to be consistent with the name

fedepaol:
LGTM

Issue #1770, opened 2023-01-10
serviceAnnounced for no reason and excesive ServiceReconciler logs after update to 0.13.7 from 0.12.1


pupseba:
After updating MetalLB in a few kubernetes clusters (v1.23.13 with calico cni and kube-proxy using iptables), moving from 0.12.1 to 0.13.7, we started getting these messages for no apparent reason. We only use L2 mode:

{"caller":"main.go:344","event":"serviceAnnounced","ips":["10.147.53.6"],"level":"info","msg":"service has IP, announcing","pool":"kafka-platform","protocol":"layer2","ts":"2023-01-10T13:41:25Z"}

They come in bursts, where all the IPs associated with one particular kubernetes node get announced. They are not moved from one server to another, is just that they get (re)announced by the same node they were already announced.

There is no restart, configuration change...nothing that we can see that explains why the IPs are announced "out of the blue". Yet, this is shown in logs (l2advertisement as an example, other CRs are also observed to log before the force SyncStateReprocessAll used to call ForceReload):
`
{"caller":"config_controller.go:139","controller":"ConfigReconciler","event":"force service reload","level":"info","ts":"2023-01-17T12:17:38Z"}
{"caller":"config_controller.go:51","controller":"ConfigReconciler","level":"info","start reconcile":"metallb-system/l2advertisement1","ts":"2023-01-17T12:17:38Z"}
{"caller":"config_controller.go:151","controller":"ConfigReconciler","end reconcile":"metallb-system/l2advertisement1","level":"info","ts":"2023-01-17T12:17:38Z"}
`

From one day to another, during which these sort of events are seen in logs, we are unalbe to see any change in the "under watch" resources (addresspools,bfdprofiles,bgpadvertisements,bgppeers,communities,ipaddresspools,l2advertisements). We did this comparisson by "get $i -oyaml" on different days and comparing the results, which include the .metadata.resourceVersion.

It is hard to send full logs since after the update, the amount of logging (info level) is scary. You can easily see in this graph from the last 7 days, "when" MetalLB was updated just by the amount of log entries registered. https://imgur.com/a/6Iqle6R

Most of those log entries are for "ServiceReconciler" happening over and over again. I am able to see that during ServiceAnnounced events, there are also events of type "config reloaded" and "force service reload".

pupseba:
So following a bit more the code and the logs, it seems that the event comes from the config_controller. 

[...]

PR #1791, opened 2023-01-20
Config Controller: ignore changes that are not relevant

fedepaol:
The configuration controller is triggered by multiple events, including all the secrets in the metallb's namespace. This cause a cascade effect of reprocessing all the services which is not necessary.

Here we compare the last processed configuration with the one we just produced (which takes into account of non used secrets and non used namespaces, for example), limiting configuration reload events to those that really are changing the metallb's internal configuration.

On top of that, in order to keep the node label test effective, the implementation of the reconciliation function is deferred to a local function that can be overridden by the tests.

Fixes https://github.com/metallb/metallb/issues/1770

### DIFF
--- a/internal/k8s/controllers/config_controller.go
+++ b/internal/k8s/controllers/config_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"reflect"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
@@ -44,9 +45,14 @@ type ConfigReconciler struct {
 	ValidateConfig config.Validate
 	ForceReload    func()
 	BGPType        string
+	currentConfig  *config.Config
 }
 
 func (r *ConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	return requestHandler(r, ctx, req)
+}
+
+var requestHandler = func(r *ConfigReconciler, ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	level.Info(r.Logger).Log("controller", "ConfigReconciler", "start reconcile", req.NamespacedName.String())
 	defer level.Info(r.Logger).Log("controller", "ConfigReconciler", "end reconcile", req.NamespacedName.String())
 	updates.Inc()
@@ -133,12 +139,22 @@ func (r *ConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	}
 
 	level.Debug(r.Logger).Log("controller", "ConfigReconciler", "rendered config", dumpConfig(cfg))
+	if r.currentConfig != nil && reflect.DeepEqual(r.currentConfig, cfg) {
+		level.Debug(r.Logger).Log("controller", "ConfigReconciler", "event", "configuration did not change, ignoring")
+		return ctrl.Result{}, nil
+	}
+
+	r.currentConfig = cfg
 
 	res := r.Handler(r.Logger, cfg)
 	switch res {
 	case SyncStateError:
 		configStale.Set(1)
 		updateErrors.Inc()
+		// if the configuration load failed, we reset the current config because this is gonna lead to a retry
+		// of the reconciliaton loop. If we don't reset, the retry will find the config identical and will exit,
+		// which is not what we want here.
+		r.currentConfig = nil
 		level.Error(r.Logger).Log("controller", "ConfigReconciler", "metallb CRs and Secrets", dumpClusterResources(&resources), "event", "reload failed, retry")
 		return ctrl.Result{}, retryError
 	case SyncStateReprocessAll:
--- a/internal/k8s/controllers/config_controller_test.go
+++ b/internal/k8s/controllers/config_controller_test.go
@@ -36,6 +36,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	k8sscheme "k8s.io/client-go/kubernetes/scheme"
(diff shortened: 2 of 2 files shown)


internal/k8s/controllers/config_controller.go
```diff
@@ -133,7 +139,12 @@ func (r *ConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 	}
 
 	level.Debug(r.Logger).Log("controller", "ConfigReconciler", "rendered config", dumpConfig(cfg))
+	if r.oldConfig != nil && reflect.DeepEqual(r.oldConfig, cfg) {
+		level.Error(r.Logger).Log("controller", "ConfigReconciler", "event", "configuration did not change, ignoring")
```
liornoy, line 143: Should we have level.Info here instead?

internal/k8s/controllers/config_controller_test.go
```diff
@@ -144,6 +145,81 @@ func TestConfigController(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("create failed on peer2: %v", err)
+	}
+	_, err = r.Reconcile(context.TODO(), req)
+	if err != nil {
+		t.Fatalf("reconcile failed: %v", err)
+	}
+	if !handlerCalled {
+		t.Fatalf("handler not called")
+	}
```
liornoy, line 205: Why do we need the second reconciler call with the bgppeer in this test?

internal/k8s/controllers/config_controller_test.go
```diff
@@ -144,6 +145,81 @@ func TestConfigController(t *testing.T) {
...
+
+	handlerCalled = false
+	err = fakeClient.Create(context.TODO(), &corev1.Secret{Type: corev1.SecretTypeBasicAuth, ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: testNamespace},
+		Data: map[string][]byte{"password": []byte([]byte("nopass"))}})
+
```
liornoy, line 210: nit: extra newline between the `err = ....` and the `if err != nil`

liornoy:
Left a couple of comments, regarding the node selector tests, I don't understand what you mean by `to keep the node label test effective`. Are those changes required because the changes of the reconciler broke them?

fedepaol:
> Left a couple of comments, regarding the node selector tests, I don't understand what you mean by `to keep the node label test effective`. Are those changes required because the changes of the reconciler broke them?

it didn't "broke them" but the handler wasn't being called anymore because the configuration wasn't changing

internal/k8s/controllers/config_controller.go
fedepaol: Yep, good catch

internal/k8s/controllers/config_controller_test.go
fedepaol: I wanted to ensure that if the configuration is changed, the handler is called

liornoy: Ok got it. sounds good!

liornoy:
> > Left a couple of comments, regarding the node selector tests, I don't understand what you mean by `to keep the node label test effective`. Are those changes required because the changes of the reconciler broke them?
> 
> it didn't "broke them" but the handler wasn't being called anymore because the configuration wasn't changing

Thanks for the clarification.

pupseba:
Hi. Is there any estimated date for when this improvement will hit master and be released in a new version?

internal/k8s/controllers/config_controller.go
```diff
@@ -133,7 +139,12 @@ func (r *ConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
...
+		return ctrl.Result{}, nil
+	}
 
+	r.oldConfig = cfg
```
oribon, line 147: maybe I'm missing something, but isn't setting this before the Handler problematic in the sense that if the Handler fails with a retry error the config would be ignored in the next reconcile (because of the new DeepEqual conditional)?

internal/k8s/controllers/config_controller.go
oribon: this could even be debug imo, unless you expect this to not show so often?

internal/k8s/controllers/config_controller.go
```diff
@@ -44,9 +45,14 @@ type ConfigReconciler struct {
 	ValidateConfig config.Validate
 	ForceReload    func()
 	BGPType        string
+	oldConfig      *config.Config
```
oribon, line 48: nit: isn't this more currentConfig?

internal/k8s/controllers/config_controller.go
fedepaol: no the consideration is totally fair. Let me change it so we set the current only when the current is processed successfuly

internal/k8s/controllers/config_controller_test.go
fedepaol: the line above is on multiple lines, having it attached seems a bit cramped.

internal/k8s/controllers/config_controller.go
fedepaol: done

internal/k8s/controllers/config_controller.go
fedepaol: done

internal/k8s/controllers/config_controller.go
fedepaol: done

fedepaol:
> Hi. Is there any estimated date for when this improvement will hit master and be released in a new version?

@pupseba no promises, but aiming at merging by end of this week and doing a new release by end of next week

oribon (approved):
lgtm

PR #1793, opened 2023-01-23
Modify only LB IP within LB Ingress

### LINKED ISSUE #1733: controller: leaving `service.status.loadBalancer.ingress[].hostname` intact
We have a working setup with metallb+k8s_external coredns plugin for service type loadbalancer discovery and now we can resolve svc-lb hostnames. However I need to be able to set this hostname on service.Status so that it is shown when listing services:

```
$ oc get svc
example-svc      LoadBalancer   172.30.230.151   172.21.34.237,my-svc.my-ns.svc.example.com   8080:31153/TCP      25m
```

I have another controller which can write `service.status.loadBalancer.ingress[].hostname` however it is reverted when metallb controller is running. As a feature request, I would like metallb to ignore hostnames on status and only work with  `ingress[].ip` instead, and leaves `hostnames` intact.

mlguerrero12:
Do not wipe out the current status of the LB. Instead, modify only the IP of the LB ingress point.
Users might have set values for LB Ingress hostname/ports.

Fixed #1733

### DIFF
--- a/controller/service.go
+++ b/controller/service.go
@@ -37,6 +37,7 @@ const (
 
 func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service) bool {
 	lbIPs := []net.IP{}
+	var lbHostnames []string
 	var err error
 	// Not a LoadBalancer, early exit. It might have been a balancer
 	// in the past, so we still need to clear LB state.
@@ -59,10 +60,15 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 	// The assigned LB IP(s) is the end state of convergence. If there's
 	// none or a malformed one, nuke all controlled state so that we
 	// start converging from a clean slate.
+	// Store hostnames for further processing.
 	for i := range svc.Status.LoadBalancer.Ingress {
 		ip := svc.Status.LoadBalancer.Ingress[i].IP
+		hostname := svc.Status.LoadBalancer.Ingress[i].Hostname
 		if len(ip) != 0 {
 			lbIPs = append(lbIPs, net.ParseIP(ip))
+		}
+		if hostname != "" {
+			lbHostnames = append(lbHostnames, hostname)
 		}
 	}
 	if len(lbIPs) == 0 {
@@ -157,11 +163,16 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 
 	// At this point, we have an IP selected somehow, all that remains
 	// is to program the data plane.
-	lbIngressIPs := []v1.LoadBalancerIngress{}
+	ipHostname := getIPHostnameRelation(lbHostnames, lbIPs)
+	svc.Status.LoadBalancer = v1.LoadBalancerStatus{}
 	for _, lbIP := range lbIPs {
-		lbIngressIPs = append(lbIngressIPs, v1.LoadBalancerIngress{IP: lbIP.String()})
+		ip := lbIP.String()
+		svc.Status.LoadBalancer.Ingress = append(svc.Status.LoadBalancer.Ingress, v1.LoadBalancerIngress{
+			IP:       ip,
+			Hostname: ipHostname[ip],
+		})
 	}
-	svc.Status.LoadBalancer.Ingress = lbIngressIPs
+
 	if svc.Annotations == nil {
 		svc.Annotations = make(map[string]string)
 	}
@@ -265,4 +276,18 @@ func isEqualIPs(ipsA, ipsB []net.IP) bool {
 		return ipsB[i].String() < ipsB[j].String()
 	})
 	return reflect.DeepEqual(ipsA, ipsB)
+}
+
+// getIPHostnameRelation creates a map with the ips that have a hostname set externally.
+func getIPHostnameRelation(lbHostnames []string, lbIPs []net.IP) map[string]string {
+	ipHostname := make(map[string]string)
+	numIPs := len(lbIPs)
+	for i, hostname := range lbHostnames {
+		if i >= numIPs {
+			break
+		}
+		ipHostname[lbIPs[i].String()] = hostname
+	}
+
+	return ipHostname
 }
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -236,6 +236,82 @@ func TestControllerMutation(t *testing.T) {
 		},
 
 		{
+			desc: "simple LoadBalancer with a hostname ",
+			in: &v1.Service{
+				Spec: v1.ServiceSpec{
+					Type:       "LoadBalancer",
+					ClusterIPs: []string{"1.2.3.4"},
+				},
+				Status: v1.ServiceStatus{
+					LoadBalancer: v1.LoadBalancerStatus{
+						Ingress: []v1.LoadBalancerIngress{
+							{
+								Hostname: "foo.bar.local",
+							},
+						},
+					},
+				},
+			},
+			want: &v1.Service{
+				Spec: v1.ServiceSpec{
+					ClusterIPs: []string{"1.2.3.4"},
+					Type:       "LoadBalancer",
+				},
+				Status: v1.ServiceStatus{
+					LoadBalancer: v1.LoadBalancerStatus{
+						Ingress: []v1.LoadBalancerIngress{
+							{
+								IP:       "1.2.3.0",
+								Hostname: "foo.bar.local",
+							},
+						},
+					},
+				},
+			},
+		},
+
+		{
+			desc: "simple LoadBalancer with a single IP and many hostnames ",
+			in: &v1.Service{
+				Spec: v1.ServiceSpec{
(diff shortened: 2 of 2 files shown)


controller/service.go
```diff
@@ -157,11 +157,15 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 
 	// At this point, we have an IP selected somehow, all that remains
 	// is to program the data plane.
-	lbIngressIPs := []v1.LoadBalancerIngress{}
-	for _, lbIP := range lbIPs {
-		lbIngressIPs = append(lbIngressIPs, v1.LoadBalancerIngress{IP: lbIP.String()})
+	lenLB := len(svc.Status.LoadBalancer.Ingress)
+	for i, lbIP := range lbIPs {
+		if i >= lenLB {
```
fedepaol, line 162: can you explain the reasoning behind this logic? why do we append only if the lenght is bigger? What happens if the service is processed twice?

I think the issue raised is about the hostname part of an ingress, which in case of loadbalancer services must be still associated with an ip provided by metallb.

mlguerrero12: Since the main idea is to only modify the IP, the cleaning/adding part will only remove/add an ingress when the other fields (hostname and ports) are unset. If there exist an ingress with a hostname for example, the IP will be overwritten.

To replicate the issue 1733, I created a controller that adds an ingress with a hostname value. When metallb controller assigns the IP, it will use the same ingress. The IP will be set. The intention is that both, hostname and IP, are in the same ingress. They are related.

When the service is processed twice, the IP will be added in the same place.

So, when there exist ingresses, the controller will set the IP on them. Otherwise, it will append a new ingress with the IP.

That's how I understood the request from the issue 1733, am I missing something?

fedepaol: Ok, now I get it. What I find hard-ish to follow is that we are trying to recycle the current set of ips, tweaking the lenghts.
What I think it'd be a bit more straightful, is to use the newly produced values as the entry points and consume the hostnames added to the status part of the service.

So, what I think it'd be a bit more straightforward is to build a map out of the service we get from the k8s api, that relates each ip to its hostname, and use it to feed the hostname part of the ingress ip.
WDYT?

controller/service.go
```diff
@@ -174,7 +178,15 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 func (c *controller) clearServiceState(key string, svc *v1.Service) {
 	c.ips.Unassign(key)
 	delete(svc.Annotations, annotationIPAllocateFromPool)
-	svc.Status.LoadBalancer = v1.LoadBalancerStatus{}
+
```
fedepaol, line 181: MetalLB is still the one in charge of the service, so I think it's fair to erase everything. The hostname makes sense only in association with the ip

controller/service.go
mlguerrero12: I'm not tweaking the length. I just find out what the length of the array is, so that I don't index out of range. If it is possible to safely index the array, then the ip is set. Otherwise, a new ingress is created. I think that using a map will require more code. Not sure if it is simpler to understand.

controller/service.go
mlguerrero12: I see issues in some places. For example when a user requests an ip from a different pool than the one assigned. We would wipe out everything and then if everything goes well, a new ip will be assigned. The hostname will be lost.

We could say that the controller of the user should be able to see this and add the hostname to the newly created ingress. But I don't think we should rely on that. I'm fine if you prefer to simply delete everything. Please let me know.

mlguerrero12: I gave it another thought and I don't think we should go for delete always, especially since we must avoid having two controllers fighting.

Say a user creates an invalid service, such as requesting an ip from a non existing pool, if we delete everything, the user's controller will add the hostname back. This will cause our controller to delete everything again. The cycle will start over. I've tested it.

As you said before, a hostname only makes sense when an IP is allocated. Therefore, I  propose to leave the "Deletion" as it is right now (do not delete ingress with a hostname - clear ip only) and for the "Addition", remove ingresses with hostnames that do not have an ip allocated. For this, I will follow the approach you're suggesting (map + always add IP). This way, we will avoid cycles for erroneous requests and remove invalid hostnames.

What do you think?

fedepaol: what I don't like is leaving elements of the ingress slice without ip. I am afraid that this will bite us in the future revealing corner cases that we don't know of today. And in general, add unnecessary complexity to handle hypothetical scenarios that shouldn't be under metallb's scope.

A lb service doesn't work if it doesn't have an exposed ip, so I think we should wipe the loadbalancer.ingress part of the service, which is under metallb's responsability.

If the controller sets the hostname's part of a service that is not exposed (meaning, it doesn't have a lb ip), I think this is a bad behaviour of that controller which wouldn't have any way to resolve that name to an ip that does not exists.

controller/service.go
```diff
@@ -59,11 +60,16 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 	// The assigned LB IP(s) is the end state of convergence. If there's
 	// none or a malformed one, nuke all controlled state so that we
 	// start converging from a clean slate.
+	// Store hostnames for further processing.
 	for i := range svc.Status.LoadBalancer.Ingress {
 		ip := svc.Status.LoadBalancer.Ingress[i].IP
+		hostname := svc.Status.LoadBalancer.Ingress[i].Hostname
```
fedepaol, line 66: not related to this change, but if we are using i only to access the array, we can convert this to _, ingress and access the element directly.

controller/service.go
```diff
@@ -266,3 +277,17 @@ func isEqualIPs(ipsA, ipsB []net.IP) bool {
 	})
 	return reflect.DeepEqual(ipsA, ipsB)
 }
+
+// getIPHostnameRelation creates a map with the ips that have a hostname set externally.
+func getIPHostnameRelation(lbHostnames []string, lbIPs []net.IP) map[string]string {
+	ipHostname := make(map[string]string)
```
fedepaol, line 283: We can fill the map while we iterate the ingress structure (either in the loop above, or by re iterating it as it's gonna be a few elements anyway). By doing that we have more clear the relationship between ip and the name, while here we need to reason about the fact that we must process in the same order or if there's an ip in the list that does not have an hostname associated.

controller/service.go
```diff
@@ -266,3 +277,17 @@ func isEqualIPs(ipsA, ipsB []net.IP) bool {
 	})
 	return reflect.DeepEqual(ipsA, ipsB)
 }
+
+// getIPHostnameRelation creates a map with the ips that have a hostname set externally.
+func getIPHostnameRelation(lbHostnames []string, lbIPs []net.IP) map[string]string {
```
fedepaol, line 282: I'd just call this hostnamesForIPs  , after what it returns

controller/service.go
mlguerrero12: Yes, the idea was to iterate the ingress structure again, but everything is deleted in line 75 when a service has no lb ips previously assigned (first time allocations). Therefore, the hostnames are lost.

Which loop are you referring to exactly?

fedepaol: What I mean is, you can build the map of the desired association where you collect `lbHostnames`, and use it later.

mlguerrero12: how can I build the map there if there are no ips (except when there is a reassignment)? The key of the map is the IP and the value the hostname. Sorry, I don't understand.

fedepaol: Sorry, let's stop for a sec. I might have misunderstood the whole thing.

My thought was that in lb services, any external controller would associate an hostname to an ip that is being set by metallb. So the flow would've been: metallb sets the ip, the controller associates the hostname to the ip.

Looking at the api docs they seem mutually exclusives:
```
// LoadBalancerIngress represents the status of a load-balancer ingress point:
// traffic intended for the service should be sent to an ingress point.
type LoadBalancerIngress struct {
	// IP is set for load-balancer ingress points that are IP based
	// (typically GCE or OpenStack load-balancers)
	// +optional
	IP string `json:"ip,omitempty" protobuf:"bytes,1,opt,name=ip"`

	// Hostname is set for load-balancer ingress points that are DNS based
	// (typically AWS load-balancers)
	// +optional
	Hostname string `json:"hostname,omitempty" protobuf:"bytes,2,opt,name=hostname"`

	// Ports is a list of records of service ports
	// If used, every port defined in the service should have an entry in it
	// +listType=atomic
	// +optional
	Ports []PortStatus `json:"ports,omitempty" protobuf:"bytes,4,rep,name=ports"`
}
```

So, if that is the case, the logic is even simpler. Sorry again, I probably misunderstood the whole thing.

fedepaol:
Let's park this for a bit, I want to make sure I understand what is the requirement and if this is a legit use case. Sorry, I lightheartedly thought that this was the way to have metallb integrated with dns providers, but I am not sure anymore given the api description here https://github.com/metallb/metallb/pull/1793#discussion_r1088082701

PR #1794, opened 2023-01-23
Remove return value for convergeBalancer

mlguerrero12:
The method convergeBalancer returns always True. This boolean is not
needed, therefore it is removed.

### DIFF
--- a/controller/main.go
+++ b/controller/main.go
@@ -71,9 +71,8 @@ func (c *controller) SetBalancer(l log.Logger, name string, svcRo *v1.Service, _
 	svc := svcRo.DeepCopy()
 	successRes := controllers.SyncStateSuccess
 	wasAllocated := c.isServiceAllocated(name)
-	if !c.convergeBalancer(l, name, svc) {
-		return controllers.SyncStateError
-	}
+	c.convergeBalancer(l, name, svc)
+
 	if wasAllocated && !c.isServiceAllocated(name) { // convergeBalancer may deallocate our service and this means it did it.
 		// if the service was deallocated, it may have have left room
 		// for another one, so we reprocess
--- a/controller/service.go
+++ b/controller/service.go
@@ -35,7 +35,7 @@ const (
 	annotationIPAllocateFromPool = "metallb.universe.tf/ip-allocated-from-pool"
 )
 
-func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service) bool {
+func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service) {
 	lbIPs := []net.IP{}
 	var err error
 	// Not a LoadBalancer, early exit. It might have been a balancer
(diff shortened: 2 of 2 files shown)


controller/main.go
```diff
@@ -71,38 +71,26 @@ func (c *controller) SetBalancer(l log.Logger, name string, svcRo *v1.Service, _
...
 		// for another one, so we reprocess
 		level.Info(l).Log("event", "serviceUpdated", "msg", "removed loadbalancer from service, services will be reprocessed")
 		successRes = controllers.SyncStateReprocessAll
 	}
-	if reflect.DeepEqual(svcRo, svc) {
-		level.Debug(l).Log("event", "noChange", "msg", "service converged, no change")
-		return successRes
-	}
-
-	toWrite := svcRo.DeepCopy()
```
fedepaol, line 88: why this logic is removed? This version was defensive and made sure that we are only updating the status part of the service. The new version on the other hand depends on the behaviour of convergeBalancer, making this a bit more fragile.

controller/main.go
```diff
@@ -71,38 +71,26 @@ func (c *controller) SetBalancer(l log.Logger, name string, svcRo *v1.Service, _
...
 
-	if !reflect.DeepEqual(toWrite, svcRo) {
+	// Only the Status and Annotations are modified in convergeBalancer.
+	if !reflect.DeepEqual(svcRo.Status, svc.Status) || !reflect.DeepEqual(svcRo.Annotations, svc.Annotations) {
 		if err := c.client.UpdateStatus(svc); err != nil {
 			level.Error(l).Log("op", "updateServiceStatus", "error", err, "msg", "failed to update service")
 			return controllers.SyncStateError
 		}
 		level.Info(l).Log("event", "serviceUpdated", "msg", "updated service object")
+	} else {
```
fedepaol, line 90: I'd rather have an early return than an else here

fedepaol:
Please keep the commits independent and atomic. The subject should reflect what you did, and the body why you did it.
This https://cbea.ms/git-commit/ is a good guideline on writing effective commit messages.

controller/main.go
mlguerrero12: Because it's cleaner in my opinion. Do we need to add this kind of protection of our own code?

controller/main.go
mlguerrero12: Agree, it looks better. However, I didn't put it because it's so close to the end that it is not an early return anymore. Will change it.

controller/main.go
fedepaol: It's definitely cleaner. What I am asking here is if you are 100% sure we don't need it and you checked it.

If you did, then it's fine (despite it should still belong to a different commit and not to the "remove return value" one). If not, then please check before :-) I really don't have the answer because this part has more or less been stable and I thought that being defensive was fine.

mlguerrero12: I'm sure. I'm also relying on the ci (assuming the failing test is not related). Nevertheless, you're right. This is out of the scope of this PR. Will remove it and create a new one.

fedepaol:
lgtm, please fix the lint error

mlguerrero12:
it's complaining about the return. It seems that the original version with an "else" is better.

PR #1795, opened 2023-01-23
Refactor table test in controller_test

mlguerrero12:
Run cases of the table test as subtests. This causes IDEs such as Goland to recognize each case separately. It is then possible to run each case directly. It also improves logging.


controller/controller_test.go
```diff
@@ -897,20 +895,7 @@ func TestControllerMutation(t *testing.T) {
...
-		}
-
-		if t.Failed() {
-			// Don't run more test cases if we've already failed, to
-			// keep the output readable.
-			break
-		}
-
-		// Shuffle the input vector, and run again.
-		for x := range tests {
```
fedepaol, line 909: why this part was removed? I don't think it's part of this change and I guess we still want it.
If a commit says "move to subtests", it should do that and nothing more.

controller/controller_test.go
```diff
@@ -897,20 +895,7 @@ func TestControllerMutation(t *testing.T) {
 					panic("bad wanted IP in loadbalancer status")
 				}
 			}
-		}
-
-		if t.Failed() {
```
fedepaol, line 902: why this was removed?

controller/controller_test.go
mlguerrero12: Right, out of the scope of this PR. Will add it back.

fedepaol:
lgtm

PR #1797, opened 2023-01-24
Enable adding of additional Labels to speaker+controller

DerFels:
Hi Metallb Team,
we have several labels that we like to add to our k8s resources. Some are purely for humans, while others are technically relevant (f.e. tracing, NetworkPolicies).
I noticed that your helm chart does not offer this functionality, so I wanted to help by adding a little code that enables users to specify additionaLabels in the values.yaml which will be picked up by the speaker+controller Daemonsets.

If there is anything missing or you disagree with the solution, I'm always open to feedback :) 

Also: Thanks for providing a helm chart in the first place!

### DIFF
--- a/charts/metallb/templates/controller.yaml
+++ b/charts/metallb/templates/controller.yaml
@@ -6,6 +6,9 @@ metadata:
   labels:
     {{- include "metallb.labels" . | nindent 4 }}
     app.kubernetes.io/component: controller
+    {{- range $key, $value := .Values.controller.labels }}
+    {{ $key }}: {{ $value | quote }}
+    {{- end }}
 spec:
   {{- if .Values.controller.strategy }}
   strategy: {{- toYaml .Values.controller.strategy | nindent 4 }}
@@ -29,6 +32,9 @@ spec:
       labels:
         {{- include "metallb.selectorLabels" . | nindent 8 }}
         app.kubernetes.io/component: controller
+        {{- range $key, $value := .Values.controller.labels }}
+        {{ $key }}: {{ $value | quote }}
+        {{- end }}
     spec:
       {{- with .Values.controller.runtimeClassName }}
       runtimeClassName: {{ . | quote }}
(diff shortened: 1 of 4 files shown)


DerFels:
@fedepaol could you please take a look? 
Thank you!

charts/metallb/templates/controller.yaml
```diff
@@ -6,6 +6,11 @@ metadata:
   labels:
     {{- include "metallb.labels" . | nindent 4 }}
     app.kubernetes.io/component: controller
+    {{- if .Values.controller.additionalLabels }}
```
mlguerrero12, line 9: I don't think you need this check. If I remember well, "with" and "range" do not generate output if the array is empty.

fedepaol: right!

charts/metallb/values.yaml
```diff
@@ -233,6 +233,7 @@ controller:
   runtimeClassName: ""
   affinity: {}
   podAnnotations: {}
+  additionalLabels: {}
```
fedepaol, line 236: just "labels" as they are in the controller subsection

charts/metallb/values.yaml
```diff
@@ -294,6 +295,7 @@ speaker:
   ## Selects which runtime class will be used by the pod.
   runtimeClassName: ""
   podAnnotations: {}
+  additionalLabels: {}
```
fedepaol, line 298: same

charts/metallb/templates/controller.yaml
DerFels: Thanks @mlguerrero12 , I wasn't aware of this!

fedepaol:
@DerFels please run inv helmdocs to regenerate the helm docs (ci is annoying but prevents errors :-) )

DerFels:
> @DerFels please run inv helmdocs to regenerate the helm docs (ci is annoying but prevents errors :-) )

Done :)

PR #1799, opened 2023-01-24
Add return after updating service

mlguerrero12:
Two opposite messages were logged after updating the service object.


liornoy:
good catch, lgtm

PR #1802, opened 2023-01-25
Bump dependencies

fedepaol:
Bumping a few dependencies and the go version to benefit from fixes.


frr-tools/metrics/exporter.go
```diff
@@ -24,9 +23,8 @@ import (
 )
```
oribon, line 24: nit: metrics binary was added by mistake?

fedepaol: yep, my bad

fedepaol: removed

oribon (approved):
lgtm

PR #1804, opened 2023-01-26
Design proposal: Support specifying node name to indicate the preferred leader for layer2

dongjieh:
The implementation: https://github.com/metallb/metallb/pull/1773

### DIFF
--- a/design/layer2-prefer-leader.md
+++ b/design/layer2-prefer-leader.md
@@ -1,0 +1,103 @@
+# Configuring Preferred Node To Be Layer2 Leader
+
+## Summary
+
+The purpose of the ehancement is to support configuring preferred node to be the leader for L2 mode.
+
+With this enhancement, the leader election will be artificially controllable.
+
+## Motivation
+
+MetalLB’s layer2 mode is simliar to `Keepalived`(https://metallb.universe.tf/concepts/layer2/#comparison-to-keepalived) which using `VRRP`.
+
+This enhancement implement functionality similar to [Priority and Preemption](https://www.cisco.com/assets/sol/sb/Switches_Emulators_v2_3_5_xx/help/350_550/index.html#page/tesla_350_550_olh/ts_vrrp_18_09.html) configuration of `VRRP`.
+
+### Goals
+
+- In L2 mode, being able to specify a node as the preferred leader for a service.
+- When a node is configured as the preferred leader, it replaces the current leader.
+
+### Non-Goals
+
+- Change the current leader mechanism election using memberlist.
+- Change the data models of Metallb's CRDs.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+When using Metallb's Layer2 mode, user wants to specify one node as the leader for a service and let all traffic for this service's IP goes to this sepcified node.
+
+#### Story 2
+
+When using Metallb's Layer2 mode, the current leader node requires maintenance, user want to manually switchover to a sepcified node.
+
+## Design Details
+
+### Preferred Node Design proposal
+
+The proposal here is to:
+
+1. Add preferred node configuration for a service. In order not to change exsiting data model of the CRDs, 
+   the configuration can be placed in an annotation of the service. This can also make the configuration 
+   service-oriented and more flexible.
+
+2. Enhance Metallb's layer2 leader election process, choose the preferred node as leader.
+
+#### Preferred Node configuration
+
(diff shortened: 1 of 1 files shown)


liornoy:
The design looks good to me.

design/layer2-prefer-leader.md
```diff
@@ -0,0 +1,103 @@
...
+- Change the current leader mechanism election using memberlist.
+- Change the data models of Metallb's CRDs.
+
+## Proposal
+
+### User Stories
+
+#### Story 1
+
+When using Metallb's Layer2 mode, user wants to specify one node as the leader for a service and let all traffic for this service's IP goes to this sepcified node.
```
fedepaol, line 31: This still doesn't tell anything about the specific use case. What I would like to understand is, is there a reason why a user would want to prefer a node over another but not strictly, that justifies the addition of the feature?

design/layer2-prefer-leader.md
```diff
@@ -0,0 +1,103 @@
...
+
+#### Story 2
+
+When using Metallb's Layer2 mode, the current leader node requires maintenance, user want to manually switchover to a sepcified node.
```
fedepaol, line 35: How this would be achieved with this feature? By changing the leader manually?
I the node is not available metallb should automatically detect it and move the ip to another node. If we want to exclude a node, this can be done either via the node selectors we now have and / or playing with the speaker node selector.

design/layer2-prefer-leader.md
```diff
@@ -0,0 +1,103 @@
...
+
+1. Add preferred node configuration for a service. In order not to change exsiting data model of the CRDs, 
+   the configuration can be placed in an annotation of the service. This can also make the configuration 
+   service-oriented and more flexible.
+
+2. Enhance Metallb's layer2 leader election process, choose the preferred node as leader.
+
+#### Preferred Node configuration
+
+If user prefer one node to serve the Load Balancer IP, add the `metallb.universe.tf/preferredNode` annotation to service,
```
fedepaol, line 51: We still need to settle the discussion about the rationale, but I'd rather have this part of the l2advertisement which is in control of the cluster admin, rather than being part of the service definition which is handled by the application.

design/layer2-prefer-leader.md
chenwng: We'd like to make MetalLB provide the similar function as the one provided by priority in VRRP, which gives user the feasibility to select the node they prefer to serve the traffic. Traditionally, this can be done by setting priorities. We consider the same requirement applies to VIP function provided by MetallB. The node preferred may have some special characters (resources/connectivity/service affinity/...) defined by user to make it the best candidate during an election. Hash-based election may not be able to reflect user's wish. 

We can also define a list of nodes to define the preferences of them. But so far a single one can also satisfy our needs. Not sure if someone would need it.

design/layer2-prefer-leader.md
chenwng: Maintenance doesn't mean the node is not available. For example, if a troubleshooting step requires, we may need the traffic for a service only passing through certain node to make capturing possible. As far as I understand, the nodeselector would affect all services served by the L2Advertisement, which may affect a much larger scope.

design/layer2-prefer-leader.md
chenwng: It is a matter of granularity. We consider giving user a choice to assign the preferred node is more feasible, so different users won't have to share the same configuration of l2advertisement.  User A may prefer to use node X as his preferred node for this service,  while user B may prefer to use node Y as his, and both services can be served by the same l2advertisement/IPAddressPool.

design/layer2-prefer-leader.md
fedepaol: If troubleshooting is needed, you can have (or it will be possible in the next release) an ipaddresspool tied to a given service with a label selector, tie the advertisement to that pool only and so affect only a single service.

design/layer2-prefer-leader.md
fedepaol: Yes, but this choice can't be left to the user, because you'd risk to have all the services pinned to a node. This setting must be under control of the administrator.

design/layer2-prefer-leader.md
fedepaol: This is what I am asking. Changing the code "to make it work like vrrp" is not a valid reason. 
We need to document and discuss what is the use case and what benefit will this change provide to metallb's users, and if that benefit is worth the extra complexity added to the codebase.

PR #1806, opened 2023-01-26
FRR Mode: allow overlapping ips in case of vrfs

fedepaol:
With FRR, it's fair to have the same ip for different peers if they belong to different vrfs.

### DIFF
--- a/internal/bgp/frr/frr.go
+++ b/internal/bgp/frr/frr.go
@@ -44,8 +44,12 @@ var osHostname = os.Hostname
 
 // sessionName() defines the format of the key of the 'sessions' map in
 // the 'frrState' struct.
-func sessionName(myAddr string, myAsn uint32, addr string, asn uint32) string {
-	return fmt.Sprintf("%d@%s-%d@%s", asn, addr, myAsn, myAddr)
+func sessionName(s session) string {
+	baseName := fmt.Sprintf("%d@%s-%d@%s", s.PeerASN, s.PeerAddress, s.MyASN, s.SourceAddress)
+	if s.VRFName == "" {
+		return baseName
+	}
+	return baseName + "/" + s.VRFName
 }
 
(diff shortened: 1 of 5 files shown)


liornoy:
Maybe we can also add a test:
"duplicate bgp address - same vrf" 
where `mustFail: true` ?

fedepaol:
> Maybe we can also add a test to validation_test.go: "duplicate bgp address - same vrf" where `mustFail: true` ?

In theory this is the same as "duplicate bgp address" one (where we have the default vrf), but I can add another one.

internal/bgp/frr/frr.go
```diff
@@ -44,8 +44,11 @@ var osHostname = os.Hostname
 
 // sessionName() defines the format of the key of the 'sessions' map in
 // the 'frrState' struct.
-func sessionName(myAddr string, myAsn uint32, addr string, asn uint32) string {
-	return fmt.Sprintf("%d@%s-%d@%s", asn, addr, myAsn, myAddr)
+func sessionName(s session) string {
+	if s.VRFName == "" {
+		return fmt.Sprintf("%d@%s-%d@%s", s.PeerASN, s.PeerAddress, s.MyASN, s.SourceAddress)
```
mlguerrero12, line 49: perhaps strings.builder can be used instead of having 2 similar fmt.sprintf?

fedepaol: This function and the sprintf are compact enough that I wouldn't mind having a bit of duplication, however if that is a concern, I'd just go with string concatenation here.

I don't think this is so performance sensible that the benefit of using strings.Builder balances the slight penalty in readability, compared to just using +

Let me change it.

fedepaol: done

mlguerrero12:
lgtm

oribon (approved):
lgtm

PR #1809, opened 2023-01-28
l2 ndp: ignore link local unicast address of ipv6

cyclinder:
we should ingore link local unicast address of ipv6, Or when a interface no ipv6 global address and have dadfailed link local unicast address of ipv6, speaker stop works.

Signed-off-by: cyclinder <qifeng.guo@daocloud.io>

we hit an issue that speaker stop works, due to ipv6 dadfailed linkLocalUnicast address of ipv6, see https://blog.tankywoo.com/2013/09/27/ipv6-dadfailed-problem.html

I knew https://github.com/metallb/metallb/pull/1726 can be fix this issue, But in fact, we should really ignore the ipv6 local link address, speaker don't need to focus on ipv6 linkLocalUnicast.

### DIFF
--- a/internal/layer2/announcer.go
+++ b/internal/layer2/announcer.go
@@ -98,7 +98,7 @@ func (a *Announce) updateInterfaces() {
 			if !ok {
 				continue
 			}
-			if ipaddr.IP.To4() != nil || !ipaddr.IP.IsLinkLocalUnicast() {
+			if ipaddr.IP.To4() != nil || ipaddr.IP.IsLinkLocalUnicast() {
 				continue
 			}
 			keepNDP[ifi.Index] = true


cyclinder:
wecom-temp-ab74b5fabd7f139e8a5d4b924f209a4a

cyclinder:
So I think this PR fix is vaild, we really should ignore the local link address, I believe it is a bug.

fedepaol:
My concern is about flipping the current logic, which assumes we should create a responder when at least one linklocal address exists. I am afraid we are fixing it by incident here, because there's no way to parse the dadfailed status of a linklocal.

My question is, if an interface has no global address but only linklocal, should it work or not? And how about viceversa?
The current logic makes me think linklocal is what we care about, and that matches the linklocal definition I find here https://www.cisco.com/c/en/us/support/docs/ip/ip-version-6-ipv6/113328-ipv6-lla.html

Given that, I am not sure this is the right fix to the problem or if we should have a way to get the dadfailed information.

cyclinder:
@fedepaol Thanks for the explanation.

Yes, this fix needs more testing and verification I think,  Now is not the right time to merge.  

Global Address or Link Local, We discussed it on slack， No conclusion yet.

It would be nice if we should have a way to get the `dadfailed` information. I open an [issue](https://github.com/vishvananda/netlink/issues/848) in `netlink`  to find a way. but no response so far.

I will go to verify that if  _Global Address_ or _Link Local_ first.

PR #1813, opened 2023-01-31
Fix l2 ip pool assignment test

pperiyasamy:
While creating test ip pool for test namespace by referring to its f.Namespace.Labels, then it contains old labels due to race condition which causes ip pool never get used for service as intended and makes test to fail. Hence this retrives latest namespace object and using some of its labels for the pool creation.

### DIFF
--- a/e2etest/l2tests/assignment.go
+++ b/e2etest/l2tests/assignment.go
@@ -200,13 +200,20 @@ var _ = ginkgo.Describe("IP Assignment", func() {
 					AllocateTo: &metallbv1beta1.ServiceAllocation{Priority: 20, Namespaces: []string{f.Namespace.Name}},
 				},
 			}
+			testNs, err := cs.CoreV1().Namespaces().Get(context.Background(), f.Namespace.Name, metav1.GetOptions{})
+			framework.ExpectNoError(err)
+			testNs.Labels["foo1"] = "bar1"
+			testNs.Labels["foo2"] = "bar2"
+			_, err = cs.CoreV1().Namespaces().Update(context.Background(), testNs, metav1.UpdateOptions{})
+			framework.ExpectNoError(err)
 			namespaceLabelPoolWithHigherPriority := metallbv1beta1.IPAddressPool{
 				ObjectMeta: metav1.ObjectMeta{Name: "test-ns-label-pool-2"},
 				Spec: metallbv1beta1.IPAddressPoolSpec{
 					Addresses: []string{
 						"192.168.10.0/32",
 					},
-					AllocateTo: &metallbv1beta1.ServiceAllocation{Priority: 10, NamespaceSelectors: []metav1.LabelSelector{{MatchLabels: f.Namespace.Labels}}},
+					AllocateTo: &metallbv1beta1.ServiceAllocation{Priority: 10,
+						NamespaceSelectors: []metav1.LabelSelector{{MatchLabels: map[string]string{"foo1": "bar1", "foo2": "bar2"}}}},
 				},
 			}
 			namespacePoolNoPriority := metallbv1beta1.IPAddressPool{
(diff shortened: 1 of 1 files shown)


e2etest/l2tests/assignment.go
```diff
@@ -200,13 +200,24 @@ var _ = ginkgo.Describe("IP Assignment", func() {
 					AllocateTo: &metallbv1beta1.ServiceAllocation{Priority: 20, Namespaces: []string{f.Namespace.Name}},
 				},
 			}
+			testNs, err := cs.CoreV1().Namespaces().Get(context.Background(), f.Namespace.Name, metav1.GetOptions{})
```
fedepaol, line 203: I'd just add a couple of fixed labels to the current one (like foo=aaa, bar=bbb), update the service and use those fixed that we control as selectors.

pperiyasamy: done, I hope that's what you intended to do.

e2etest/l2tests/assignment.go
```diff
@@ -222,12 +237,18 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
 				Pools: []metallbv1beta1.IPAddressPool{namespacePoolWithLowerPriority, namespaceLabelPoolWithHigherPriority, namespacePoolNoPriority},
 			}
-			err := ConfigUpdater.Update(resources)
+			err = ConfigUpdater.Update(resources)
 			framework.ExpectNoError(err)
 
-			svc1, _ := testservice.CreateWithBackend(cs, f.Namespace.Name, "svc-test-ns-label-pool-1")
-			svc2, _ := testservice.CreateWithBackend(cs, f.Namespace.Name, "svc-test-ns-label-pool-2")
-			svc3, _ := testservice.CreateWithBackend(cs, f.Namespace.Name, "svc-test-ns-label-pool-3")
+			updateSvcLabelsFn := func(svc *corev1.Service) {
```
fedepaol, line 243: why do we need to update the labels to the service too? The selector is only against the namespace, right?

e2etest/l2tests/assignment.go
```diff
@@ -200,13 +204,24 @@ var _ = ginkgo.Describe("IP Assignment", func() {
 					AllocateTo: &metallbv1beta1.ServiceAllocation{Priority: 20, Namespaces: []string{f.Namespace.Name}},
 				},
 			}
+			testNs, err := cs.CoreV1().Namespaces().Get(context.Background(), f.Namespace.Name, metav1.GetOptions{})
+			framework.ExpectNoError(err)
+			if testNs.Labels == nil {
```
fedepaol, line 209: given that the framework always append some labels, I'd remove this and assume the map is set and empty, and I would just do the assignement here without the function

e2etest/l2tests/assignment.go
```diff
@@ -200,13 +204,24 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+				testNs.Labels = map[string]string{}
+			}
+			updateLabels(testNs.Labels)
+			_, err = cs.CoreV1().Namespaces().Update(context.Background(), testNs, metav1.UpdateOptions{})
+			framework.ExpectNoError(err)
+			nsLabelSelector := make(map[string]string)
```
fedepaol, line 215: same, just add the map to the selector below. I think the result will be clearer even if it adds repetition on the labels to be added

e2etest/l2tests/assignment.go
fedepaol: yep, this will allow us to have more control assuming nothing will modify our labels

e2etest/l2tests/assignment.go
pperiyasamy: ok, i thought you're also asking to update service with those labels to confirm no clash b/w ns and svc label selectors.

fedepaol: ah no no, we just need to make the labels we are using in the ns selector stable.

PR #1815, opened 2023-02-01
Add govulncheck action

fedepaol:
Adding the action to asses vulnerabilities with govulncheck.

### DIFF
--- a/.github/workflows/vulncheck.yaml
+++ b/.github/workflows/vulncheck.yaml
@@ -1,0 +1,19 @@
+name: Vuln check
+on:
+  pull_request:
+    branches:
+      - "main"
+      - "v**"
+  schedule: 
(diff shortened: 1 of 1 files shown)


fedepaol:
I suspect the action is failing because we don't have write permissions from branches

fedepaol:
"You can use the permissions key to add and remove read permissions for forked repositories, but typically you can't grant write access. The exception to this behavior is where an admin user has selected the Send write tokens to workflows from pull requests option in the GitHub Actions settings." from https://docs.github.com/en/actions/using-jobs/assigning-permissions-to-jobs

fedepaol:
Merging, will revert / fix if the failure wasn't due to my considerations above. The lane is not blocking

PR #1816, opened 2023-02-02
Make the exit more robust

fedepaol:
When we clear, we need to understand if the multihop container is running, because otherwise we'll panic in inspecting it (and in removing the multihop wiring). Checking the error depends on the container runtime, so it's a poor choice. Here we check if the container is among those we are processing, and if not we don't tear down the multihop setup.


pperiyasamy:
LGTM. Thanks @fedepaol for fixing this!

PR #1820, opened 2023-02-06
Make localPref uniform among updates

### LINKED ISSUE #1739: localPref behavior not uniform
seen using an frr mode dev-env by creating 2 advertisements, deleting one and reapplying it.

given 2 bgpadvertisements targeting all of the pools (there's a single pool in the cluster):
```
kubectl get bgpadvertisements -n metallb-system -oyaml
apiVersion: v1
items:
- apiVersion: metallb.io/v1beta1
  kind: BGPAdvertisement
  metadata:
    name: empty
    namespace: metallb-system
  spec:
    aggregationLength: 32
    aggregationLengthV6: 128
    localPref: 200
- apiVersion: metallb.io/v1beta1
  kind: BGPAdvertisement
  metadata:
    name: empty500
    namespace: metallb-system
  spec:
    aggregationLength: 32
    aggregationLengthV6: 128
    localPref: 500
kind: List
metadata:
  resourceVersion: ""
```

a service using the only pool in the cluster is being advertised with localPref 200 from one speaker and with 500 from the others (show ip bgp json on the only peer):
```
ca1a1cf304b6# show ip bgp json
{
 "vrfId": 0,
 "vrfName": "default",
 "tableVersion": 3,
 "routerId": "172.18.0.5",
 "defaultLocPrf": 100,
 "localAS": 64512,
 "routes": { "192.168.10.0/32": [
  {
    "valid":true,
    "multipath":true,
    "pathFrom":"internal",
    "prefix":"192.168.10.0",
    "prefixLen":32,
    "network":"192.168.10.0\/32",
    "metric":0,
    "locPrf":500,
    "weight":0,
    "peerId":"172.18.0.4",
    "path":"",
    "origin":"IGP",
    "nexthops":[
      {
        "ip":"172.18.0.4",
        "hostname":"kind-worker2",
        "afi":"ipv4",
        "used":true
      }
    ]
  },
  {
    "valid":true,
    "bestpath":true,
    "pathFrom":"internal",
    "prefix":"192.168.10.0",
    "prefixLen":32,
    "network":"192.168.10.0\/32",
    "metric":0,
    "locPrf":500,
    "weight":0,
    "peerId":"172.18.0.2",
    "path":"",
    "origin":"IGP",
    "nexthops":[
      {
        "ip":"172.18.0.2",
        "hostname":"kind-control-plane",
        "afi":"ipv4",
        "used":true
      }
    ]
  },
  {
    "valid":true,
    "pathFrom":"internal",
    "prefix":"192.168.10.0",
    "prefixLen":32,
    "network":"192.168.10.0\/32",
    "metric":0,
    "locPrf":200,
    "weight":0,
    "peerId":"172.18.0.3",
    "path":"",
    "origin":"IGP",
    "nexthops":[
      {
        "ip":"172.18.0.3",
        "hostname":"kind-worker",
        "afi":"ipv4",
        "used":true
      }
    ]
  }
] }  }
```

this might be related to the order of the route-maps:
```
a speaker advertising with 500:

kind-worker2# show running-config 
Building configuration...

Current configuration:
!
frr version 7.5.1_git
frr defaults traditional
hostname kind-worker2
log timestamp precision 3
no ipv6 forwarding
!
router bgp 64512
 no bgp ebgp-requires-policy
 no bgp default ipv4-unicast
 no bgp network import-check
 neighbor 172.18.0.5 remote-as 64512
 neighbor 172.18.0.5 timers 30 90
 !
 address-family ipv4 unicast
  network 192.168.10.0/32
  neighbor 172.18.0.5 activate
  neighbor 172.18.0.5 route-map 172.18.0.5-in in
  neighbor 172.18.0.5 route-map 172.18.0.5-out out
 exit-address-family
 !
 address-family ipv6 unicast
  neighbor 172.18.0.5 activate
  neighbor 172.18.0.5 route-map 172.18.0.5-in in
  neighbor 172.18.0.5 route-map 172.18.0.5-out out
 exit-address-family
!
ip prefix-list 172.18.0.5-200-ipv4-localpref-prefixes seq 5 permit 192.168.10.0/32
ip prefix-list 172.18.0.5-pl-ipv4 seq 5 permit 192.168.10.0/32
ip prefix-list 172.18.0.5-pl-ipv4 seq 10 deny any
ip prefix-list 172.18.0.5-500-ipv4-localpref-prefixes seq 5 permit 192.168.10.0/32
!
ipv6 prefix-list 172.18.0.5-pl-ipv4 seq 5 deny any
!
route-map 172.18.0.5-in deny 20
!
route-map 172.18.0.5-out permit 1
 match ip address prefix-list 172.18.0.5-200-ipv4-localpref-prefixes
 on-match next
 set local-preference 200
!
route-map 172.18.0.5-out permit 2
 match ip address prefix-list 172.18.0.5-500-ipv4-localpref-prefixes
 on-match next
 set local-preference 500
!
route-map 172.18.0.5-out permit 3
 match ip address prefix-list 172.18.0.5-pl-ipv4
!
route-map 172.18.0.5-out permit 4
 match ipv6 address prefix-list 172.18.0.5-pl-ipv4
!
ip nht resolve-via-default
!
ipv6 nht resolve-via-default
!
line vty
!
end
```

```
a speaker advertising with 200

kind-worker# show running-config 
Building configuration...

Current configuration:
!
frr version 7.5.1_git
frr defaults traditional
hostname kind-worker
log timestamp precision 3
no ipv6 forwarding
!
router bgp 64512
 no bgp ebgp-requires-policy
 no bgp default ipv4-unicast
 no bgp network import-check
 neighbor 172.18.0.5 remote-as 64512
 neighbor 172.18.0.5 timers 30 90
 !
 address-family ipv4 unicast
  network 192.168.10.0/32
  neighbor 172.18.0.5 activate
  neighbor 172.18.0.5 route-map 172.18.0.5-in in
  neighbor 172.18.0.5 route-map 172.18.0.5-out out
 exit-address-family
 !
 address-family ipv6 unicast
  neighbor 172.18.0.5 activate
  neighbor 172.18.0.5 route-map 172.18.0.5-in in
  neighbor 172.18.0.5 route-map 172.18.0.5-out out
 exit-address-family
!
ip prefix-list 172.18.0.5-500-ipv4-localpref-prefixes seq 5 permit 192.168.10.0/32
ip prefix-list 172.18.0.5-pl-ipv4 seq 5 permit 192.168.10.0/32
ip prefix-list 172.18.0.5-pl-ipv4 seq 10 deny any
ip prefix-list 172.18.0.5-200-ipv4-localpref-prefixes seq 5 permit 192.168.10.0/32
!
ipv6 prefix-list 172.18.0.5-pl-ipv4 seq 5 deny any
!
route-map 172.18.0.5-in deny 20
!
route-map 172.18.0.5-out permit 1
 match ip address prefix-list 172.18.0.5-500-ipv4-localpref-prefixes
 on-match next
 set local-preference 500
!
route-map 172.18.0.5-out permit 2
 match ip address prefix-list 172.18.0.5-200-ipv4-localpref-prefixes
 on-match next
 set local-preference 200
!
route-map 172.18.0.5-out permit 3
 match ip address prefix-list 172.18.0.5-pl-ipv4
!
route-map 172.18.0.5-out permit 4
 match ipv6 address prefix-list 172.18.0.5-pl-ipv4
!
ip nht resolve-via-default
!
ipv6 nht resolve-via-default
!
line vty
!
end
```

mlguerrero12:
These changes ensure that the local preference is unique among BGP ADVs with the same type of BGP update. It is still possible to have BGP ADVs with different values of localPref if the set of BGP peers are different. BGP ADVs with common BGP peers must not have common nodes. Checks are performed for BGP ADVs that have the same aggregation length.

Fixed #1739


internal/config/config.go
```diff
@@ -935,9 +936,48 @@ func validateBGPAdvPerPool(adv *BGPAdvertisement, pool *Pool) error {
...
+						"with common pools and aggregation lengths", adv.LocalPref, bgpAdv.LocalPref)
+				}
+			}
+		}
+	}
+
 	return nil
 }
 
+func uniqueConfig(newAdv, adv *BGPAdvertisement) bool {
```
fedepaol, line 956: let's call this "advertisementsAreCompatible"

internal/config/config.go
```diff
@@ -935,9 +936,48 @@ func validateBGPAdvPerPool(adv *BGPAdvertisement, pool *Pool) error {
...
+	// BGP ADVs with different set of BGP peers do not collide.
+	if len(newAdv.Peers) != 0 && len(adv.Peers) != 0 {
+		equalPeer := false
+		for _, peer := range newAdv.Peers {
+			if slices.Contains(adv.Peers, peer) {
```
fedepaol, line 961: you can early return here by checking the negative

internal/config/config.go
```diff
@@ -935,9 +936,48 @@ func validateBGPAdvPerPool(adv *BGPAdvertisement, pool *Pool) error {
...
+				break
+			}
+		}
+		if !equalPeer {
+			return true
+		}
+	}
+
+	// BGP ADVs with different set of nodes do not collide.
+	for node := range newAdv.Nodes {
```
fedepaol, line 972: let's handle the case when one of the two maps is empty. I don't remember if when we get here, the nodes are already unrolled in the map but I wouldn't rely on that.

internal/config/config.go
```diff
@@ -935,9 +936,48 @@ func validateBGPAdvPerPool(adv *BGPAdvertisement, pool *Pool) error {
 				"this pool is more specific than the aggregation length for addresses %s", adv.AggregationLength, lowest, addr)
 		}
 	}
+
+	// Verify that BGP ADVs set a unique local preference value per BGP update.
+	for _, bgpAdv := range pool.BGPAdvertisements {
+		if adv.LocalPref != bgpAdv.LocalPref {
+			if adv.AggregationLength == bgpAdv.AggregationLength || adv.AggregationLengthV6 == bgpAdv.AggregationLengthV6 {
```
fedepaol, line 943: Let's embed this if in the "are compatible" function

internal/config/config.go
fedepaol: This is totally sane and makes sense.
I am a bit afraid of a scenario where a new node is added (or a label is changed to a node), and that change makes the configuration invalid. That won't be caught by the webhook (because we don't have webhooks on nodes and we can't have them), so from a user's point of view the configuration will be invalid without an explicit reason.
I think that the most common use case is to have the localpref be differentiated on peer basis rather than on node basis, so I am tempted to say let's be more restrictive and not allow advertisements on different nodes to have different prefixes.

Thoughts? Also invoking @oribon

config/rbac/role.yaml
```diff
@@ -14,6 +14,12 @@ rules:
       - get
       - list
       - watch
+  - apiGroups:
```
fedepaol, line 17: In case we decide we want to validate against nodes, we'll have to add this to the charts too.

internal/config/config.go
mlguerrero12: done

internal/config/config.go
mlguerrero12: Nodes are already unrolled and it already covers the case when one of the maps is empty. It will simply not collide.

internal/config/config.go
fedepaol: right, I misread it

internal/config/config.go
```diff
@@ -935,9 +936,48 @@ func validateBGPAdvPerPool(adv *BGPAdvertisement, pool *Pool) error {
...
+					"with common pools and aggregation lengths", adv.LocalPref, bgpAdv.LocalPref)
+			}
+		}
+	}
+
 	return nil
 }
 
+func advertisementsAreCompatible(newAdv, adv *BGPAdvertisement) bool {
+	if adv.AggregationLength == newAdv.AggregationLength || adv.AggregationLengthV6 == newAdv.AggregationLengthV6 {
```
fedepaol, line 955: this can be flipped and return (I hope :D )

internal/config/config.go
mlguerrero12: I'm with you in your concern about adding a node. We need to come up with a solution here.

Regarding the use of localPref, I think it is all the opposite. The most common use case is with nodes. What's the point of setting the same localPref for all speakers against a specific peer? We could choose any value for localPref since it won't add anything to the updates. It will be useless in my opinion.

The way I've used localPref before was with the aim of giving preference to some updates in iBGP. In this case, users might select a set of nodes with higher preference to receive traffic.

internal/config/config.go
mlguerrero12: done

config/rbac/role.yaml
mlguerrero12: yes, I saw the tests failing due to this

mlguerrero12: I will create a PR on the operator for the same

internal/config/config.go
mlguerrero12: done

internal/config/config.go
```diff
@@ -935,9 +936,48 @@ func validateBGPAdvPerPool(adv *BGPAdvertisement, pool *Pool) error {
 				"this pool is more specific than the aggregation length for addresses %s", adv.AggregationLength, lowest, addr)
 		}
 	}
+
+	// Verify that BGP ADVs set a unique local preference value per BGP update.
+	for _, bgpAdv := range pool.BGPAdvertisements {
+		if adv.LocalPref != bgpAdv.LocalPref {
```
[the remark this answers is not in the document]
mlguerrero12: I think this is fine. No need to insert the attribute in the function. We might use it in the future for something else.

internal/config/config.go
oribon: sorry for the late response
I'm a bit concerned about this approach, mainly because a change in the cluster which we can't listen for that  suddenly makes the config bad feels wrong to me, especially when we (still) don't have an accessible way to expose if the running configuration is good or not.

there's a thing I'd like to get your opinion that I'm still not sure of - for all of the other fields that overlap between different resources targeting the same service (like communities) we union/add them together and suddenly with localpref we'd take an "intersection" approach, saying that they all have to be the same value for a given advertisement.
is it wrong to treat overlapping localprefs similarly, where "adding" them together means choosing the highest value for the advertisement?

mlguerrero12: Yes, we shouldn't go this way if we can't expose the state of the running configuration easily.

Indeed, community is another (and I think the only other) attribute that should ideally be unique. I think that our design was simply applying what it was present in the BGP Adv regardless of the attributes' values. It fit perfect for communities since it can only have 2 states (present or not), but not so well for localpref, as you saw it in the issue you reported. We could choose the highest value but I think we will need to implement some additional logic in the speaker (bgp controller). It might get complicated there. See the TODO in updateAds.

Another option is to act as a mutating webhook by simply overriding the value of localpref if there is another one present (or using the one with the highest value). This way we won't need changes in the speaker. We can return a warning saying that the localpref was changed so that users can be aware of it.

What do you think?

mlguerrero12: @fedepaol, any thoughts on this?

fedepaol: even if we had a mutating webhook, we wouldn't be able to be always on the safe side because we might have concurrent add operations that sneak in together.
We'd have to invalidate the config anyway in this scenario.

I am almost convinced that the "newly added node" invalidating the config is not that terrible.

If the current configuration is valid, it means that the user took care of labelling the nodes in such a way that they are partiotioned wrt localpref (or that they just don't care).
If a new node comes in and invalidates the configuration, it means it wasn't labeled properly,

What is going to happen is:
- the existing speakers will keep the old configuration, without the extra node, but they will keep running
- the speaker running on the new node won't advertise

also, any new config change won't have any effect.
Having the same node aimed by two localprefs is not right, the current behaviour is not predictable and resorting to priorities will just hide a misconfiguration under the carpet, because what a user really wants is to apply one single localpref from one node. They will find metallb is behaving, it's just that the overridden localpref might be more subtle to find, compared to metallb not working at all on one node (plus we have metrics for stale configs and we will land https://github.com/metallb/metallb/pull/1685/files soon or later)

mlguerrero12:
@fedepaol, is there a way to link a pr from the operator to this pr so that the ci doesn't fail?

fedepaol:
lgtm, thanks and sorry again for the delay!

PR #1821, opened 2023-02-06
Add documentation for ip pool service allocation

pperiyasamy:
This adds web content for ip pool with service allocation feature and a sample yaml file for such configuration.

### DIFF
--- a/internal/allocator/allocator.go
+++ b/internal/allocator/allocator.go
@@ -284,7 +284,7 @@ func (a *Allocator) Allocate(svcKey string, svc *v1.Service, serviceIPFamily ipf
 		}
 		return alloc.ips, nil
 	}
-	pinnedPools := a.sortedPoolsForService(svc)
+	pinnedPools := a.pinnedPoolsForService(svc)
 	for _, pool := range pinnedPools {
 		if ips, err := a.AllocateFromPool(svcKey, svc, serviceIPFamily, pool.Name, ports, sharingKey, backendKey); err == nil {
 			return ips, nil
@@ -305,44 +305,29 @@ func (a *Allocator) Allocate(svcKey string, svc *v1.Service, serviceIPFamily ipf
 // This method returns sorted ip pools which are allocatable for given service.
 // When ip pool is not set with priority, then just append it after sorted
 // priority ip pools.
-func (a *Allocator) sortedPoolsForService(svc *v1.Service) []*config.Pool {
-	var priorityPools, noPriorityPools []*config.Pool
+func (a *Allocator) pinnedPoolsForService(svc *v1.Service) []*config.Pool {
+	var pools []*config.Pool
 	if svc == nil {
-		return priorityPools
+		return pools
 	}
 	for _, nsPoolName := range a.pools.ByNamespace[svc.Namespace] {
 		if nsPool, ok := a.pools.ByName[nsPoolName]; ok {
-			if !nsPool.AutoAssign {
+			if !nsPool.AutoAssign || !a.isPoolCompatibleWithService(nsPool, svc) {
 				continue
 			}
-			if nsPool.ServiceAllocations.Priority > 0 {
-				priorityPools = append(priorityPools, nsPool)
-				continue
-			}
-			noPriorityPools = append(noPriorityPools, nsPool)
+			pools = append(pools, nsPool)
 		}
 	}
-	svcLabels := labels.Set(svc.Labels)
 	for _, svcPoolName := range a.pools.ByServiceSelector {
 		if svcPool, ok := a.pools.ByName[svcPoolName]; ok {
-			if !svcPool.AutoAssign {
+			if !svcPool.AutoAssign || !a.isPoolCompatibleWithService(svcPool, svc) {
 				continue
 			}
-			for _, svcSelector := range svcPool.ServiceAllocations.ServiceSelectors {
-				isMatch := svcSelector.Matches(svcLabels)
-				if isMatch && svcPool.ServiceAllocations.Priority > 0 {
-					priorityPools = append(priorityPools, svcPool)
-				} else if isMatch {
-					noPriorityPools = append(noPriorityPools, svcPool)
-				}
-			}
+			pools = append(pools, svcPool)
 		}
 	}
-	sort.Slice(priorityPools, func(i, j int) bool {
-		return priorityPools[i].ServiceAllocations.Priority <
-			priorityPools[j].ServiceAllocations.Priority
-	})
-	return append(priorityPools, noPriorityPools...)
+	sortPools(pools)
+	return pools
 }
 
 func (a *Allocator) isPoolCompatibleWithService(p *config.Pool, svc *v1.Service) bool {
@@ -372,6 +357,24 @@ func (a *Allocator) Pool(svc string) string {
 		}
 	}
 	return ""
+}
+
+func sortPools(pools []*config.Pool) {
+	// A lower value for pool priority equals a higher priority and sort
+	// pools from higher to low priority. when no priority (0) set on
+	// the pool, then that is considered as lowest priority.
+	sort.Slice(pools, func(i, j int) bool {
+		if pools[i].ServiceAllocations.Priority > 0 &&
+			pools[j].ServiceAllocations.Priority > 0 {
+			return pools[i].ServiceAllocations.Priority <
+				pools[j].ServiceAllocations.Priority
+		}
+		if pools[i].ServiceAllocations.Priority == 0 &&
+			pools[j].ServiceAllocations.Priority > 0 {
+			return false
+		}
+		return true
+	})
 }
 
 func sharingOK(existing, new *key) error {
--- a/internal/allocator/allocator_test.go
+++ b/internal/allocator/allocator_test.go
@@ -83,6 +83,28 @@ func TestAssignment(t *testing.T) {
 				ipnet("1000::6:0/120"),
 			},
 		},
+		"test5": {
+			Name:          "test5",
+			AvoidBuggyIPs: true,
+			AutoAssign:    true,
+			ServiceAllocations: &config.ServiceAllocation{Priority: 20, Namespaces: sets.New("test-ns1"),
+				ServiceSelectors: []labels.Selector{selector("team=metallb")}},
+			CIDR: []*net.IPNet{
+				ipnet("1.2.7.0/24"),
+				ipnet("1000::7:0/120"),
+			},
+		},
+		"test6": {
+			Name:          "test6",
+			AvoidBuggyIPs: true,
+			AutoAssign:    true,
+			ServiceAllocations: &config.ServiceAllocation{Priority: 20, Namespaces: sets.New("test-ns2"),
+				ServiceSelectors: []labels.Selector{selector("foo=bar")}},
+			CIDR: []*net.IPNet{
+				ipnet("1.2.8.0/24"),
+				ipnet("1000::9:0/120"),
+			},
+		},
 	}}); err != nil {
 		t.Fatalf("SetPools: %s", err)
 	}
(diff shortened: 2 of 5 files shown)


website/content/configuration/_advanced_ipaddresspool_configuration.md
```diff
@@ -48,6 +48,52 @@ To specify a single IP address in a pool, use `/32` in the CIDR notation
...
+          foo: bar
+    serviceSelectors:
+      - matchExpressions:
+          - {key: app, operator: In, values: [bar]}
+```
+
+The above IPAddressPool example is pinned to service(s) with has a label
+matching expression `key: app, operator: In, values: [bar]` and created
+either in `namespace-a` or `namespace-b` or any other namespace(s) which
+has a matching label `foo: bar`.
```
fedepaol, line 95: I'd also add a note to say that if a service chooses a pool (or an ip belonging to a pool) which doesn't match the service will stay in pending.

website/content/configuration/_advanced_ipaddresspool_configuration.md
```diff
@@ -48,6 +48,52 @@ To specify a single IP address in a pool, use `/32` in the CIDR notation
...
+      - matchLabels:
+          foo: bar
+    serviceSelectors:
+      - matchExpressions:
+          - {key: app, operator: In, values: [bar]}
+```
+
+The above IPAddressPool example is pinned to service(s) with has a label
+matching expression `key: app, operator: In, values: [bar]` and created
+either in `namespace-a` or `namespace-b` or any other namespace(s) which
```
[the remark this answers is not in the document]
pperiyasamy: yes, fixed with this commit https://github.com/metallb/metallb/pull/1821/commits/cd7f7fb5588e54cac51322580c787f61af1494ac.

website/content/configuration/_advanced_ipaddresspool_configuration.md
```diff
@@ -48,6 +48,52 @@ To specify a single IP address in a pool, use `/32` in the CIDR notation
...
 {{% /notice %}}
 
+### Reduce scope of address allocation to specific Namespace and Service
+
+This option can be used to reduce the scope of particular IPAddressPool
+to set of namespaces and services, by adding an optional namespace/service
+selector.
+
+The use the pool with lowest value of priority in case of multiple matches.
+This is deterministic and will make the behaviour more clear to the users.
```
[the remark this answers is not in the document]
pperiyasamy: removed.

website/content/configuration/_advanced_ipaddresspool_configuration.md
pperiyasamy: done.

internal/allocator/allocator.go
```diff
@@ -345,13 +337,57 @@ func (a *Allocator) sortedPoolsForService(svc *v1.Service) []*config.Pool {
...
 func (a *Allocator) isPoolCompatibleWithService(p *config.Pool, svc *v1.Service) bool {
-	if p.ServiceAllocations != nil && p.ServiceAllocations.Namespaces.Len() > 0 &&
-		!p.ServiceAllocations.Namespaces.Has(svc.Namespace) {
-		return false
+	svcLabels := labels.Set(svc.Labels)
+	if p.ServiceAllocations != nil && p.ServiceAllocations.Namespaces.Len() > 0 {
+		if !p.ServiceAllocations.Namespaces.Has(svc.Namespace) {
+			return false
+		}
+		hasSvcLabelMatch := false
```
fedepaol, line 378: Is this necessary? If we don't return false here we'll fall back in the check below, and if the service doesn't match we'll return false anyway

internal/allocator/allocator.go
```diff
@@ -310,17 +310,9 @@ func (a *Allocator) sortedPoolsForService(svc *v1.Service) []*config.Pool {
...
-				continue
-			}
-			if nsPool.ServiceAllocations.Priority > 0 {
-				priorityPools = append(priorityPools, nsPool)
-				continue
-			}
-			noPriorityPools = append(noPriorityPools, nsPool)
-		}
+	if len(a.pools.ByNamespace[svc.Namespace]) > 0 {
+		priorityPools, noPriorityPools = a.sortedNamespacePoolsForService(svc)
```
fedepaol, line 314: thinking again about this, now the priority / no priority is useless as long as we handle the compare function properly, because at the end we exit from here with one single list

fedepaol: Also, instead of reimplementing it, I'd leverage the isPoolCompatibleWithService in both this loop and the one below. Basically the idea is "we see that this matches, let's see if it's compatible.

I'd do the same in the loop below even if it's logically reduntant now and burns a few cycles, because in the future we might extend the "isPoolCompatible" function with more checks.

internal/allocator/allocator_test.go
```diff
@@ -520,6 +531,29 @@ func TestAssignment(t *testing.T) {
 			ips:     []string{"1.2.6.10"},
 			wantErr: true,
 		},
+		{
+			desc:   "attempt to assign ip from pool pinned with namespace and service label",
```
fedepaol, line 535: can we add two tests that should fail because the labels are compatible but the ns is not and viceversa?

internal/config/config.go
```diff
@@ -561,7 +561,8 @@ func poolsByNamespace(pools map[string]*Pool) map[string][]string {
 func poolsByServiceSelector(pools map[string]*Pool) []string {
 	var poolsByServiceSelector []string
 	for _, pool := range pools {
-		if pool.ServiceAllocations == nil || len(pool.ServiceAllocations.ServiceSelectors) == 0 {
+		if pool.ServiceAllocations == nil || len(pool.ServiceAllocations.Namespaces) > 0 ||
```
fedepaol, line 564: is this change necessary? I think we should avoid the optimization of "checking both first, serviceselector only later" because it might bite us in the future. I guess it's cleaner if we have two separate sets, and each time we check for the whole compatiibility.

website/content/configuration/_advanced_ipaddresspool_configuration.md
```diff
@@ -48,6 +48,42 @@ To specify a single IP address in a pool, use `/32` in the CIDR notation
...
+          foo: bar
+    serviceSelectors:
+      - matchExpressions:
+          - {key: app, operator: In, values: [bar]}
+```
+
+The above IPAddressPool example is pinned to service(s) with has a label
+matching expression `key: app, operator: In, values: [bar]` and created
+either in `namespace-a` or `namespace-b`.
+When a service explicitly chooses an IPAddressPool (or an IP Address
```
fedepaol, line 83: Nit: add the note modifier here (just because the pr needs to be reworked anyway)

pperiyasamy: done.

internal/allocator/allocator_test.go
pperiyasamy: done

internal/allocator/allocator.go
pperiyasamy: yes, this check is needed for service allocations which has only namespaces, but not service selectors.

internal/allocator/allocator.go
pperiyasamy: yes, correct, we don't need `noPriorityPools` and it can be done with sort function itself.
for other comment, though these methods has some duplicate piece of code and each one is used for different purpose. I think existing structure looks to be cleaner.

internal/config/config.go
pperiyasamy: ok, reverted it.

internal/allocator/allocator.go
fedepaol: but if a service has no service selectors we'll go to the final return, right?

internal/allocator/allocator.go
fedepaol: I think having two loops (one for namespaces and the other for services) that check the match on each side, and then the compatibility is more symmetric and easier to reason about. 

Every loop should be like "check if this pool matches upon criteria x, and check if matches completely". This is easier to follow because we don't need to think that we already checked the namespaces in the previous loop, and  will make things easier if we add a third criteria.


So, what I am proposing is to use the `isPoolCompatibleWithService` function while we check in `getNamespacePoolsForService` and also in the `range a.pools.ByServiceSelector` loop. Also, since the former will be slimmer because of that, we can consider either removing the wrapper function (or wrapping the service selector logic in another function too, it feels asymmetric).

pperiyasamy: OK, I see your point. changed it accordingly, Hope it looks better now.

website/content/configuration/_advanced_ipaddresspool_configuration.md
fedepaol: what I meant is, use the `{{% notice note %}}` wrapper to highlight this.

fedepaol: and since we are going to touch this, I'd rephrase with When a service explicitly chooses an IPAddressPool via the xxx annotation or an ip via the spec.loadbalancerip...

internal/allocator/allocator.go
fedepaol: yes! Much better

fedepaol:
one final nit about the doc part, and lgtm

internal/allocator/allocator.go
```diff
@@ -305,44 +305,29 @@ func (a *Allocator) Allocate(svcKey string, svc *v1.Service, serviceIPFamily ipf
...
 // priority ip pools.
-func (a *Allocator) sortedPoolsForService(svc *v1.Service) []*config.Pool {
-	var priorityPools, noPriorityPools []*config.Pool
+func (a *Allocator) pinnedPoolsForService(svc *v1.Service) []*config.Pool {
+	var pools []*config.Pool
 	if svc == nil {
-		return priorityPools
+		return pools
 	}
 	for _, nsPoolName := range a.pools.ByNamespace[svc.Namespace] {
```
oribon, line 313: why isn't it enough to just iterate over all of the pools ByName instead of multiple fors? the `isPoolCompatible` would filter out the unrelated pools anyways? (sorry if I'm confusing stuff, not familiar with this part and the Pools struct)

fedepaol: yep, but here we save some loops by skipping the non pinned loops, as "byName" contains all of them

oribon: not sure about saving loops, as it seems the current implementation can add the same pool twice to the `pools` slice. I think it would be more readable to just iterate once over everything and the performance hit (if at all) shouldn't be too big

pperiyasamy: we can avoid pool duplicates in the `pools` slice with a change in config.go, but that was reverted based on this [comment](https://github.com/metallb/metallb/pull/1821#discussion_r1107011002). I think iterating over all pools may be somewhat costlier than current one as it needs to validate ns and svc labels for every ns/svc pools.

pperiyasamy: oh! no, the isPoolCompatibleWithService ensures no duplicates get added into `pools` slice, i think we're better here.

oribon: what I meant (and I might be wrong), that in the current way byNS/bySelector might overlap (or contain everything?) and that way we aren't really saving any iterations by not going once over everything

internal/allocator/allocator.go
```diff
@@ -374,6 +359,21 @@ func (a *Allocator) Pool(svc string) string {
 	return ""
 }
 
+func sortPools(pools []*config.Pool) {
+	sort.Slice(pools, func(i, j int) bool {
+		if pools[i].ServiceAllocations.Priority > 0 &&
```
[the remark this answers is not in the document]
pperiyasamy: added comment now.

website/content/configuration/_advanced_ipaddresspool_configuration.md
```diff
@@ -48,6 +48,44 @@ To specify a single IP address in a pool, use `/32` in the CIDR notation
 (e.g. `42.176.25.64/32`).
 {{% /notice %}}
 
+### Reduce scope of address allocation to specific Namespace and Service
+
+This option can be used to reduce the scope of particular IPAddressPool
+to set of namespaces and services, by adding an optional namespace
+and/or service selectors.
+
```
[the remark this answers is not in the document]
pperiyasamy: ok, done

website/content/configuration/_advanced_ipaddresspool_configuration.md
```diff
@@ -48,6 +48,44 @@ To specify a single IP address in a pool, use `/32` in the CIDR notation
...
+kind: IPAddressPool
+metadata:
+  name: ippool-ns-service-alloc-sample
+  namespace: metallb-system
+spec:
+  addresses:
+    - 192.168.20.0/24
+  avoidBuggyIPs: true
+  serviceAllocation:
+    priority: 50
```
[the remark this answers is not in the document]
pperiyasamy: done [here](https://github.com/metallb/metallb/pull/1821/files#diff-02a9536c1258b76a7825571c2016dd4807f0f8ffb66485d5fb3260eb5b01d806R88-R94).

internal/allocator/allocator.go
pperiyasamy: byNS/bySelector may contain subset or all of ip pools. but it depends on user's pool config. If we do it by iterating over all pools via `a.pools.ByName`, the lookup cost for desired ns/svc pools are high. @fedepaol your thoughts ?

website/content/configuration/_advanced_ipaddresspool_configuration.md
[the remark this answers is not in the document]
oribon: I think it should be documented here as well for the users + the fact that not specifying a priority / setting priority 0 is the lowest priority.

internal/allocator/allocator.go
oribon: iiuc, if most of the people won't use this feature the byNS/bySelector will always contain all of the pools.
having said that, I wouldn't want this to block this PR if we intend to have it in the next release as it's not too crucial.

fedepaol: I can't think about a scenario where there are more than... 100 pools, to say a number.
On the other hand, I think the current version is more than understandable so I'd say let's go with this version, I don't mind cleaning up later if it becomes too messy / if we add another discriminator.

website/content/configuration/_advanced_ipaddresspool_configuration.md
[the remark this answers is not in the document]
fedepaol: :+1: , and I think this is the last standing comment right?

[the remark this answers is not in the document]
oribon: yep, other than that the pr looks good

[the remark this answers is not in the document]
pperiyasamy: added [here](https://github.com/metallb/metallb/pull/1821/files#diff-02a9536c1258b76a7825571c2016dd4807f0f8ffb66485d5fb3260eb5b01d806R92-R94).

PR #1826, opened 2023-02-07
Use errors.Is to check for a specific error

Fish-pro:
Signed-off-by: Fish-pro <zechun.chen@daocloud.io>

comparing with == will fail on wrapped errors. Use errors.Is to check for a specific error


fedepaol:
lgtm

PR #1827, opened 2023-02-07
Clean up nil error returns for the functions

Fish-pro:
Signed-off-by: Fish-pro <zechun.chen@daocloud.io>

An nil error returned in a function that added a useless judgment when called. This pr optimization parameter returns a value defined


fedepaol:
question: did you ran metallb through a linter to catch these or just manually?
If the former, mind adding it to ci?

Issue #1753, opened 2022-12-22
Bump the frr version


fedepaol:
We should jump on the latest frr version, even if using the 7.5 format.

cyclinder:
I can do this 😄

fedepaol:
it's more complex than it seems, bumping alone won't work I think. There is some hammering to be done here and there, but if you want to contribute I'd be happy

cgoncalves:
Agreed. Latest stable 8.x version.

PR #1829, opened 2023-02-13
Update frr to 8.4.2

fedepaol:
Bumping the frr version to 8.4.2
Also, changing the image revealed a misconfiguration we had in propagating the fixed ips to frr, so with this change we get rid of those fixed ips.

Fixes https://github.com/metallb/metallb/issues/1753

### DIFF
--- a/charts/metallb/values.yaml
+++ b/charts/metallb/values.yaml
@@ -318,7 +318,7 @@ speaker:
     enabled: false
     image:
       repository: quay.io/frrouting/frr
-      tag: 7.5.1
+      tag: 8.4.2
       pullPolicy:
     metricsPort: 7473
     resources: {}
--- a/internal/bgp/frr/frr.go
+++ b/internal/bgp/frr/frr.go
@@ -286,8 +286,10 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
 			switch family {
 			case ipfamily.IPv4:
 				rout.ipV4Prefixes[prefix] = prefix
+				neighbor.HasV4Advertisements = true
 			case ipfamily.IPv6:
 				rout.ipV6Prefixes[prefix] = prefix
+				neighbor.HasV6Advertisements = true
 			}
 		}
 	}
--- a/internal/config/validation.go
+++ b/internal/config/validation.go
@@ -7,6 +7,7 @@ import (
 
 	"github.com/pkg/errors"
 	metallbv1beta2 "go.universe.tf/metallb/api/v1beta2"
+	"go.universe.tf/metallb/internal/ipfamily"
 )
 
 type Validate func(ClusterResources) error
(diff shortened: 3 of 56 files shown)


internal/config/validation.go
```diff
@@ -96,6 +97,53 @@ func DiscardNativeOnly(c ClusterResources) error {
 	return nil
 }
 
+// validateConfig is meant to validate all the inter-dependencies of a parsed configuration.
+// In this case, we ensure that bfd echo is not enabled on a pool.
```
oribon, line 101: nit: on a v6 pool?
also, this being called "validateConfig" but being dedicated to a specific dependency feels a bit odd, wdyt about adding another function specifically for the bfd echo and calling it from validateConfig?
it would make validateConfig seem silly (just calling another function) but for long term when we find more inter-dependencies it makes sense to keep this function clean

internal/config/validation.go
```diff
@@ -96,6 +97,53 @@ func DiscardNativeOnly(c ClusterResources) error {
...
+func validateConfig(cfg *Config) error {
+outer:
+	for _, p := range cfg.Pools.ByName {
+		containsV6 := false
+		for _, cidr := range p.CIDR {
+			if ipfamily.ForCIDR(cidr) == ipfamily.IPv6 {
+				containsV6 = true
+				break
+			}
+			if !containsV6 { // we only care about v6 advertisements
```
oribon, line 111: not sure I understand, if the first cidr is ipv4 and the next is ipv6 we continue to the next pool?
I thought any pool that has an ipv6 cidr should be processed - meaning removing the outer loop, having this if outside of the cidrs loop? can you leave a comment there what exactly is a config we're trying to avoid?

internal/config/validation.go
```diff
@@ -96,6 +97,53 @@ func DiscardNativeOnly(c ClusterResources) error {
...
+				continue outer
+			}
+		}
+		for _, a := range p.BGPAdvertisements {
+			if len(a.Peers) == 0 {
```
oribon, line 116: can you add a comment here please (why do we care about an advertisement that will not be advertised)?

internal/config/validation.go
```diff
@@ -96,6 +97,53 @@ func DiscardNativeOnly(c ClusterResources) error {
...
+
+func hasBFDEcho(peer *Peer, bfdProfiles map[string]*BFDProfile) bool {
+	profile, ok := bfdProfiles[peer.BFDProfile]
+	if !ok {
+		return false
+	}
+	if profile.EchoMode {
+		return true
+	}
+	return false
```
oribon, line 144: super nit: return profile.EchoMode instead?

charts/metallb/values.yaml
```diff
@@ -318,7 +318,7 @@ speaker:
     enabled: false
     image:
       repository: quay.io/frrouting/frr
-      tag: 7.5.1
```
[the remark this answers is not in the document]
fedepaol: this is required because we want to keep the 7.5 format of the configuration.
The frr versoin statement declares the version of the configuration we want to apply

internal/bgp/frr/frr.go
```diff
@@ -286,8 +286,10 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
 			switch family {
 			case ipfamily.IPv4:
 				rout.ipV4Prefixes[prefix] = prefix
+				neighbor.HasV4Advertisements = true
```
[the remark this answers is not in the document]
fedepaol: It's the same as appending to advertisements. This flag is a shortcut to avoid looping over advertisements in order to understand if we need to add the deny any section or not, but it is not adding any information per se.
The map is being built every time we call this function, and what we care about is whether the peer has at least one v4 (and v6) advertisement in order to do the proper rendering

internal/config/validation.go
fedepaol: it will be advertised. len(peers) == 0 means all peers. Commenting

internal/config/validation.go
fedepaol: not sure I like it more, but I can change it

internal/config/validation.go
fedepaol: no this is wrong, let me fix it

oribon (approved):
lgtm

PR #1840, opened 2023-02-20
Pin the controller-gen version used by tasks

fedepaol:
Bumping the manifests relies on the version that is installed on the
local machine. This means, calling it on different machines might lead
to different results, having inconsistently generated files.
Here we install the binary to a local path with a pinned version and we
consume it from there.

Also, bumping the release version in the release notes to avoid the need
of having another pr queued.


liornoy:
lgtm!

PR #1849, opened 2023-02-28
Trigger info event for svc ip cleanup

pperiyasamy:
When IPAddressPool is modified to be compatible with another namespace or service, then existing assigned IPs are no more valid for existing namespace/service. The convergeBalancer kicks in for service which cleans up assigned IP and attempts to allocate IPs from another compatible IP address pool. This commit ensures informational event is triggered to notify about cause for clean up of assigned IP for the service.


fedepaol:
This won't hurt, although it can happen (afaik) only after a controller restart.

PR #1857, opened 2023-03-09
ServiceMonitor and PodMonitor should accept string

caruccio:
Hello there.

This is what happens when setting `prometheus.serviceMonitor.interval=300`:

```
Helm install failed: ServiceMonitor.monitoring.coreos.com "metallb-speaker-monitor" is invalid: spec.endpoints[0].interval: Invalid value: "integer": spec.endpoints[0].interval in body must be of type string: "integer"
```

Because of this:

```
$ kubectl explain ServiceMonitor.spec.endpoints.interval
KIND:     ServiceMonitor
VERSION:  monitoring.coreos.com/v1

FIELD:    interval <string>

DESCRIPTION:
     Interval at which metrics should be scraped
```


caruccio:
This is what happens when setting `prometheus.serviceMonitor.interval=300`:

    Helm install failed: ServiceMonitor.monitoring.coreos.com "metallb-speaker-monitor" is invalid: spec.endpoints[0].interval: Invalid value: "integer": spec.endpoints[0].interval in body must be of type string: "integer"

fedepaol:
thanks!

fedepaol:
are you sure that with this change it will work? Asking because I am not sure the rendering of the chart would be different using an integer or a string (meaning, not sure that with a string helm will add surrounding "" )

caruccio:
At least on my test it works. Looks like the yaml parser gives precedency to string values.

```
$ git diff
diff --git a/charts/metallb/values.schema.json b/charts/metallb/values.schema.json
index 9d3516d0..51a19f73 100644
--- a/charts/metallb/values.schema.json
+++ b/charts/metallb/values.schema.json
@@ -199,7 +199,7 @@
             "jobLabel": { "type": "string" },
             "interval": {
               "anyOf": [
-                { "type": "integer" },
+                { "type": "string" },
                 { "type": "null" }
               ]
             },
@@ -225,7 +225,7 @@
             "jobLabel": { "type": "string" },
             "interval": {
               "anyOf": [
-                { "type": "integer" },
+                { "type": "string" },
                 { "type": "null" }
               ]
             },

$ helm version
version.BuildInfo{Version:"v3.11.2", GitCommit:"912ebc1cd10d38d340f048efaf0abda047c3468e", GitTreeState:"clean", GoVersion:"go1.18.10"}

$ helm template metal ./ --set prometheus.serviceMonitor.enabled=true --set prometheus.serviceAccount=sa  --set prometheus.namespace=monitoring --set prometheus.serviceMonitor.interval=999s |grep 999 -C2
    - port: "metrics"
      honorLabels: true
      interval: 999s
  jobLabel: "app.kubernetes.io/name"
  namespaceSelector:
--
  endpoints:
    - port: "metrics"
      interval: 999s
      honorLabels: true
```

caruccio:
My assumption on string precedency was wrong:

```
$ helm template metal ./ --set prometheus.serviceMonitor.enabled=true --set prometheus.serviceAccount=sa  --set prometheus.namespace=monitoring --set prometheus.serviceMonitor.interval=999 |grep 999 -C2
metallb:
- prometheus.serviceMonitor.interval: Must validate at least one schema (anyOf)
- prometheus.serviceMonitor.interval: Invalid type. Expected: string, given: integer
```

Looks like seting it to accept both integer and string, and forcing quotes on templates is the best approach. Do you agree?

fedepaol:
> Looks like seting it to accept both integer and string, and forcing quotes on templates is the best approach. Do you agree?

yep

PR #1863, opened 2023-03-17
FRR Mode: provide a way to append a piece of configuration

fedepaol:
Currently it is not possible to run multiple FRR instances on the same
node, due to port clashing in first instance.
In order to provide a way to overcome this issue, we provide a _non
supported - use at your own risk_ way to append a fixed configuration
right after the one generated by metallb. This can be used in short term
to mitigate some of the issues reported.

This solution will be deprecated in the long term, being replaced by a
more cohesive one.


e2etest/bgptests/bgp.go
```diff
@@ -977,70 +977,98 @@ var _ = ginkgo.Describe("BGP", func() {
...
-		resources := metallbconfig.ClusterResources{
-			Pools: []metallbv1beta1.IPAddressPool{
-				{
-					ObjectMeta: metav1.ObjectMeta{
-						Name: "rejectroutes",
-					},
-					Spec: metallbv1beta1.IPAddressPoolSpec{
-						Addresses: []string{
-							addressesRange,
+	ginkgo.Context("FRR validate reload feedback", func() {
```
oribon, line 980: nit: probably unrelated(ish) name

internal/k8s/k8s.go
```diff
@@ -422,3 +423,11 @@ func UseEndpointSlices(kubeClient kubernetes.Interface) bool {
 	}
 	return true
 }
+
+func (c *Client) GetConfigMap(name, namespace string) (map[string]string, error) {
```
oribon, line 427: nit: I think this is unused

internal/bgp/bgp.go
```diff
@@ -78,4 +78,5 @@ type SessionParameters struct {
 type SessionManager interface {
 	NewSession(logger log.Logger, args SessionParameters) (Session, error)
 	SyncBFDProfiles(profiles map[string]*config.BFDProfile) error
+	SyncExtraInfo(extras string) error
```
oribon, line 81: nit: why not ExtraConfig?

fedepaol: because this is implementation independent, so in the frr case we use it for adding a piece of configuration, but it might not always be the case. I wanted to keep the name a bit higher level

PR #1878, opened 2023-03-29
Add an assignement test that uses the namespace labels

fedepaol:
All the current assignement tests leverage the namespace list. Here we add a test validating that the label selector works too.


fedepaol:
@pperiyasamy please take a look

e2etest/pkg/k8s/namespace.go
```diff
@@ -0,0 +1,47 @@
...
+	backoff := wait.Backoff{
+		Steps:    5,
+		Duration: 10 * time.Second,
+		Factor:   2,
+	}
+	err := cs.CoreV1().Namespaces().Delete(context.Background(), name, metav1.DeleteOptions{})
+	if err != nil {
+		return err
+	}
+	err = wait.ExponentialBackoff(backoff, func() (bool, error) {
```
pperiyasamy, line 36: why is backoff needed for ns delete ? is it because of service delete take a while in the test ?

pperiyasamy:
LGTM

PR #1881, opened 2023-03-31
E2E Tests: rewrite the check routes injected test

fedepaol:
We want to verify that each container is able to send routes and that changing the configmap before the configuration is applied and after the configuration is applied still works.

So, we split the all in one test in a baseline where we check metallb should not accept incoming routes, and to another one where we validate each external container is able to send routes.


liornoy:
Looks like the Routes function needs to be refactored to query VRFs as well.

fedepaol:
> Looks like the Routes function needs to be refactored to query VRFs as well.

yes I am working on the fix

liornoy:
lgtm. splitting the tests was a good idea IMO.

PR #1900, opened 2023-04-20
Pin the kubectl version

fedepaol:
Kubectl is used to generate the manifests (along with the kustomize extension).
Here we pin to a version and we consume a local copy to make the dev environment consistent across machines (and CI)


config/manifests/metallb-native.yaml
```diff
@@ -533,6 +534,7 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.11.1
+  creationTimestamp: null
```
mlguerrero12, line 537: introduced with the v1.27 version of kubectl I suppose?

tasks.py
```diff
@@ -293,7 +288,7 @@ def generate_manifest(ctx, crd_options="crd:crdVersions=v1", bgp_type="native",
         layer=bgp_type
         if with_prometheus:
             layer = "prometheus-" + layer
-        res = run("kubectl kustomize config/{} > {}".format(layer, output))
+        res = run("{} kustomize config/{} > {}".format(kubectl_path, layer, output))
```
mlguerrero12, line 291: I would use f-strings instead. Code is more readable. But this is fine as well of course.

danielmellado (commented):
Just double-checking this. Should we assume that the arch will always be amd64 and for linux? Are there any other supported archs for metallb? If so, this need to be changed.

danielmellado (commented):
Just some minor nit, stated above ;)

tasks.py
```diff
@@ -873,3 +868,23 @@ def deployprometheus(ctx):
     print("Waiting for prometheus pods to be running")
     run("kubectl -n monitoring wait --for=condition=Ready --all pods --timeout 300s")
 
+
+def _fetch_kubectl():
+    if not os.path.exists(build_path):
+        os.makedirs(build_path, mode=0o750)
+    curl_command = "curl -o {} -LO https://dl.k8s.io/release/{}/bin/linux/amd64/kubectl".format(kubectl_path, kubectl_version)
```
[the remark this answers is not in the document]
fedepaol: this is about the dev machine and you are right, maybe this is not a good assumption because people might use a mac (or win) to work on metallb.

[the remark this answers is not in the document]
fedepaol: done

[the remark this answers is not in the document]
danielmellado: Thanks, lgtm ;)

tasks.py
```diff
@@ -873,3 +868,22 @@ def deployprometheus(ctx):
...
+    version = run("{} version --short".format(kubectl_path), warn=True).stdout
+    for line in version.splitlines():
+        if line.startswith("Client Version:"):
+            version = line.split(":")[1].strip()
+            if version == kubectl_version:
+                return
+    run(curl_command)
+    run("chmod +x {}".format(kubectl_path))
+
+_fetch_kubectl()
```
mlguerrero12, line 888: why are you calling this function here? I see that other helper functions are called where needed. To keep consistency, I would expect this to be called inside dev_env itself.

fedepaol: because I don't want to bloat every function with a new line, as opposed to check the version every time we call env (and not doing anything if there is nothing to do)

tasks.py
```diff
@@ -873,3 +868,22 @@ def deployprometheus(ctx):
     print("Waiting for prometheus pods to be running")
     run("kubectl -n monitoring wait --for=condition=Ready --all pods --timeout 300s")
 
+def _fetch_kubectl():
+    if not os.path.exists(build_path):
+        os.makedirs(build_path, mode=0o750)
+    curl_command = "curl -o {} -LO https://dl.k8s.io/release/{}/bin/$(go env GOOS)/$(go env GOARCH)/kubectl".format(kubectl_path, kubectl_version)
```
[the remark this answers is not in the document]
fedepaol: as you said, it's "mainly". generatemanifests uses it too for example. And I don't want to ask the user to put hte architecture every time as opposed to automatically discovering it.

tasks.py
fedepaol: as a second thought, let me recheck how many functions are calling it

fedepaol: changed

tasks.py
```diff
@@ -284,6 +279,7 @@ def validate_kind_version():
         raise Exit(message="kind version >= {} required".format(min_version))
 
 def generate_manifest(ctx, crd_options="crd:crdVersions=v1", bgp_type="native", output=None, with_prometheus=False):
+    _fetch_kubectl()
```
mlguerrero12, line 282: I'd add it in the task generatemanifests instead.

tasks.py
```diff
@@ -422,6 +419,7 @@ def dev_env(ctx, architecture="amd64", name="kind", protocol=None, frr_volume_di
 # Configure MetalLB in the dev-env for layer2 testing.
 # Identify the unused network address range from kind network and used it in configmap.
 def layer2_dev_env():
+    _fetch_kubectl()
```
mlguerrero12, line 422: no needed. Part of dev_env task

tasks.py
```diff
@@ -432,12 +430,13 @@ def layer2_dev_env():
...
         f.write(layer2_config)
     # Apply the MetalLB ConfigMap
-    run("kubectl apply -f %s/config.yaml" % dev_env_dir)
+    run("{} apply -f {}/config.yaml".format(kubectl_path, dev_env_dir))
 
 # Configure MetalLB in the dev-env for BGP testing. Start an frr based BGP
 # router in a container and configure MetalLB to peer with it.
 # See dev-env/bgp/README.md for some more information.
 def bgp_dev_env(ip_family, frr_volume_dir):
+    _fetch_kubectl()
```
mlguerrero12, line 439: same as above

tasks.py
fedepaol: I thought about it, but I prefer to be on the safe side and this is idempotent, so doesn't hurt.
This will make this function future proof in case we start using it from elsewhere.

tasks.py
fedepaol: same

PR #1910, opened 2023-04-26
E2E: extend the "inject" test with prefix lists

fedepaol:
Users requiring routes to come from outside may need to filter only allowed routes. Here we validate that it works.

### DIFF
--- a/e2etest/bgptests/bgp.go
+++ b/e2etest/bgptests/bgp.go
@@ -1036,7 +1036,7 @@ var _ = ginkgo.Describe("BGP", func() {
 
 			for _, c := range FRRContainers {
 				err := frrcontainer.PairWithNodes(cs, c, pairingIPFamily, func(frr *frrcontainer.FRR) {
-					frr.NeighborConfig.ToAdvertise = toInject
+					frr.NeighborConfig.ToAdvertise = []string{toInject}
 				})
 				framework.ExpectNoError(err)
 			}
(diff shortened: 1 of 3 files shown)


e2etest/bgptests/bgp.go
```diff
@@ -1108,11 +1113,17 @@ var _ = ginkgo.Describe("BGP", func() {
 					if err == nil {
 						return fmt.Errorf("route not injected from %s", c.Name)
 					}
+					err = checkRouteInjected(speakerPods, pairingIPFamily, toFilter, c.RouterConfig.VRF)
```
mlguerrero12, line 1116: not in the scope of this PR but I'd change the name to isRouteInjected and the return value to be a boolean. Based on that you can return an error in the closure. It's very misleading the way it is (return error when route is injected).

mlguerrero12:
lgtm

e2etest/bgptests/bgp.go
```diff
@@ -1108,11 +1113,17 @@ var _ = ginkgo.Describe("BGP", func() {
...
+					if err != nil {
+						return fmt.Errorf("to filter route injected from %s, %w", c.Name, err)
```
liornoy, line 1118: tiny nit
```suggestion
						return fmt.Errorf("failed to filter route injected from %s, %w", c.Name, err)
```

liornoy:
I wonder if we should be more explicit with the test description/title to remember that we're testing this use case of filtering addresses.
Maybe worth adding a small ginkgo.By to be more explicit?

e2etest/bgptests/bgp.go
fedepaol: No you are right, I am fixing it. This started simpler but didn't evolve in a readable way.

e2etest/bgptests/bgp.go
fedepaol: done

liornoy:
lgtm

Issue #1911, opened 2023-04-26
Remove deprecated kustomization vars


conblem:
Greetings

I just installed metallb using the kustomization variant deschribe in the docs.
This is how my config looks:
```yaml
namespace: metallb-system

resources:
  - github.com/metallb/metallb/config/native?ref=v0.13.9
  - pool.yaml
  - advertisement.yaml
```
Now my Namespace is always out of sync because of this comment added by kustomize:
<img width="1141" alt="grafik" src="https://user-images.githubusercontent.com/19853857/234691501-ec4d698e-f582-48b8-a8f8-33f6c9710beb.png">

This seems to stem from the following line in the native kustomize yaml:
[https://github.com/metallb/metallb/blob/1284a44ecb2e6bdb606285907d6d2d72008834ce/config/native/kustomization.yaml#L13](https://github.com/metallb/metallb/blob/1284a44ecb2e6bdb606285907d6d2d72008834ce/config/native/kustomization.yaml#L13)

As far as I can tell however these variables are never referenced which is why the error comment is produced. Furthermore this feature is deprecated by kustomize as can be seen here: [https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/vars/](https://kubectl.docs.kubernetes.io/references/kustomize/kustomization/vars/)

Thanks alot for the great project!
Cheers

conblem:
Sure I'll do it!

PR #1914, opened 2023-04-28
Remove deprecated kustomization vars

conblem:
This removes the deprecated kustomization vars discussed in #1911 . I used `kustomize build` and compared the output before and after this change and it seems to output the same code.


liornoy:
Looks good to me.

PR #1919, opened 2023-05-02
Update Dockerfile

wongcai:
$TZ does not work in the container, fix it

### DIFF
--- a/speaker/Dockerfile
+++ b/speaker/Dockerfile
@@ -45,7 +45,7 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
   go.universe.tf/metallb/speaker
 
 FROM docker.io/alpine:latest
-
+RUN apk add --no-cache tzdata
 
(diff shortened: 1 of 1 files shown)


fedepaol:
Can you expand on the reason why we would need $TZ in the container?

wongcai:
> Can you expand on the reason why we would need $TZ in the container?

yes, UTC always works

I found this issue since I injected env TZ though the container engine cri-o

we'd better have correct timestamp just for std logs

fedepaol:
> > Can you expand on the reason why we would need $TZ in the container?
> 
> yes, UTC always works
> 
> I found this issue since I injected env TZ though the container engine cri-o
> 
> we'd better have correct timestamp just for std logs

So this would align the logs to the local tz? Or what?
Can you add a clearer commit message? And also, it'd make sense to change the controller's dockerfile too if we decide we need this

PR #1920, opened 2023-05-03
Verify LB ip and address pool annotations compatibility

### LINKED ISSUE #1916: opposite loadBalancerIPs and address-pool annotations on a service
A service can set an ip with the annotation loadBalancerIPs that doesn't correspond to the pool set with the annotation address-pool.

From the code, when the ip is first allocated, it only takes into account the value on the loadBalancerIPs annotation. It doesn't verify if an address pool is set with the annotation address-pool.

When the svc (with an allocated ip) is updated, the controller verifies that the pool where the ip comes from is the same as the one specified in the address-pool annotation. Since the ip doesn't belong to this pool, it logs the error "user requested a different pool than the one currently assigned". It then proceeds to clear the ip (internally) and performs again an allocation. The result will be the same as when the allocation was executed from the first time.

So, the end result is deterministic, the loadBalancerIPs annotation takes precedence. However, the code and logs of the controller are not. Also, the docs do not mention this either.

Question now is, is this behavior intended? If yes, we should modify the code to prevent undesired logs and clearly state this on the docs. If not, we will need to verify that these two annotations are compatible.

mlguerrero12:
When a service requests a specific LB IP(s) and address pool via annotations, these must be compatible.

Fixes #1916

### DIFF
--- a/controller/service.go
+++ b/controller/service.go
@@ -193,6 +193,8 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error) 
 		return nil, err
 	}
 
+	desiredPool := svc.Annotations[annotationAddressPool]
+
 	// If the user asked for a specific IPs, try that.
 	if len(desiredLbIPs) > 0 {
 		if serviceIPFamily != desiredLbIPFamily {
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -6,3 +6,5 @@ dev-env/unittest/bin
 
 # generated by helm chart-releaser
 .cr-*
+
+.idea/
```
fedepaol, line 10: please add this as a separate commit

controller/service.go
```diff
@@ -201,10 +203,17 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error)
 		if err := c.ips.Assign(key, svc, desiredLbIPs, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc)); err != nil {
 			return nil, err
 		}
+
+		// Verify that ip and address pool annotations are compatible.
+		if desiredPool != "" && c.ips.Pool(key) != desiredPool {
+			c.ips.Unassign(key)
+			return nil, fmt.Errorf("requested loadBalancer IP(s) %q is not compatible with requested address pool %s", desiredLbIPs, desiredPool)
```
fedepaol, line 210: we might want to return a custom error we handle outside here. Specifically, when this happens it makes sense to clear the service. We might get here with a user adding the annotation after the lb ip. In that case we need to wipe.

.gitignore
mlguerrero12: ok

fedepaol:
lgtm

.gitignore
fedepaol: note that this doesn't necessarily have to be a separate pr, but just a separate commit in the same pr for trivial things such as this is fine.

PR #1923, opened 2023-05-04
E2E L2: Fix the NetworkUnavailable test

liornoy:
This PR includes 2 commits that modify the L2 NetworkUnavailable test
which currently uses the ARP for fetching the announcing node,
and make it use the GetSvcNode function instead, which relies on 
inspecting a "nodeAssigned" event. 

This is necessary for us to be able to test this use case on the Openshift CI
where we use a host-networked setup. 

The second commit changes the GetSvcNode function to sort the events by
the timestamp and thus getting the latest node that was assigned to the service.


e2etest/pkg/k8s/services.go
```diff
@@ -30,17 +31,22 @@ func GetSvcNode(cs clientset.Interface, svcNS string, svcName string, allNodes *
...
-	if msg == "" {
+	if len(svcEvents) == 0 {
 		return nil, errors.New("service doesn't be assigned node")
 	}
 
+	sort.Slice(svcEvents, func(i, j int) bool {
+		return svcEvents[i].LastTimestamp.Before(&svcEvents[j].LastTimestamp)
+	})
+
+	msg := svcEvents[len(svcEvents)-1].Message
```
fedepaol, line 48: can we sort the other way around and pick svcEvents[0]? I think it'd be clearer

liornoy: done

liornoy:
I am still investigating why d/s sync fails, I see the networkunavailable test case failing again.

liornoy:
/hold
I'm investigating if this is still failing the openshift CI

liornoy:
closing this one for a while until we'll get to work on the openshift/metallb sync

liornoy:
This fix is working for the downstream sync: https://github.com/openshift/metallb/pull/137

PR #1945, opened 2023-05-19
Trigger reprocessall event only when an ip has been unassigned

mlguerrero12:
This prevents the unnecessary reprocess of all services when a service with no LB IP assigned is deleted.

### DIFF
--- a/controller/main.go
+++ b/controller/main.go
@@ -51,11 +51,15 @@ func (c *controller) SetBalancer(l log.Logger, name string, svcRo *v1.Service, _
 	defer level.Debug(l).Log("event", "endUpdate", "msg", "end of service update")
 
 	if svcRo == nil {
-		c.deleteBalancer(l, name)
-		// There might be other LBs stuck waiting for an IP, so when
-		// we delete a balancer we should reprocess all of them to
-		// check for newly feasible balancers.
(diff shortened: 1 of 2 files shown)


controller/main.go
```diff
@@ -51,11 +51,14 @@ func (c *controller) SetBalancer(l log.Logger, name string, svcRo *v1.Service, _
 	defer level.Debug(l).Log("event", "endUpdate", "msg", "end of service update")
 
 	if svcRo == nil {
-		c.deleteBalancer(l, name)
-		// There might be other LBs stuck waiting for an IP, so when
-		// we delete a balancer we should reprocess all of them to
-		// check for newly feasible balancers.
-		return controllers.SyncStateReprocessAll
+		if c.ips.Unassign(name) {
```
fedepaol, line 54: Can we have a "IsAllocated" function on the allocator and have it guarding this if? In this way it is going to be clearer why we do the reprocessall (as opposed to having to rely on the fact that unassign returns false if the service is not there).

The effect would be the same but this function would be easier to understand.

fedepaol:
Nice catch!

controller/main.go
mlguerrero12: Done. More expensive to run tough.

fedepaol:
LGTM

Issue #1943, opened 2023-05-18
Add a separate go mod for tests


fedepaol:
Because we use the e2e framework, we need to pin the right k8s api version every time.
Having a separate go mod would probably result in a cleaner metallb go.mod

PR #1946, opened 2023-05-19
Move the e2e tests to a standalone, self dependent go module

fedepaol:
Currently the e2e test framework contains a lot of logic, but has some dependencies that root into metallb/internal.
This makes it unpractical (and sometimes impossible) to export the framework to be consumed by other projects (such as the frr daemon being discussed here https://github.com/metallb/metallb/pull/1933 ).

On top of that, this means that the metallb's go.mod file is polluted by dependencies that are needed only by tests. Because of this, it makes sense to split the e2e tests in a standalone module that now depends only from the metallb api (and only for the specific metallb tests). This is done by copy-pasting / reimplementing some of the dependencies from the metallb's internal logic, under the spirit that is better to have something self contained ([a little copy is better than a little dependency](https://www.youtube.com/watch?v=PAAkCSZUG1c&t=568s))

Also, fixes https://github.com/metallb/metallb/issues/1943


go.mod
```diff
@@ -48,45 +45,38 @@ require (
 	github.com/Nvveen/Gotty v0.0.0-20120604004816-cd527374f1e5 // indirect
```
mlguerrero12, line 45: there is a comment above "// pinning to specific versions is required because we are consuming
// the k8s.io/kubernetes module from the tests (via the "k8s.io/kubernetes/test/e2e/framework" package)
// See https://github.com/kubernetes/kubernetes/issues/90358#issuecomment-617859364"

Was this not supposed to delete this require block?

fedepaol: removed

e2etest/go.work
```diff
@@ -0,0 +1,6 @@
+go 1.19
+
+use (
+	./
```
mlguerrero12, line 4: I don't think you need this one?

fedepaol: we do, if we remove it we'll get
`main module (go.universe.tf/metallb) does not contain package go.universe.tf/metallb/e2etest`

So we need the current module (e2e) and the metallb one in order to consume the api not from the most recent tag but from the version we have in the repo.

mlguerrero12:
LGTM

PR #1948, opened 2023-05-19
Remove IP verifications in SetPools

### LINKED ISSUE #462: Remove the config reloading fail-static behavior

mlguerrero12:
Pool changes might invalidate the LB IPs of services. When this happens, the pool controller ignores the new configuration and continues operating with the current one. This PR modifies this behavior by always accepting the new configuration and letting the service controller to clear the IPs of the affected services.

Fixes #462

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -1040,12 +1040,23 @@ func TestControllerConfig(t *testing.T) {
 		t.Errorf("SetBalancer produced unexpected mutation (-want +got)\n%s", diff)
 	}
 
-	// Now that an IP is allocated, removing the IP pool is not allowed.
-	if c.SetPools(l, &config.Pools{ByName: map[string]*config.Pool{}}) != controllers.SyncStateError {
-		t.Fatalf("SetPools that deletes allocated IPs was accepted")
+	// Removing the IP pool should be accepted even if the IP is allocated
+	// This triggers a reprocess all event to force the service to get a new IP
+	if c.SetPools(l, &config.Pools{ByName: map[string]*config.Pool{}}) != controllers.SyncStateReprocessAll {
+		t.Fatalf("SetPools that deletes allocated IPs was not accepted")
 	}
 
-	// Deleting the config also makes MetalLB sad.
+	// Balancer should clear the service. Not reprocess all because previous ip is not longer assignable.
+	if c.SetBalancer(l, "test", gotSvc, epslices.EpsOrSlices{}) != controllers.SyncStateErrorNoRetry {
+		t.Fatalf("SetBalancer failed")
+	}
+
+	gotSvc = k.gotService(gotSvc)
+	if diff := diffService(svc, gotSvc); diff != "" {
+		t.Errorf("SetBalancer produced unexpected mutation (-want +got)\n%s", diff)
+	}
+
+	// Deleting the config makes MetalLB sad.
 	if c.SetPools(l, nil) != controllers.SyncStateErrorNoRetry {
 		t.Fatalf("SetPools that deletes the config was accepted")
 	}
@@ -1097,7 +1108,7 @@ func TestDeleteRecyclesIP(t *testing.T) {
 		},
 	}
 	if c.SetBalancer(l, "test2", svc2, epslices.EpsOrSlices{}) == controllers.SyncStateError {
-		t.Fatal("SetBalancer svc2 failed")
+		t.Fatal("SetBalancer svc2 was supposed to fail with no retry")
 	}
 	if k.gotService(svc2) != nil {
 		t.Fatal("SetBalancer svc2 mutated svc2 even though it should not have allocated")
@@ -1121,7 +1132,7 @@ func TestDeleteRecyclesIP(t *testing.T) {
 		t.Fatal("svc2 didn't get an IP")
 	}
 }
-func TestControllerDualStackConfig(t *testing.T) {
+func TestControllerReassing(t *testing.T) {
 	k := &testK8S{t: t}
 	c := &controller{
 		ips:    allocator.New(),
@@ -1129,40 +1140,15 @@ func TestControllerDualStackConfig(t *testing.T) {
 	}
 
 	l := log.NewNopLogger()
+
 	svc := &v1.Service{
 		Spec: v1.ServiceSpec{
 			Type:       "LoadBalancer",
-			ClusterIPs: []string{"1.2.3.4", "1000::"},
+			ClusterIPs: []string{"1.2.3.4"},
 		},
 	}
-	if c.SetBalancer(l, "test", svc, epslices.EpsOrSlices{}) == controllers.SyncStateError {
-		t.Fatalf("SetBalancer failed")
-	}
 
-	gotSvc := k.gotService(svc)
-	if gotSvc != nil {
-		t.Errorf("SetBalancer with no configuration mutated service (-in +out)\n%s", diffService(svc, gotSvc))
-	}
-	if k.loggedWarning {
-		t.Error("SetBalancer with no configuration logged an error")
-	}
-
-	// Set an empty config. Balancer should still not do anything to
-	// our unallocated service, and return an error to force a
-	// retry after sync is complete.
-	if c.SetPools(l, &config.Pools{ByName: map[string]*config.Pool{}}) == controllers.SyncStateError {
-		t.Fatalf("SetPools with empty config failed")
-	}
-
-	gotSvc = k.gotService(svc)
-	if gotSvc != nil {
-		t.Errorf("unsynced SetBalancer mutated service (-in +out)\n%s", diffService(svc, gotSvc))
-	}
-	if k.loggedWarning {
-		t.Error("unsynced SetBalancer logged an error")
-	}
-
-	// Set a config with some IPs. Still no allocation, not synced.
+	// Set a config with some IPs.
 	pools := &config.Pools{ByName: map[string]*config.Pool{
 		"default": {
 			Name:       "default",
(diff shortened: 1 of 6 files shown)


controller/main.go
```diff
@@ -74,16 +75,26 @@ func (c *controller) SetBalancer(l log.Logger, name string, svcRo *v1.Service, _
...
 	if c.convergeBalancer(l, name, svc) != nil {
 		syncStateRes = controllers.SyncStateErrorNoRetry
 	}
 
-	if wasAllocated && !c.isServiceAllocated(name) { // convergeBalancer may deallocate our service and this means it did it.
-		// if the service was deallocated, it may have have left room
-		// for another one, so we reprocess
-		level.Info(l).Log("event", "serviceUpdated", "msg", "removed loadbalancer from service, services will be reprocessed")
-		syncStateRes = controllers.SyncStateReprocessAll
+	if wasAllocated && !c.isServiceAllocated(name) {
```
fedepaol, line 89: not related to this pr, but the deepequal below should probably go before this.

controller/controller_test.go
```diff
@@ -1040,12 +1040,22 @@ func TestControllerConfig(t *testing.T) {
 		t.Errorf("SetBalancer produced unexpected mutation (-want +got)\n%s", diff)
 	}
 
-	// Now that an IP is allocated, removing the IP pool is not allowed.
-	if c.SetPools(l, &config.Pools{ByName: map[string]*config.Pool{}}) != controllers.SyncStateError {
-		t.Fatalf("SetPools that deletes allocated IPs was accepted")
+	// Now that an IP is allocated, removing the IP pool should be accepted.
```
fedepaol, line 1043: nit: removing the IP pool should be accepted even if the IP is allocated, and triggers a reprocess to force the service to get a new IP

controller/controller_test.go
```diff
@@ -1040,12 +1040,22 @@ func TestControllerConfig(t *testing.T) {
...
+		t.Fatalf("SetBalancer failed")
+	}
+
+	gotSvc = k.gotService(gotSvc)
+	if diff := diffService(svc, gotSvc); diff != "" {
+		t.Errorf("SetBalancer produced unexpected mutation (-want +got)\n%s", diff)
 	}
 
-	// Deleting the config also makes MetalLB sad.
+	// Deleting the config makes MetalLB sad.
```
fedepaol, line 1058: If we allow removing one pool, we should allow removing all of them and having the services becoming empty.
Not sure if the easier way is to fill the config.Pools with empty lists / maps or getting rid of the check in SetPools

controller/controller_test.go
```diff
@@ -981,7 +981,7 @@ func TestControllerConfig(t *testing.T) {
 			ClusterIPs: []string{"1.2.3.4"},
 		},
 	}
-	if c.SetBalancer(l, "test", svc, epslices.EpsOrSlices{}) == controllers.SyncStateError {
+	if c.SetBalancer(l, "test", svc, epslices.EpsOrSlices{}) != controllers.SyncStateSuccess {
```
fedepaol, line 984: makes sense to change the comment and log the different result (same below)?

fedepaol:
Did a first pass. Still need to go over part of the tests, plus it will probably make sense to cover this with some e2e (still need to think about it).

controller/controller_test.go
mlguerrero12: Yes, this needs to be applied to all places where setBalancer is used. I'm planning to create a PR only for this, addressing the various return codes and logs. The unit tests above (not affected by setPool changed) use setBalancer as well.

controller/controller_test.go
```diff
@@ -1040,12 +1040,22 @@ func TestControllerConfig(t *testing.T) {
...
+		t.Fatalf("SetPools that deletes allocated IPs was not accepted")
+	}
+
+	// Balancer should clear the service. Not reprocess all because previous ip is not longer assignable.
+	if c.SetBalancer(l, "test", gotSvc, epslices.EpsOrSlices{}) != controllers.SyncStateErrorNoRetry {
+		t.Fatalf("SetBalancer failed")
+	}
+
+	gotSvc = k.gotService(gotSvc)
+	if diff := diffService(svc, gotSvc); diff != "" {
```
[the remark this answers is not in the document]
mlguerrero12: gotSvc should be cleared indeed. The svc var is always cleared.

controller/main.go
```diff
@@ -74,16 +75,26 @@ func (c *controller) SetBalancer(l log.Logger, name string, svcRo *v1.Service, _
 	// a reason.
 	svc := svcRo.DeepCopy()
 	syncStateRes := controllers.SyncStateSuccess
+
+	var prevIPs []net.IP
```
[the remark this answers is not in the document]
mlguerrero12: The current logic does that. If the IPs are erased, it forces a retry all. What I want to do here is to prevent a unnecessary retry all if the previous IP is not valid anymore. It was impossible to have this scenario before because of the logic in setPools, but now it is.

Now, when the service finds another IP, setBalancer does not force a retry all. This is the case in the current logic as well. So, the comparison of IPs is something we could do to improve the code since in theory, the previous ip can be allocated somewhere else. But note that this is out of scope of this PR.

controller/main.go
mlguerrero12: indeed

controller/service.go
```diff
@@ -50,6 +50,13 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 		return nil
 	}
 
+	// Return if pools are empty.
+	if len(c.pools.ByName) == 0 {
```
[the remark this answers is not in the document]
mlguerrero12: it is, I just wanted to have a fast return now that it is possible to empty the pools.

controller/service.go
```diff
@@ -234,6 +241,14 @@ func (c *controller) isServiceAllocated(key string) bool {
 	return c.ips.Pool(key) != ""
 }
 
+func (c *controller) allocatedIPs(svc string) []net.IP {
```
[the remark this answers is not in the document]
mlguerrero12: not really, I removed it. Thanks.

controller/service.go
```diff
@@ -234,6 +241,14 @@ func (c *controller) isServiceAllocated(key string) bool {
...
+	return c.ips.IPs(svc)
+}
+
+func (c *controller) isIPAssignable(ips []net.IP) bool {
```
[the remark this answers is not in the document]
mlguerrero12: correct, I should use another word instead of assignable. But the idea is to verify that the ip is still part of a pool. Will delete the wrapper. I wanted to have something similar to the method 'isServiceAllocated' but it's not really needed.

e2etest/l2tests/assignment.go
```diff
@@ -146,6 +147,94 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+
+			gomega.Eventually(func() error {
+				svc1, err := cs.CoreV1().Services(svc1.Namespace).Get(context.Background(), svc1.Name, metav1.GetOptions{})
+				framework.ExpectNoError(err)
+				err = config.ValidateIPInRange([]metallbv1beta1.IPAddressPool{pool2}, e2eservice.GetIngressPoint(
+					&svc1.Status.LoadBalancer.Ingress[0]))
+				return err
+			}, time.Minute, 1*time.Second).ShouldNot(gomega.HaveOccurred())
+		})
+
```
liornoy, line 235: nit: extra newline

e2etest/l2tests/assignment.go
```diff
@@ -146,6 +147,94 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+				&svc1.Status.LoadBalancer.Ingress[0]))
+			framework.ExpectNoError(err)
+
+			ginkgo.By("deleting pool 1")
+			p := &metallbv1beta1.IPAddressPool{}
+			err = ConfigUpdater.Client().Get(context.Background(), client.ObjectKey{Namespace: ConfigUpdater.Namespace(), Name: pool1.Name}, p)
+			framework.ExpectNoError(err)
+			err = ConfigUpdater.Client().Delete(context.Background(), p)
+			framework.ExpectNoError(err)
+
```
liornoy, line 226: Maybe for clarity add here:
`ginkgo.By("validate LoadBalancer IP is re-allocated from pool2")`

e2etest/l2tests/assignment.go
```diff
@@ -146,6 +147,94 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+			ginkgo.By("validate LoadBalancer IP is allocated from pool1")
+			err := config.ValidateIPInRange([]metallbv1beta1.IPAddressPool{pool1}, e2eservice.GetIngressPoint(
+				&svc1.Status.LoadBalancer.Ingress[0]))
+			framework.ExpectNoError(err)
+
+			ginkgo.By("deleting all pools")
+			err = ConfigUpdater.Client().DeleteAllOf(context.Background(), &metallbv1beta1.IPAddressPool{}, client.InNamespace(ConfigUpdater.Namespace()))
+			framework.ExpectNoError(err)
+
```
liornoy, line 201: Maybe for clarity add here:
ginkgo.By("validate LoadBalancer IP is removed from the svc")

liornoy (commented):
small nits.

e2etest/l2tests/assignment.go
mlguerrero12: thanks

mlguerrero12:
> small nits.

@liornoy, thanks. I addressed them.

controller/main.go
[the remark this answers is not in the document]
fedepaol: I understand that. I am talking about untangling the code because now we have wasAllocated and prevIPs. prevIPs already contains the information that hte service wasAllocated, so we don't need it anymroe. We can  make the logic only on prevIPs + PoolForIP

controller/main.go
fedepaol: Let's do it then :-) A separate commit in this PR is fine, even if it's out of scope.

What we normally do is, if we find some cleanup out of scope, it is fine to do it contextually as long as it belongs to a separate commit. This is possible because we don't squash so each entry will be separated.

controller/service.go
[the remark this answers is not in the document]
fedepaol: Fair enough

controller/main.go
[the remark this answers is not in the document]
mlguerrero12: ok, I see what you mean. Will change it. Thanks.

controller/controller_test.go
mlguerrero12: will revert these changes and create another PR to improve the unit tests of controller_test.go

controller/main.go
mlguerrero12: ok

controller/controller_test.go
mlguerrero12: done

controller/main.go
mlguerrero12: done

fedepaol (changes requested):

fedepaol:
Sorry for the delay in the review, I left one last comment about keeping the map aligned when we reprocess the configuration. Also, @oribon can you have a look?

controller/controller_test.go
```diff
@@ -1121,87 +1132,98 @@ func TestDeleteRecyclesIP(t *testing.T) {
 		t.Fatal("svc2 didn't get an IP")
 	}
 }
-func TestControllerDualStackConfig(t *testing.T) {
```
liornoy, line 1124: don't we want to keep testing the dual-stack scenario?
I think that with the changes you introduce, we're removing the 
check that the controller assigns dual stack ip for a service.

oribon:
lgtm other than the unresolved comments

PR #1950, opened 2023-05-22
Simplify pool allocated search

mlguerrero12:
The pool is directly obtained from the alloc structure instead of looping through all pools.

### DIFF
--- a/internal/allocator/allocator.go
+++ b/internal/allocator/allocator.go
@@ -350,10 +350,7 @@ func (a *Allocator) isPoolCompatibleWithService(p *config.Pool, svc *v1.Service)
 // service has no IP allocated, "" is returned.
 func (a *Allocator) Pool(svc string) string {
 	if alloc := a.allocated[svc]; alloc != nil {
-		pool := poolFor(a.pools.ByName, alloc.ips)
-		if pool != nil {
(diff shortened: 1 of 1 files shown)


liornoy:
lgtm

internal/allocator/allocator.go
```diff
@@ -350,10 +350,8 @@ func (a *Allocator) isPoolCompatibleWithService(p *config.Pool, svc *v1.Service)
 // service has no IP allocated, "" is returned.
 func (a *Allocator) Pool(svc string) string {
 	if alloc := a.allocated[svc]; alloc != nil {
-		pool := poolFor(a.pools.ByName, alloc.ips)
-		if pool != nil {
-			return pool.Name
-		}
+		// Should never be an allocation without a pool.
```
fedepaol, line 353: Let's remove this comment. I think it's safe to assume this is always the case, so the comment will just add friction to the reader who doesn't know the previous version and will doubt there are extra cases.

If we don't completely trust alloc always containing a pool we might consider logging or even panicking.

mlguerrero12: will remove it

mlguerrero12: done

PR #1951, opened 2023-05-22
Trigger reprocess all event only when pool changes

mlguerrero12:
This prevents the pool controller to trigger the reprocess all event when there are no changes in the pools.

### DIFF
--- a/internal/k8s/controllers/pool_controller.go
+++ b/internal/k8s/controllers/pool_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"reflect"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
@@ -41,6 +42,7 @@ type PoolReconciler struct {
 	Handler        func(log.Logger, *config.Pools) SyncState
 	ValidateConfig config.Validate
 	ForceReload    func()
+	currentConfig  *config.Config
 }
 
 func (r *PoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
@@ -89,6 +91,10 @@ func (r *PoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 	}
 
 	level.Debug(r.Logger).Log("controller", "PoolReconciler", "rendered config", dumpConfig(cfg))
+	if reflect.DeepEqual(r.currentConfig, cfg) {
+		level.Debug(r.Logger).Log("controller", "PoolReconciler", "event", "configuration did not change, ignoring")
+		return ctrl.Result{}, nil
+	}
 
 	res := r.Handler(r.Logger, cfg.Pools)
 	switch res {
@@ -106,6 +112,8 @@ func (r *PoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
 		level.Error(r.Logger).Log("controller", "PoolReconciler", "metallb CRs and Secrets", dumpClusterResources(&resources), "event", "reload failed, no retry")
 		return ctrl.Result{}, nil
 	}
+
+	r.currentConfig = cfg
 
 	configLoaded.Set(1)
 	configStale.Set(0)


controller/main.go
```diff
@@ -125,8 +125,13 @@ func (c *controller) SetPools(l log.Logger, pools *config.Pools) controllers.Syn
 		level.Error(l).Log("op", "setConfig", "error", err, "msg", "applying new configuration failed")
 		return controllers.SyncStateError
 	}
-	c.pools = pools
-	return controllers.SyncStateReprocessAll
+
+	if !reflect.DeepEqual(c.pools, pools) {
+		c.pools = pools
+		return controllers.SyncStateReprocessAll
+	}
```
liornoy, line 132: Maybe we should check the other way around, if is DeepEqual then return SyncStateSuccess,
with a log similar to what we have in the SetBalancer:
`level.Debug(l).Log("event", "noChange", "msg", "service converged, no change")`
else continue to execute the c.ips.SetPools and return Reprocess.
This is just to avoid running the c.ips.SetPools on the same/unchanged pools

mlguerrero12: I had a reason to always run c.ips.SetPools but I can't remember what it was.

Will change it as you suggest. Thank you.

mlguerrero12: done

liornoy: it's just a suggestion, but maybe don't negate the DeepEqual if statement,
so we won't have to branch if inside an if and make it easier to follow.
i.e. if the pools are equal we throw a log and return success, else
continue like before.

mlguerrero12: I prefer the last return to be a success

fedepaol: +1, the happy path must be the leftmost indented one unless exceptional cases.

controller/main.go
```diff
@@ -121,12 +121,17 @@ func (c *controller) SetPools(l log.Logger, pools *config.Pools) controllers.Syn
 		return controllers.SyncStateErrorNoRetry
 	}
 
-	if err := c.ips.SetPools(pools); err != nil {
-		level.Error(l).Log("op", "setConfig", "error", err, "msg", "applying new configuration failed")
-		return controllers.SyncStateError
+	if !reflect.DeepEqual(c.pools, pools) {
```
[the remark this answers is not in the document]
mlguerrero12: yeah, that will require more changes. I would need to add a new parameter in the PoolReconciler structure to keep track of the pools. Perhaps in another PR.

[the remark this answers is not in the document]
mlguerrero12: the config controller has a filed to store the current config, the pool controller does not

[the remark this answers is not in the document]
fedepaol: Right, but it's about copying the config controller logic. 
If we want this kind of optimization, we should be consistent.

controller/main.go
liornoy: you can also say that the Reprocess state is the happy path, as it was previously,
anyways it is unclear to me in what scenario would we reconcile and call the pool controller handler when the pool didn't change

fedepaol: > it is unclear to me in what scenario would we reconcile and call the pool controller handler when the pool didn't change

This is a valid point too. The only scenario I can think of is when we do that is because a community entry changes but is not part of a legacy addresspool, which is a pretty cornery case.

mlguerrero12: yeah, mainly changes on community and namespaces not related to existing pools

liornoy: but the community resource is translated into a BGPadv (both legacy and non) and that's part of the pool, no?

mlguerrero12: community affects the legacy pools. BGPadv are part of the pools indeed but in the pools controller they are always empty

controller/main.go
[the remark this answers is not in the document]
mlguerrero12: I agree with the consistency. But perhaps the config controller should be the one that needs to be changed. I don't know. I'll study it soon. The only thing I don't like is to have (and maintain) many variables with pool. We have pools in the controller structure, allocator structure and now we are thinking about adding a new one for the pool controller structure.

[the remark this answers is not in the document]
fedepaol: I am suggesting not to keep only the pool, but do exactly the same as we do in the config controller, which is keeping an old version of the config and compare the rendered one (which yes, it will contain the pools).

I don't think it's a problem from readibility point of view, the config controller seems straightforward. We maintain a current config, check if changed and discard if it did not.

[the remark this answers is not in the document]
mlguerrero12: I changed it as you suggested

fedepaol:
lgtm, thanks!

PR #1965, opened 2023-06-01
fix: add missing namespace to helm chart templates

### LINKED ISSUE #1964: Helm chart templates are missing the namespace
I recently migrated from kustomize to helm charts for metallb. While the kutomize manifests had namespace (https://github.com/metallb/metallb/blob/e708fc9010afb80c379bdfadd5d543931164ca5e/config/native/kustomization.yaml#L10), the helm chart templates are missing namespaces for namespace scoped resources when generating manifests using helm templates (We generate manifests using helm and then apply using kubectl).
For example: https://github.com/metallb/metallb/blob/e708fc9010afb80c379bdfadd5d543931164ca5e/charts/metallb/templates/rbac.yaml#L109-L111

ankitm123:
Fixes #1964

### DIFF
--- a/charts/metallb/templates/controller.yaml
+++ b/charts/metallb/templates/controller.yaml
@@ -3,6 +3,7 @@ apiVersion: apps/v1
 kind: Deployment
 metadata:
   name: {{ template "metallb.fullname" . }}-controller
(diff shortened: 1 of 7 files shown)


ankitm123:
I will update the PR this weekend.

PR #1968, opened 2023-06-04
Add "ginkgo_params" flag to the e2etests task

### LINKED ISSUE #1955: e2etest task: provide additional "ginkgo_params" flag
Add additional "ginkgo_params" flag in the e2etest task in tasks.py
to allow developers to insert custom Ginkgo flags to pass the e2e tests.

An example for a use case is: as a developer, I want to invoke the e2e test with ginkgo's `--until-it-fails` flag
so that I could validate that my test case is constantly passing.

liornoy:
This commit adds a "ginkgo_params" flag to the e2etests task. This addition is to enable developers to insert ginkgo params that are not already covered by existing parameters (focus and skip)

A simple example of this feature is running the e2etests with the --until-it-fails flag, which is handy for catching flaky tests.

The implementation injects the value of this new parameter in the ginkgo test run command. and It's possible to add numerous ginkgo params for example:
`invoke e2etest --ginkgo-params="--until-it-fails -v --focus=L2"`

Also, added a mention of this new flag in the e2etests/README.md.

Fixed #1955

### DIFF
--- a/e2etest/README.md
+++ b/e2etest/README.md
@@ -24,6 +24,11 @@ Run only L2 test suite:
 inv e2etest --focus L2
 ```
 
+Run with additional ginkgo parameters for example:
+```
+inv e2etest --ginkgo-params="--until-it-fails -v"
+```
+
 The test suite will run the appropriate tests against the cluster.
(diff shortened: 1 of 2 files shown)


fedepaol:
LGTM

PR #1973, opened 2023-06-06
BGP: invoke SyncBFDProfiles on empty lists too

fedepaol:
When a bfd profile is deleted, the sync is not called and we leave the bfd profile filled even if it was deleted on the CR side.

Additionally, we add a unit test to validate the deletion works properly.


oribon (approved):
lgtm

PR #1974, opened 2023-06-06
Release: override the tag of the copy init containers

fedepaol:
When doing the release process, the sidecar containers *-cp were not being tagged properly, with the result of having the kustomize - manifests based deployment using the "main" tag.

Here we fix and we change the bumprelease command to verify we didn't leave any :main tag around.


oribon (approved):
lgtm

PR #1979, opened 2023-06-13
Fix go list errors

mlguerrero12:


mlguerrero12:
yes, we have broken it 2 times. It makes sense to have a check in CI

mlguerrero12:
will do it

fedepaol:
> will do it

thanks!

PR #1988, opened 2023-06-22
Add go list check in CI

mlguerrero12:


tasks.py
```diff
@@ -889,6 +889,13 @@ def gomodtidy(ctx):
     if not res.ok:
         raise Exit(message="go mod tidy failed")
 
+@task
+def golist(ctx):
+    """Runs go list"""
+    res = run("go list -modfile=go.mod -m -json -mod=mod all", hide="out")
```
fedepaol, line 895: I'd remove the -json flag here

mlguerrero12: done

mlguerrero12:
@fedepaol, any idea why this is failing? This doesn't seem to be correct.

#### Uncommitted files found, you may need to run inv generatemanifests ####

.github/workflows/ci.yaml
```diff
@@ -78,6 +78,7 @@ jobs:
           inv verifylicense
           inv gomodtidy
           inv checkchanges --action="run inv gomodtidy"
+          inv golist
```
fedepaol, line 81: you must follow this with an `inv checkchanges --action="run inv golist"`

fedepaol:
> @fedepaol, any idea why this is failing? This doesn't seem to be correct.
> 
> #### Uncommitted files found, you may need to run inv generatemanifests

The comment is wrong because we missed a check (left a comment). The output shows that go.sum in not committed (and if I try to run inv golist locally it shows a changed go.sum)

mlguerrero12:
> > @fedepaol, any idea why this is failing? This doesn't seem to be correct.
> > #### Uncommitted files found, you may need to run inv generatemanifests
> 
> The comment is wrong because we missed a check (left a comment). The output shows that go.sum in not committed (and if I try to run inv golist locally it shows a changed go.sum)

Go.sum didn't change for me. Will check again. Thanks.

mlguerrero12:
Still the same. However, when I deleted the content of the go.sum and ran again inv golist, there were some changes. Let's see.

PR #1990, opened 2023-06-27
Make the configuration conversion stable

fedepaol:
When translating the configuration coming from the k8s api to the internal one, the outcome depends on the order in which the cache returns us the various items. This may cause unnecessary reloads.

In order to avoid this, we pre-order the list of items to feed the conversion with.
The sorting is done in a conversion function in the controller (rather than the config package) in order to make it harder to forget to add the sorting action.

### DIFF
--- a/internal/bgp/frr/frr.go
+++ b/internal/bgp/frr/frr.go
@@ -298,8 +298,8 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
 			advConfig := advertisementConfig{
 				IPFamily:         family,
 				Prefix:           prefix,
-				Communities:      communities,
-				LargeCommunities: largeCommunities,
+				Communities:      sort.StringSlice(communities),
+				LargeCommunities: sort.StringSlice(largeCommunities),
 				LocalPref:        adv.LocalPref,
 			}
 
@@ -313,6 +313,7 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
 				neighbor.HasV6Advertisements = true
 			}
 		}
+		sortAdvertiesements(neighbor.Advertisements)
 	}
 
 	for _, r := range sortMap(routers) {
@@ -457,4 +458,36 @@ func sortMap[T any](toSort map[string]T) []T {
 		res = append(res, toSort[k])
 	}
 	return res
+}
+
+func sortAdvertiesements(toSort []*advertisementConfig) {
+	sort.Slice(toSort, func(i, j int) bool {
+		if toSort[i].IPFamily != toSort[j].IPFamily {
+			return toSort[i].IPFamily < toSort[j].IPFamily
+		}
+		if toSort[i].Prefix != toSort[j].Prefix {
+			return toSort[i].Prefix < toSort[j].Prefix
+		}
+		if toSort[i].LocalPref != toSort[j].LocalPref {
+			return toSort[i].LocalPref < toSort[j].LocalPref
+		}
+		if len(toSort[i].Communities) != len(toSort[j].Communities) {
+			return len(toSort[i].Communities) < len(toSort[j].Communities)
+		}
+		for k := range toSort[i].Communities {
+			if toSort[i].Communities[k] != toSort[j].Communities[k] {
+				return toSort[i].Communities[k] < toSort[j].Communities[k]
+			}
+		}
+		if len(toSort[i].LargeCommunities) != len(toSort[j].LargeCommunities) {
+			return len(toSort[i].LargeCommunities) < len(toSort[j].LargeCommunities)
+		}
+
+		for k := range toSort[i].LargeCommunities {
+			if toSort[i].LargeCommunities[k] != toSort[j].LargeCommunities[k] {
+				return toSort[i].LargeCommunities[k] < toSort[j].LargeCommunities[k]
+			}
+		}
+		return false
+	})
 }
--- a/internal/bgp/frr/frr_test.go
+++ b/internal/bgp/frr/frr_test.go
@@ -5,6 +5,7 @@ package frr
 import (
 	"flag"
 	"fmt"
+	"math/rand"
 	"net"
 	"os"
 	"os/exec"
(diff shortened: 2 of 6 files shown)


internal/bgp/frr/frr.go
```diff
@@ -458,3 +459,31 @@ func sortMap[T any](toSort map[string]T) []T {
...
+		for k := range toSort[i].Communities {
+			if toSort[i].Communities[k] != toSort[j].Communities[k] {
+				return toSort[i].Communities[k] < toSort[j].Communities[k]
+			}
+		}
+		for k := range toSort[i].LargeCommunities {
+			if toSort[i].LargeCommunities[k] != toSort[j].LargeCommunities[k] {
+				return toSort[i].LargeCommunities[k] < toSort[j].LargeCommunities[k]
+			}
+		}
```
liornoy, line 486: Should we check at `LargeCommunities` that the length is equal first?

internal/bgp/frr/frr_test.go
```diff
@@ -878,12 +889,19 @@ func TestTwoAdvertisementsTwoSessions(t *testing.T) {
...
+	rand.Shuffle(len(advs), func(i, j int) {
+		advs[i], advs[j] = advs[j], advs[i]
+	})
+	err = session.Set(advs[0], advs[1])
 	if err != nil {
 		t.Fatalf("Could not advertise prefix: %s", err)
 	}
-	err = session1.Set(adv1, adv2)
+
+	rand.Shuffle(len(advs), func(i, j int) {
```
liornoy, line 901: Can we add more attempts to shuffle and check like in `config_conversion_test.go` so we'll be more confident?

internal/k8s/controllers/config_conversion_test.go
```diff
@@ -0,0 +1,235 @@
...
+		},
+	}
+
+	oldConfig, err := toConfig(config.DontValidate,
+		pools,
+		peers,
+		bfdProfiles,
+		l2Advs,
+		bgpAdvs,
+		[]v1beta1.AddressPool{},
```
liornoy, line 186: Why do we have the `AddressPool` empty here?
We do run sortedCopy against it in `toConfig` I think.

internal/k8s/controllers/config_conversion_test.go
```diff
@@ -0,0 +1,235 @@
...
+		communities,
+		map[string]v1.Secret{},
+		nodes,
+		namespaces,
+		v1.ConfigMap{})
+
+	if err != nil {
+		t.Fatalf("conversion failed")
```
liornoy, line 194: err print missing

internal/k8s/controllers/config_conversion_test.go
```diff
@@ -0,0 +1,235 @@
...
+			bgpAdvs,
+			[]v1beta1.AddressPool{},
+			communities,
+			map[string]v1.Secret{},
+			nodes,
+			namespaces,
+			v1.ConfigMap{})
+
+		if err != nil {
+			t.Fatalf("conversion failed")
```
liornoy, line 223: err print missing

internal/k8s/controllers/config_conversion_test.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+
+		if !reflect.DeepEqual(config, oldConfig) {
+			t.Fatalf("conversion is not stable, seed %d, %v\n %v\n", seed, spew.Sdump(config), spew.Sdump(oldConfig))
```
liornoy, line 227: nit: let's add these small details so the output will be clearer
```suggestion
			t.Fatalf("conversion is not stable, seed %d, got: %v\n expected: %v\n", seed, spew.Sdump(config), spew.Sdump(oldConfig))
```

internal/k8s/controllers/config_conversion_test.go
```diff
@@ -0,0 +1,235 @@
...
+		}
+		oldConfig = config
```
liornoy, line 229: I think we can omit this assignment as the value doesn't change, and keep comparing to the same oldConfig.

internal/k8s/controllers/pool_controller.go
```diff
@@ -83,7 +84,19 @@ func (r *PoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.
...
+		[]metallbv1beta2.BGPPeer{},
+		[]metallbv1beta1.BFDProfile{},
+		[]metallbv1beta1.L2Advertisement{},
+		[]metallbv1beta1.BGPAdvertisement{},
+		addressPools.Items,
+		communities.Items,
+		map[string]corev1.Secret{},
+		[]corev1.Node{},
+		namespaces.Items,
+		corev1.ConfigMap{})
```
liornoy, line 98: That's a little hard to read, plus we'll need to update the parameters list when a new CRD is added. why not reuse the `ClusterResources` struct, consume it as the input, and re-assign its fields with the sorted copy?
I would go even a step further and split `toConfig` into two methods, first only the logic of sorting the resources, so we'll have a `sortedResources`, and then call the `config.For` on it.

fedepaol: > That's a little hard to read, plus we'll need to update the parameters list when a new CRD is added. 

What I was totally afraid of was to forget the sorting in case a new CRD is added. I wanted to make it really hard to forget, and that's why I came up with the long list. But I guess I can fix that by doing the sorting in the toConfig as you said. Let me try to get a stab at it

internal/bgp/frr/frr.go
fedepaol: Yep, good catch

internal/bgp/frr/frr_test.go
fedepaol: This will require a bit of reworking, as I don't want to pay the FRR extra effort every iteration. Let me try to fix it.

internal/k8s/controllers/config_conversion_test.go
fedepaol: adding

fedepaol: done

internal/k8s/controllers/config_conversion_test.go
fedepaol: done

internal/k8s/controllers/config_conversion_test.go
fedepaol: changed

PR #1994, opened 2023-07-04
E2E: allow to set the host container as ebgp or ibgp

fedepaol:
Running the e2e tests with host container allows only of the 4 degrees at the time (assuming the host is one hop away from the cluster). Here we allow setting the container as ebgp or ibgp to cover at least two.

### DIFF
--- a/e2etest/bgptests/infra_setup.go
+++ b/e2etest/bgptests/infra_setup.go
@@ -52,6 +52,13 @@ var (
 	}
 )
 
+type HostBGPMode string
+
+const (
+	EBGPMode HostBGPMode = "ebgp"
+	IBGPMode HostBGPMode = "ibgp"
+)
+
 func init() {
 	if ip := os.Getenv("PROVISIONING_HOST_EXTERNAL_IPV4"); len(ip) != 0 {
 		hostIPv4 = ip
@@ -98,8 +105,8 @@ func ExternalContainersSetup(externalContainers string, cs *clientset.Clientset)
 	return res, nil
 }
 
-func HostContainerSetup(image string) ([]*frrcontainer.FRR, error) {
-	config := hostnetContainerConfig(image)
+func HostContainerSetup(image string, bgpMode HostBGPMode) ([]*frrcontainer.FRR, error) {
+	config := hostnetContainerConfig(image, bgpMode)
 	res, err := frrcontainer.Create(config)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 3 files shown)


liornoy:
is the go.work.sum change due to leftovers from previous commits?
because I don't see a reason why it should update from the changes here.

fedepaol:
> is the go.work.sum change due to leftovers from previous commits? because I don't see a reason why it should update from the changes here.

I reverted it and I am not able to trigger the update :shrug:

tasks.py
```diff
@@ -786,9 +786,10 @@ def helmdocs(ctx, env="container"):
     "external_containers": "a comma separated list of external containers names to use for the test. (valid parameters are: ibgp-single-hop / ibgp-multi-hop / ebgp-single-hop / ebgp-multi-hop)",
     "native_bgp": "tells if the given cluster is deployed using native bgp mode ",
     "external_frr_image": "overrides the image used for the external frr containers used in tests",
-    "ginkgo_params": "additional ginkgo params to run the e2e tests with"
+    "ginkgo_params": "additional ginkgo params to run the e2e tests with",
+    "host_bgpmode": "tells wether to run the host container in ebgp or ibgp mode"
```
liornoy, line 790: nit: host_bgp_mode

e2etest/bgptests/infra_setup.go
```diff
@@ -336,26 +343,46 @@ func externalContainersConfigs() map[string]frrcontainer.Config {
...
+				ASN:      metalLBASN,
+				MultiHop: false,
+			},
+			Router: frrconfig.RouterConfig{
+				ASN:     externalASN,
+				BGPPort: 179,
+			},
+			Network:  "host",
+			HostIPv4: hostIPv4,
+			HostIPv6: hostIPv6,
```
liornoy, line 383: maybe to check that the input is valid we can use `switch` here and 
on default return an error?

because now, if a user wants "ibgp" and misspelled as "obgp" he will get "ebgp"

tasks.py
fedepaol: done

e2etest/bgptests/infra_setup.go
fedepaol: done, and added validation on the translation

liornoy (commented):
just 3 small comments

liornoy:
lgtm

PR #1997, opened 2023-07-09
Multiip mode, metallb.universe.tf/enforced-single-node annotation

Protryon:
This PR does two things:
* Adds `metallb.universe.tf/enforced-single-node=yes` annotation for single(ish) node deployments, which allow a user to bypass the restrictions on shared ips with `externalTrafficPolicy: Local`. This of course has a risk of blackholing traffic, which is why it's just an opt-in. It's useful for small clusters with strict node assignment.
* Allows the `metallb.universe.tf/loadBalancerIPs` annotation to specify multiple IPs of the same type and bind to all of them, even in different pools.
  * As a subitem, I ported _some_ of the usage of `net.IP` to `netip.Addr`.

Most of the PR is the second item.
I have not written many additional tests. I'm running with IPv4 only in L2 mode in my two test clusters successfully. I _may_ go ahead and write those tests, but that's contingent on this having any chance of getting merged -- I don't know of any philosophical conflicts with what I'm doing, but that's up to the maintainers here of course.

### DIFF
--- a/controller/service.go
+++ b/controller/service.go
@@ -25,6 +25,7 @@ import (
 	"github.com/go-kit/log/level"
 	v1 "k8s.io/api/core/v1"
 
+	"go.universe.tf/metallb/internal/allocator"
 	"go.universe.tf/metallb/internal/allocator/k8salloc"
 	"go.universe.tf/metallb/internal/ipfamily"
 )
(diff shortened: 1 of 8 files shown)


fedepaol:
For something that requires an API change, I'd prefer discussing it over a design proposal. I am especially interested to hear what is the use case (and if it's worth the extra complexity the implementation brings with it).
You can find examples of the proposals here https://github.com/metallb/metallb/tree/main/design

fedepaol:
For something that requires an API change, I'd prefer discussing it over a design proposal. I am especially interested to hear what is the use case (and if it's worth the extra complexity the implementation brings with it).
You can find examples of the proposals here https://github.com/metallb/metallb/tree/main/design

PR #2001, opened 2023-07-14
E2E: allow to specify v4 / v6 prefixes to inject

fedepaol:
Until now we were testing injection with single stack, so the "toAnnounce" prefixes where added either to the v4 or to the v6 family section. Here we change it to allow the user to specify both ipv4 prefixes and ipv6 prefixes.


oribon:
lgtm

PR #2004, opened 2023-07-17
WIP: Fix controller reboot

liornoy:
This is WIP, validating against CI. will add e2etest after.


fedepaol:
I haven't checked the tests yet but a few comments on the implementation. The implementation works, but it's a bit hacky and could be made more explicit by:

- letting the pool handler understand if it's the first time we receive a configuration (https://github.com/metallb/metallb/blob/1a49c3045237a3934bb4510c0a21961c841400f1/controller/main.go#L115)
- adding a new SyncState value (reprocessAssignedFirst)
- propagating the information (reloadAssignedFirst) through the channel adding a field to the reloadEvent https://github.com/metallb/metallb/blob/5385d5c4aa96da3aeb6a72ef6b0d69e1a47cb38a/internal/k8s/controllers/service_controller_reload.go#L34
- handling the flag accordingly in the reprocessAll function

This has the advantage that the flow is clear(er), and the controller implementation is in charge of the mechanism

internal/k8s/controllers/service_controller_reload.go
```diff
@@ -71,7 +71,11 @@ func (r *ServiceReconciler) reprocessAll(ctx context.Context, req ctrl.Request)
 	for _, service := range services.Items {
 		service := service // so we can use &service
 		if filterByLoadBalancerClass(&service, r.LoadBalancerClass) {
-			level.Debug(r.Logger).Log("controller", "ServiceReconciler", "filtered service", req.NamespacedName)
+			level.Debug(r.Logger).Log("controller", "ServiceReconciler", "filtered service", service.Name, "reason", "not type LoadBalance")
```
fedepaol, line 74: When doing this first pass where we filter the services without ip, we must return errRetry so the reconciliation loop will be triggered again

liornoy:
> I haven't checked the tests yet but a few comments on the implementation. The implementation works, but it's a bit hacky and could be made more explicit by:
> 
> * letting the pool handler understand if it's the first time we receive a configuration (https://github.com/metallb/metallb/blob/1a49c3045237a3934bb4510c0a21961c841400f1/controller/main.go#L115
>   )
> * adding a new SyncState value (reprocessAssignedFirst)
> * propagating the information (reloadAssignedFirst) through the channel adding a field to the reloadEvent https://github.com/metallb/metallb/blob/5385d5c4aa96da3aeb6a72ef6b0d69e1a47cb38a/internal/k8s/controllers/service_controller_reload.go#L34
> * handling the flag accordingly in the reprocessAll function
> 
> This has the advantage that the flow is clear(er), and the controller implementation is in charge of the mechanism

@fedepaol without using some kind of manual boolean "isFirstReload" flag, how would you make the pool handler understand if it's the first time we receive a configuration?

fedepaol:
> > This has the advantage that the flow is clear(er), and the controller implementation is in charge of the mechanism
> 
> @fedepaol without using some kind of manual boolean "isFirstReload" flag, how would you make the pool handler understand if it's the first time we receive a configuration?

Having the flag is legit I guess, something like "firstConfiguration" to be checked and reset together when we get the first non nil set of pools.

liornoy:
> > > This has the advantage that the flow is clear(er), and the controller implementation is in charge of the mechanism
> > 
> > 
> > @fedepaol without using some kind of manual boolean "isFirstReload" flag, how would you make the pool handler understand if it's the first time we receive a configuration?
> 
> Having the flag is legit I guess, something like "firstConfiguration" to be checked and reset together when we get the first non nil set of pools.

Ok, thanks for the reply. I was just making sure that this flag wasn't the hacky element of the implementation.

fedepaol:
> > > > This has the advantage that the flow is clear(er), and the controller implementation is in charge of the mechanism
> > > 
> > > 
> > > @fedepaol without using some kind of manual boolean "isFirstReload" flag, how would you make the pool handler understand if it's the first time we receive a configuration?
> > 
> > 
> > Having the flag is legit I guess, something like "firstConfiguration" to be checked and reset together when we get the first non nil set of pools.
> 
> Ok, thanks for the reply. I was just making sure that this flag wasn't the hacky element of the implementation.

Nope, what I think is hacky is doing a sequence of actions from the wrapping site (the controller) which has no control on how the items are handled

fedepaol:
As discussed offline, another possible solution is to always loop over the services with an IP first, and those without the IP afterwards. This will remove a lot of the complexity of understanding when to do what, but it has the downside of having this ad hoc logic in the wrapping controller instead of inside the business logic.
Thoughts?
Also, @oribon wdyt?

liornoy:
I'll expand on this and say that we drift from the approach of signaling the "firstReload" from within the ReloadEvent struct because it is not normally reachable from the reconcile function and forcing it there will add unnecessary complexity.

internal/k8s/controllers/service_controller.go
```diff
@@ -40,13 +41,14 @@ import (
...
-	LoadBalancerClass string
-	Reload            chan event.GenericEvent
+	Logger             log.Logger
+	Scheme             *runtime.Scheme
+	Namespace          string
+	Handler            func(log.Logger, string, *v1.Service, epslices.EpsOrSlices) SyncState
+	Endpoints          NeedEndPoints
+	LoadBalancerClass  string
+	Reload             chan event.GenericEvent
+	FirstConfiguration atomic.Bool
```
oribon, line 51: nit: I think this is better as a private field (the reason being this would have been hidden if we had a NewController func)

internal/k8s/controllers/service_controller_reload.go
```diff
@@ -67,6 +68,12 @@ func (r *ServiceReconciler) reprocessAll(ctx context.Context, req ctrl.Request)
...
 	}
 
+	// Make it process the services with allocated IP first
+	sortedServices := services.Items
+	sort.Slice(sortedServices, func(i, j int) bool {
+		return len(sortedServices[i].Status.LoadBalancer.Ingress) > len(sortedServices[j].Status.LoadBalancer.Ingress)
+	})
+
 	retry := false
 	for _, service := range services.Items {
```
oribon, line 78: nit: I think it would be clearer to loop over `sortedServices`

internal/k8s/controllers/service_controller_reload.go
```diff
@@ -101,6 +108,9 @@ func (r *ServiceReconciler) reprocessAll(ctx context.Context, req ctrl.Request)
 		level.Info(r.Logger).Log("controller", "ServiceReconciler - reprocessAll", "event", "force service reload")
 		return ctrl.Result{}, errRetry
 	}
+	if r.FirstConfiguration.Load() {
+		r.FirstConfiguration.Store(false)
+	}
```
oribon, line 113: nit: I think this is idempotent -`r.FirstConfiguration.Store(false)` should be enough (no need for the conditional)

internal/k8s/controllers/service_controller_test.go
oribon, line 1: nit/personal preference: I think unit tests should go with the same commit as the implementation, but I guess this is okay

oribon, line 1: is there a place we can add a unit test that covers the scenario (rather than checking the flag flipped)?

oribon (commented):
few nits, will go over the e2es later

internal/k8s/controllers/service_controller_test.go
liornoy: resolved privately.

liornoy:
@oribon the ci failures here are being worked on:
https://github.com/actions/runner-images/issues/8080#issuecomment-1680088869

e2etest/pkg/metallb/metallb.go
```diff
@@ -74,3 +78,24 @@ func SpeakerPodInNode(cs clientset.Interface, node string) (*corev1.Pod, error)
...
+// RestartController restarts metallb's controller pod and waits for it to be running and ready.
+func RestartController(cs clientset.Interface) {
+	controllerPod, err := ControllerPod(cs)
+	framework.ExpectNoError(err)
+
+	err = cs.CoreV1().Pods(controllerPod.Namespace).Delete(context.TODO(), controllerPod.Name, metav1.DeleteOptions{})
+	framework.ExpectNoError(err)
+
+	err = wait.PollImmediate(5*time.Second, 3*time.Minute, func() (bool, error) {
+		pod, err := ControllerPod(cs)
```
fedepaol, line 91: Let's ensure this pod's name is different from the one we just deleted

internal/k8s/nodes/nodes.go
```diff
@@ -6,8 +6,8 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-// ConditionStatus returns the status of the condition for a given node.
-func ConditionStatus(n *corev1.Node, ct corev1.NodeConditionType) corev1.ConditionStatus {
+// conditionStatus returns the status of the condition for a given node.
+func conditionStatus(n *corev1.Node, ct corev1.NodeConditionType) corev1.ConditionStatus {
```
fedepaol, line 10: Can we have this in a separate commit?

e2etest/pkg/service/validate.go
```diff
@@ -25,3 +27,19 @@ func ValidateL2(svc *corev1.Service) error {
...
+
+func ValidateDesiredLB(svc *corev1.Service) {
+	desiredLbIPs := svc.Annotations["metallb.universe.tf/loadBalancerIPs"]
+	if desiredLbIPs == "" {
+		return
+	}
+	framework.ExpectEqual(desiredLbIPs, strings.Join(GetIngressIPs(svc.Status.LoadBalancer.Ingress), ","))
+}
+
+func GetIngressIPs(ingresses []corev1.LoadBalancerIngress) []string {
```
fedepaol, line 39: this can be private I think

e2etest/l2tests/assignment.go
```diff
@@ -515,4 +518,57 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+
+				specificService, err = cs.CoreV1().Services(specificService.Namespace).Get(context.TODO(), specificService.Name, metav1.GetOptions{})
+				framework.ExpectNoError(err)
+
+				if len(specificService.Status.LoadBalancer.Ingress) == 0 {
+					return fmt.Errorf("ip assigned to the LB service %s was removed", specificService.Name)
+				}
+
+				return nil
+			}, 2*time.Minute, 1*time.Second).ShouldNot(gomega.HaveOccurred())
```
fedepaol, line 571: 1 second here means we'll try to restart the controller (and wait 5 secs) every second. I'd relax it to 30 seconds / 1 minute.

e2etest/l2tests/assignment.go
```diff
@@ -515,4 +518,57 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+			specificIP := "192.168.10.100"
+
+			resources := config.Resources{
+				Pools: []metallbv1beta1.IPAddressPool{
+					{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: "assignment-controller-reset-test-pool",
+						},
+						Spec: metallbv1beta1.IPAddressPoolSpec{
+							Addresses: []string{"192.168.10.100/32"},
```
fedepaol, line 533: Can we have two ips and ensure that a service with an assigned ip gets the same ip after the restart (which should be the case, given the new logic)

internal/k8s/controllers/service_controller.go
```diff
@@ -63,6 +65,11 @@ func (r *ServiceReconciler) reconcileService(ctx context.Context, req ctrl.Reque
 
 	var service *v1.Service
 
+	if r.firstConfiguration.Load() {
```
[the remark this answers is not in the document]
liornoy: I tested this implementation without this if statement and it is necessary.
We have a corner case where the configuration is loaded, and a service reconcile occurs not as part of the reprocessAll (i.e. unassigned service is handled first and steals the IP).
What do you mean by expose from the handler implementation?
The handler doesn't know when did the reprocessAll ended, only the service controller knows and that's why we store it in a member of the ServiceReconciler.

internal/k8s/nodes/nodes.go
liornoy: sure, will do.

e2etest/pkg/metallb/metallb.go
liornoy: done

e2etest/l2tests/assignment.go
```diff
@@ -515,4 +518,57 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+
+			svc, err = jig.CreateLoadBalancerServiceWaitForClusterIPOnly(func(svc *v1.Service) { svc.ObjectMeta.Name = "service-c" })
+			framework.ExpectNoError(err)
+			defer service.Delete(cs, svc)
+
+			ginkgo.By("validating metallb keeps the specific ip after controller restart")
+			gomega.Consistently(func() error {
+				metallb.RestartController(cs)
+
+				// sleep test to let the controller time to initialize work
```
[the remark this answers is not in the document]
liornoy: The logic is solid inside restartcontroller.
We put sleep here to let the controller component to do his work allocating IP / reconciling and preventing potential flakes.

e2etest/pkg/service/validate.go
liornoy: right. as it was before. I don't know why I changed it. thanks!

e2etest/l2tests/assignment.go
liornoy: Yes. done.

internal/k8s/controllers/service_controller.go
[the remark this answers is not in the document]
fedepaol: What I mean is, the flag is called firstConfiguration but this code knows nothing about configuration.

Something in the middle might be:
- rename the flag as initialLoadPerformed
- leave a comment that says we care about loading the already assigned services first, the other later when the controller restart, and that the full load is triggered whenever we receive the first configuration

e2etest/l2tests/assignment.go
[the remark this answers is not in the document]
fedepaol: If you want to avoid flakes you can have an eventually first and a consistently later, instead of a sleep that depends on how powerful the runner is. As a rule of thumb, we should avoid sleeps and have a very solid justification for not implementing polling logic against some state, otherwise it will bite us.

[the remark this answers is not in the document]
liornoy: I agree. fixing.

internal/k8s/controllers/service_controller.go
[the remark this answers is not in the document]
liornoy: done.

e2etest/l2tests/assignment.go
```diff
@@ -145,6 +148,60 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+
+			svc, err = jig.CreateLoadBalancerServiceWaitForClusterIPOnly(func(svc *v1.Service) { svc.ObjectMeta.Name = "service-c" })
+			framework.ExpectNoError(err)
+			defer service.Delete(cs, svc)
+
+			ginkgo.By("restarting the controller and validating that the service keeps the same ip")
+			gomega.Consistently(func() error {
+				metallb.RestartController(cs)
+
+				gomega.Eventually(func() error {
```
fedepaol, line 188: My suggestion here was to check the condition with eventually first, and then consistently, so it stabilizes.
This implementation takes for granted the result of the eventually right after the restart, not checking if it keeps being true.
Also, we are not checking the ip for the non pinned service to remain the same.

internal/k8s/controllers/service_controller.go
```diff
@@ -47,6 +47,9 @@ type ServiceReconciler struct {
 	Endpoints         NeedEndPoints
 	LoadBalancerClass string
 	Reload            chan event.GenericEvent
+	// initialLoadPerformed turns true after the first reprocessAll. while false don't reconcile services
```
fedepaol, line 50:
```suggestion
	// initialLoadPerformed is set after the first time we call reprocessAll. This is required because we want the first time we load the services to follow the assigned first, non assigned later order.
	// This allows avoiding to have services with already assigned IP to get their IP stolen by other services.
```

internal/k8s/controllers/service_controller.go
```diff
@@ -63,6 +66,11 @@ func (r *ServiceReconciler) reconcileService(ctx context.Context, req ctrl.Reque
 
 	var service *v1.Service
 
+	if !r.initialLoadPerformed {
+		level.Debug(r.Logger).Log("controller", "ServiceReconciler", "message", "filtered service, still in the first configurations phase")
```
fedepaol, line 70:
```suggestion
		level.Debug(r.Logger).Log("controller", "ServiceReconciler", "message", "filtered service, still in waiting for the initial load to be performed")
```

e2etest/l2tests/assignment.go
liornoy: I had the outer block as `Consistently` because I want the controller restart to be performed multiple times.
Because from my experience reproducing this issue manually, with all the moving parts here,
I had cases where only after the second controller restart the bug appeared.
Does it makes sense? Maybe we should have another inner consistently instead of the eventually?

fedepaol: to validate a consistently after some time, you need to have both against the same condition.
One thing you can do is to:
- have the body in a local function (assigned to a variable)
- do the restart - eventually - consistently inside a controlled loop instead of anotehr eventually, because at the end of the day you want to try it multiple times

liornoy: good idea. I think it might be ok to remove the eventually even. and just have 5-times loop with restart and consistently asserting the services preserve their IPs.

fedepaol: if you do consistently only after the restart, you might hit those "align" issue you put the sleep for though

internal/k8s/controllers/service_controller.go
liornoy: done

internal/k8s/controllers/service_controller.go
liornoy: done

internal/k8s/controllers/service_controller.go
```diff
@@ -47,8 +47,9 @@ type ServiceReconciler struct {
 	Endpoints         NeedEndPoints
 	LoadBalancerClass string
 	Reload            chan event.GenericEvent
-	// initialLoadPerformed turns true after the first reprocessAll. while false don't reconcile services
-	// because we care about loading the already assigned services first.
+	// initialLoadPerformed is set after the first time we call reprocessAll.
```
fedepaol, line 50: this should belong to the previous commit

internal/k8s/controllers/service_controller.go
```diff
@@ -67,7 +68,7 @@ func (r *ServiceReconciler) reconcileService(ctx context.Context, req ctrl.Reque
 	var service *v1.Service
 
 	if !r.initialLoadPerformed {
-		level.Debug(r.Logger).Log("controller", "ServiceReconciler", "message", "filtered service, still in the first configurations phase")
+		level.Debug(r.Logger).Log("controller", "ServiceReconciler", "message", "filtered service, still waiting for the initial load to be performed")
```
fedepaol, line 71: same

e2etest/l2tests/assignment.go
```diff
@@ -145,6 +148,63 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+			defer service.Delete(cs, svc)
+
+			svc, err = jig.CreateLoadBalancerServiceWaitForClusterIPOnly(func(svc *v1.Service) { svc.ObjectMeta.Name = "service-d" })
+			framework.ExpectNoError(err)
+			defer service.Delete(cs, svc)
+
+			restartAndAssert := func() {
+				metallb.RestartController(cs)
+				gomega.Consistently(func() error {
+					specificService, err = cs.CoreV1().Services(specificService.Namespace).Get(context.TODO(), specificService.Name, metav1.GetOptions{})
```
fedepaol, line 191: Sorry I wasn't clear in https://github.com/metallb/metallb/pull/2004#discussion_r1299981050

What I was suggesting there is to ensure that even a service without a specific IP maintains it's IP after the restart. In order to do that you'll have to loosen the cidr (24, for example) and assert that afer the restart the ip remains the same for at least one of the others

e2etest/l2tests/assignment.go
```diff
@@ -145,6 +148,63 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+			specificService, err := jig.CreateLoadBalancerService(30*time.Second, func(svc *v1.Service) {})
+			framework.ExpectNoError(err)
+			defer service.Delete(cs, specificService)
+			service.ValidateDesiredLB(specificService)
+
+			svc, err := jig.CreateLoadBalancerServiceWaitForClusterIPOnly(func(svc *v1.Service) { svc.ObjectMeta.Name = "service-b" })
+			framework.ExpectNoError(err)
+			defer service.Delete(cs, svc)
+
+			svc, err = jig.CreateLoadBalancerServiceWaitForClusterIPOnly(func(svc *v1.Service) { svc.ObjectMeta.Name = "service-c" })
```
fedepaol, line 180: if you override svc, then we'll delete only one service at the end

e2etest/l2tests/assignment.go
liornoy: I don't see how this will work, because when I loosen to /24 all of the services will get allocated and thus it wouldn't create the conditions for the bug to occur from the first place, where we have more LB services than available IPs and the controller assigns an used IP to a different service.

fedepaol: The bug will happen anyway, because the controller will assign the ips in the order the services are processed.
So, before this fix, it was possible to have svcA, svcB and svcC to get ipA, ipB and ipC, restart the contorller and have them scrambled.

fedepaol: Right but we want to ensure that a service that got an IP preserves it regardless it's asking or getting assigned. What we can do then is to have two ips and 4 services, and check that both services that got an IP keep their IPs.

e2etest/l2tests/assignment.go
liornoy: fixed thanks. Notice that I must create new Jig object for each service.

internal/k8s/controllers/service_controller.go
liornoy: fixed

internal/k8s/controllers/service_controller.go
liornoy: fixed

e2etest/l2tests/assignment.go
liornoy: fixed

e2etest/l2tests/assignment.go
liornoy: ok. done.

fedepaol:
lgtm, thanks for fixing this!

PR #2022, opened 2023-07-25
Find default log level from env

### LINKED ISSUE #2007: Allow setting speaker log level from environment variable
The speaker has a default log level of `info` and this is only changeable using additional command line arguments. It would be nice if this was configurable using an environment variable. I'd suggest `METALLB_SPEAKER_LOG_LEVEL`?

I'd prefer not to override all arguments of a container in order to set one additional argument, as that means I'll likely lose out on some updated defaults in the future.

It's about this line:
https://github.com/metallb/metallb/blob/22ffd63a2942a4cd5afd28926fa6a124cecf8ed3/speaker/main.go#L80

bh-tt:
Fixed #2007

### DIFF
--- a/speaker/main.go
+++ b/speaker/main.go
@@ -68,6 +68,10 @@ type service interface {
 func main() {
 	prometheus.MustRegister(announcing)
 
+        logLevelDefault := os.Getenv("METALLB_SPEAKER_LOG_LEVEL")
+        if logLevelDefault == "" {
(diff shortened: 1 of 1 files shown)


bh-tt:
Right, perhaps combine this with removing the cmd flag? It specifies the default value anyway.

bh-tt:
As discussed in #2007 this is not necessary here.

PR #2023, opened 2023-07-25
Fixing typo for Advanced IPAddressPool

JoshuaCooper:
The link on the page was 404'ing, I changed it to the link the side bar has.


fedepaol:
Hey, sorry for the late reply and thanks for the contribution! Would you mind signing the commit?

Thanks again!

PR #2031, opened 2023-08-02
Add sleep to the reloader validation test

liornoy:
Add 5 seconds sleep to give the speaker time to run the reload logic and produce logs, to prevent flakes in the "FRR validate reload feedback" test case.

### DIFF
--- a/e2etest/bgptests/bgp.go
+++ b/e2etest/bgptests/bgp.go
@@ -1235,6 +1235,9 @@ var _ = ginkgo.Describe("BGP", func() {
 			err := ConfigUpdater.Update(resources)
 			framework.ExpectNoError(err)
 
+			// Sleep to let the speaker time to reload the config and produce logs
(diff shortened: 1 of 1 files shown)


e2etest/bgptests/bgp.go
```diff
@@ -1235,6 +1235,9 @@ var _ = ginkgo.Describe("BGP", func() {
 			err := ConfigUpdater.Update(resources)
 			framework.ExpectNoError(err)
 
+			// Sleep to let the speaker time to reload the config and produce logs
+			time.Sleep(5 * time.Second)
```
oribon, line 1239: how does this solve something when the eventually afterwards runs for 2 minutes?

liornoy: I thought that the logs are taken from the pod we fetch outside of the eventually block,
but in second look you're right and it just uses the pod name and fetches the latest logs.
Closing this one for now.

PR #2034, opened 2023-08-04
add more livenessProbe configurability in helm chart for frr container

PeterGrace:
This PR implements the additional fields available to a `livenessProbe` for the `frr` container.  The rationale is that I run k8s nodes on SOQuartz arm64 boards and they sometimes can take >1 second to respond to `livenessProbe` queries.  Because the fields were previously undefined, `timeoutSeconds` was defaulting to `1`, leading to the `frr` container falling into CrashLoopBackOff.

### DIFF
--- a/charts/metallb/templates/speaker.yaml
+++ b/charts/metallb/templates/speaker.yaml
@@ -364,7 +364,10 @@ spec:
           httpGet:
             path: /livez
             port: {{ .Values.speaker.frr.metricsPort }}
+          initialDelaySeconds: {{ .Values.speaker.livenessProbe.initialDelaySeconds }}
           periodSeconds: {{ .Values.speaker.livenessProbe.periodSeconds }}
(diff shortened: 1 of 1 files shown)


fedepaol:
mind squashing the commits?

PeterGrace:
> mind squashing the commits?

Hi @fedepaol , I've squashed the commits which should be reflected in 2a609b6.

I see the CI shows some unit test and static analysis errors, I don't think that this change would impact those but let me know if I need to do some further investigation there.  Thanks!

fedepaol:
> > mind squashing the commits?
> 
> Hi @fedepaol , I've squashed the commits which should be reflected in [2a609b6](https://github.com/metallb/metallb/commit/2a609b66a56abe5f5abee3381b26fb2e5e7b4cc1).
> 
> I see the CI shows some unit test and static analysis errors, I don't think that this change would impact those but let me know if I need to do some further investigation there. Thanks!

yep, the failure is not related to this change (and it was already addressed).
LGTM, thanks!

PR #2046, opened 2023-08-18
Improve logging around config status

### LINKED ISSUE #1568: Improve logging around config status
Opening issue as requested by @fedepaol .

May I suggest as an improvement to the controller to help people that logging be improved around the config.

I was only able to sort out that I was missing a config because of this line after enabling debugging: https://github.com/metallb/metallb/blob/main/controller/main.go#L64

Which pointed me in the right direction, but as far as I can tell that's the only place where any mention of a config is made in the logs (and at a deeper debug level). 

Instead, on startup (and/or configuration/reconfiguration) some logging in the controller about the current state of the config (present, not present, correct, incorrect) be logged to `info` with details dumped to `debug` levels.

_Originally posted by @djryanj in https://github.com/metallb/metallb/issues/1473#issuecomment-1213226989_

kevydotvinu:
- Add logging in the controller about the current state of the config
- Dump the details to debug levels

Fixes #1568

### DIFF
--- a/controller/main.go
+++ b/controller/main.go
@@ -125,14 +125,14 @@ func (c *controller) SetPools(l log.Logger, pools *config.Pools) controllers.Syn
 	level.Debug(l).Log("event", "startUpdate", "msg", "start of config update")
 	defer level.Debug(l).Log("event", "endUpdate", "msg", "end of config update")
 
-	if pools == nil || pools.ByName == nil {
-		level.Error(l).Log("op", "setConfig", "error", "no MetalLB configuration in cluster", "msg", "configuration is missing, MetalLB will not function")
+	if pools == nil || len(pools.ByName) == 0 {
+		level.Error(l).Log("op", "setConfig", "error", fmt.Sprintf("%+v", pools), "msg", "configuration is missing, MetalLB will not function")
 		return controllers.SyncStateErrorNoRetry
 	}
 
 	c.ips.SetPools(pools)
 	c.pools = pools
-
+	level.Info(l).Log("op", "setConfig", "msg", "MetalLB configuration applied")
 	return controllers.SyncStateReprocessAll
 }
 


fedepaol:
@kevydotvinu mind fixing the conflict and signing the commit?

kevydotvinu:
@fedepaol, I have resolved the conflict and signed the commit.

controller/main.go
```diff
@@ -123,14 +123,17 @@ func (c *controller) SetPools(l log.Logger, pools *config.Pools) controllers.Syn
 	level.Debug(l).Log("event", "startUpdate", "msg", "start of config update")
 	defer level.Debug(l).Log("event", "endUpdate", "msg", "end of config update")
 
-	if pools == nil || pools.ByName == nil {
-		level.Error(l).Log("op", "setConfig", "error", "no MetalLB configuration in cluster", "msg", "configuration is missing, MetalLB will not function")
+	if pools == nil || len(pools.ByName) == 0 {
+		level.Error(l).Log("op", "setConfig", "error", fmt.Sprintf("%+v", pools), "msg", "MetalLB configuration is not present")
```
fedepaol, line 127: I'd keep the "MetalLB will not function" message as it's quite clear

controller/main.go
```diff
@@ -123,14 +123,17 @@ func (c *controller) SetPools(l log.Logger, pools *config.Pools) controllers.Syn
 	level.Debug(l).Log("event", "startUpdate", "msg", "start of config update")
 	defer level.Debug(l).Log("event", "endUpdate", "msg", "end of config update")
 
-	if pools == nil || pools.ByName == nil {
-		level.Error(l).Log("op", "setConfig", "error", "no MetalLB configuration in cluster", "msg", "configuration is missing, MetalLB will not function")
+	if pools == nil || len(pools.ByName) == 0 {
```
fedepaol, line 126: this is not related to "changing the logging". If there is a reason for this change, please add it as a separate commit with the reasoning behind the change.

controller/main.go
```diff
@@ -123,14 +123,17 @@ func (c *controller) SetPools(l log.Logger, pools *config.Pools) controllers.Syn
...
+		level.Error(l).Log("op", "setConfig", "error", fmt.Sprintf("%+v", pools), "msg", "MetalLB configuration is not present")
 		return controllers.SyncStateErrorNoRetry
+	} else {
+		level.Info(l).Log("op", "setConfig", "msg", "MetalLB configuration is present")
```
fedepaol, line 130: No need for the else, as we are returning above.
Also, I find this log redundant with the "applied below, let's keep only the "applied" one. I'd also remove the dump of the pools. When dump the whole struct in debug as soon as we receive it, I think it's enough.

kevydotvinu:
@fedepaol, I have added the suggested changes.

PR #2048, opened 2023-08-20
feat(helm): Enables the creation of L2Advertisement at IpAddressPool objects during deployment

c-dst:
add helm template ipaddresspool and l2advertisement
update helm values add ipaddresspool and l2advertisement definition


gclawes:
This needs a `values.schema.json` to accompany the changes to `values.yaml`

fedepaol:
Hey, thanks for the PR!
I'd like to address the request coming from multiple users for creating the resources contextually to helm.

However, this brings a few issues:

- having to maintain a template for ALL the resources (here we have only the L2Advertisement / IPAddressPool, how about the BGP stack)?
- this will likely fail the first time the chart is applied, because the webhook won't be running and thus the application of the CRs will fail. Did you try this?

c-dst:
Hi, 

> Hey, thanks for the PR! I'd like to address the request coming from multiple users for creating the resources contextually to helm.
> 
> However, this brings a few issues:
> 
>     * having to maintain a template for ALL the resources (here we have only the L2Advertisement / IPAddressPool, how about the BGP stack)?

Today I have no way to test a BGP stack, it's why I only added the CRDs L2Advertisement / IPAddressPool. I can update PR with the BGPs stack but it would not be tested.  

> 
>     * this will likely fail the first time the chart is applied, because the webhook won't be running and thus the application of the CRs will fail. Did you try this?

No, I didn't try this.

ginn13:
> * this will likely fail the first time the chart is applied, because the webhook won't be running and thus the application of the CRs will fail. Did you try this?

@fedepaol on first install webhook won't fail because is not installed.
However it would help to deploy MetalLB at once with CRs. It is very useful on day-0 operations and makes single-step installation instead on 2-step, when you have to create СR IPAddressPool and L2Advertisement (or BGP stack) only after MetalLB chart is installed.

PR #2052, opened 2023-08-22
Fix Gosec error in the configmaptocrs

liornoy:
Fix Gosec error inside the configmaptocrs:
G601 (CWE-118): Implicit memory aliasing in for loop. (Confidence: MEDIUM, Severity: MEDIUM)

This is also a bug fix because it didn't create
the correct echoMode and passiveMode values in the CRD. Updated the golden files.


fedepaol:
LGTM. thanks

PR #2054, opened 2023-08-23
Rephrase docs about `IPAddressPool` priority

### LINKED ISSUE #2008: Ambiguity in documentation regarding priority
It says in [the documentation](https://metallb.org/configuration/_advanced_ipaddresspool_configuration/#reduce-scope-of-address-allocation-to-specific-namespace-and-service):

> Given a service, if multiple matching IPAddressPool are available MetalLB will check for the availability of IPs sorting the matching IPAddressPool by priority, starting from the highest to the lowest. A lower number for priority field equals a higher priority. If multiple IPAddressPool have the same priority, the choice will be random. When not specifying a priority / setting priority 0 is considered as lowest priority and will be used for assignment only if the pools with priority can’t be used.

The first sentence sounds ambiguous to me. Starting from the highest to the lowest "what" exactly? Highest number or highest priority? Also "starting from x to y" sounds strange, even if I can understand what it is supposed to mean.

And then it says that "a lower number for priority field equals a higher priority" but then also that "setting priority 0 is considered as lowest priority". What??

This must be rephrased. Here's my suggestion:

> If multiple `IPAddressPool` objects are available to a `Service`, MetalLB will check for the availability of IPs by sorting the matching `IPAddressPool` objects by priority. It will first select the `IPAddressPool` with the highest `priority` number. If multiple `IPAddressPool` objects have the same priority, the choice will be random. If the `priority` field is not set, that will be treated as if it was set to 0, meaning that it will be the last pool to be used.

**NOTE:** Please double check that I wrote the right thing. I might have gotten it the wrong way around since I don't know how to interpret the current documentation. 😄 

Cheers!

lindhe:
Fixes #2008

### DIFF
--- a/website/content/configuration/_advanced_ipaddresspool_configuration.md
+++ b/website/content/configuration/_advanced_ipaddresspool_configuration.md
@@ -90,7 +90,7 @@ objects by priority.
 It will first select the `IPAddressPool` with the lowest `priority` number
 (i.e., `priority=1` is the highest priority).
 If the `priority` field is unset or set to `0`, it will have the lowest
-priority (meaning that it will be the last pool to be used).
+priority (i.e., it will be the last pool to be used).
 If multiple `IPAddressPool` objects have the same priority, the choice will be
 random.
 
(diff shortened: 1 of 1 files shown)


fedepaol:
Thanks!

fedepaol:
lgtm

PR #2066, opened 2023-09-06
Verify syncState in controller tests

mlguerrero12:
Improve unit tests of te controller by explicitly verifying the returned syncState value.

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -209,13 +209,15 @@ func TestControllerMutation(t *testing.T) {
 	// pure function that reliably produces the same end state
 	// regardless of past controller state.
 	tests := []struct {
-		desc    string
-		in      *v1.Service
-		want    *v1.Service
-		wantErr bool
+		desc              string
+		in                *v1.Service
+		prevPool          string
+		want              *v1.Service
+		expectedSyncState controllers.SyncState
 	}{
 		{
-			desc: "deleted balancer",
+			desc:              "deleted balancer",
+			expectedSyncState: controllers.SyncStateSuccess,
 		},
 
 		{
@@ -226,6 +228,7 @@ func TestControllerMutation(t *testing.T) {
 					ClusterIPs: []string{"1.2.3.4"},
 				},
 			},
+			expectedSyncState: controllers.SyncStateSuccess,
 		},
 
 		{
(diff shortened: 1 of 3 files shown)


internal/allocator/allocator.go
```diff
@@ -504,3 +504,7 @@ func (a *Allocator) checkSharing(svc string, ip string, ports []Port, sk *key) e
 	}
 	return nil
 }
+
+func (a *Allocator) AllocatePool(svc, pool string) {
```
fedepaol, line 508: This is not related I think

e2etest/go.work.sum
```diff
@@ -172,6 +170,7 @@ github.com/peterbourgon/diskv v2.0.1+incompatible/go.mod h1:uqqh8zWWbv1HBMNONnaR
 github.com/pkg/diff v0.0.0-20210226163009-20ebb0f2a09e/go.mod h1:pJLUxLENpZxwdsKMEsNbx1VGcRFpLqf3715MtcvvzbA=
 github.com/prometheus/client_model v0.3.0/go.mod h1:LDGWKZIo7rky3hgvBe+caln+Dr3dPggB5dvjtD7w9+w=
 github.com/rogpeppe/go-internal v1.9.0/go.mod h1:WtVeX8xhTBvf0smdhujwtBcq4Qrzq/fJaraNFVN+nFs=
+github.com/rogpeppe/go-internal v1.10.0/go.mod h1:UQnix2H7Ngw/k4C5ijL5+65zddjncjaFoBhdsK/akog=
```
fedepaol, line 173: This is not related

controller/controller_test.go
```diff
@@ -912,20 +940,23 @@ func TestControllerMutation(t *testing.T) {
...
-			wantErr: true,
+			expectedSyncState: controllers.SyncStateSuccess,
 		},
 	}
 
 	for i := 0; i < 100; i++ {
 		for _, test := range tests {
 			t.Run(test.desc, func(t *testing.T) {
 				k.reset()
+				c.ips.Unassign("test")
```
fedepaol, line 951: By unassigning here, we are making the tests stateless. The outer loop is there so the execution of the various SetBalancer is tested with shuffled order. 

Is there a reason for clearing the state and enforcing the previous allocation for some tests?

Issue #2005, opened 2023-07-17
Improve speaker logging when externalTrafficPolicy is Cluster and memberlist is disabled.


vijaySamanuri:
Version of MetalLB: v0.12.1  (valid for even latest metallb version)
protocol: L2
Version of Kubernetes: v1.27

when we set `externalTrafficPolicy` of service to `Cluster` and `memberlist` is disabled.
the respective node speaker is withdrawing service with `notOwner` reason and not responding to ARP requests
```
{"caller":"level.go:63","event":"startUpdate","level":"debug","msg":"start of service update","service":"tca-cp-ns0/istio-ingressgateway","ts":"2023-07-15T11:48:30.102841867Z"}
{"caller":"level.go:63","event":"serviceWithdrawn","ip":null,"ips":["10.X.X.X"],"level":"info","msg":"withdrawing service announcement","pool":"tca-cp-ns0","protocol":"layer2","reason":"notOwner","service":"tca-cp-ns0/istio-ingressgateway","ts":"2023-07-15T11:48:30.102933399Z"}
{"caller":"level.go:63","event":"endUpdate","level":"debug","msg":"end of service update","service":"tca-cp-ns0/istio-ingressgateway","ts":"2023-07-15T11:48:30.10296086Z"}
```

and the external ip set through metallb is no more accessible from outside
```
03:54:56.800506 eth0 B  ARP, Request who-has 10.X.X.X (ff:ff:ff:ff:ff:ff) tell X.X.X.X, length 46
03:54:58.806450 eth0 B  ARP, Request who-has 10.X.X.X (ff:ff:ff:ff:ff:ff) tell X.X.X.X, length 46
```
no ARP replies and we are getting **no route to host** from external clients.
but when we use externalTrafficPolicy `Local` we are able to access from outside.

when you check the [code](https://github.com/metallb/metallb/blob/dbe1a20bb820e2b99337fd658ee40a2bbb53df42/speaker/layer2_controller.go#L103) 
if `externalTrafficPolicy` is `Cluster` it is depending on the memberlist, if disabled it will be emptylist and raises the above withdrawing error.
whereas if it set to `Local` even if the member list is empty it is quering k8s endpoints and getting the node details.

i feel we should improve logs and give some clue in the speaker logs (at least at debug level) mentioning **memberlist is empty or quorum is disabled** so that user would retry fixing memberlist
instead of directly withdrawing service.
reason `notOwner` is not very clear.

PR #2069, opened 2023-09-08
improve speaker logging when no available nodes

cyclinder:
we should skip announcing L2 service and log it when no available nodes

Fixes https://github.com/metallb/metallb/issues/2005


fedepaol:
lgtm, thanks

Issue #2074, opened 2023-09-12
Metallb should use Endpointslices `serving` property, not `ready`


zerkms:
### MetalLB Version

0.13.9

### Deployment method

Manifests

### Main CNI

kube-router

### Kubernetes Version

1.26.4

### Describe the bug

UPD: all this mostly relevant to `ExternalTrafficPolicy: Cluster` services, but in my opinion all `eps.ready` uses should be suppressed.

At the moment throughout the code

https://github.com/metallb/metallb/blob/a5f74ed4c45c3592be171f983053dca79b1f09f9/speaker/layer2_controller.go#L67-L71

the "readiness" of an endpoint slice is determined by the `ready` property.

Now let's have a look at the EPS conditions documentation:

```
FIELDS:
   ready	<boolean>
     ready indicates that this endpoint is prepared to receive traffic,
     according to whatever system is managing the endpoint. A nil value
     indicates an unknown state. In most cases consumers should interpret this
     unknown state as ready. For compatibility reasons, ready should never be
     "true" for terminating endpoints.

   serving	<boolean>
     serving is identical to ready except that it is set regardless of the
     terminating state of endpoints. This condition should be set to true for a
     ready endpoint that is terminating. If nil, consumers should defer to the
     ready condition.

   terminating	<boolean>
     terminating indicates that this endpoint is terminating. A nil value
     indicates an unknown state. Consumers should interpret this unknown state
     to mean that the endpoint is not terminating.

```

So, as you can see the better suited field is in fact `serving`.

Why it's important: with the current implementation it's impossible to implement a service that implement any graceful shutdown strategy whatsoever. As long as the pod becomes in terminating state - its `eps.ready` condition becomes `False`.

If `eps.serving` was used instead - then on pod shutdown it had chance to stay running and complete serving clients gracefully

### To Reproduce

1. Create a pod
2. Expose it via a service (with `externalTrafficPolicy: Local`)
3. Describe the `endpointslice` for the service and see the status of all 3 condition fields.
4. Now delete the pod
5. Quickly describe the `endpointslice` again

### Expected Behavior

Metallb should not immediately stop announcing the pod that is currently being terminated, but instead refer to the `serving` field.

So the flow should look like this:

1. The pod receives delete request. CRI sends signal to the process to shutdown. At this point Pod is still in Ready state
2. Endpointslice.ready=False, Endpointslice.serving=True
3. metallb continues announcing the node ip with the pod
[...]

PR #2088, opened 2023-09-20
Switched from eps.ready to eps.serving

zerkms:
Fixes https://github.com/metallb/metallb/issues/2074

   accepted.
   the PR is merged.

This PR switches from using `eps.ready` to `eps.serving`.

The idea behind it is to give pods time to gracefully shutdown.

### DIFF
--- a/internal/k8s/epslices/endpoint_slices.go
+++ b/internal/k8s/epslices/endpoint_slices.go
@@ -26,13 +26,16 @@ const (
 
 const SlicesServiceIndexName = "ServiceName"
 
-// IsConditionReady tells if the conditions represent a ready state, interpreting
-// nil ready as ready.
-func IsConditionReady(conditions discovery.EndpointConditions) bool {
-	if conditions.Ready == nil {
-		return true
+// IsConditionServing tells if the conditions represent a serving state, deferring
+// to ready state if serving == nil.
+func IsConditionServing(conditions discovery.EndpointConditions) bool {
+	if conditions.Serving == nil {
+		if conditions.Ready == nil {
+			return true
+		}
+		return *conditions.Ready
 	}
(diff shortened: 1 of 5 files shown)


fedepaol:
Mind signing the commit?

zerkms:
@fedepaol committed from the wrong account :-)

internal/k8s/epslices/endpoint_slices.go
```diff
@@ -35,6 +35,15 @@ func IsConditionReady(conditions discovery.EndpointConditions) bool {
 	return *conditions.Ready
 }
 
+// IsConditionServing tells if the conditions represent a serving state, deferring
```
fedepaol, line 38: Do we still need IsConditionReady now?

zerkms: @fedepaol yep, the documentation instructs to defer to the readiness state, if serving is nil, so I followed it literally too

```
   serving	<boolean>
     serving is identical to ready except that it is set regardless of the
     terminating state of endpoints. This condition should be set to true for a
     ready endpoint that is terminating. If nil, consumers should defer to the
     ready condition.
```

fedepaol:
I have another ask: mind expanding the body of the commit with the reason why we are doing this change?

internal/k8s/epslices/endpoint_slices.go
zerkms: Okay, merged the `IsConditionReady` logic into `IsConditionServing`

zerkms:
@fedepaol 

> body of the commit with the reason why we are doing this change?

fair request, I did my best in the last commit :-)

fedepaol:
LGMT and thanks a lot!

Issue #1226, opened 2022-02-16
Custom local AS per peer basis with FRR


fedepaol:
We did not use it, but it's possible to specify a custom as per peer basis using [local-as](https://docs.frrouting.org/en/latest/bgp.html#clicmd-neighbor-PEER-local-as-AS-NUMBER-no-prepend-replace-as)

Currently, metallb rejects custom `localAs`s if they are not all equal. Using that option, we could match the native behaviour.

manikanth04:
@fedepaol I will work on this issue.

/assign

Amila-Rukshan:
@fedepaol, I would like to work on this issue.

I could build the project and can run a dev-env. but when I tried to run e2e tests but got this error. Any clue how to sort out this error? I'm on a mac with amd processor.

Thanks!

```
[BeforeSuite] [FAILED] [4.343 seconds]
[BeforeSuite] 
/Users/amilarukshansenadheera/Desktop/Contributions/metallb/e2etest/e2etest_suite_test.go:118

  Begin Captured GinkgoWriter Output >>
    Sep  2 18:51:27.493: INFO: >>> kubeConfig: /var/folders/q1/r0lbdq0x7cl8t4q1_n0nlp840000gn/T/tmpy0e6dvb2
    Sep  2 18:51:31.836: INFO: Unexpected error: 
        <*errors.withStack | 0xc000352b70>: 
        failed to create vrf red : Error: Unknown device type.
        : exit status 2
        {
            error: <*errors.withMessage | 0xc00052e260>{
                cause: <*exec.ExitError | 0xc00052e240>{
                    ProcessState: {
                        pid: 7306,
                        status: 512,
                        rusage: {
                            Utime: {
                                Sec: 0,
                                Usec: 49217,
                                Pad_cgo_0: [0, 0, 0, 0],
                            },
                            Stime: {
                                Sec: 0,
                                Usec: 24498,
                                Pad_cgo_0: [0, 0, 0, 0],
                            },
                            Maxrss: 23949312,
                            Ixrss: 0,
                            Idrss: 0,
                            Isrss: 0,
                            Minflt: 10927,
                            Majflt: 0,
                            Nswap: 0,
                            Inblock: 0,
                            Oublock: 0,
                            Msgsnd: 6,
                            Msgrcv: 22,
                            Nsignals: 60,
                            Nvcsw: 31,
                            Nivcsw: 876,
                        },
                    },
                    Stderr: nil,
                },
[...]

PR #2090, opened 2023-09-21
Custom local ASN per peer basis with FRR

Amila-Rukshan:
Fixes: https://github.com/metallb/metallb/issues/1226

This allows using different myASNs from metallb nodes when peering with outside routers. 

Example of two BGP peers with different local ASNs:
```
apiVersion: metallb.io/v1beta2
kind: BGPPeer
metadata:
  name: peer1
  namespace: metallb-system
spec:
  myASN: 64514
  peerASN: 64513
  peerAddress: 172.30.0.2
```

```
apiVersion: metallb.io/v1beta2
kind: BGPPeer
metadata:
  name: peer2
  namespace: metallb-system
spec:
  myASN: 64515
  peerASN: 64513
  peerAddress: 172.30.0.3
```

### DIFF
--- a/internal/bgp/frr/frr.go
+++ b/internal/bgp/frr/frr.go
@@ -224,7 +224,7 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
 		var exist bool
 		var rout *router
 
-		routerName := routerName(s.RouterID.String(), s.MyASN, s.VRFName)
+		routerName := routerName(s.RouterID.String(), s.VRFName)
 		if rout, exist = routers[routerName]; !exist {
 			rout = &router{
 				myASN:        s.MyASN,
@@ -256,6 +256,7 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
 			neighbor = &neighborConfig{
 				IPFamily:       family,
 				ASN:            s.PeerASN,
+				LocalASN:       s.MyASN,
 				Addr:           host,
 				Port:           uint16(portUint),
 				HoldTime:       uint64(s.HoldTime / time.Second),
--- a/internal/bgp/frr/frr.go
+++ b/internal/bgp/frr/frr.go
@@ -219,7 +219,16 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
 		config.Loglevel = frrLogLevel
 	}
 
-	for _, s := range sm.sessions {
+	// to get the sessions in some predictable order, we need to sort the keys
+	// this is needed for testing purpose
+	var sessionKeys []string
+	for key := range sm.sessions {
+		sessionKeys = append(sessionKeys, key)
+	}
+	sort.Strings(sessionKeys)
+
+	for _, sKey := range sessionKeys {
+		s := sm.sessions[sKey]
 		var neighbor *neighborConfig
 		var exist bool
 		var rout *router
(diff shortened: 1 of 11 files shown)


Amila-Rukshan:
@fedepaol, I tried to make a change to int tests by changing ASNs in the `vrfContainersConfig`. When running tests, single test ("configure peers one by one and validate FRR paired with nodes") times out from time to time and at a  different place in tests (check [run1 in fork](https://github.com/Amila-Rukshan/metallb/actions/runs/6263621606) and [run2 in PR](https://github.com/metallb/metallb/actions/runs/6259086660) ). I could not see an already added test for native mode. shall I increase the relevant time out and check?

Thanks!

internal/bgp/frr/frr.go
```diff
@@ -219,12 +219,21 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
 		config.Loglevel = frrLogLevel
 	}
 
-	for _, s := range sm.sessions {
+	// to get the sessions in some predictable order, we need to sort the keys
+	// this is needed for testing purpose
+	var sessionKeys []string
```
fedepaol, line 224: you can use sortmap here, or have a variation for pointers
 https://github.com/metallb/metallb/blob/651427b19fcc9771f03bf5bff5eb3305229d0c10/internal/bgp/frr/frr.go#L450

internal/config/validation.go
```diff
@@ -143,14 +143,6 @@ func DiscardNativeOnly(c ClusterResources) error {
 			peerAddr[peerKey] = true
 		}
 	}
-	for _, p := range c.Peers {
-		for _, p1 := range c.Peers[1:] {
-			if p.Spec.MyASN != p1.Spec.MyASN &&
-				p.Spec.VRFName == p1.Spec.VRFName {
-				return fmt.Errorf("peer %s has myAsn different from %s, in FRR mode all myAsn must be equal for the same VRF", p.Spec.Address, p1.Spec.Address)
```
fedepaol, line 150: this applies to ebgp only right?

fedepaol:
Couple of other comments:

- I'd rather go with the extra field. As much as clunky it sounds, it is the most straightforward and obvious approach. We can't really take a random one as the "main" asn, because then metallb will behave differently with those routers that are really meant to be in the same asn (and we might pick a different one). Consider the scenario where we have both ebgp and ibgp peers.

- In any case, we need to have e2e tests to validate that what we are merging works and won't regress in the future. What I suggest is either spinning up a new frr container, or change the configuration on the flight to consume another asn for the nodes (because from its perspective, it will change).

Amila-Rukshan:
> I'd rather go with the extra field

Okay, From where we should read it? Is "reading ain asn from node annotation" the best way to do as suggested [here](https://github.com/metallb/metallb/issues/1226#issuecomment-1713523508)? Is there any better way we can do it?

Amila-Rukshan (commented):
> I'd rather go with the extra field

Okay, From where we should read it? Is "reading ain asn from node annotation" the best way to do as suggested [here](https://github.com/metallb/metallb/issues/1226#issuecomment-1713523508)? Is there any better way we can do it?

fedepaol:
> > I'd rather go with the extra field
> 
> Okay, From where we should read it? Is "reading main asn from node annotation" the best way to do as I suggested [here](https://github.com/metallb/metallb/issues/1226#issuecomment-1713523508)? Is there any better way we can do it?

sorry for the late reply (again). I'd keep the current logic, with the myASN we have now forced to be the same for all the nodes, and a new field "customASN" or "peerASN" that defines the override. It's not elegant, but will be super clear, won't break the api and we can still deprecate the myASN field if we defer the logic to somewhere else.

PR #2099, opened 2023-10-02
Add support for speaker securityContext

mjnagel:
Small addition to support setting the pod level security context on the speaker daemonset.

This is useful in some environments where explicit securityContext must be set for pods, or potentially when using a rebuilt image with a different (user/group could be different).

### DIFF
--- a/charts/metallb/README.md
+++ b/charts/metallb/README.md
@@ -144,6 +144,7 @@ Kubernetes: `>= 1.19.0-0`
 | speaker.reloader.resources | object | `{}` |  |
 | speaker.resources | object | `{}` |  |
 | speaker.runtimeClassName | string | `""` |  |
+| speaker.securityContext | object | `{}` |  |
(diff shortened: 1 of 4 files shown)


fedepaol:
> Not sure why the one e2e test is failing - my changes didn't affect the manifests and shouldn't have even effected the default helm behavior.

probably a flake (we got some of them, although quite rarely).
Would you mind aligning the [value.schema.json](https://github.com/metallb/metallb/blob/main/charts/metallb/values.schema.json) file too?

mjnagel:
@fedepaol done - just left as an object in the schema. Happy to add the properties but it didn't seem like other things were that verbose in the schema. Let me know if you want any adjustments.

fedepaol:
> @fedepaol done - just left as an object in the schema. Happy to add the properties but it didn't seem like other things were that verbose in the schema. Let me know if you want any adjustments.

yep makes sense, thanks!

PR #2117, opened 2023-10-14
feat(docker): add distroless solution

### LINKED ISSUE #2110: Use distroless images for the speaker and the controller

Dupe of https://github.com/metallb/metallb/issues/676 , raising this to make it more evident.
We might want to get the images built from:scratch instead of alpine.

The only pitfall is the speaker container that needs to move binaries around, to fill the frr-metrics or the reloader containers.

Distroless containers for the speaker and the controller.

Also, the troubleshooting doc updated to understand how to ssh into those containers using ephemeral containers.

### Additional context

_No response_

georgettica:

inspired by #677
FIxes #2110

### DIFF
--- a/internal/bgp/frr/docker_test.go
+++ b/internal/bgp/frr/docker_test.go
@@ -54,7 +54,7 @@ func TestMain(m *testing.M) {
 		log.Fatalf("failed to move vtysh.conf to %s - %s - %s", frrDir, err, res)
 	}
 	buf := new(bytes.Buffer)
-	resCode, err := containerHandle.Debug([]string{"cp", "/etc/tempfrr/vtysh.conf", "/etc/frr/vtysh.conf"},
+	resCode, err := containerHandle.Exec([]string{"cp", "/etc/tempfrr/vtysh.conf", "/etc/frr/vtysh.conf"},
 		dockertest.ExecOptions{
 			StdErr: buf,
 		})


oribon:
> why does this need approval again?

because you've made changes since the last approval and this is your first pr :sweat_smile: 

Also, you should use `inv dev-env` (https://github.com/metallb/metallb/blob/main/tasks.py) to check your changes locally if you're not already using it

oribon:
It seems that you got the native part working and the problem lies with the frr mode.
Are you familiar with https://metallb.universe.tf/concepts/bgp/#frr-mode ? 
Anyways it seems that the ci fails on the deployment part which you can replicate on your machine using 
`inv dev-env -b frr`
and you'll see that the deployment fails with something like:
```
        message: 'failed to create containerd task: failed to create shim task: OCI
          runtime create failed: runc create failed: unable to start container process:
          exec: "/bin/sh": stat /bin/sh: no such file or directory: unknown'
```
which makes sense since distroless doesn't have /bin/sh, and it hits the part in the init container where we copy.

you'll probably need to modify `config/frr/speaker-patch.yaml` and the speaker's Dockerfile more.

Thanks a lot for your efforts!

oribon:
The first init container (`cp-frr-files`) is running frr's image while the other two (`cp-reloader`,`cp-metrics`) use the speaker's image, which explains why the first succeeds.
As for the other two, I played with it a bit and the problem is `sh` and `cp` missing like I assumed. something that seems to work is:
(1) adding the `cp` binary to the speaker's image:
```
FROM gcr.io/distroless/static:latest

COPY --from=busybox:1.36-uclibc /bin/cp /cp
COPY --from=builder /build/speaker /speaker
...
```
(2) updating `speaker-patch.yaml` so the 2 init containers use it, e.g for `cp-reloader` it'd be:
```
command: ["/cp","-f","/frr-reloader.sh","/etc/frr_reloader/"]
```
these might not be optimal but I think it is the right direction.

since we're updating the patch file, you'll need to run `inv generatemanifests` before pushing again.
Thanks again :)

oribon:
> PS, busybox is one binary (also cp) that mocks all of the commands, so this binary you copied can give full access to the machine

TIL :) so we should probably use something else for copying :sweat_smile: 

> it's still unclear to me why we need to pass the cp binary to the base image for this to work
> 
> initcontainers should fix this issue.

the way stuff currently works is we have these "frr-tools" as part of the speaker image and copy them to a shared volume that is used later by the `frr-metrics` and `reloader` containers (and they are running frr's image). we do that to not maintain another set of images for reloader/metrics or modify frr's image (and it is convenient).

oribon:
hey @georgettica , are you still interested working on this?

oribon:
sorry I am not sure I understand what you meant, just asking if you need help working on this

oribon:
np, can you explain what you tried and the error you got?
I tried doing what I suggested above and it seemed to get a cluster running (although I did not run the e2es)

oribon:
Doing a short recap:
what we currently do for frr mode is building the "frr-tools" images as part of the speaker image and copying them to shared volumes as part of the initialization process (using init containers). when switching to distroless we can't copy inside the init containers using "cp" (because it is missing). I'm not aware of an alternative to copy files between images as part of initialization.

As a first step, I suggest taking busybox's cp as mentioned earlier so we can get past the phase where we can't even run metallb on a cluster and see if there are more gaps or that's it - that is we'll see how CI behaves and it will unblock you. you can use https://github.com/oribon/metallb/commit/fe0e1731764be50d7caf8324da464b442e126678 which works on my machine (`inv dev-env -b frr` gives me a "healthy" cluster).

When CI is green with busybox's cp we can start optimizing that part  (as you said, using busybox is wrong because it is "all encompassing") - which could be by finding a suitable cp binary or even writing a simple go binary that does it.

oribon:
np (and no rush), also notice that you shouldn't modify the `cp-frr-files` init container because it is based on frr and not the speaker

oribon:
sorry, just noticed you modified the configmaptocrs image instead of the speaker (also there is no need to carry all of these commits, rebasing your commits on top of main should be enough)

oribon:
you're actually progressing :smile: the ones that failed quickly belong to the helm lanes (which makes sense since you haven't changed them yet) and can be left like this for now.
the relevant failed lanes are the native with prometheus enabled, and frr deployed via manifests

oribon:
I'd focus now on making the non-helm lanes pass, which is probably related to changes in `e2etest`

oribon:
right it probably relates to that change, also notice that it happens in `(ipv4, native, manifests, enabled)` but not the one with prometheus disabled, so focusing on that might help you.

oribon:
I'll try giving this a look in the following days, but I see that you fixed some lanes (such as `(ipv4, native, manifests, enabled)`) which is good! thanks again!

oribon:
first of all thanks again for the efforts you put here, it's not trivial at all :) 

the idea of embedding the debug mechanism as part of the Executor is correct imo, but what I think makes more sense is: keeping the interface only with Exec() (as I think it is cleaner),
adding a new struct that implements it specifically for debug containers (e.g podDebugExecutor) and using it specifically where it is needed. That way the only places in the code where we will use debug containers are where they are needed, and what I mean by that is that in some places we really should not use debug containers - e.g when we run `vtysh` we really want to exec against the frr container in the pod (that is, Execing the podExecutor).

This should also solve the timeout that we see in the frr/frr-k8s lanes, as it hangs on:
```sh
~ kubectl --namespace=metallb-system debug speaker-5p95t --image=busybox --container=pod-executor-debugger --stdin --tty -- vtysh -c show bgp vrf all ipv4 json
Warning: container pod-executor-debugger: failed to create containerd task: failed to create shim task: OCI runtime create failed: runc create failed: unable to start container process: exec: "vtysh": executable file not found in $PATH: unknown
```
(also if there's a way to have a timeout on `kubectl debug` it'll be useful here)

wdyt?

oribon:
In general we create an executor whenever it is needed in a test, there isn't a "global" one (except the "Host" one).
Since there have been quite a few changes to the e2e suite since we started looking at this, I suggest restoring the changes you made so far to `e2etest` (but keep a local copy) and rebasing with the current main, then we'll run ci and see what's the current state of which tests are failing and reintroduce the changes incrementally if needed. That will let us see more clearly which tests really need a "Debug" and a better picture of the current state.
wdyt?

oribon:
hey @georgettica , just checking if you are still interested working on this?

oribon:
np :) and like before I'm asking if you need any help working on this - if you're not interested anymore ofc it's fine to unassign you from this (again, thanks a lot for your efforts so far)

PR #2119, opened 2023-10-14
Get rid of circleci refs in docs

gionn:
Not sure is still relevant and/or up to date given the current https://github.com/metallb/metallb/blob/main/.github/workflows/publish.yaml


fedepaol:
thanks!

PR #2121, opened 2023-10-15
Allow custom bind address for memberlist

Rablet:
The bind address for memberlist is currently always set to the pod's IP address:
` - name: METALLB_ML_BIND_ADDR
          valueFrom:
            fieldRef:
              fieldPath: status.podIP`
This PR allows this to be customised. 

In my scenario I wanted it to listen to a different address as the speaker pods were connecting to each other via a VIP which didn't work unless I changed the bind address.

The current behaviour is kept as the default behaviour, but it can be overridden using the new `speaker.memberlist.mlBindAddrOverride` configuration in values.yaml.


fedepaol:
@Rablet need to update the helm docs, please run `inv helm-docs` and repush (and squash the commits, please!)
Thanks!

Rablet:
I believe I have done this now - please let me know if this is not the case!

Issue #284, opened 2018-07-28
Layer 2 mode on RPi


vkorn:
**Is this a bug report or a feature request?**:
bug

**What happened**:
I'm trying to use layer 2 routing with RPi setup. However, I can't reach exposed service from outside of cluster. 

 BGP works as expected but it's a home network and I don't want to overcomplicate things. 

More details:
 
- from pod to pod, using service IP, everything works 
- from node to pod, using service IP, everything works 
- from outside of the cluster I'm unable to access exposed service

arping + tcpdump: 

- from outside -- no luck. Details below*.
- from router -- no luck. Router is definitely not blocking ARP traffic and I'm able to arping physical RPi. 
- from pod to service -- works 
- from node to service -- works only through networks' plugin interface, "weave" in my case

I've checked that IPVS is not enabled, tried flannel and WeaveNet, tried 1.9.3 and 1.11 k8s versions with the same result.

Speakers are correctly responding when requests are actually coming. During the startup, all interfaces are used for ARP responders, so thins seems to be correct.  No errors in logs either. 

iptables are updated 
```
-A KUBE-SERVICES -d 192.168.0.110/32 -p tcp -m comment --comment "kube-system/traefik:http loadbalancer IP" -m tcp --dport 80 -j KUBE-FW-X3WUOHPTYIG7AA3Q
-A KUBE-SERVICES -d 192.168.0.110/32 -p tcp -m comment --comment "kube-system/traefik:https loadbalancer IP" -m tcp --dport 443 -j KUBE-FW-IKNZCF5XJQBTG3KZ
```

*More details about outside arping: 

From speaker logs: 
```
{"caller":"announcer.go:87","event":"createARPResponder","interface":"wlan0","msg":"created ARP responder for interface","ts":"2018-07-28T00:20:15.560229854Z"}
```

From physical RPi: 
```
pi@k8s-agent-1:~ $ sudo tcpdump -i wlan0 arp
tcpdump: verbose output suppressed, use -v or -vv for full protocol decode
listening on wlan0, link-type EN10MB (Ethernet), capture size 262144 bytes
00:53:01.746441 ARP, Request who-has 192.168.0.110 tell vlad-mp, length 44
00:53:01.750945 ARP, Request who-has k8s-agent-1 tell DD-WRT, length 28
00:53:01.750993 ARP, Reply k8s-agent-1 is-at b8:27:eb:01:a6:73 (oui Unknown), length 28
[...]

PR #2124, opened 2023-10-17
Docs: Add promiscuous mode workaround for L2 on WiFi

Rablet:
There is currently an issue with some devices where they do not properly work when using WiFi.

I wan into this on my Raspberry Pi and it took me quite some time troubleshooting until I ran into this issue: https://github.com/metallb/metallb/issues/284

After enabling promiscuous mode it has worked flawlessly.

This PR adds this information to the troubleshooting section of the website to save others the pain of hunting this down via other sources.


fedepaol:
Thanks!

PR #2132, opened 2023-10-19
Disable http2 for the controller webhook

fedepaol:
This is optional, and should remediate the http2 fast reset bug.


oribon (approved):
lgtm

PR #2140, opened 2023-10-25
Use AES-256 for memberlist message encryption

### LINKED ISSUE #1982: Message encryption for Memberlist hardcoded to AES-128
[Hashicorp Memberlist](https://github.com/hashicorp/memberlist) supports AES-128, AES-192 and AES-256 for message encryption. This can be found here - https://github.com/hashicorp/memberlist/blob/v0.5.0/config.go#L187.

A review of codebase suggests, hardcoding this to AES-128 only. This is essentially controlled by the size of the secret being used. For the latest code it can be observed that [the size is 16](https://github.com/metallb/metallb/blob/main/internal/speakerlist/speakerlist.go#L79).
```
		sha := sha256.New()
		mconfig.SecretKey = sha.Sum([]byte(secret))[:16]
```

Is this choice on purpose ? Can this be made configurable so that the choice of symmetric encryption algorithm can be made ? AES-192 & AES-256 might be a better option if given a choice.

AzraelSec:
`SpeakerList` currently configures memberlist to use AES-128 to encrypt the messages involved in the cluster membership management. This PR increases the size of `SecretKey` to make memberlist use AES-256 instead.

Closes #1982.

### DIFF
--- a/internal/speakerlist/speakerlist.go
+++ b/internal/speakerlist/speakerlist.go
@@ -76,7 +76,7 @@ func New(logger log.Logger, nodeName, bindAddr, bindPort, secret, namespace, lab
 		level.Warn(logger).Log("op", "startup", "warning", "no ml-secret-key set, memberlist traffic will not be encrypted")
 	} else {
 		sha := sha256.New()
-		mconfig.SecretKey = sha.Sum([]byte(secret))[:16]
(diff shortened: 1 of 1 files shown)


fedepaol:
Thanks, queued for merging!

AzraelSec:
Happy to contribute! Thanks 😄

PR #2144, opened 2023-10-26
Add option to configure BGP connect time

### LINKED ISSUE #2093: Make BGP connect retry time configurable

With the current configuration options, I can't express that I would like FRR running under control of metallb to (re-)try connecting to my upstream bgp peer more frequently than every 120 seconds.

I would like another field under `spec` for `BGPPeer`: `connectTime` that get's translated into the correct setting in `frr.conf`.

### Additional context

See [an issue relevant to connect time in frr](https://github.com/FRRouting/frr/issues/4745).

AlinaSecret:
This commit introduces an option for users to modify the BGP connect time setting, which controls the time BGP waits between connection attempts to a neighbor.
Fixes #2093


apache_license.go.txt
```diff
@@ -0,0 +1,15 @@
+/*
```
AlinaSecret, line 1: wasn't sure where to put this file

api/v1beta2/bgppeer_types.go
```diff
@@ -17,10 +17,14 @@ limitations under the License.
 package v1beta2
 
 import (
+	"time"
+
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+const DefaultConnectTime = 120 * time.Second
```
AlinaSecret, line 26: this constant isn't used here but is used in other places for example in the conversion function to the v1beta1 bgpeer version, the reason i chose to put it in this file because this value already appears in this file as part of the kubebuilder comments and logically its not an feature in the v1beta1 version so i thought it be better to put in here even if its not used in this file

apache_license.go.txt
fedepaol: where does it come from?

api/v1beta2/bgppeer_types.go
```diff
@@ -55,6 +59,11 @@ type BGPPeerSpec struct {
 	// +optional
 	KeepaliveTime metav1.Duration `json:"keepaliveTime,omitempty"`
 
+	// Requested BGP connect time, controls how long BGP waits between connection attempts to a neighbor.
+	// +kubebuilder:default:="120s"
```
AlinaSecret, line 63: what i don`t like about this file that i didn't find a way to that the default value will appear only once in this file instead of twice.

e2etest/bgptests/bgp.go
```diff
@@ -433,6 +433,7 @@ var _ = ginkgo.Describe("BGP", func() {
 			Peers: metallb.PeersForContainers(FRRContainers, ipFamily, func(p *metallbv1beta2.BGPPeer) {
 				p.Spec.PasswordSecret = corev1.SecretReference{Name: metallb.GetBGPPeerSecretName(p.Spec.ASN, p.Spec.Port, p.Spec.VRFName)}
 				p.Spec.Password = ""
+				p.Spec.ConnectTime = metav1.Duration{Duration: time.Second * 5}
```
AlinaSecret, line 436: this change isn't necessary so the test will pass, but i found that is shortens the test in almost 2 minutes, and i think its nice so thought its worth mentioning it. I didn't give any thought if this can shorten more tests

apache_license.go.txt
AlinaSecret: i use it as a header for the generation of the the deepcopy functions

fedepaol: I'd say the less terrible place is under dev-env with a proper name (such as deepcopy_header.txt or something like that)

AlinaSecret:
general comment:
as i mentioned in the commit message:
"Various unit tests, are updated to include an initial value for the
connectTime field. This is necessary because, in GoLang structs,
numerical and duration fields that are not set contain a value of zero.
The config updater does not distinguish between unset fields and fields
explicitly set to zero. In the case of the connectTime field, this
cannot be zero, so it needs to be explicitly set to include a different
value in all end-to-end test structs."

just wanted to add that i don't think that the best solution for not updating unset values so they are created with there default values is  just setting the values explicitly as i did in the pr. i think maybe need to add a flag or a parameter with which fields not to update so fields with the value of zero wont be updated   should be added to the conifgupdater update function. but i think its out of scope from this pr

AlinaSecret:
another general comment:

while working on this, some of the times i got a race on this test: 

```
--- FAIL: TestNodeReconciler_SetupWithManager (12.76s)
    testing.go:1446: race detected during execution of test
```
i think i didn't add anything that could cause a race but thought it may be worth mentioning.

here is an example:

https://github.com/AlinaSecret/metallb/actions/runs/6616096555/job/17969668010

configsamples/bgppeer_withsecret.yaml
```diff
@@ -6,6 +6,7 @@ metadata:
 spec:
   holdTime: 9m0s
   keepaliveTime: 0s
+  connectTime: 120s
```
fedepaol, line 9: this seems to be a bit hidden, let's add a new file "_with_session_options"

internal/config/config.go
```diff
@@ -418,6 +421,16 @@ func peerFromCR(p metallbv1beta2.BGPPeer, passwordSecrets map[string]corev1.Secr
 		return nil, fmt.Errorf("invalid keepaliveTime %q", p.Spec.KeepaliveTime)
 	}
 
+	// by frr docs connect time should be between 1 - 65535 (max value for unsigned integer)
+	if 1 > p.Spec.ConnectTime.Duration.Seconds() || p.Spec.ConnectTime.Duration.Seconds() > math.MaxUint16 {
```
fedepaol, line 425: This can be added as a validation annotation in the crd. The reason why we have other fields validated here is because originally the configuration was coming from a configmap

api/v1beta2/bgppeer_types.go
```diff
@@ -55,6 +59,11 @@ type BGPPeerSpec struct {
...
+	// +optional
+	ConnectTime metav1.Duration `json:"connectTime,omitempty"`
```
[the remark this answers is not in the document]
AlinaSecret: why does it matter if it will land in all the frr configurations?
i don't really understand why this is a bad thing.

i do think that its a better user experience to have the default values as part of the documentation and well known, instead of making the users search for it by themselves in frr docs.  Also if we already support a feature i think its better to have full control over the value, and not be deepened by frr, i think if frr do change the value one day it can change the value for the users without notice. i think its better to  have full control for the value.

how ever i do think using a pointer can give  a neat implementation to check easily if the value is set for native mode.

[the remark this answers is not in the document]
fedepaol: I understand the general sentiment, and I wrote that too (when I mentioned that it's on the defensive side), and that's why I asked for @oribon 's view. We have good and comprehensive CI that should guarantee things will work.

> why does it matter if it will land in all the frr configurations? i don't really understand why this is a bad thing.
> 

For example, we hit rendering problems with default values: 
https://github.com/FRRouting/frr/issues/14242


> i do think that its a better user experience to have the default values as part of the documentation and well known, instead of making the users search for it by themselves in frr docs. Also if we already support a feature i think its better to have full control over the value, and not be deepened by frr, i think if frr do change the value one day it can change the value for the users without notice. i think its better to have full control for the value.
> 

I totally agree with this part, I am just afraid of side effects.

> how ever i do think using a pointer can give a neat implementation to check easily if the value is set for native mode.

Let's wait a bit before moving on, I am open to both alternatives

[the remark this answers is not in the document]
oribon: talked with @fedepaol , I think going forward any field that is FRR specific should be pointerized (and omitted from frr.conf) like we do with bfdprofiles, the main reason being them understanding these concepts better than us and at this point we're frr consumers, so it feels natural to defer it to whatever they do and generate a simpler configuration (although I agree we do lose some "control/visibility"). 
about the documentation part I think writing what frr assigns as default is enough as I believe a user who wouldn't set this field doesn't really care whether we set the default or not.

internal/config/config.go
AlinaSecret: Sure! did it

configsamples/bgppeer_withsecret.yaml
AlinaSecret: done :)

api/v1beta2/bgppeer_types.go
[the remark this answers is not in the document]
AlinaSecret: Tanks @oribon and @fedepaol for the great explanations! its always nice to understand why this is the chosen approach :)   changed connect time to be pointerized

tasks.py
```diff
@@ -898,10 +911,20 @@ def generatemanifests(ctx):
...
+    generate_deepcopy()
+
     _align_helm_crds(
         source='config/manifests/metallb-frr.yaml',
         output='charts/metallb/charts/crds/templates/crds.yaml')
 
+
+def generate_deepcopy():
+    install_controller_gen()
+    res = run(f'{build_path}/bin/controller-gen object:headerFile="{LICENSE_PATH}" paths="./api/..."')
```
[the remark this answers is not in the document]
AlinaSecret: i think the comment is not relevant for this line(the version is used only in installation of the command not in usage), but anyway i changed the usage so its similar to kubectl

tasks.py
```diff
@@ -292,21 +296,30 @@ def validate_kind_version():
...
         if with_prometheus:
             layer = "prometheus-" + layer
         res = run("{} kustomize config/{} > {}".format(kubectl_path, layer, output))
         if not res.ok:
             raise Exit(message="Failed to kustomize manifests")
 
+
+@cache
+def install_controller_gen():
+    run("GOPATH={} go install sigs.k8s.io/controller-tools/cmd/controller-gen@{}".format(build_path,
```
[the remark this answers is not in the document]
AlinaSecret: made fetch_kubectl more generic and did the same for controller gen

api/v1beta2/bgppeer_types.go
AlinaSecret: Not relevant anymore due to making connect time pointer field

internal/bgp/bgp.go
fedepaol, line 1: nit on the commit message: there's no configmap. Also, this commit does the conversion from the API to the internal configuration. I'd set this as the commit subject. The validation is an additional thing here.

internal/bgp/frr/frr_test.go
```diff
@@ -125,6 +126,7 @@ func TestSingleEBGPSessionMultiHop(t *testing.T) {
 			PeerASN:       200,
 			HoldTime:      time.Second,
 			KeepAliveTime: time.Second,
+			ConnectTime:   ptr.To(time.Second),
```
fedepaol, line 129: Instead of adding it to all the tests, just add it to a couple so we know the default case still works.

internal/bgp/frr/config.go
```diff
@@ -71,6 +71,7 @@ type neighborConfig struct {
 	Port                uint16
 	HoldTime            uint64
 	KeepaliveTime       uint64
+	ConnectTime         *uint64
```
fedepaol, line 74: no need for it to be a pointer, I'd defer the ne 0 logic to the template. It's kind of equivalent but probably more straightforward

AlinaSecret: i understand, changed it.

internal/bgp/frr/frr_test.go
AlinaSecret: done :)

internal/bgp/bgp.go
AlinaSecret: Totally, sounds a lot better!

fedepaol:
@AlinaSecret mind fixing the conflicts here?

fedepaol:
LGTM, sent to merge queue

fedepaol:
@AlinaSecret mind checking the CI failures?

PR #2148, opened 2023-11-04
Publish metallb-crds chart

rmvangun:
This PR publishes the CRDs subchart located at `charts/metallb/charts/crds`. The reason for this is to support separating the installation of CRDs from the installation of other metallb resources.

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -141,15 +141,22 @@ jobs:
 
           exists=$(curl -s -H "Accept: application/vnd.github.v3+json" https://github.com/$GITHUB_REPOSITORY/releases/tag/$repo-chart-$tag -w %{http_code} -o /dev/null)
           if [[ $exists != "200" ]]; then
-            echo "Creating release..."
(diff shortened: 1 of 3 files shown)


rmvangun:
I haven't had an opportunity to test this without a few credentials, but I'll get to it soon. Let me know if there's some reason the CRD chart hasn't already been published! It would be convenient for my current work but will happily change course if this isn't the way to go.

fedepaol:
Thanks for the PR!

I am not super expert on how helm chart releaser work, it'd be great if you could test this on a separate repo and show the results. I normally tweak a couple of names / secrets to make this work.
In particular, I have a couple of points of concern:
- the co-existence of the currently released helm chart, which embeds the crds (and it does, otherwise metallb wouldn't be able to deploy from helm). Will we require also a chart for standalone metallb - without crds? If so, seems too much of a complication to me
- how to publish both charts to github pages, where we currently serve them

Also also, what is the reason why you need to deploy the crds in a different phase from metallb?

Issue #2150, opened 2023-11-08
add ipv4/ipv6 PoolActive and PoolCapacity in metrics


woodgear:
### Is your feature request related to a problem?

in dualstack k8s and service with RequireDualStack, if a pool could not assign both ipv4/ipv6 address, it could not assign ip to this service at all.
we are try to add a prometheus alert rule to monitor  this. but the current exported metrics does not distinguish with ipv4/ipv6.
so. if metallb export ipv4PoolCapacity,ipv6PoolCapacity,ipv4PoolActive,ipv6PoolActive metrics.that could be better.

### Describe the solution you'd like

export ipv4PoolCapacity,ipv6PoolCapacity,ipv4PoolActive,ipv6PoolActive metrics

### I've read and agree with the following

- [x] I've checked all open and closed issues and my request is not there.
- [x] I've checked all open and closed pull requests and my request is not there.

woodgear:
here is a demo impl https://github.com/metallb/metallb/pull/2151. without compile or test, just to show what im try to impl

PR #2151, opened 2023-11-08
Metrics: add ipv4/6 addresses_in_use_total and addresses_total

woodgear:
issue: https://github.com/metallb/metallb/issues/2150

### DIFF
--- a/e2etest/l2tests/l2.go
+++ b/e2etest/l2tests/l2.go
@@ -502,7 +502,7 @@ var _ = ginkgo.Describe("L2", func() {
 				L2Advs: []metallbv1beta1.L2Advertisement{emptyL2Advertisement},
 			}
 
-			poolCount, err := config.PoolCount(resources.Pools[0])
+			total, ipv4, ipv6, err := config.PoolCount(resources.Pools[0])
 			framework.ExpectNoError(err)
 
 			err = ConfigUpdater.Update(resources)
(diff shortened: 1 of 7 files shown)


fedepaol:
Sorry for the delay. I think this makes sense, and would even align with https://github.com/metallb/metallb/blob/main/design/crd-status.md

WDYT @oribon ?

oribon:
sounds good

woodgear:
cool。 i’will start working on this。cr status and metrics are different。 like prometheus  could not monitor via crd status，so i thought we should add it both。in this pr，i will still focus on the metrics part。

fedepaol:
> cool。 i’will start working on this。cr status and metrics are different。 like prometheus could not monitor via crd status，so i thought we should add it both。in this pr，i will still focus on the metrics part。

Yep it's totally fine to work only on the metrics part.

fedepaol:
Hey @woodgear , just pinging to check if you are still interested to work on this

woodgear:
yes. im try to fighting the ci..

woodgear:
@fedepaol

internal/allocator/allocator.go
```diff
@@ -391,14 +430,17 @@ func sharingOK(existing, new *key) error {
 }
 
 // poolCount returns the number of addresses in the pool.
-func poolCount(p *config.Pool) int64 {
+func poolCount(p *config.Pool) (int64, int64, int64) {
```
fedepaol, line 433: Can you adjust the godoc?

internal/k8s/controllers/reconciliation_test.go
```diff
@@ -179,7 +179,7 @@ var _ = AfterSuite(func() {
 	cancel()
 	err := testEnv.Stop()
 	Expect(err).NotTo(HaveOccurred())
-	Eventually(mgrDone.Load(), 5*time.Second, 200*time.Millisecond).Should(BeTrue())
+	Eventually(mgrDone.Load, 5*time.Second, 200*time.Millisecond).Should(BeTrue())
```
fedepaol, line 182: Why this change?

e2etest/l2tests/l2.go
```diff
@@ -502,7 +502,7 @@ var _ = ginkgo.Describe("L2", func() {
 				L2Advs: []metallbv1beta1.L2Advertisement{emptyL2Advertisement},
 			}
 
-			poolCount, err := config.PoolCount(resources.Pools[0])
+			total, _, _, err := config.PoolCount(resources.Pools[0])
```
fedepaol, line 505: Can we check v4 and v6 numbers too depending on ipFamily?

fedepaol:
Thanks, a couple of nits but looks really neat!

internal/k8s/controllers/reconciliation_test.go
woodgear: golangci-lint complains about
`
internal/k8s/controllers/reconciliation_test.go:182:2: ginkgo-linter: use a function call in Eventually. This actually checks nothing, because Eventually receives the function returned value, instead of function itself, and this value is never changed; consider using `Eventually(mgrDone.Load, 5*time.Second, 200*time.Millisecond).Should(BeTrue())` instead (ginkgolinter)
`

e2etest/l2tests/l2.go
woodgear: ok

internal/allocator/allocator.go
woodgear: ok

e2etest/l2tests/l2.go
```diff
@@ -514,21 +514,39 @@ var _ = ginkgo.Describe("L2", func() {
...
+					expect int64
 				}
-				err = metrics.ValidateGaugeValue(int(poolCount), "metallb_allocator_addresses_total", map[string]string{"pool": poolName}, controllerMetrics)
-				if err != nil {
-					return err
-				}
-				err = metrics.ValidateOnPrometheus(promPod, fmt.Sprintf(`metallb_allocator_addresses_in_use_total{pool="%s"} == 0`, poolName), metrics.There)
-				if err != nil {
-					return err
+				cases := Cases{
```
fedepaol, line 521: nit: given it's local, you can use an anonymous struct here (and no need to have it exported anyway).

Something like cases := []struct{name,
                                                          expect}{
                                            }

internal/k8s/controllers/reconciliation_test.go
fedepaol: Not sure you saw my previous comment, but can you move this to a separate commit (still in this pr)?

e2etest/l2tests/l2.go
woodgear: nice trick.

internal/k8s/controllers/reconciliation_test.go
woodgear: ok

woodgear:
@fedepaol what should i do next to make it merge?

fedepaol:
LGTM, one final ask: https://github.com/metallb/metallb/pull/2151#discussion_r1450404190
Can you move that change in a separate commit (still in this PR)? Thanks a lot!

woodgear:
@fedepaol updated.

fedepaol:
Awesome, sending to the merge queue!

PR #2152, opened 2023-11-08
feat(helmchart): add extraContainers to speaker

DrBu7cher:
In the environment we are using we would like to deploy a container right next to every MetalLB Speaker. The easiest and most reliable way for us would be to use the functionality this PR would add.

If you want me to add this to the `controller.yaml` too, feel free to tell me.

### DIFF
--- a/charts/metallb/values.yaml
+++ b/charts/metallb/values.yaml
@@ -249,6 +249,8 @@ controller:
     successThreshold: 1
     timeoutSeconds: 1
 
+  extraContainers: []
+
 # speaker contains configuration specific to the MetalLB speaker
 # daemonset.
 speaker:
@@ -338,6 +340,8 @@ speaker:
 
   frrMetrics:
     resources: {}
+
+  extraContainers: []
 
 crds:
   enabled: true
--- a/charts/metallb/README.md
+++ b/charts/metallb/README.md
@@ -24,6 +24,7 @@ Kubernetes: `>= 1.19.0-0`
 |-----|------|---------|-------------|
 | controller.affinity | object | `{}` |  |
 | controller.enabled | bool | `true` |  |
+| controller.extraContainers | list | `[]` |  |
 | controller.image.pullPolicy | string | `nil` |  |
 | controller.image.repository | string | `"quay.io/metallb/controller"` |  |
 | controller.image.tag | string | `nil` |  |
@@ -111,6 +112,7 @@ Kubernetes: `>= 1.19.0-0`
 | speaker.affinity | object | `{}` |  |
 | speaker.enabled | bool | `true` |  |
 | speaker.excludeInterfaces.enabled | bool | `true` |  |
+| speaker.extraContainers | list | `[]` |  |
 | speaker.frr.enabled | bool | `true` |  |
 | speaker.frr.image.pullPolicy | string | `nil` |  |
 | speaker.frr.image.repository | string | `"quay.io/frrouting/frr"` |  |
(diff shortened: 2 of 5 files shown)


fedepaol:
> In the environment we are using we would like to deploy a container right next to every MetalLB Speaker. The easiest and most reliable way for us would be to use the functionality this PR would add.
> 
> If you want me to add this to the `controller.yaml` too, feel free to tell me.

Can you expand a bit the use case, to understand why and how it could help other metallb users.

DrBu7cher:
Of course.
We have a container that is going to configure the announced IPs on the interface of the nodes that metallb is announcing from. We could create a seperate DaemonSet with the same configuration parallel to metallb, but it would be easier and more reliable to have the container be part of the speaker-pod(s).

fedepaol:
> Of course. We have a container that is going to configure the announced IPs on the interface of the nodes that metallb is announcing from. We could create a seperate DaemonSet with the same configuration parallel to metallb, but it would be easier and more reliable to have the container be part of the speaker-pod(s).

And what's the reason again (sorry if I insist, trying to understand if it makes sense as a metallb feature)? 
This is for l2 I guess, as you can't have multiple nodes with the same ip. Also, this goes under the assumption that each node has only one interface?
And also also, what is the relationship between the extra container and the speaker? It is just for deployment commodity I think, as it won't be able to inspect what ips are being announced from a given node.

DrBu7cher:
If you ping the BGP IP, the ICMP packets go into a loop between node and router and you never get a response.
This is working as expected from your side, as far as I know,
> Also, please be aware that pinging the service IP won’t work.
[source](https://metallb.universe.tf/troubleshooting/#general-concepts)
but we want our IPs to be pingable.

We configured/added the IPs to the node interfaces on the host that metallb is running on and ping works now. I built a container/controller that configures the IP on the host, so we can deploy and manage it through Kubernetes.

It can handle multiple interfaces, but since I only care about ping, it would not be a problem if I would configure the IP on the loopback interface always.

fedepaol:
Please let's add this to the controller too, for symmetry. I am not sure I like it but it is nearly to not invasive.
Also, CI will complain about the helm docs I think

cc @oribon

charts/metallb/values.yaml
```diff
@@ -249,6 +249,8 @@ controller:
     successThreshold: 1
     timeoutSeconds: 1
 
+  extraContainers: {}
```
oribon, line 252: shouldn't this be an array?

oribon (commented):
specific use-case aside, I think this parameter is common enough for us to include as well (and shouldn't cause any harm I guess) and +1 for having it in the controller also.

charts/metallb/values.yaml
DrBu7cher: You are correct.

DrBu7cher: Fixed

DrBu7cher:
Please approve it again :rocket:

oribon:
sorry I missed it, mind adding an array validation in `values.schema.json` before?

DrBu7cher:
Can you check again? Is this what you meant?
Is something else missing?

oribon:
> How can I test if it is correct?

by changing the value to something that is not an array:
`extraContainers: "abc"`
you get:
```
$ helm template .
metallb:
- speaker.extraContainers: Invalid type. Expected: array, given: string
- speaker: Must validate all the schemas (allOf
```

DrBu7cher:
The `values.scheme.json` was not correct.
Please approve again @oribon

DrBu7cher:
Like so?

oribon:
close :) put the `controller placement` commit first and all the other changes in another commit (so there are 2 commits in total)

DrBu7cher:
Please approve @oribon

PR #2162, opened 2023-11-13
Enable a frr-k8s based bgp backend

fedepaol:
Following the [frr splitting](https://github.com/metallb/metallb/blob/main/design/splitfrr-proposal.md) design proposal, here we add a new BGP implementation, whcih is based on [FRR-K8s](https://github.com/metallb/frr-k8s). What this implemetation does is to generate the corresponding frr-k8s configuration so that the FRR-k8s daemon can configure FRR.

This will allow external actors to add additional FRR knobs that are not related to MetalLB.


speaker/main.go
```diff
@@ -187,6 +189,7 @@ func main() {
 		},
 		ValidateConfig:    validateConfig,
 		LoadBalancerClass: *loadBalancerClass,
+		WithFRRK8s:        listenFRRK8s,
```
oribon, line 196: nit: unrelated to this commit

internal/k8s/controllers/frrk8s_config_controller.go
```diff
@@ -0,0 +1,139 @@
...
+	r.Lock()
+	desired := r.desiredConfiguration.DeepCopy()
+	r.Unlock()
+
+	if reflect.DeepEqual(current.Spec, desired.Spec) {
+		return ctrl.Result{}, nil
+	}
+	desired.Spec.DeepCopyInto(&current.Spec)
+	current.Name = desired.Name
+	current.Namespace = desired.Namespace
```
oribon, line 93: nit: this is a bit confusing/unnecessary , also I think it might make sense to let the session manager send a spec instead of the whole resource, and the ObjectMeta will be handled here. wdyt?

internal/k8s/controllers/frrk8s_config_controller_test.go
```diff
@@ -0,0 +1,169 @@
...
+				Routers: []frrv1beta1.Router{
+					{
+						ASN: 25,
+					},
+				},
+			},
+		},
+	}
+
+	err = m.GetClient().Create(ctx, &frrConfig)
```
oribon, line 103: can you reload and eventually instead to see it creates properly on first time? also currently this might break the controller a bit as the desiredConfig is empty and it would attempt to reset everything (including the name/ns)?

internal/k8s/k8s.go
```diff
@@ -218,6 +227,24 @@ func New(cfg *Config) (*Client, error) {
 		}
 	}
 
+	if cfg.WithFRRK8s {
+		fmt.Println("FEDE setup")
```
oribon, line 231: :smile:

internal/k8s/k8s.go
```diff
@@ -218,6 +227,24 @@ func New(cfg *Config) (*Client, error) {
...
+			Client:    mgr.GetClient(),
+			Logger:    cfg.Logger,
+			Scheme:    mgr.GetScheme(),
+			Namespace: cfg.Namespace,
+			NodeName:  cfg.NodeName,
+			Reload:    make(chan event.GenericEvent),
+		}
+		if err := frrk8sController.SetupWithManager(mgr); err != nil {
+			level.Error(c.logger).Log("error", err, "unable to create controller", "node")
+			fmt.Println("FEDE error")
```
oribon, line 242: :smile:

internal/k8s/k8s.go
```diff
@@ -218,6 +227,24 @@ func New(cfg *Config) (*Client, error) {
...
+		frrk8sController := controllers.FRRK8sReconciler{
+			Client:    mgr.GetClient(),
+			Logger:    cfg.Logger,
+			Scheme:    mgr.GetScheme(),
+			Namespace: cfg.Namespace,
+			NodeName:  cfg.NodeName,
+			Reload:    make(chan event.GenericEvent),
+		}
+		if err := frrk8sController.SetupWithManager(mgr); err != nil {
+			level.Error(c.logger).Log("error", err, "unable to create controller", "node")
```
oribon, line 241: nit: "node" -> "frr-k8s"

config/frr-k8s/kustomization.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: kustomize.config.k8s.io/v1beta1
+kind: Kustomization
+resources:
+- https://github.com/metallb/frr-k8s/config/default/?timeout=120&ref=main
```
oribon, line 4: we should remember setting this to != main before merging this

config/frr-k8s/controller-patch.yaml
oribon, line 1: is this necessary?

config/frr-k8s/frr-cm.yaml
oribon, line 1: same?

internal/bgp/frrk8s/frrk8s.go
```diff
@@ -0,0 +1,380 @@
...
+	"go.universe.tf/metallb/internal/bgp/frr"
+	metallbconfig "go.universe.tf/metallb/internal/config"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type sessionManager struct {
+	sessions         map[string]*session
+	bfdProfiles      []frr.BFDProfile
+	metallbNamespace string
+	currentNode      string
```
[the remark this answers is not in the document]
fedepaol: I'll change them to what they mean: targetNamespace and nodeToConfigure or something along that line
wdyt?

internal/bgp/frrk8s/frrk8s.go
```diff
@@ -0,0 +1,380 @@
...
+		}
+		newConfig.Spec.BGP.Routers = append(newConfig.Spec.BGP.Routers, toAdd)
+	}
+
+	for _, bfd := range sm.bfdProfiles {
+		bfd := bfd
+		toAdd := frrv1beta1.BFDProfile{
+			Name: bfd.Name,
+		}
+		if bfd.ReceiveInterval != nil {
```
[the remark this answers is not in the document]
fedepaol: they prevent the speaker from panicking if the value is nil

internal/bgp/frrk8s/frrk8s.go
```diff
@@ -0,0 +1,380 @@
...
+			ASN:       r.myASN,
+			ID:        r.routerID,
+			VRF:       r.vrf,
+			Neighbors: sortMap(r.neighbors),
+			Prefixes:  sortMap(r.prefixes),
+		}
+		newConfig.Spec.BGP.Routers = append(newConfig.Spec.BGP.Routers, toAdd)
+	}
+
+	for _, bfd := range sm.bfdProfiles {
```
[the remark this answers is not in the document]
fedepaol: I'll move the sorting here, I think it makes more sense

internal/bgp/frrk8s/frrk8s.go
```diff
@@ -0,0 +1,380 @@
...
+	sessions         map[string]*session
+	bfdProfiles      []frr.BFDProfile
+	metallbNamespace string
+	currentNode      string
+	sync.Mutex
+	desiredConfig         frrv1beta1.FRRConfiguration
+	configChangedCallback func(interface{})
+}
+
+func (sm *sessionManager) SetEventListener(callback func(interface{})) {
```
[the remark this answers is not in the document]
fedepaol: I think you saw in the next commit how the session manager is buried into the speaker abstraction and it's not easy to extract (unless we want to have a singleton variable as part of the module, which I am not sure I like).

Here we set a listener that listens to event. The name does not seem terrible to me, but I can switch to callback if you want.

internal/bgp/frrk8s/frrk8s.go
```diff
@@ -0,0 +1,380 @@
...
+			toAdd.MinimumTTL = bfd.MinimumTTL
+		}
+		toAdd.EchoMode = &bfd.EchoMode
+		toAdd.PassiveMode = &bfd.PassiveMode
+
+		newConfig.Spec.BGP.BFDProfiles = append(newConfig.Spec.BGP.BFDProfiles, toAdd)
+	}
+	sm.desiredConfig = newConfig
+	toNotify := sm.desiredConfig.DeepCopy()
+	sm.configChangedCallback(*toNotify)
```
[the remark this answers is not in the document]
fedepaol: I thought about that but I also thought that we have the debouncer in frrk8s, plus some squashing will be introduced because of the api delays. I'd say let's try without and see if that fits the bill.

internal/k8s/controllers/frrk8s_config_controller.go
fedepaol: it's not unnecessary, but I can provide a better way to make it not confusing.
It is required for the scenario where the object was not created yet, and so at this point current is an empty object (because the key was not set).
I don't like having the object half filled on one side and the other half (despite only name + namespace) on the other, I think it makes sense for the underlying implementation to generate the object as as whole and to decide for its name / namespace

internal/k8s/k8s.go
fedepaol: ops

config/frr-k8s/kustomization.yaml
fedepaol: we'll have to do a tag first

config/frr-k8s/controller-patch.yaml
fedepaol: yes, but adapted the type. We need it because the controller serves as webhook endpoint and it must know what needs to validate.

config/frr-k8s/frr-cm.yaml
fedepaol: removed

config/frr-k8s/frr-k8s-image-patch.yaml
oribon, line 1: this?

internal/k8s/controllers/frrk8s_config_controller.go
oribon: now it's better, what I didn't like was that both tried "taking control" over the metadata

internal/bgp/frrk8s/frrk8s.go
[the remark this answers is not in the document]
fedepaol: As you prefer, I see listener and callback as synonims, but I won't mind changing it

internal/bgp/frrk8s/frrk8s.go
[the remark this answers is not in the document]
fedepaol: ah dang you are right!

internal/bgp/frrk8s/frrk8s.go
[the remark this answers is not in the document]
fedepaol: fair enough, I'll add a debouncer to the controller

internal/k8s/controllers/frrk8s_config_controller.go
```diff
@@ -0,0 +1,142 @@
...
+	err := r.Get(ctx, client.ObjectKey{Name: r.desiredConfiguration.Name, Namespace: r.desiredConfiguration.Namespace}, &current)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return ctrl.Result{}, err
+	}
+
+	if reflect.DeepEqual(current.Spec, r.desiredConfiguration.Spec) {
+		return ctrl.Result{}, nil
+	}
+
+	toApply := &frrv1beta1.FRRConfiguration{ObjectMeta: metav1.ObjectMeta{Name: r.desiredConfiguration.Name, Namespace: r.desiredConfiguration.Namespace}}
```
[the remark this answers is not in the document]
fedepaol: done

internal/k8s/controllers/frrk8s_config_controller_test.go
```diff
@@ -0,0 +1,176 @@
...
+	}()
+	err = v1beta1.AddToScheme(k8sscheme.Scheme)
+	g.Expect(err).ToNot(HaveOccurred())
+	err = v1beta2.AddToScheme(k8sscheme.Scheme)
+	g.Expect(err).ToNot(HaveOccurred())
+	err = frrv1beta1.AddToScheme(k8sscheme.Scheme)
+	g.Expect(err).ToNot(HaveOccurred())
+	m, err := manager.New(cfg, manager.Options{Metrics: metricsserver.Options{BindAddress: "0"}})
+	g.Expect(err).ToNot(HaveOccurred())
+
```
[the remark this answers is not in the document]
fedepaol: done

oribon:
lgtm :smile:

PR #2163, opened 2023-11-14
Some fixes around reflect.DeepEqual

oribon:
Not a bug in particular, just fixing some stuff around DeepEqual where it's mostly wrong. Further improvements are still needed, just a step towards better comparisons.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1109,7 +1109,7 @@ func bfdIntFromConfig(value *uint32, min, max uint32) (*uint32, error) {
 func validateDuplicateBGPAdvertisements(ads []metallbv1beta1.BGPAdvertisement) error {
 	for i := 0; i < len(ads); i++ {
 		for j := i + 1; j < len(ads); j++ {
-			if reflect.DeepEqual(ads[i], ads[j]) {
+			if reflect.DeepEqual(ads[i].Spec, ads[j].Spec) { // TODO: proper equal considering the slices order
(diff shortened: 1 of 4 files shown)


internal/config/config.go
```diff
@@ -1099,7 +1109,7 @@ func bfdIntFromConfig(value *uint32, min, max uint32) (*uint32, error) {
 func validateDuplicateBGPAdvertisements(ads []metallbv1beta1.BGPAdvertisement) error {
 	for i := 0; i < len(ads); i++ {
 		for j := i + 1; j < len(ads); j++ {
-			if reflect.DeepEqual(ads[i], ads[j]) {
+			if reflect.DeepEqual(ads[i].Spec, ads[j].Spec) { // TODO: proper equal considering the slices order
```
fedepaol, line 1112: I'd rather fix it now or file an issue, instead of a todo in the code

fedepaol:
We should stop and think what is the best way to ensure we won't hit this again when we add a new slice to an object that is relying on the order. I am afraid if we don't enforce in some way, we'll hit it again because it's really hard to see the big picture and remember/foresee if a sort is needed.

Some random thoughts:
- we either enforce the ordering at the source (when we convert the API to the internal config) or at the end (when we generate the frr configuration)
- maps are useful for checking duplicates, but we should probably have only slices when it comes to propagating the data from one layer to another
- we could even think about some "shuffler" utility that leverages reflection, navigates the object tree and shuffles whenever it finds a slice. This could be useful to validate that the conversion is stable

PR #2165, opened 2023-11-15
E2E: set the controller-runtime logger in BeforeSuite

fedepaol:
Since we updated the controller-runtime, e2e execution produces an annoying "[controller-runtime] log.SetLogger(...) was never called" message. Here we set the logger to the gingko writer to make it happy.


oribon (approved):
lgtm

PR #2176, opened 2023-11-21
tasks.py: Fix regex errors

liornoy:
Fix "SyntaxWarning" regex errors that Python is complaining about in the latest version, by adding a "r" character before strings with escape sequence.


fedepaol:
LGTM

PR #2178, opened 2023-11-22
Speaker: Allow configuring MemberList timeouts for WAN environments

pbasov:
In WAN environments timeouts and intervals for Probes and Gossip in Memberlist can be too tight, since MetalLB uses [DefaultLANConfig](https://github.com/hashicorp/memberlist/blob/master/config.go#L296) by default. 

This PR allows using [DefaultWANConfig](https://github.com/hashicorp/memberlist/blob/master/config.go#L340) instead through a chart value.


charts/metallb/templates/speaker.yaml
```diff
@@ -235,6 +235,9 @@ spec:
         {{- if .Values.loadBalancerClass }}
         - --lb-class={{ .Values.loadBalancerClass }}
         {{- end }}
+        {{- if .Values.speaker.wanConfig }}
+        - --ml-wan-config={{ .Values.speaker.wanConfig }}
```
fedepaol, line 239: given it's a bool you can do directly =true (or, I don't remember if being a flag sets it automatically to true when the flag is specified)

internal/speakerlist/speakerlist.go
```diff
@@ -56,7 +56,12 @@ func New(logger log.Logger, nodeName, bindAddr, bindPort, secret, namespace, lab
 		return &sl, nil
 	}
 
-	mconfig := memberlist.DefaultLANConfig()
+	var mconfig *memberlist.Config
```
fedepaol, line 59: let's init to DefaultWanConfig and override if the parameter is there

mconfig := memberlist.DefaultLANConfig()
if xxx {
   mconfig = memberlist.DefaultWANConfig()
}

fedepaol:
@pbasov mind changing the commit body to explain the rationale behind the change (basically what you put in the pr text)

charts/metallb/templates/speaker.yaml
pbasov: Yep, works without an explicit `true`, fixed

pbasov:
Updated with your suggestions, PTAL

fedepaol:
thanks, sending to merge queue!

PR #2180, opened 2023-11-26
E2E: Enable frr-k8s mode

oribon:
We adjust the e2e tests to be able to run against an frr-k8s mode deployment, this ensures us we don't regress from what is working on the other modes.


e2etest/pkg/config/frrprovider.go
```diff
@@ -0,0 +1,118 @@
...
+	"fmt"
+
+	"go.universe.tf/e2etest/pkg/executor"
+	"go.universe.tf/e2etest/pkg/metallb"
+	corev1 "k8s.io/api/core/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+type FRRProvider interface {
```
fedepaol, line 15: how about an frrprovider package (and renaming everything accordingly?)

e2etest/pkg/config/frrprovider.go
```diff
@@ -0,0 +1,118 @@
...
+	FRRForSpeaker(namespace, name string) (executor.Executor, error)
```
fedepaol, line 16: can you add a godoc for these two functions?

fedepaol, line 16: FRRExecutorForSpeaker or just FRRExecutorFor(namespace, speakerName string)

e2etest/pkg/config/frrprovider.go
```diff
@@ -0,0 +1,118 @@
...
+	BGPMetricsPodForSpeaker(namespace, name string) (pod *corev1.Pod, metricsPrefix string, err error)
```
fedepaol, line 17: Same

e2etest/pkg/metrics/metrics.go
```diff
@@ -31,21 +31,31 @@ type prometheusResponseData struct {
 
 // MetricsForPod returns the parsed metrics for the given pod, scraping them
 // from the source pod.
-func ForPod(controller, target *corev1.Pod, namespace string) ([]map[string]*dto.MetricFamily, error) {
+func ForPod(promPod, target *corev1.Pod, namespace string) ([]map[string]*dto.MetricFamily, error) {
```
fedepaol, line 34: why did you change from controller to promPod?

e2etest/pkg/config/frrprovider.go
oribon: done (moved to `e2etest/pkg/frr/provider`)

e2etest/pkg/config/frrprovider.go
oribon: done

oribon: done (the latter)

e2etest/pkg/config/frrprovider.go
oribon: done

e2etest/pkg/metrics/metrics.go
oribon: I don't remember if I had a strict reason, might be just as part of copying from what we did in frr-k8s without thinking too much if it's necessary (in frr-k8s it was). In any case, I prefer leaving this with the promPod as it's more natural to scrape from it and shouldn't change anything, wdyt?

fedepaol: Yep, the reason we scrape from the controller is, originally we weren't deploying prometheus. I was curious if there was another reason behind moving other than "it's cleaner" (which is totally fine).

PR #2183, opened 2023-11-27
Adding frrk8s deployment method to helm

fedepaol:
Extending the charts / CI / tasks to support frrk8s via helm


charts/metallb/values.yaml
```diff
@@ -346,3 +347,6 @@ speaker:
 crds:
   enabled: true
   validationFailurePolicy: Fail
+
+frrk8s:
```
oribon, line 351: can you add a short comment about what this is about + that it is mutually exclusive with frr mode?

.github/workflows/ci.yaml
```diff
@@ -146,6 +146,15 @@ jobs:
           helm repo add frrk8s https://metallb.github.io/frr-k8s
           ct lint --check-version-increment=false --validate-maintainers=false --target-branch ${{ github.event.repository.default_branch }}
 
+      - name: Check if deps are updated
+        run: |
+          helm repo add frrk8s https://metallb.github.io/frr-k8s
+          helm dependency update charts/metallb
```
oribon, line 152: this creates the tgz?

PR #2188, opened 2023-11-29
Kustomize: pin the frr-k8s overlay to 0.0.1

fedepaol:
main is a moving target, we want to pin to a stable version.


oribon (approved):
LGTM

PR #2189, opened 2023-11-29
unit: update envtests

oribon:
We move the tests that rely on envtest to a single ginkgo suite as it is a more standard approach.
envtest-setup is now installed using go install as in newer controller-runtime.


fedepaol:
lgtm, thanks!

PR #2190, opened 2023-11-30
frr-k8s: support passwords

oribon:
We add support for passing the secret ref for peers that specify it to frr-k8s and adjust the single e2e to use all containers.


internal/bgp/frrk8s/frrk8s.go
```diff
@@ -233,6 +235,10 @@ func (sm *sessionManager) updateConfig() error {
 			}
 			portUint16 := uint16(portUint)
 
+			if reflect.DeepEqual(s.PasswordRef, corev1.SecretReference{}) && s.Password != "" {
```
fedepaol, line 238: My idea was to support both and have the password translated to a secret, but it's also true that a smoother approach would be just to add plain passwords support in frrk8s.
I am not sure I like this, because it would require users to change their configurations if they want to switch to the new mode.

internal/config/config.go
```diff
@@ -468,36 +470,43 @@ func peerFromCR(p metallbv1beta2.BGPPeer, passwordSecrets map[string]corev1.Secr
...
 		Password:      password,
+		PasswordRef:   passwordRef,
 		BFDProfile:    p.Spec.BFDProfile,
 		EBGPMultiHop:  p.Spec.EBGPMultiHop,
 		VRF:           p.Spec.VRFName,
 	}, nil
 }
 
-func passwordForPeer(p metallbv1beta2.BGPPeer, passwordSecrets map[string]corev1.Secret) (string, error) {
+func passwordForPeer(p metallbv1beta2.BGPPeer, passwordSecrets map[string]corev1.Secret) (string, corev1.SecretReference, error) {
```
fedepaol, line 480: At the cost of doing the same thing twice, I'd split this in "passwordForPeer" and "passwordSecretForPeer".
The two return values is not super clear

internal/config/validation.go
```diff
@@ -205,3 +211,12 @@ func hasBFDEcho(peer *Peer, bfdProfiles map[string]*BFDProfile) bool {
 func peerAddressKey(peer metallbv1beta2.BGPPeerSpec) string {
 	return fmt.Sprintf("%s-%s", peer.Address, peer.VRFName)
 }
+
+func discardPeersWithRawPasswords(peers []metallbv1beta2.BGPPeer) error {
```
fedepaol, line 215: See my other comment, this feels clunky

internal/config/config.go
oribon: went for something a bit different, defaulting to the peer's password and unpacking the secret when specified

internal/bgp/frrk8s/frrk8s.go
```diff
@@ -233,6 +235,11 @@ func (sm *sessionManager) updateConfig() error {
 			}
 			portUint16 := uint16(portUint)
 
+			password := s.Password
```
fedepaol, line 238: still not sure I like this.

Possible alternatives:

- we don't translate secret -> password in frrk8s mode
- we have two different fields, one with the "to be used in local bgp (frr / native)", and the other which is the password to propagate

fedepaol:
LGTM

PR #2193, opened 2023-12-01
Move webhooks out of API package

### LINKED ISSUE #2184: Move webhooks out of the API definitions

We import the metallb API in our controllers to create the various CRDs. Controller-runtime has made a breaking change making the API of 0.15 webhooks incompatible with 0.14 leading us to be unable to import metallb as a result without conflicts until it is bumped. We can always create our own folder and import the APIs solely without the webhooks, but if the API changes we are vulnerable to losing track of it.

An example of the issues we run into at the moment.

```
# go.universe.tf/metallb/api/v1beta2
vendor/go.universe.tf/metallb/api/v1beta2/bgppeer_webhook.go:39:27: cannot use &BGPPeer{} (value of type *BGPPeer) as admission.Validator value in variable declaration: *BGPPeer does not implement admission.Validator (wrong type for method ValidateCreate)
                have ValidateCreate() error
                want ValidateCreate() (admission.Warnings, error)
```

What we've done with a lot of our packages is pull out the webhooks and have them in their own package so that version change doesn't matter, however that requires a refactor from the webhook as a method on the API type since that cannot be pulled out of the package. Instead we could use a generic handle function.

For the time being, an upgrade to 0.15 with the corresponding interface changes, such as validations being switched from returning an `err` to `(admission.Warnings, error)` should suffice.

### Additional context

I can work on this, feel free to assign to me and I can send a pull request in later this week or next week.

michaelasp:
Fixed #2184 

Moves webhooks out of API by creating a handler for each type rather than linking to API types. This ensures that consumers can pull APIs without relying on any controller-runtime specific dependencies.


michaelasp:
A fairly large change unfortunately, the process is fairly basic though. We move all webhooks into their own package and create a validator object rather than base it off the API type. Behavior should be nearly identical before/after.

webhooks/webhookv1beta1/addresspool_webhook.go
```diff
@@ -0,0 +1,163 @@
+/*
```
fedepaol, line 1: I'd move this under internal/k8s/webhooks as it's not necessary to have this exported

webhooks/validate/cluster_resources.go
fedepaol, line 1: Moving the package out of api removes the need of the validator interface (which was in place to break the circular dependency between api and config).
At this point I'd try to understand if it makes sense to remove it and use the config package directly from the webhooks.

fedepaol:
> A fairly large change unfortunately, the process is fairly basic though. We move all webhooks into their own package and create a validator object rather than base it off the API type. Behavior should be nearly identical before/after.

Yep, no worries. CI should give us confidence to understand if something breaks. I see there are some linter checks to hammer.

webhooks/validate/cluster_resources.go
michaelasp: It seems like there is some dynamic behavior linked to this validator. Based on the type of cluster(BGP, FRR, ...) it either allows or disallows some types of objects. I think this may require further look, and may not make sense to bundle with this change. I can look into removing the interface, but it looks like a bit more work than just removing the interface altogether.

michaelasp:
Not sure about the changelog, let me look into why the e2e is failing.

webhooks/validate/cluster_resources.go
fedepaol: fair enough, this can be done in a separate PR (but I think it's doable)

fedepaol:
> Looks like all tests are passing except for the changelog test. Looks like that was added a few days ago, let me know what needs to be done there. Otherwise looks fairly good to go!

Yep this is me trying to find a better way than building the release notes manually. Just change the release notes file under https://github.com/metallb/metallb/blob/main/website/content/release-notes/_index.md with an entyr under chores

e2etest/e2etest_suite_test.go
```diff
@@ -65,15 +65,7 @@ var (
...
 	e2econfig.CopyFlags(e2econfig.Flags, flag.CommandLine)
 	framework.RegisterCommonFlags(flag.CommandLine)
-	/*
-		Using framework.RegisterClusterFlags(flag.CommandLine) results in a panic:
-		"flag redefined: kubeconfig".
-		This happens because controller-runtime registers the kubeconfig flag as well.
-		To solve this we set the framework's kubeconfig directly via the KUBECONFIG env var
-		instead of letting it call the flag. Since we also use the provider flag it is handled manually.
-	*/
-	flag.StringVar(&framework.TestContext.Provider, "provider", "", "The name of the Kubernetes provider (gce, gke, local, skeleton (the fallback if not set), etc.)")
```
fedepaol, line 75: This change is not related to the commit. Mind moving to another commit?

michaelasp: Done!

fedepaol:
Sending to merge queue, thanks!

PR #2196, opened 2023-12-04
Change default BGP mode to frr in dev-env command

liornoy:
Change the default BGP mode in the tasks.py dev-env command to FRR.

The rationale for this is that we no longer develop features for the native mode, and it makes sense to set the development environment to default to FRR.


fedepaol:
@liornoy can you add an entry in the changelog file?

liornoy:
@fedepaol why merge queue cancelled if  `This branch has no conflicts with the base branch`

Issue #2158, opened 2023-11-09
CRD Status: Implement L2 Service Status


oribon:
### Is your feature request related to a problem?

https://github.com/metallb/metallb/blob/main/design/crd-status.md

### Describe the solution you'd like

As described in https://github.com/metallb/metallb/blob/main/design/crd-status.md#service-announcement---l2 we want to implement the `ServiceL2Status` resource

### I've read and agree with the following

- [X] I've checked all open and closed issues and my request is not there.
- [X] I've checked all open and closed pull requests and my request is not there.

lwabish:
I'd love to implement this

oribon:
thanks! assigned, let us know if you need any help in #metallb-dev or here

lwabish:
The very first problem I met is: 
```
❯ kubebuilder create api --version v1beta1 --kind ServiceL2Status --group metallb.io
Error: failed to create API: unable to load configuration file: 
unable to load the configuration: 
unable to read "PROJECT" file: open PROJECT: no such file or directory
```
On the other hand, it seems not possible to init a kubebuilder project based on a non empty folder.
```
❯ kubebuilder init --domain metallb.io
Error: failed to initialize project: unable to run pre-scaffold tasks of "base.go.kubebuilder.io/v4": 
target directory is not empty (only go.mod, go.sum, files and directories with the prefix ".", 
files with the suffix ".md" or capitalized files name are allowed); found existing file "api"
```
So now there seems no way to generate the xxx_type.go ? I wonder how was the xxx_types.go written before?

fedepaol:
> The very first problem I met is:
> 
> ```
> ❯ kubebuilder create api --version v1beta1 --kind ServiceL2Status --group metallb.io
> Error: failed to create API: unable to load configuration file: 
> unable to load the configuration: 
> unable to read "PROJECT" file: open PROJECT: no such file or directory
> ```
> 
> On the other hand, it seems not possible to init a kubebuilder project based on a non empty folder.
> 
> ```
> ❯ kubebuilder init --domain metallb.io
> Error: failed to initialize project: unable to run pre-scaffold tasks of "base.go.kubebuilder.io/v4": 
> target directory is not empty (only go.mod, go.sum, files and directories with the prefix ".", 
> files with the suffix ".md" or capitalized files name are allowed); found existing file "api"
> ```
> 
> So now there seems no way to generate the xxx_type.go ? I wonder how was the xxx_types.go written before?

[...]

PR #2198, opened 2023-12-05
Implement ServiceL2Status CRD to expose service announcing status

lwabish:
This is an implement of what I was assigned in  #2158.
The core implementation have almost been done and I have test the  feature in my local dev-env. 
I will continue doing some detailed work such as the helm chart files, etc.
Please do some early code review if possible, any suggestions would be appreciated.

/kind feature

```release-note
Implement ServiceL2Status CRD to expose service announcing status in layer2 mode
```


oribon:
Thanks a lot for opening this!
I think we should take a bit different approach regarding how we manage these resources.
Instead of an imperative approach where this is embedded in the existing layer2_controller and we take action according to the current step, we should go for a more declarative one where we have a new k8s controller that can react to general events, fetching the desired state of a service and building the status accordignly.
More specifically - we can make the layer2_controller expose the state of advertised services (e.g by a func that accepts a namespaced name for a service and return its advertisement state), add a new k8s controller that listens to a channel and feed that channel from the layer2_controller every time a state of an advertised service changes with an event that service x has changed. When the new k8s controller gets the event that service x has changed it queries its state and builds the relevant status.
You can take inspiration from how we did it in frr-k8s, where the nodestate controller fetches the different pieces to create the state from other components:
https://github.com/metallb/frr-k8s/blob/main/internal/controller/frrstate_controller.go
https://github.com/metallb/frr-k8s/blob/8ea5c2e33beaab49af944d14452c85993095e844/cmd/main.go#L197

This will allow us to delegate the concerns of managing the statuses to a single point which should be easy to follow.
Also, there are still some corners we need to make sure we're covering such as not leaving stale statuses around and making sure the user doesn't edit them but we can leave them for later once we settle on the general picture. wdyt?

Again, thanks a lot for your efforts and please let me know if you need any clarifications or help!

lwabish:
@oribon Please review the new implement

internal/k8s/controllers/layer2_status_controller.go
oribon, line 1: can you add some tests under `reconciliation_test.go?

oribon (commented):
Thanks a lot for the PR!
did a first pass, can you please add an e2e as well so we have a general sense if it's working? Please let me know if you need anything :)

lwabish:
Thanks for your reviewing and sugguestions.I'll keep working on it.

lwabish:
> Thanks a lot for the PR! did a first pass, can you please add an e2e as well so we have a general sense if it's working? Please let me know if you need anything :)

I met this problem while trying to process go module dependencies in e2etest directory.

```
❯ cd e2etest
❯ go mod download
go: k8s.io/endpointslice@v0.0.0: invalid version: unknown revision v0.0.0
❯ go mod why k8s.io/endpointslice
go: downloading k8s.io/endpointslice v0.0.0
# k8s.io/endpointslice
(main module does not need package k8s.io/endpointslice)
```

Somehow this didn't affect `inv e2etest`.
Adding `k8s.io/endpointslice => k8s.io/endpointslice v0.28.2` to go.mod could solve this problem but I'm not quite sure if this is a env specific problem.
Should I commit my e2etest/go.mod and related go module files?
Of course this is not a fatal or urgent problem but a little confusing.

oribon:
it's a bit hard for me to say, but in general committing whatever `go mod tidy` changes should be fine 😅

internal/k8s/controllers/layer2_status_controller.go
```diff
@@ -0,0 +1,97 @@
...
+func (r *Layer2StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	level.Info(r.Logger).Log("controller", "Layer2StatusReconciler", "start reconcile", req.NamespacedName.String())
+	defer level.Info(r.Logger).Log("controller", "Layer2StatusReconciler", "end reconcile", req.NamespacedName.String())
+
+	state := &v1beta1.ServiceL2Status{}
+	err := r.Client.Get(ctx, types.NamespacedName{Name: req.Name, Namespace: req.Namespace}, state)
+	if errors.IsNotFound(err) {
+		state.Name, state.Namespace = req.Name, req.Namespace
+		err = r.Client.Create(ctx, state)
+	}
```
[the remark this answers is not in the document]
lwabish: It seems that CreateOrUpdate can't update status subresource, am I doing something wrong? I have checked  both `frrk8s_config_controller` and [example_test](https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/controller/controllerutil/example_test.go), they are doing update for spec, not for status subresource.

```go
// even with this, the cr created first time doesn't contain the latest status
state.Status = desiredStatus
_, err = controllerutil.CreateOrUpdate(ctx, r.Client, state, func() error {
// not working, the status remains empty unless a manual update
state.Status = desiredStatus
return nil
})
```

internal/k8s/controllers/layer2_status_controller.go
```diff
@@ -0,0 +1,97 @@
...
+
+	ipAdvS := r.StatusFetcher(req.NamespacedName.String())
+	if len(ipAdvS) < 1 {
+		level.Debug(r.Logger).Log("controller", "Layer2StatusReconciler", "delete serviceL2status", req.NamespacedName.String())
+		if err = r.Client.Delete(ctx, state); err != nil {
+			level.Error(r.Logger).Log("controller", "Layer2StatusReconciler", "failed to delete", err)
+			return ctrl.Result{}, err
+		}
+		return ctrl.Result{}, nil
+	}
```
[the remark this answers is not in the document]
lwabish: The channel some speaker is watching gets an element only from the same speaker after it ensures that some service belongs to itself. So It seems that this is not a problem.

lwabish:
@oribon I have made the changes according to your sugguestions, please review them

internal/k8s/controllers/layer2_status_controller.go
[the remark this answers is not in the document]
oribon: oh, then sorry about that let's revert to what you initially did (just calling create and update in the end)

internal/k8s/controllers/layer2_status_controller.go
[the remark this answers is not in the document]
oribon: not sure I understand, the controller listens for both events that come from the channel and those that come from a change to an object (the `For(&v1beta1.ServiceL2Status{})`). what prevents speakers from deleting eachother's resources?

oribon:
> @oribon I have made the changes according to your sugguestions, please review them

thanks a lot, I'll get to this in the following days :) 
in the meanwhile I ran CI (and will rerun it again whenever you push)

lwabish:
I ran some git rebase commands according to git help page to fix DCO test.Now The DCO test is ok.
But It seems that my commit log is mixed with some other commits.Not sure if this is a problem.
Besides, the CI workflow seems to be waiting for approval to run
@oribon

lwabish:
> > I ran some git rebase commands according to git help page to fix DCO test.Now The DCO test is ok. But It seems that my commit log is mixed with some other commits.Not sure if this is a problem. Besides, the CI workflow seems to be waiting for approval to run @oribon
> 
> approved again, about the commit log it might cause some problems but anyways I'll ask you sooner or later to clean it up a bit 😅 the easiest way imo would be squashing your commits together (for example using `rebase -i`, cherry-picking them on top of main and overriding this branch. Also, it'll be easier to review when the commits are organized into a few (I guess it shouldn't be more than ~4-5 commits for this PR: API / Manifests / implementation / E2E)

No problem.This is what I planned to to after solving other problems.

speaker/layer2_controller.go
```diff
@@ -145,6 +146,7 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce
 
 func (c *layer2Controller) SetBalancer(l log.Logger, name string, lbIPs []net.IP, pool *config.Pool, client service, svc *v1.Service) error {
 	ifs := c.announcer.GetInterfaces()
+	statusUpdated := false
```
oribon, line 149: nit: rename to "updateStatus"

speaker/main.go
```diff
@@ -158,6 +159,9 @@ func main() {
 		SList:                  sList,
 		bgpType:                bgpImplementation(bgpType),
 		InterfaceExcludeRegexp: interfacesToExclude,
+		Layer2StatusChange: func(svc *v1.Service) {
```
oribon, line 162: I think it's cleaner if this would receive a namespacedname instead of a service (because we don't do anything meaningful with the service object)

internal/k8s/controllers/layer2_status_controller.go
```diff
@@ -0,0 +1,134 @@
...
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/source"
+)
+
+type StatusFetcher func(namespace, name string) []layer2.IPAdvertisement
```
oribon, line 25: sorry in a second thought maybe it's better to have this one namespaced name instead separated as it's more common in the code (sorry for the hassle)

internal/k8s/controllers/reconciliation_test.go
```diff
@@ -426,3 +445,34 @@ var _ = Describe("FRRK8S Controller", func() {
...
+			layer2StatusUpdateChan <- NewL2StatusEvent(testNamespace, name)
+			Eventually(func() string {
+				toCheck := v1beta1.ServiceL2Status{}
+				err := k8sClient.Get(context.TODO(), client.ObjectKey{Name: name, Namespace: testNamespace}, &toCheck)
+				if err != nil {
+					return ""
+				}
+				return toCheck.Status.Node
+			}, 5*time.Second, 200*time.Millisecond).Should(Equal(testNodeName))
+
```
oribon, line 467: can you add an update to the interface and verify they are reflected in the status?

e2etest/l2tests/interface_selector.go
```diff
@@ -83,6 +85,40 @@ var _ = ginkgo.Describe("L2-interface selector", func() {
 			framework.ExpectNoError(err)
 		})
 
+		ginkgo.It("Validate ServiceL2Status interface", func() {
+			ginkgo.By("generate a random int to choose some interface for announcing")
+			i := rand.Intn(len(NodeNics))
```
oribon, line 90: let's just choose the first no need for this rand (and also fail the test if len(NodeNics) == 0

e2etest/l2tests/interface_selector.go
```diff
@@ -83,6 +85,40 @@ var _ = ginkgo.Describe("L2-interface selector", func() {
...
+				err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
+				framework.ExpectNoError(err)
+			}()
+
+			var l2Status *metallbv1beta1.ServiceL2Status
+			gomega.Eventually(func() error {
+				var err error
+				l2Status, err = status.GetL2Status(ConfigUpdater.Client(), svc)
+				return err
+			}, 2*time.Minute, 1*time.Second).ShouldNot(gomega.HaveOccurred())
```
oribon, line 117: can you add a 5s `gomega.Consistently` after this to make sure the speakers do not fight?

e2etest/l2tests/l2.go
```diff
@@ -128,6 +129,26 @@ var _ = ginkgo.Describe("L2", func() {
...
+			ginkgo.By("validating the node in ServiceL2Status is the one who is announcing for the service")
+			allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+			framework.ExpectNoError(err)
+			gomega.Eventually(func() string {
+				node, err := k8s.GetSvcNode(cs, svc.Namespace, svc.Name, allNodes)
+				if err != nil {
+					return err.Error()
+				}
+				return node.Name
+			}, time.Minute, time.Second).Should(gomega.Equal(l2Status.Status.Node))
```
oribon, line 151: same (adding a consistently)

oribon (commented):
looks a lot better thanks a lot!

internal/k8s/controllers/layer2_status_controller.go
```diff
@@ -0,0 +1,134 @@
...
+
+	return ctrl.Result{}, nil
+}
+
+func (r *Layer2StatusReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1beta1.ServiceL2Status{}).
+		WatchesRawSource(&source.Channel{Source: r.Chan}, &handler.EnqueueRequestForObject{}).
+		// only trigger reconcile function when the source object is not a ServiceL2Status
+		// this prevents reconciling cycle because we operate ServiceL2Status itself in the reconcile function
```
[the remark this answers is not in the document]
lwabish: Yes I got you but in our case I think it's slightly different.
With`ctrl.NewControllerManagedBy(mgr).For().WatchesRawSource()`, both ServiceL2Status and GenericEvent objects can trigger the reconcile and the reconciler itself may create ServiceL2Status object, which leads to a cycle reconciling.
Meanwhile, I have tried deleting the `For()` but the controller stops working.So We have to keep the `For()`  but add an `EventFilter` to filter out ServiceL2Status from triggering the reconciling.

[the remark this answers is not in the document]
oribon: I understand, I'm talking about us wanting to reconcile on ServiceL2Status objects to make sure they always match the speaker. For example, if the user (or an unrelated speaker) modifies `status.node`, you'd expect the relevant speaker to override it to the correct value.
Regarding the cycle reconciliation, we can compare the current object with the desired status and updating only if it's needed, that way we won't trigger another reconciliation, e.g in `frrk8s_config_controller.go` we do:
```go
	if reflect.DeepEqual(current.Spec, r.desiredConfiguration.Spec) {
		level.Debug(r.Logger).Log("controller", "FRRK8sReconciler", "event", "not reconciling because of no change")
		return ctrl.Result{}, nil
	}
```

lwabish:
@oribon All done except for the proposal to add advertise ips. Please rerun the ci workflow 😄

internal/k8s/controllers/layer2_status_controller.go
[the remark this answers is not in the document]
lwabish: > I understand, I'm talking about us wanting to reconcile on ServiceL2Status objects to make sure they always match the speaker. For example, if the user (or an unrelated speaker) modifies `status.node`, you'd expect the relevant speaker to override it to the correct value. 

Indeed.In this case, it seems that I have to remove the EventFilter to allow ServiceL2Status object triggering the reconcile funtion.I'll work on it to see if there is any problem.

lwabish:
@oribon  
I think there's something about the reconcile funtion logic we should dive a little deeper.
Here are the two different approaches I have organized regarding how the reconcile function works.
1. The way that I wrote before:
With an eventfilter that filter out all serviceL2Status objects, the reconciler only got triggered by the channel elements, which means only the relavent speaker reconcile for a single service, no confilcts occur. But just like you mentioned recently, if some other controller or even users change the serviceL2Status, our reconciler can't get triggered.
This is also how the frr controller got implemented.
2. Without an eventfilter:
Every speaker can reconcile for a single service and external operations to the serviceL2Status object is considerd, which is more cloud-native. The difficulty lies in: the most direct status is stored only in one particular  speaker, others can't determine if a serviceL2Status should be delete or not from the status inside itself.
In my latest reconcile funtion, besides removing the eventfilter, I added some conditions to help reconciler determine if a serviceL2Status should be deleted, which solved the confilcts and things seems ok right now according to the unit tests and e2e tests. 
But I'm not sure if there is any other edge situation about this approach, wdyt ?

oribon:
> @oribon I think there's something about the reconcile funtion logic we should dive a little deeper. Here are the two different approaches I have organized regarding how the reconcile function works.
> 
> 1. The way that I wrote before:
>    With an eventfilter that filter out all serviceL2Status objects, the reconciler only got triggered by the channel elements, which means only the relavent speaker reconcile for a single service, no confilcts occur. But just like you mentioned recently, if some other controller or even users change the serviceL2Status, our reconciler can't get triggered.
>    This is also how the frr controller got implemented.
> 2. Without an eventfilter:
>    Every speaker can reconcile for a single service and external operations to the serviceL2Status object is considerd, which is more cloud-native. The difficulty lies in: the most direct status is stored only in one particular  speaker, others can't determine if a serviceL2Status should be delete or not from the status inside itself.
>    In my latest reconcile funtion, besides removing the eventfilter, I added some conditions to help reconciler determine if a serviceL2Status should be deleted, which solved the confilcts and things seems ok right now according to the unit tests and e2e tests.
>    But I'm not sure if there is any other edge situation about this approach, wdyt ?

I see what you're saying, our problem is that currently a speaker can't know if it's responsible for a given status resource. I think a good approach (and the most bulletproof one) would be embedding the node's name as part of the resource object like we intend doing in the [bgp status resource](https://github.com/metallb/metallb/blob/main/design/crd-status.md#service-announcement---bgp).
That way the question "is this speaker responsible for the object" boils down to "is the resource's name has the node's name as a suffix" (leveraging the fact that a resource's name is immutable) and reflecting this in the event filter.
Doing that, we can revert to your original implementation without the new auxiliary function, deleting the object if it has 0 advs but still having the advantage of listening for objects changes.
wdyt? cc @fedepaol

lwabish:
> @oribon @lwabish haven't checked the implementation yet.
> 
> My gut feeling is, if a given l2 status is shared among different speakers (i.e. the service moves from one speaker to the other, for example because an endpoint moved or a node selector changed), the old owner only knows it doesn't own the service anymore unless we add extra logic to understand if no one should advertise the service at all. So a given speaker will have to choose wether delete the status or not, just because it's not advertising the service, and getting this right looking at the current status is not easy (what if the speaker that should delete crashes for example)?
> 
> So, given we are doing the same for BGP, I am not against to not naming the resource as the service but having a "per node" l2status instance (where the service name / node name can be added as labels as in BGP). The name can be the concatenation of service and node, or something autogenerated, but by doing this any speaker will take care of the lifecycle of a single instance and its life (and ours) will be easier.
> 
> Hope I got things right

With this design, I think the controller logic could be simpler.
As a user , if I want to investigate my service annoucement status, my command would be like:
```bash
kubectl get servicel2status -n $NS --selector=metallb.io/service=$SVC-NAME
```
whose output would contains a couple of  servicel2status objects.
If I am understanding this right, maybe we should optimize the servicel2status status object design.For example the `status.node` as a string would be empty except for the only one which is advertising, in this case a boolean type may be more clear?

fedepaol:
> > @oribon @lwabish haven't checked the implementation yet.
> > My gut feeling is, if a given l2 status is shared among different speakers (i.e. the service moves from one speaker to the other, for example because an endpoint moved or a node selector changed), the old owner only knows it doesn't own the service anymore unless we add extra logic to understand if no one should advertise the service at all. So a given speaker will have to choose wether delete the status or not, just because it's not advertising the service, and getting this right looking at the current status is not easy (what if the speaker that should delete crashes for example)?
> > So, given we are doing the same for BGP, I am not against to not naming the resource as the service but having a "per node" l2status instance (where the service name / node name can be added as labels as in BGP). The name can be the concatenation of service and node, or something autogenerated, but by doing this any speaker will take care of the lifecycle of a single instance and its life (and ours) will be easier.
> > Hope I got things right
> 
> With this design, I think the controller logic could be simpler. As a user , if I want to investigate my service annoucement status, my command would be like:
> 
> ```shell
> kubectl get servicel2status -n $NS --selector=metallb.io/service=$SVC-NAME
> ```

And this would be the same as we are proposing for BGP (because then, we'd have multiple nodes advertising the service)

> 
> whose output would contains a couple of servicel2status objects. If I am understanding this right, maybe we should optimize the servicel2status status object design.For example the `status.node` as a string would be empty except for the only one which is advertising, in this case a boolean type may be more clear?

What I am suggesting here is, if a given speaker is not advertising the service anymore, then it should delete its own instance. So at a given time, there would be only one l2 adv per service (assuming one speaker is advertising).

lwabish:
> > > @oribon @lwabish haven't checked the implementation yet.
> > > My gut feeling is, if a given l2 status is shared among different speakers (i.e. the service moves from one speaker to the other, for example because an endpoint moved or a node selector changed), the old owner only knows it doesn't own the service anymore unless we add extra logic to understand if no one should advertise the service at all. So a given speaker will have to choose wether delete the status or not, just because it's not advertising the service, and getting this right looking at the current status is not easy (what if the speaker that should delete crashes for example)?
> > > So, given we are doing the same for BGP, I am not against to not naming the resource as the service but having a "per node" l2status instance (where the service name / node name can be added as labels as in BGP). The name can be the concatenation of service and node, or something autogenerated, but by doing this any speaker will take care of the lifecycle of a single instance and its life (and ours) will be easier.
> > > Hope I got things right
> > 
> > 
> > With this design, I think the controller logic could be simpler. As a user , if I want to investigate my service annoucement status, my command would be like:
> > ```shell
> > kubectl get servicel2status -n $NS --selector=metallb.io/service=$SVC-NAME
> > ```
> 
> And this would be the same as we are proposing for BGP (because then, we'd have multiple nodes advertising the service)
> 
> > whose output would contains a couple of servicel2status objects. If I am understanding this right, maybe we should optimize the servicel2status status object design.For example the `status.node` as a string would be empty except for the only one which is advertising, in this case a boolean type may be more clear?
> 
> What I am suggesting here is, if a given speaker is not advertising the service anymore, then it should delete its own instance. So at a given time, there would be only one l2 adv per service (assuming one speaker is advertising).

Got it.Sounds good to me. I'll start implement it. By the way, do we need to update the design doc before implementing?

fedepaol:
> Got it.Sounds good to me. I'll start implement it. By the way, do we need to update the design doc before implementing?

I'd say it's not necessary. It's totally natural for an implementation to drift from a design, as reality differs from our speculations :P

lwabish:
@oribon Hi, the latest ci workflow got an [error](https://github.com/metallb/metallb/actions/runs/7675640366/job/20932986745?pr=2198#step:10:250) because of lacking my new crd in metallb operator, which is another repo.

I'm not sure what is the best approach to deal with this situation: Skip some specific test, add my new crd to operator repo, or anything else?

oribon:
> @oribon Hi, the latest ci workflow got an [error](https://github.com/metallb/metallb/actions/runs/7675640366/job/20932986745?pr=2198#step:10:250) because of lacking my new crd in metallb operator, which is another repo.
> 
> I'm not sure what is the best approach to deal with this situation: Skip some specific test, add my new crd to operator repo, or anything else?

We can ignore it for now (this needs to be changed in the operator repo, known chicken-egg problem), is this ready for another round of review other than that?

lwabish:
> > @oribon Hi, the latest ci workflow got an [error](https://github.com/metallb/metallb/actions/runs/7675640366/job/20932986745?pr=2198#step:10:250) because of lacking my new crd in metallb operator, which is another repo.
> > I'm not sure what is the best approach to deal with this situation: Skip some specific test, add my new crd to operator repo, or anything else?
> 
> We can ignore it for now (this needs to be changed in the operator repo, known chicken-egg problem), is this ready for another round of review other than that (if so can you please rebase your commits like last time)?

@oribon Rebase done.
By the way, I found metallb imports lint rules seems a little different from what I usually do. I'm used to  linting imports with [goimports-reviser](https://github.com/incu6us/goimports-reviser), which splits imports to three zones(go packages/external packages/project packages). This leads to my commits contain some imports modifications, which increased the likelihood of code conflicts.
So I wonder what is the best practice of go imports you guys are following. I guess `inv lint` is enough?

oribon:
right `inv lint` should be enough (and we're forcing it in ci iirc)
thanks, I'll give this a look in the following days :)

e2etest/pkg/status/l2status.go
```diff
@@ -0,0 +1,39 @@
...
+func GetSvcPossibleL2Status(cs client.Client, svc *v1.Service, nodes *v1.NodeList) ([]*v1beta1.ServiceL2Status, error) {
+	var l2Statuses []*v1beta1.ServiceL2Status
+	for _, node := range nodes.Items {
+		if s, err := GetL2Status(cs, svc, node.Name); err == nil {
+			l2Statuses = append(l2Statuses, s)
+		} else if errors.IsNotFound(err) {
+			continue
+		} else {
+			return nil, err
+		}
```
oribon, line 36: same as before, let's avoid elses 😅 
```go
		s, err := GetL2Status(cs, svc, node.Name)
		if err != nil && errors.IsNotFound(err) {
			continue
		}
		if err != nil {
			return nil, err
		}
		l2Statuses = append(l2Statuses, s)
```

oribon (commented):
left a few comments but this looks really good and almost ready!
I have a few concerns now about something related to the naming we chose in regards to the cleanup, but we'll defer it to another PR after we merge this (which ofc you can take if you'd like) - I'll open a separate issue for that. thanks again!

internal/k8s/controllers/layer2_status_controller.go
```diff
@@ -0,0 +1,127 @@
...
+	level.Debug(r.Logger).Log("controller", "Layer2StatusReconciler", "updated state", dumpResource(state))
+
+	return ctrl.Result{}, nil
+}
+
+func (r *Layer2StatusReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&v1beta1.ServiceL2Status{}).
+		WatchesRawSource(&source.Channel{Source: r.ReconcileChan}, &handler.EnqueueRequestForObject{}).
+		Complete(r)
```
oribon, line 105: we can now add a `WithEventFilter` to reconcile only those that end with the node's name

internal/k8s/controllers/layer2_status_controller.go
```diff
@@ -0,0 +1,127 @@
...
+	StatusFetcher StatusFetcher
+}
+
+func (r *Layer2StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	level.Info(r.Logger).Log("controller", "Layer2StatusReconciler", "start reconcile", req.NamespacedName.String())
+	defer level.Info(r.Logger).Log("controller", "Layer2StatusReconciler", "end reconcile", req.NamespacedName.String())
+
+	ipAdvS := r.StatusFetcher(req.NamespacedName)
+
+	objName := fmt.Sprintf("%s-%s", req.Name, r.NodeName)
```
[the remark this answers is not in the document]
lwabish: Something wrong related to this did occured when I was debugging the tests.But later I found this worked just because wrong objNames fetched none statues so the reconcile stopped.
Of course the optimization you are offering is really good, will do it

lwabish:
@oribon Done :)

oribon (approved):
lgtm! thanks a lot, I know this took a while but the effort you put here is going to be very useful with the next status resources we want to implement :smile:
can you please rebase your commits again (also on top of main)?
ccing @fedepaol to also give this a look

lwabish:
@oribon Wonderful experience working with you guys and I learned a lot from your reviews.Thank you, too.
Should I still rebase the commits to four like before or squash them all?

fedepaol:
Folks, this looks really neat!
Sending to merge queue. Thanks a lot @lwabish for doing this, and @oribon for assisting!

fedepaol:
@lwabish I think you need to regenerate the manifests?

lwabish:
> @lwabish I think you need to regenerate the manifests?

Sorry I will fix that tonight

fedepaol:
Please rebase to the right commit

lwabish:
> Please rebase to the right commit

Ok, just finished

fedepaol:
@lwabish apologies, I was sure I sent this to be merged.
Would you be so kind to rebase again and fix the conflicts? Thanks a lot!

lwabish:
> @lwabish apologies, I was sure I sent this to be merged. Would you be so kind to rebase again and fix the conflicts? Thanks a lot!

No problem.

lwabish:
> > @lwabish apologies, I was sure I sent this to be merged. Would you be so kind to rebase again and fix the conflicts? Thanks a lot!
> 
> No problem.

@fedepaol Just finished.

PR #2199, opened 2023-12-05
FRRK8s mode: log more info under debug

fedepaol:
Adding more logs for better triaging.
Also, removing the desiredConfig field which is now useless.


internal/bgp/frrk8s/frrk8s.go
```diff
@@ -367,3 +383,20 @@ func sortMap[T any](toSort map[string]T) []T {
...
 	return res
 }
+
+func ConfigToDump(config frrv1beta1.FRRConfiguration) (string, error) {
+	toDump := config.DeepCopy()
+	for _, r := range toDump.Spec.BGP.Routers {
+		for i := range r.Neighbors {
+			if r.Neighbors[i].Password != "" {
+				r.Neighbors[i].Password = "<retracted>"
+			}
```
oribon, line 393: nit: I think we should put \<retracted\> anyways

fedepaol: I'd prefer to know when a password is set or it is empty as it may influence the behaviour.

oribon: I guess it is not crucial, my reasoning is more about the password field should not carry any info when dumped (that is having no password is also a form of info), and that's also what we do when dumping the BGPPeer.

fedepaol: fair enough, I'll change it

fedepaol: done

oribon (approved):
lgtm

PR #2201, opened 2023-12-06
E2E: change the way we enable the vrf container

fedepaol:
Instead of binding it to the bgp mode, we set an explicit flag that says if we want it or not, and we default it to false. This is useful when doing local tests to reduce the complexity, and for those dev environments which do not support VRFs.


oribon (approved):
lgtm

PR #2202, opened 2023-12-06
Webhook does not process transient errors properly

### LINKED ISSUE #2173: Webhook does not process transient errors properly
### MetalLB Version

latest

### Deployment method

Not relevant

### Main CNI

N/A

### Kubernetes Version

_No response_

### Cluster Distribution

_No response_

### Describe the bug

Using the webhooks, we do not want to block resources depending on the order they are received - e.g we don't want to block a peer specifying a bfdProfile that doesn't exist yet because it could eventually converge. today we achieve that using the TransientError  type, returning it only when an error that relates to dependencies between resources occur and we ignore this type of err in the webhook's validator.
The problem with that approach is that in some cases this causes the webhook to miss an actual error, because in the webhook we pass multiple objects of the same type to the validation and return early when hitting any error, which for example if a TransientError happens for the first resource the rest of the resources are not being validated at all.
As an example, consider the following resources:
```yaml
apiVersion: metallb.io/v1beta2
kind: BGPPeer
metadata:
  name: peer1
  namespace: metallb-system
spec:
  peerAddress: 172.16.30.2
  peerASN: 65433
  myASN: 65439
  bfdProfile: "non-existing"
---
apiVersion: metallb.io/v1beta2
kind: BGPPeer
metadata:
  name: peer2
  namespace: metallb-system
spec:
  peerAddress: 172.16.30.3
  peerASN: 0
  myASN: 65439
```
peer2 is invalid since it specifies peerASN: 0  so we would expect the webhook to block it anyways, but the outcome of the webhook changes as a result of the order: if peer2 is created without peer1 it is blocked, if it is created after peer1 it passes.
A possible solution is instead of ignoring transient errors in the webhook's validator, it would pass the resources to `For` without fields that can cause a transient error (resetting them), essentially not allowing the webhook to receive a transient error at all.
TransientErrors could stay in the codebase (serving as a "doc"), we would just not act on them in the webhooks.

### To Reproduce

described above

### Expected Behavior

described above

### Additional Context

described above

shimritproj:
Fixed  #2173 

There are TransientErrors that can occasionally result in overlooking genuine issues. The solution is to reset the fields that might be causing a TransientError.

### DIFF
--- a/internal/config/validator.go
+++ b/internal/config/validator.go
@@ -3,7 +3,7 @@
 package config
 
 import (
-	"errors"
+	"strings"
 
 	metallbv1beta1 "go.universe.tf/metallb/api/v1beta1"
 	metallbv1beta2 "go.universe.tf/metallb/api/v1beta2"
@@ -58,13 +58,43 @@ func (v *validator) Validate(resources ...client.ObjectList) error {
 			clusterResources.Nodes = append(clusterResources.Nodes, list.Items...)
 		}
 	}
+	clusterResources = resetTransientErrorsFields(clusterResources)
 	_, err := For(clusterResources, v.validate)
-	if errors.As(err, &TransientError{}) { // we do not want to make assumption on ordering in webhooks.
-		return nil
-	}
 	return err
 }
 
 func NewValidator(validate Validate) apivalidate.ClusterObjects {
 	return &validator{validate: validate}
+}
+
+// Returns the given ClusterResources without the fields that can cause a TransientError.
+// We use this so the webhooks do not make assumptions based on the ordering of objects.
+func resetTransientErrorsFields(clusterResources ClusterResources) ClusterResources {
+	for i := range clusterResources.Peers {
+		clusterResources.Peers[i].Spec.BFDProfile = ""
+		clusterResources.Peers[i].Spec.PasswordSecret = v1.SecretReference{}
+	}
+	for i, bgpAdv := range clusterResources.BGPAdvs {
+		var communities []string
+		for _, community := range bgpAdv.Spec.Communities {
+			// We want to pass only communities that are potentially explicit values.
+			if strings.Contains(community, ":") {
+				communities = append(communities, community)
+			}
+		}
+		clusterResources.BGPAdvs[i].Spec.Communities = communities
+	}
+	for i, legacyAddrPool := range clusterResources.LegacyAddressPools {
+		for j, bgpAdv := range legacyAddrPool.Spec.BGPAdvertisements {
+			var communities []string
+			for k := range bgpAdv.Communities {
+				// We want to pass only communities that are potentially explicit values.
+				if strings.Contains(bgpAdv.Communities[k], ":") {
+					communities = append(communities, bgpAdv.Communities[k])
+				}
+			}
+			clusterResources.LegacyAddressPools[i].Spec.BGPAdvertisements[j].Communities = communities
+		}
+	}
+	return clusterResources
 }
--- a/internal/config/validator_test.go
+++ b/internal/config/validator_test.go
@@ -5,16 +5,19 @@ package config
 import (
 	"testing"
 
-	"go.universe.tf/metallb/api/v1beta2"
+	"github.com/google/go-cmp/cmp"
+	metallbv1beta1 "go.universe.tf/metallb/api/v1beta1"
+	metallbv1beta2 "go.universe.tf/metallb/api/v1beta2"
+	v1 "k8s.io/api/core/v1"
 )
 
 func TestValidator(t *testing.T) {
 	v := validator{DontValidate}
 
-	bgpPeerList := v1beta2.BGPPeerList{
-		Items: []v1beta2.BGPPeer{
+	bgpPeerList := metallbv1beta2.BGPPeerList{
+		Items: []metallbv1beta2.BGPPeer{
 			{
-				Spec: v1beta2.BGPPeerSpec{
+				Spec: metallbv1beta2.BGPPeerSpec{
 					MyASN:      42,
 					ASN:        42,
 					Address:    "1.2.3.4",
@@ -26,5 +29,133 @@ func TestValidator(t *testing.T) {
 	err := v.Validate(&bgpPeerList)
 	if err != nil {
 		t.Error("The validator should not fail for non existing bfd profile")
+	}
+}
+
+func TestResetTransientErrorsFields(t *testing.T) {
+	tests := []struct {
+		desc             string
+		clusterResources ClusterResources
+		expected         ClusterResources
+	}{
+		{
+			desc:             "empty resource",
+			clusterResources: ClusterResources{},
+			expected:         ClusterResources{},
+		},
+		{
+			desc: "BFDProfiles, PasswordSecrets and Community references are reset",
+			clusterResources: ClusterResources{
+				Peers: []metallbv1beta2.BGPPeer{
+					{
+						Spec: metallbv1beta2.BGPPeerSpec{
+							BFDProfile: "myBFDProfile",
+							MyASN:      64512,
+							ASN:        64512,
+							Address:    "172.30.0.3",
+							PasswordSecret: v1.SecretReference{
+								Name:      "myName",
+								Namespace: "myNamespace",
+							},
+						},
+					},
+				},
+				Communities: []metallbv1beta1.Community{
+					{
+						Spec: metallbv1beta1.CommunitySpec{
+							Communities: []metallbv1beta1.CommunityAlias{
+								{
+									Name:  "myCommunityAlias",
+									Value: "65000:100",
+								},
+							},
+						},
+					},
+				},
+				BGPAdvs: []metallbv1beta1.BGPAdvertisement{
+					{
+						Spec: metallbv1beta1.BGPAdvertisementSpec{
+							Communities: []string{
+								"11111:11aaaa",
+								"larg:12345:12345:12345",
+								"alias",
+								"myCommunityAlias",
+							},
+						},
+					},
+				},
+				LegacyAddressPools: []metallbv1beta1.AddressPool{
+					{
+						Spec: metallbv1beta1.AddressPoolSpec{
+							BGPAdvertisements: []metallbv1beta1.LegacyBgpAdvertisement{
+								{
+									Communities: []string{
+										"11111:11aaaa",
+										"larg:12345:12345:12345",
+										"alias",
+										"myCommunityAlias",
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			expected: ClusterResources{
+				Peers: []metallbv1beta2.BGPPeer{
+					{
+						Spec: metallbv1beta2.BGPPeerSpec{
+							BFDProfile:     "",
+							MyASN:          64512,
+							ASN:            64512,
+							Address:        "172.30.0.3",
+							PasswordSecret: v1.SecretReference{},
+						},
+					},
+				},
+				Communities: []metallbv1beta1.Community{
+					{
+						Spec: metallbv1beta1.CommunitySpec{
+							Communities: []metallbv1beta1.CommunityAlias{
+								{
+									Name:  "myCommunityAlias",
+									Value: "65000:100",
+								},
+							},
+						},
+					},
+				},
+				BGPAdvs: []metallbv1beta1.BGPAdvertisement{
+					{
+						Spec: metallbv1beta1.BGPAdvertisementSpec{
+							Communities: []string{
+								"11111:11aaaa",
+								"larg:12345:12345:12345",
+							},
+						},
+					},
+				},
+				LegacyAddressPools: []metallbv1beta1.AddressPool{
+					{
+						Spec: metallbv1beta1.AddressPoolSpec{
+							BGPAdvertisements: []metallbv1beta1.LegacyBgpAdvertisement{
+								{
+									Communities: []string{
+										"11111:11aaaa",
+										"larg:12345:12345:12345",
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+	for _, test := range tests {
+		result := resetTransientErrorsFields(test.clusterResources)
+		if !cmp.Equal(result, test.expected) {
+			t.Fatalf("test %s failed, unexpected clusterResources (-want +got)\n%s", test.desc, cmp.Diff(result, test.expected))
+		}
 	}
 }
--- a/website/content/release-notes/_index.md
+++ b/website/content/release-notes/_index.md
@@ -17,6 +17,7 @@ Chores:
 - Images updated to Go 1.20.12 ([PR 2213](https://github.com/metallb/metallb/pull/2213))
 - CI/E2E: Relabel the frr metrics from frr-k8s to show as MetalLB's ([PR 2210](https://github.com/metallb/metallb/pull/2210))
 - Dev-env: change the default BGP mode to FRR ([PR 2196](https://github.com/metallb/metallb/pull/2196))
+- Webhooks: avoid transient errors ([PR 2202](https://github.com/metallb/metallb/pull/2202)), [ISSUE 2173](https://github.com/metallb/metallb/issues/2173))
 
 ## Version 0.13.12
 


internal/config/validator.go
```diff
@@ -44,14 +44,35 @@ func (v *validator) Validate(resources ...client.ObjectList) error {
...
+			for i := range (clusterResources.LegacyAddressPools) {
+				for j := range(clusterResources.LegacyAddressPools[i].Spec.BGPAdvertisements) {
+					for k := range(clusterResources.LegacyAddressPools[i].Spec.BGPAdvertisements[j].Communities){
+						if !existString(clusterResources.LegacyAddressPools[i].Spec.BGPAdvertisements[j].Communities[k], clusterResources.Communities) {
+							clusterResources.LegacyAddressPools[i].Spec.BGPAdvertisements[j].Communities = removeElement(clusterResources.LegacyAddressPools[i].Spec.BGPAdvertisements[j].Communities,
+								 clusterResources.LegacyAddressPools[i].Spec.BGPAdvertisements[j].Communities[k])
+						}
+					}
+				}
+			}
```
liornoy, line 75: Using indexes to reference slice items make it very hard to read.
You can the second returned value (the object itself) from range as following:
```
for _, legacyAddrPool := range clusterResources.LegacyAddressPools {
    for _, bgpAdv := range legacyAddrPool {
    ...
    
```

internal/config/validator.go
```diff
@@ -68,3 +89,32 @@ func (v *validator) Validate(resources ...client.ObjectList) error {
 func NewValidator(validate Validate) apivalidate.ClusterObjects {
 	return &validator{validate: validate}
 }
+
+func existString(c string, community []metallbv1beta1.Community ) bool {
```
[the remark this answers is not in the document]
liornoy: can we please also rename the function because "existString" is too generic, and that's specifically to check if a community exists

liornoy (commented):
Thanks for the PR!
Left some comments, and please write a commit message :)

internal/config/validator.go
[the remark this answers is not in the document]
shimritproj: I changed the name of the function and the arguments to be like this :  `func ExistCommunity(community string, communities []metallbv1beta1.Community )`

internal/config/validator.go
shimritproj: I agree and changed it.

internal/config/validator.go
```diff
@@ -44,14 +44,35 @@ func (v *validator) Validate(resources ...client.ObjectList) error {
 			clusterResources.Pools = append(clusterResources.Pools, list.Items...)
```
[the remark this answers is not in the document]
shimritproj: I disagree because I don't understand why it is better than the existing one.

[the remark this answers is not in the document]
liornoy: I think it would be better to move the override logic to a function because this will make the switch-case block less cluttered so it is easier to read and understand.
We do extra logic for peers, bgpadv, and addresspool, and it might confuse the reader.
With a function the code will look cleaner, and we can add a small comment explaining why we do this non-trivial override.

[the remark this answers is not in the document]
shimritproj: I agree that an additional function should be created whose role will only be to reset the fields that cause TransientError.

internal/config/validator.go
```diff
@@ -58,13 +58,69 @@ func (v *validator) Validate(resources ...client.ObjectList) error {
 			clusterResources.Nodes = append(clusterResources.Nodes, list.Items...)
 		}
 	}
+	clusterResources = Resetfields(clusterResources)
```
liornoy, line 61: Let's keep this function private to this package

liornoy:
Sorry for not saying this before, but now that I think of this, we also don't need to export the other 2 functions - `RemoveElement` and `ExistCommunity`. 
And can you please run "go fmt" on it? this has some formatting issues.

shimritproj:
> Sorry for not saying this before, but now that I think of this, we also don't need to export the other 2 functions - `RemoveElement` and `ExistCommunity`. And can you please run "go fmt" on it? this has some formatting issues.

Yes, sure! you can see now that the file is updated.

internal/config/validator.go
```diff
@@ -58,13 +58,69 @@ func (v *validator) Validate(resources ...client.ObjectList) error {
...
 	}
+	clusterResources = resetfields(clusterResources)
 	_, err := For(clusterResources, v.validate)
 	if errors.As(err, &TransientError{}) { // we do not want to make assumption on ordering in webhooks.
 		return nil
 	}
 	return err
 }
 
+func resetfields(clusterResources ClusterResources) ClusterResources {
```
fedepaol, line 69: resetfields is a bit too generic. Please add a proper name and a comment that says what are the fields that we are resetting.

internal/config/validator.go
```diff
@@ -58,13 +58,69 @@ func (v *validator) Validate(resources ...client.ObjectList) error {
...
+	for i := range clusterResources.Peers {
+		clusterResources.Peers[i].Spec.BFDProfile = ""
+		clusterResources.Peers[i].Spec.PasswordSecret.Name = ""
+	}
+	for _, bgpAdv := range clusterResources.BGPAdvs {
+		for _, community := range bgpAdv.Spec.Communities {
+			if !ExistCommunity(community, clusterResources.Communities) {
+				bgpAdv.Spec.Communities = RemoveElement(bgpAdv.Spec.Communities, community)
```
fedepaol, line 77: I am not sure about what happens if we remove an item / reassing while we iterate.
A better approach is to build the new communities slice while iterate, and replace bgpAdv.Spec.Communities with the new slice after the loop

fedepaol:
Do we need transient error after these changes?

fedepaol:
Also, please change the commit message explaining the why we are doing this change.

internal/config/validator.go
shimritproj: I agree , I changed the name to be `resetTransientErrorsFields`.

shimritproj:
> Do we need transient error after these changes?
> 
> And, can you add unit tests?

On the Webhooks side, resetting the fields that cause Transienterrors, and there will be no Transienterrors after the changes. Transienterrors prevent us from seeing real errors. However, on the Speaker side, the Transienterrors will remain as usual. This is because, in the Speaker, the Transienterrors help us document the code. 
As for unit tests - yes, I will do them.
@fedepaol

fedepaol:
> > Do we need transient error after these changes?
> > And, can you add unit tests?
> 
> On the Webhooks side, resetting the fields that cause Transienterrors, and there will be no Transienterrors after the changes. Transienterrors prevent us from seeing real errors. However, on the Speaker side, the Transienterrors will remain as usual. This is because, in the Speaker, the Transienterrors help us document the code.> Do we need transient error after these changes?
> And, can you add unit tests?

> However, on the Speaker side, the Transienterrors will remain as usual. This is because, in the Speaker, the 
> Transienterrors help us document the code. As for unit tests - yes, I will do them. @fedepaol

Not sure I understood this. Can you expand a bit?

> As for unit tests - yes, I will do them. @fedepaol

liornoy:
> > > Do we need transient error after these changes?
> > > And, can you add unit tests?
> > 
> > 
> > On the Webhooks side, resetting the fields that cause Transienterrors, and there will be no Transienterrors after the changes. Transienterrors prevent us from seeing real errors. However, on the Speaker side, the Transienterrors will remain as usual. This is because, in the Speaker, the Transienterrors help us document the code.> Do we need transient error after these changes?
> > And, can you add unit tests?
> 
> > However, on the Speaker side, the Transienterrors will remain as usual. This is because, in the Speaker, the
> > Transienterrors help us document the code. As for unit tests - yes, I will do them. @fedepaol
> 
> Not sure I understood this. Can you expand a bit?
> 
> > As for unit tests - yes, I will do them. @fedepaol

Regarding transient errors, it might make sense to maintain the transient error as a custom error type. We can then return this error type in certain functions in config.go. This approach will make it clearer why we override some fields in the webhook.

internal/config/validator.go
```diff
@@ -58,13 +56,51 @@ func (v *validator) Validate(resources ...client.ObjectList) error {
...
+		clusterResources.Peers[i].Spec.PasswordSecret.Name = ""
+	}
+	for i, bgpAdv := range clusterResources.BGPAdvs {
+		var communities []string
+		for _, community := range bgpAdv.Spec.Communities {
+			if existCommunity(community, clusterResources.Communities) {
+				communities = append(communities, community)
+			}
+		}
+		clusterResources.BGPAdvs[i].Spec.Communities = communities
```
liornoy, line 77: now that you're using the second value of the range you don't need to use indexes,
same for all loops here

internal/config/validator.go
```diff
@@ -58,13 +56,51 @@ func (v *validator) Validate(resources ...client.ObjectList) error {
...
+// resetting the fields that lead to Transienterror
+func resetTransientErrorsFields(clusterResources ClusterResources) ClusterResources {
+	for i := range clusterResources.Peers {
+		clusterResources.Peers[i].Spec.BFDProfile = ""
+		clusterResources.Peers[i].Spec.PasswordSecret.Name = ""
+	}
+	for i, bgpAdv := range clusterResources.BGPAdvs {
+		var communities []string
+		for _, community := range bgpAdv.Spec.Communities {
+			if existCommunity(community, clusterResources.Communities) {
```
liornoy, line 73: wouldn't this make us miss real failures like a community with an invalid value?
like "234.4"

internal/config/validator.go
liornoy: we still use indexes in existCommunity and resetTransientErrorsFields

liornoy:
Can you please place "resetTransientErrorsFields" below "NewValidator" ?
it's a common practice to place the exported functions first and the private one after.

shimritproj:
> > > > Do we need transient error after these changes?
> > > > And, can you add unit tests?
> > > 
> > > 
> > > On the Webhooks side, resetting the fields that cause Transienterrors, and there will be no Transienterrors after the changes. Transienterrors prevent us from seeing real errors. However, on the Speaker side, the Transienterrors will remain as usual. This is because, in the Speaker, the Transienterrors help us document the code.> Do we need transient error after these changes?
> > > And, can you add unit tests?
> > 
> > 
> > > However, on the Speaker side, the Transienterrors will remain as usual. This is because, in the Speaker, the
> > > Transienterrors help us document the code. As for unit tests - yes, I will do them. @fedepaol
> > 
> > 
> > Not sure I understood this. Can you expand a bit?
> > > As for unit tests - yes, I will do them. @fedepaol
> 
> Regarding transient errors, it might make sense to maintain the transient error as a custom error type. We can then return this error type in certain functions in config.go. This approach will make it clearer why we override some fields in the webhook.

A transient error was preventing me from identifying real errors, so we decided to remove them. If it is crucial for us to retain the information contained in errors, we can proceed with Lior's suggestion. 

@liornoy  @fedepaol

internal/config/validator.go
liornoy: sorry, my mistake, using this value wont change the real slice, you are correct

internal/config/validator.go
shimritproj: In my opinion, no, because the verification of whether the values are correct or not should occur in the preceding functions. In the functions here, I only delete or change to an empty string. 
@liornoy

shimritproj: After discussing with Lior, I agree.
Thanks @liornoy  for your attention here.

internal/config/validator.go
```diff
@@ -58,13 +58,65 @@ func (v *validator) Validate(resources ...client.ObjectList) error {
...
+		bgpAdv.Spec.Communities = communities
+	}
+	for i, legacyAddrPool := range clusterResources.LegacyAddressPools {
+		for j, bgpAdv := range legacyAddrPool.Spec.BGPAdvertisements {
+			var communities []string
+			for k := range bgpAdv.Communities {
+				if existCommunity(bgpAdv.Communities[k], clusterResources.Communities) {
+					communities = append(communities, bgpAdv.Communities[k])
+				} else if !checkInvalidCommunityFormat(bgpAdv.Communities[k]) {
+					communities = append(communities, bgpAdv.Communities[k])
```
liornoy, line 95: Thanks, can you please include this corner case in the new unit test you wrote?

internal/config/validator.go
```diff
@@ -58,13 +58,65 @@ func (v *validator) Validate(resources ...client.ObjectList) error {
...
+		clusterResources.Peers[i].Spec.PasswordSecret.Name = ""
+	}
+	for i, bgpAdv := range clusterResources.BGPAdvs {
+		var communities []string
+		for _, community := range bgpAdv.Spec.Communities {
+			if existCommunity(community, clusterResources.Communities) {
+				communities = append(communities, community)
+			} else if !checkInvalidCommunityFormat(community) {
+				communities = append(communities, community)
+			}
```
oribon, line 83: I think this gets into the validation details too much, rather than focusing on "avoiding transient errors". What I mean is what we actually want is to not pass references to other objects (like with the bfdprofiles) without caring about the actual validation. Since it is a bit harder with this field (having multiple formats), I'd say we can check here if the string is a "valid reference" and drop it if so, and the conditions will be replaced by:
```go
	if errs := validation.IsDNS1123Subdomain(community); len(errs) != 0 {
		communities = append(communities, community)
	}
```
(using `"k8s.io/apimachinery/pkg/util/validation"`)
this still makes the assumption that a community value can not be of this format (which k8s object names follow), but I think it's better than the specific community validations.

internal/config/validator.go
```diff
@@ -58,13 +58,65 @@ func (v *validator) Validate(resources ...client.ObjectList) error {
...
+		}
+		clusterResources.BGPAdvs[i].Spec.Communities = communities
+		bgpAdv.Spec.Communities = communities
```
oribon, line 86: why is this needed?

internal/config/validator_test.go
```diff
@@ -28,3 +29,255 @@ func TestValidator(t *testing.T) {
...
+		result := existCommunity(test.community, test.communities)
+		if result != test.expect {
+			t.Errorf("%s: expect is %v, but result is %v", test.community, test.expect, result)
+		}
+	}
+}
+func TestResetTransientErrorsFields(t *testing.T) {
+	tests := []struct {
+		clusterResources       ClusterResources
+		expectclusterResources ClusterResources
```
oribon, line 89: nit: change this to just "expected"

internal/config/validator_test.go
```diff
@@ -28,3 +29,255 @@ func TestValidator(t *testing.T) {
...
+		}
+		for i := range result.LegacyAddressPools {
+			for j := range result.LegacyAddressPools[i].Spec.BGPAdvertisements {
+				if !compareStringArrays(result.LegacyAddressPools[i].Spec.BGPAdvertisements[j].Communities,
+					test.expectclusterResources.LegacyAddressPools[i].Spec.BGPAdvertisements[j].Communities) {
+					t.Errorf("Expect is %v, but result is %v", test.expectclusterResources.LegacyAddressPools[i].Spec.BGPAdvertisements[j].Communities,
+						result.LegacyAddressPools[i].Spec.BGPAdvertisements[j].Communities)
+				}
+			}
+		}
```
oribon, line 231: this can be replaced by:
```go
			if !cmp.Equal(result, expected) {
				t.Fatalf("test %s failed, unexpected clusterResources (-want +got)\n", test.desc, cmp.Diff(result, expected))
			}
```

internal/config/validator_test.go
```diff
@@ -28,3 +29,255 @@ func TestValidator(t *testing.T) {
...
+	for _, test := range tests {
+		result := existCommunity(test.community, test.communities)
+		if result != test.expect {
+			t.Errorf("%s: expect is %v, but result is %v", test.community, test.expect, result)
+		}
+	}
+}
+func TestResetTransientErrorsFields(t *testing.T) {
+	tests := []struct {
+		clusterResources       ClusterResources
```
oribon, line 88: add a "desc" field

internal/config/validator.go
shimritproj: Because I save the entire communities that don't generate a Transienterror, and when I finish the loop that occurs in lines 78-84, I update the 'communities' field. 
The goal is to update the field and ensure that the entire set of communities doesn't create a Transienterror.

internal/config/validator_test.go
shimritproj: In the struct? Why? Maybe I don't understand what you mean.
Can you give me an example?

internal/config/validator.go
oribon: I'm talking specifically about line 86, why isn't
```go
clusterResources.BGPAdvs[i].Spec.Communities = communities
```
enough?

internal/config/validator_test.go
shimritproj: Thank you!! I searched for a function that does it, but I didn't find one. So, I wrote all of this. I will change it. tnx!

internal/config/validator_test.go
oribon: look for `test.desc` in the code, together with t.Run it gives each testcase a name (description) and it makes it easier to understand which scenario failed

internal/config/validator.go
shimritproj: Oh! you're right, tnx!

website/content/release-notes/_index.md
```diff
@@ -16,6 +16,7 @@ Chores:
 - E2E: Receive prefixes using frr-k8s alongside MetalLB ([PR 2211](https://github.com/metallb/metallb/pull/2211))
 - Images updated to Go 1.20.12 ([PR 2213](https://github.com/metallb/metallb/pull/2213))
 - CI/E2E: Relabel the frr metrics from frr-k8s to show as MetalLB's ([PR 2210](https://github.com/metallb/metallb/pull/2210))
+- Webhook does not process transient errors properly ([PR 2202](https://github.com/metallb/metallb/pull/2202)), [ISSUE 2173](https://github.com/metallb/metallb/issues/2173))
```
oribon, line 19:
```suggestion
- Webhooks: avoid transient errors ([PR 2202](https://github.com/metallb/metallb/pull/2202)), [ISSUE 2173](https://github.com/metallb/metallb/issues/2173))
```

oribon (approved):
lgtm, thanks!

PR #2205, opened 2023-12-07
controller: improve docker build cache

Pitasi:
The goal of this PR is to speed up Docker builds, because I'm on a slow network and every time I ran `inv dev-env` it took a long time.

1. I moved towards the bottom lines, such as

```
ARG GIT_COMMIT=dev
ARG GIT_BRANCH=dev
```

so they don't invalidate the other docker layers when the value changes (i.e., go mod download).

2. The second thing is to reduce the number of layers by replacing instructions such as:

```
COPY go.mod go.sum ./
```

with

```
--mount=type=bind,source=go.sum,target=go.sum \
--mount=type=bind,source=go.mod,target=go.mod \
```

this way the files are mounted from the host instead of copied to a new layer.

3. I added a cache mount for downloaded go modules:

```
RUN --mount=type=cache,target=/go/pkg/mod/ \
    ...
    go mod download -x
```

### DIFF
--- a/configmaptocrs/Dockerfile
+++ b/configmaptocrs/Dockerfile
@@ -1,40 +1,38 @@
 # syntax=docker/dockerfile:1.2
 
 FROM --platform=$BUILDPLATFORM docker.io/golang:1.20.10 AS builder
-ARG GIT_COMMIT=dev
-ARG GIT_BRANCH=dev
 WORKDIR $GOPATH/go.universe.tf/metallb
 
-# Cache the downloads
-COPY go.mod go.sum ./
-RUN go mod download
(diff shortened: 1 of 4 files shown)


oribon:
thanks a lot for submitting! I still haven't gone through the details, but could the speaker benefit from similar changes as well?

Pitasi:
Yep, sure! I updated the other two Dockerfiles: speaker and configmaptocrs.

oribon (approved):
Thanks! LGTM
putting this https://docs.docker.com/build/guide/mounts/ here as reference.
ccing @fedepaol , I'll let him give this a look as well because I'm not too familiar with this 😅

fedepaol:
cool, thanks!

PR #2210, opened 2023-12-11
CI/E2E: relabel the frr metrics from frr-k8s to show as MetalLB's

oribon:
We would like to expose the frr metrics coming from frr-k8s as metallb's because it is the same set of metrics (for frr mode) and would make the integration smoother when the servicemonitors are set up in a similar way.

### DIFF
--- a/config/manifests/metallb-frr-k8s-prometheus.yaml
+++ b/config/manifests/metallb-frr-k8s-prometheus.yaml
@@ -3279,6 +3279,17 @@ spec:
     tlsConfig:
       insecureSkipVerify: true
   - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    metricRelabelings:
+    - regex: frrk8s_bgp_(.*)
+      replacement: metallb_bgp_$1
+      sourceLabels:
+      - __name__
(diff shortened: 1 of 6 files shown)


config/manifests/metallb-frr-k8s-prometheus.yaml
```diff
@@ -3279,6 +3279,17 @@ spec:
     tlsConfig:
       insecureSkipVerify: true
   - bearerTokenFile: /var/run/secrets/kubernetes.io/serviceaccount/token
+    metricRelabelings:
```
fedepaol, line 3282: Can we have these also set by default in the helm chart? Or the idea is to document the change if a user wants to keep backward compatibility?

fedepaol: If so, we should remember of doing so when we document it.

oribon: yep my idea was to document it as it depends on the pod/service monitors setup

fedepaol: then let's file an issue to document, with next release as target

PR #2212, opened 2023-12-13
RelNotes: union instead of merge

fedepaol:
The rel notes are appended, so it makes sense to do union instead of merging, otherwise each PR will cause a conflict.


fedepaol:
Yolo!

PR #2223, opened 2023-12-27
Remove pointer package and use the k8s util 'ptr'

liornoy:
Remove the pointer packages and utilize the 'ptr' utility package of Kubernetes .

The 'ptr' package employs generics, potentially eliminating the need to expand the
existing pointer package in the future.


api/v1alpha1/addresspool_conversion_test.go
```diff
@@ -30,11 +30,11 @@ func TestValidateAddressPoolConvertTo(t *testing.T) {
 			},
 
 			Protocol:   "bgp",
-			AutoAssign: pointer.BoolPtr(false),
+			AutoAssign: ptr.To[bool](false),
```
AlinaSecret, line 33: Stumbled in to this PR, and i wondered why you chose to do:
```
ptr.To[bool](false)
```
instead of:
```
ptr.To(false)
```
I personally think the type annotation is redundant , Would love to hear your opinion :) 
(comment is also relevant for the other changes in the other lines)

liornoy: You mean he can understand it even without explicitly writing Bool?

liornoy: I am not sure if that's true for type like int32 etc

liornoy: for bool you might be right. I didn't think of that

AlinaSecret: for other cases, I personally prefer:
```
ptr.To(int32(42))
```

fedepaol: So, when the input is a variable, I think the right way is to forget about declaring the type.
This covers bools, but also all the ints flavours where the argument is a variable (meaning, everywhere but the tests where we have a fixed const that would default to int ).

For the others, I don't have a strong preference and I wasn't able to find if one way is more idiomatic than the other, but I'd keep not using it (so `ptr.To(int32(42))` ) just for consistency.

fedepaol:
LGTM, thanks!

PR #2233, opened 2024-01-09
BGP: move the session matching logic one level above

fedepaol:
Sending all the advertisements to the session just to filter them after is counterintuitive. One would expect all of them to be processed. It's much better (and readable) to filter them from the calling site, before passing them to the session.

Also, removing all the *selector* tests from frr and frr-k8s since they do not make sense as the logic is happening above.


speaker/bgp_controller.go
```diff
@@ -335,13 +335,27 @@ func (c *bgpController) updateAds() error {
...
-		if err := peer.session.Set(allAds...); err != nil {
+		ads := adsPerPeer(peer.cfg.Name, allAds)
+		if err := peer.session.Set(ads...); err != nil {
 			return err
 		}
 	}
 	return nil
 }
 
+func adsPerPeer(peerName string, ads []*bgp.Advertisement) []*bgp.Advertisement {
```
oribon, line 346: nit: rename to `adsForPeer` (this reads a bit like it'll return a map or something)

speaker/bgp_controller.go
```diff
@@ -335,13 +335,27 @@ func (c *bgpController) updateAds() error {
...
+	res := []*bgp.Advertisement{}
+	for _, a := range ads {
+		if a.MatchesPeer(peerName) {
+			res = append(res, a)
+		}
+	}
+	if len(res) == 0 {
+		return nil
+	}
+	return res
```
oribon, line 356: do we actually care if the slice is nil / empty later? if not I think just returning res is easier to understand

fedepaol: This keeps the behaviour consistent with what we have today. If there are no ads, then we feed with nil. I can change if you think it doesn't matter, I kept on the conservative side (plus, I didn't want to fix a ton of tests :P )

oribon: nvm then, just keep it this way 😄

oribon (approved):
lgtm

PR #2234, opened 2024-01-09
FRR: squash advertisements with the same prefix

fedepaol:
Passing multiple advertisements with the same prefix to the template causes duplicate entries. This happens for example when the shared ip annotation is used with the services.

Since a given advertisement has the related peers as fields, the right place to squash is when we manage the advertisements related to a given neighbor.


internal/bgp/frr/frr.go
oribon, line 1: can we remove the 
```
		/* As 'session.advertised' is a map, we can be sure there are no
		   duplicate prefixes and can, therefore, just add them to the
		   'neighbor.Advertisements' list. */
```
comment now?

fedepaol: done

oribon (approved):
lgtm

Issue #2174, opened 2023-11-21
webhook_service/webhook_cert inconfigurable


rogeryuchao:
### Is your feature request related to a problem?

In controller code, some environment variables is hard coded:
1. controller => main.go => line 148, certServiceName , hardcode value: webhook-service
2. controller => main.go => line 151, webhookSecretName, hardcode value: webhook-server-cert

### Describe the solution you'd like

These 2 items naming is too common, when we want to merge the metallb system's component with other network plugin in k8s, the naming of this webhook cert and the service can be conflict with other project, it should be configurable in the variable name and use deployment manifest to config it, like the 'METALLB_ML_SECRET_NAME' and 'METALLB_DEPLOYMENT'

### I've read and agree with the following

- [X] I've checked all open and closed issues and my request is not there.
- [X] I've checked all open and closed pull requests and my request is not there.

fedepaol:
Makes sense! And thanks for filing this issue.
Would changing the name to something less common suffice? I don't think this would require any level of configurability if we choose something taht is strictly metallb related.

@rogeryuchao keen filing a pr for that? Should be quite straightforward.
If not, we'll work on this.

cyclinder:
A part of this has already finished in https://github.com/metallb/metallb/pull/2070

cyclinder:
I think they're already configurable, but we could consider renaming them to make them less common suffice.

- webhook-service -> metallb-webhook-service
- webhook-server-cert -> metallb-secret

PR #2244, opened 2024-01-15
Make the name of the webhook certs more metallb specific

cyclinder:
Rename webhook_service/webhook_cert name to avoid name conflicts with other resources in the cluster.

Fixes https://github.com/metallb/metallb/issues/2174

### DIFF
--- a/charts/metallb/templates/controller.yaml
+++ b/charts/metallb/templates/controller.yaml
@@ -185,7 +185,7 @@ spec:
       - name: cert
         secret:
           defaultMode: 420
-          secretName: webhook-server-cert
+          secretName: metallb-webhook-cert
       {{- if .Values.prometheus.controllerMetricsTLSSecret }}
       - name: metrics-certs
         secret:
--- a/charts/metallb/templates/webhooks.yaml
+++ b/charts/metallb/templates/webhooks.yaml
@@ -144,7 +144,7 @@ spec:
 apiVersion: v1
 kind: Secret
 metadata:
-  name: webhook-server-cert
+  name: metallb-webhook-cert
   namespace: {{ .Release.Namespace | quote }}
   labels:
     {{- include "metallb.labels" . | nindent 4 }}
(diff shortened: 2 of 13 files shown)


fedepaol:
@cyclinder mind checking the CI failures? I guess something went wrong.
Thanks!

cyclinder:
sure, let me checking these failures

cyclinder:
Hi @fedepaol, If we update the names of webhook_service/webhook_cert directly, this may be an incompatible change, which leads to failures with [metallb-operator](https://github.com/metallb/metallb/actions/runs/7724726313/job/21071331380?pr=2244) (the old manifests use the new docker-image, and the correct name is not specified in the controller via go flags). I think there are two ways:

1. Update metallb-operator to make it compatible with this change
2. Close this PR, webhook_service/webhook_cert are already configured.

I think 2 is enough, but I prefer 1, I like the default value to be metallb related.

DrBu7cher:
Can you fix https://github.com/metallb/metallb/blob/main/charts/metallb/charts/crds/templates/crds.yaml#L260 before everything else?
Because we see a lot of
```
cacher (bgppeers.metallb.io): unexpected ListAndWatch error: failed to list metallb.io/v1beta1, Kind=BGPPeer: conversion webhook for metallb.io/v1beta2, Kind=BGPPeer failed: Post "https://webhook-service.metallb.svc:443/convert?timeout=30s": service "webhook-service" not found; reinitializing...
```
in out api-server.

It was introduced here https://github.com/metallb/metallb/commit/5339a9a9d3742ff168a71358cc36ce4a702ed19d I think. Since the service was renamed from `metallb-webhook-service` to `webhook-service`.

fedepaol:
@cyclinder mind fixing the conflicts?

fedepaol:
Just a nit on the commit / PR title: here we are not making the name configurable, but fixing it to a metallb-only value.

fedepaol:
LGTM

Issue #439, opened 2019-06-09
Assign more than one IP


arman-sydikov:
```
apiVersion: v1
kind: ConfigMap
metadata:
  namespace: metallb-system
  name: config
data:
  config: |
    address-pools:
    - name: default
      protocol: layer2
      addresses:
      - 192.168.1.240-192.168.1.250
```

When we apply Layer 2 configuration it assigns exactly one IP from the address pool to the service.
Is it possible to assign more than one IP to the service ?
If yes, how to do it ?

---

- Version of MetalLB: v0.7.3
- Version of Kubernetes: v1.14.2
- Name and version of network addon (e.g. Calico, Weave...): flannel:v0.11.0-amd64

NicolasT:
+1. My scenario: use MetalLB to assign VIP(s) to a service, then have the MetalLB controller apply some kind of (soft) anti-affinity to distribute those per-service VIPs across multiple speakers, and set up DNS to round-robin between or multi-A answer those VIPs upon resolve requests.

I believe this could be achieved by creating multiple `Service` objects of type `LoadBalancer` with the same selector (which results in multiple VIPs assigned for the same backing service instances), but it wouldn't imply the (soft) anti-affinity between speakers so not necessarily increase availability or load balancing across multiple front-end nodes.

Does this make sense?

firefly-serenity:
Another solution would be, that Metallb implements the possibility of activating "static" IPs. The IPs would be started on one of the interfaces in the cluster even thought Metallb is not aware of a service that consumes it. The service can then activate `externalIPs` and make use of metallbs fail-over capabilities. 

This could really make the life easier if you want to distribute your applications over multiple IPs but still have the ingress controller handle tls termination.

migs35323:
+1

fedepaol:
The solution to this issue is to have multiple services sharing the same set of endpoints.
Doing this is functionally equivalent to having a service with the two ips.
The anti affinity mentioned by @NicolasT would require to have the leader election algorithm to be stateful as opposed to the current implementation, and I am not sure it's a complication we want to introduce.
Also, now we have the node selector feature that would allow to split the set of nodes the two (or more) services would be advertised from, which would help a bit in the direction of having a better balancing.

Closing this issue.

kvaps:
[...]

PR #2246, opened 2024-01-17
speaker: Added preferred L2 speaker node feature

apolovov:
## Description
The new Service annotation `metallb.universe.tf/preferredL2SpeakerNode` to force L2-announcing from specific Node if it is alive.

## Why do we need it, and what problem does it solve?
Related to https://github.com/metallb/metallb/issues/439.
For now, speakers in Layer2 mode are announcing only a single External IP per Service. This IP is being announced by a single Node, which is elected by hashing algorithm. So, unlike BGP mode, I can't load all front nodes. But i need.
@fedepaol [offered a solution](https://github.com/metallb/metallb/issues/439#issuecomment-1271664497) to use multiple Services sharing the same set of endpoints. This would work if I was confident that the announcements would be evenly distributed across all front-end nodes, but the nodes for services are selected independently of each other.
The solution in this PR is to add an annotation to specify preferred Node for announcing, if the node is dead, the election will happen as before. So that i can create multiple Services and pin them to frontend nodes manually.

## P.S.
This PR is proposal draft, if it has a right to live, I'll finish it (tests, docs, etc).

### DIFF
--- a/speaker/layer2_controller.go
+++ b/speaker/layer2_controller.go
@@ -17,6 +17,7 @@ package main
 import (
 	"bytes"
 	"crypto/sha256"
+	"k8s.io/utils/strings/slices"
 	"net"
 	"sort"
 
@@ -28,6 +29,10 @@ import (
 	"go.universe.tf/metallb/internal/layer2"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/sets"
+)
+
+const (
+	annotationPreferredL2SpeakerNode = "metallb.universe.tf/preferredL2SpeakerNode"
 )
 
 type layer2Controller struct {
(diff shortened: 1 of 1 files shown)


fedepaol:
let me think about this a bit. If the feature makes sense, I'd rather have it specified by the L2Advertisement rather than an annotation.

apolovov:
If we could pick the availableNode for current Service based on neighbouring Services (not just the first one), it could be solution.
For example, to index all the Services in order (0, 1, 2, 3...) and pick the active node with `availableNodes[serviceIndex % len(availableNodes)]`. So we'll evenly distribute IPs across the nodes without changing UX.

fedepaol:
> If we could pick the availableNode for current Service based on neighbouring Services (not just the first one), it could be solution. For example, to index all the Services in order (0, 1, 2, 3...) and pick the active node with `availableNodes[serviceIndex % len(availableNodes)]`. So we'll evenly distribute IPs across the nodes without changing UX.

I think it's a bit more complex than this. We should filter out only the services that can be announced on the nodes the service can be announced from (which includes node selectors, endpoints in case of etp local) and given the dynamic nature, this can result in a service bouncing.
Even without this and considering your example, removing a service from the list would result in rearranging all the announcing nodes.

fedepaol:
Hi! I know this may solve your short term problem, but I really want to think about this in a more broader term and see if we can do something better to improve the distribution of the L2 allocations across the cluster.
For this reason I am gonna close this PR, and will probably raise an issue to track these considerations. Thanks for the time spent here!

nabokihms:
@fedepaol, any news about the issue? The solution is required to prevent reallocations of IP addresses, and living with the patch differs from what we would like to stick with.

PR #2263, opened 2024-01-26
dev-env: wait until the lb exclusion is effective

fedepaol:
We saw a flake where the first test fails, probably because the lb exclusion is not effective yet. So here we:

- move the exclusion from the e2e command to the dev-env command, so it will work when testing / tinkering too
- active wait until all the labels are removed


oribon (approved):
lgtm

PR #2269, opened 2024-01-31
fix: health check failing for BGPPeer

budimanjojo:
This fix this error:
```
health check failed after 29m30.045208087s: timeout waiting for: [BGPPeer/networking-system/main status: 'Unknown': conversion webhook for metallb.io/v1beta2, Kind=BGPPeer failed: Post "https://webhook-service.networking-system.svc/convert?timeout=30s": service "webhook-service" not found]
```

   accepted.
   the PR is merged.


budimanjojo:
I don't know what's wrong. I used `invoke generatemanifests` in my machine to fix the error and it just complains about `yq: error: argument files: can't open 'select(.kind == "CustomResourceDefinition")': [Errno 2] No such file or directory: 'select(.kind == "CustomResourceDefinition")` and delete the whole `crds.yaml` file.

This is just a one liner change.

fedepaol:
I think there are two yq versions, the one we are consuming is https://github.com/mikefarah/yq
We should probably change inv to make sure the right yq is downloaded locally and used.
In general, the problem of your PR is that it changes the helm CRD directly, while what we do is to generate them starting fro mwhat we have under /config.
The right fix would be to change the sed used in inv.

budimanjojo:
@fedepaol thanks for the `yq-go` hint. I have edited the file patch in `./config` directory and ran `inv` command to generate the crds instead. Not sure if I should commit the generated one too or just commit the change in `./config` directory so I just commit them all.

fedepaol:
This is not good either, because the service in the kustomize configuration is named webhook-service (which exactly why this bug was introduced).
So, we either align the wh service name OR we extend the sed command when generating the charts. I think the former is cleaner and metallb-* is just a better name

fedepaol:
@budimanjojo I took the liberty to adjust your commit, to speed up the release.
Thanks again!

budimanjojo:
@fedepaol thank you!

fedepaol:
Operator lane not affected by these changes, merging.

PR #2275, opened 2024-02-04
Warn users when skipping BGP announcements

clee:
Instead of forcing users to turn on debug logging to figure out why their BGP announcements aren't going out, make these warnings.

This is pretty trivial, but I won't be upset if nobody likes it.

/kind cleanup

```release-note
The log level of the logs for skipping the BGP announcement because of the 'node.kubernetes.io/exclude-from-external-load-balancer' label and the network unavailable condition are now warning instead of debug.
```

### DIFF
--- a/speaker/bgp_controller.go
+++ b/speaker/bgp_controller.go
@@ -154,12 +154,12 @@ func (c *bgpController) ShouldAnnounce(l log.Logger, name string, _ []net.IP, po
 	}
 
 	if k8snodes.IsNetworkUnavailable(nodes[c.myNode]) {
-		level.Debug(l).Log("event", "skipping should announce bgp", "service", name, "reason", "speaker's node has NodeNetworkUnavailable condition")
+		level.Warn(l).Log("event", "skipping should announce bgp", "service", name, "reason", "speaker's node has NodeNetworkUnavailable condition")
 		return "nodeNetworkUnavailable"
 	}
 
 	if !c.ignoreExcludeLB && k8snodes.IsNodeExcludedFromBalancers(nodes[c.myNode]) {
-		level.Debug(l).Log("event", "skipping should announce bgp", "service", name, "reason", "speaker's node has labeled 'node.kubernetes.io/exclude-from-external-load-balancers'")
+		level.Warn(l).Log("event", "skipping should announce bgp", "service", name, "reason", "speaker's node has labeled 'node.kubernetes.io/exclude-from-external-load-balancers'")
 		return "nodeLabeledExcludeBalancers"
 	}
 


speaker/bgp_controller.go
```diff
@@ -148,17 +148,17 @@ func hasHealthyEndpoint(eps []discovery.EndpointSlice, filterNode func(*string)
 
 func (c *bgpController) ShouldAnnounce(l log.Logger, name string, _ []net.IP, pool *config.Pool, svc *v1.Service, epSlices []discovery.EndpointSlice, nodes map[string]*v1.Node) string {
 	if !poolMatchesNodeBGP(pool, c.myNode) {
-		level.Debug(l).Log("event", "skipping should announce bgp", "service", name, "reason", "pool not matching my node")
+		level.Warn(l).Log("event", "skipping should announce bgp", "service", name, "reason", "pool not matching my node")
```
fedepaol, line 151: My biggest concern on this is, this is gonna be spammy if your configuration is legit AND this service shouln't be announced. The networkunabailable one makes sense, because it's an exceptional condition. Also, configuration issues might become easier to spot once https://github.com/metallb/metallb/blob/main/design/crd-status.md lands

clee: My cluster doesn't issue any of these "pool not matching my node" messages when in debug mode so I assumed (probably incorrectly) that it would only show if there was a misconfiguration.

I'd be fine with leaving this one as a Debug message then. If the node label message had been a Warn, it would have saved me hours of debugging my cluster last night :)

speaker/bgp_controller.go
```diff
@@ -153,12 +153,12 @@ func (c *bgpController) ShouldAnnounce(l log.Logger, name string, _ []net.IP, po
...
 
 	if k8snodes.IsNetworkUnavailable(nodes[c.myNode]) {
-		level.Debug(l).Log("event", "skipping should announce bgp", "service", name, "reason", "speaker's node has NodeNetworkUnavailable condition")
+		level.Warn(l).Log("event", "skipping should announce bgp", "service", name, "reason", "speaker's node has NodeNetworkUnavailable condition")
 		return "nodeNetworkUnavailable"
 	}
 
 	if k8snodes.IsNodeExcludedFromBalancers(nodes[c.myNode]) {
-		level.Debug(l).Log("event", "skipping should announce bgp", "service", name, "reason", "speaker's node has labeled 'node.kubernetes.io/exclude-from-external-load-balancers'")
+		level.Warn(l).Log("event", "skipping should announce bgp", "service", name, "reason", "speaker's node has labeled 'node.kubernetes.io/exclude-from-external-load-balancers'")
```
fedepaol, line 161: I don't think we want this to be warn either, as in regular scenarios, where the label is present, this might spam the logs.
Thoughts @oribon ?

clee: This is the one in particular that, if it had been a warning instead of a debug message, would have saved me *hours* of troubleshooting. Nowhere in the MetalLB docs is this label mentioned even once, nor the effect it has.

If users don't like seeing this warning message in their logs, they can manually use other labelSelectors to limit the node selection and stop the messages. But for folks like me who had no idea the `exclude-from-external-load-balancers` label even *existed*, I strongly feel like this should not be hidden behind a debug log.

clee: Or, when #2280 lands, I'll just enable that option and be happy. :)

fedepaol: https://github.com/metallb/metallb/pull/2280 is gonna land soon and we'll release it.
I added this as part of the troubleshooting guide, and pinned an issue.
Hope it's enough, but if not a user should enable debug logs

oribon: Not sure I understand the "spam the logs" concern, as if the user has a speaker on that node they probably expect it to be able to announce lbs and would want to unlabel the node asap, but I'm fine with either approach (hoping the docs are enough 😅)

clee:
@fedepaol would you mind kicking this back into the queue again now that I've rebased and it should apply cleanly?

fedepaol:
> @fedepaol would you mind kicking this back into the queue again now that I've rebased and it should apply cleanly?

done, sorry I missed this

Issue #2274, opened 2024-02-04
L2 ARP announcing stopped working after upgrading from 0.13.12 to 0.14.2


RytisLT:
### MetalLB Version

0.4.12

### Deployment method

Charts

### Main CNI

calico

### Kubernetes Version

1.27.3

### Cluster Distribution

kubeadm

### Describe the bug

With `0.13.12` one speaker node logs the following:

    {"caller":"main.go:374","event":"serviceAnnounced","ips":["10.20.10.190"],"level":"info","msg":"service has IP, announcing","pool":"default","protocol":"layer2","ts":"2024-02-04T11:13:50Z"}

with `0.14.2` speaker pods do not contain such log message and ARP responses are not being sent. `arping` stops working

Here's my config:

    apiVersion: metallb.io/v1beta1
    kind: IPAddressPool
    metadata:
      creationTimestamp: null
      name: default
      namespace: metallb-system
    spec:
      addresses:
      - 10.20.10.190-10.20.10.190
    status: {}
    ---
    apiVersion: metallb.io/v1beta1
    kind: L2Advertisement
    metadata:
      creationTimestamp: null
      name: l2advertisement1
      namespace: metallb-system
    spec:
      ipAddressPools:
      - default
    status: {}

### To Reproduce

update from 0.13.12 to 0.14.2 use provided config

### Expected Behavior

speaker nodes continue to respond to ARP requests

### Additional Context

I have 2 clusters, production and development, both exhibit same behavior after upgrade.

I've seen only one error in logs, but it seems to be related to legacy AddressPools:

[...]

PR #2280, opened 2024-02-12
Provide a ignore-exclude-lb flag

fedepaol:
MetalLB now honors the node.kubernetes.io/exclude-from-external-load-balancers
but this might harm some deployments where there is only one node and
kubeadm still labels the control plane with such label.

Here we provide a knob to disable the behaviour and make metallb work in
such scenarios.

### DIFF
--- a/charts/metallb/values.yaml
+++ b/charts/metallb/values.yaml
@@ -268,6 +268,9 @@ speaker:
     mlSecretKeyPath: "/etc/ml_secret_key"
   excludeInterfaces:
     enabled: true
+  # ignore the exclude-from-external-loadbalancer label
+  ignoreExcludeLB: false
+
   image:
     repository: quay.io/metallb/speaker
     tag:
--- a/speaker/main.go
+++ b/speaker/main.go
@@ -82,6 +82,7 @@ func main() {
 		logLevel          = flag.String("log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", logging.Levels.String()))
 		enablePprof       = flag.Bool("enable-pprof", false, "Enable pprof profiling")
 		loadBalancerClass = flag.String("lb-class", "", "load balancer class. When enabled, metallb will handle only services whose spec.loadBalancerClass matches the given lb class")
+		ignoreLBExclude   = flag.Bool("ignore-exclude-lb", false, "ignore the exclude-from-external-load-balancers label")
 	)
 	flag.Parse()
 
@@ -157,6 +158,7 @@ func main() {
 		SList:                  sList,
 		bgpType:                bgpImplementation(bgpType),
 		InterfaceExcludeRegexp: interfacesToExclude,
+		IgnoreExcludeLB:        *ignoreLBExclude,
 	})
 	if err != nil {
 		level.Error(logger).Log("op", "startup", "error", err, "msg", "failed to create MetalLB controller")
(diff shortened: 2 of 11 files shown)


fedepaol:
Fixes https://github.com/metallb/metallb/issues/2274

charts/metallb/values.yaml
```diff
@@ -6,6 +6,8 @@ imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
 loadBalancerClass: ""
+# ignore the exclude-from-external-loadbalancer annotation
+ignoreExcludeLB: false
```
cyclinder, line 10: Would it be better to put it on speaker?

speaker/main.go
```diff
@@ -82,6 +82,7 @@ func main() {
 		logLevel          = flag.String("log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", logging.Levels.String()))
 		enablePprof       = flag.Bool("enable-pprof", false, "Enable pprof profiling")
 		loadBalancerClass = flag.String("lb-class", "", "load balancer class. When enabled, metallb will handle only services whose spec.loadBalancerClass matches the given lb class")
+		ignoreLBExclude   = flag.Bool("ignore-exclude-lb", false, "ignore the exclude-from-external-load-balancers annotation")
```
oribon, line 85: nit: annotation -> label

website/content/release-notes/_index.md
```diff
@@ -5,6 +5,10 @@ weight: 8
 
 ## Version 0.14.4
 
+New Features:
+
+- Provide a "ignore-exclude-lb" flag to ignore the `exclude-from-external-load-balancers` annotation ([PR 2280](https://github.com/metallb/metallb/pull/2280), [Issue 2274](https://github.com/metallb/metallb/issues/2274))
```
oribon, line 10: nit: annotation -> label

website/content/troubleshooting/_index.md
```diff
@@ -87,6 +87,12 @@ A given speaker won't advertise the service if:
 - there are no L2Advertisements / BGPAdvertisements matching the speaker node (if node selectors are specified)
 - the Kubernetes API reports "network not available" on the speaker's node
 
+## MetalLB is not advertising my service from my control-plane nodes or from my single node cluster
+
+Make sure your nodes are not labeled with the [node.kubernetes.io/exclude-from-external-load-balancers](https://kubernetes.io/docs/reference/labels-annotations-taints/#node-kubernetes-io-exclude-from-external-load-balancers) annotation. MetalLB honors that annotation and won't announce any service from
```
oribon, line 92: nit: annotation -> label

charts/metallb/values.yaml
fedepaol: My reasoning was, this is a property of metallb as a whole. It has direct effects only on the speaker, but it felt more like a property similar to "loadBalancerClass". Does it make sense? If not I am open to move it

speaker/main.go
fedepaol: done

website/content/release-notes/_index.md
fedepaol: done

website/content/troubleshooting/_index.md
fedepaol: done

charts/metallb/values.yaml
cyclinder: I think we already have a similar configuration, see https://github.com/metallb/metallb/blob/d9ab5ec6f5a48988ef67c314e590f01e61c2c1c5/charts/metallb/values.yaml#L269

It's more of a speaker-related property, but I'm ok with both, it's up to you :)

nit: the comment should be updated: `ignore the exclude-from-external-loadbalancer label`

oribon (approved):
lgtm (other than a dangling "annotation" instead of "label"

charts/metallb/values.yaml
fedepaol: Fair enough, I'll change it.
Also, thanks for the comment, I missed it!

fedepaol: done

cyclinder (approved):

onigoetz:
Hi, I just stumbled onto this issue and the fix that goes with it.
I can't use this for now; is a release of the new version of the helm chart planned soon ?

Thanks in advance

Issue #2161, opened 2023-11-13
Log Level Not Setting


boomam:
### MetalLB Version

0.13.12

### Deployment method

Charts

### Main CNI

Flannel

### Kubernetes Version

v1.28.3+k3s2

### Cluster Distribution

k3s

### Describe the bug

Logging level of MetalLb componnets is not setting to 'warning' as defined in the values.yaml used during deployment.  
When i observe the various pods, i can see the logging level is still set as 'info'.  
For example, on the controller pod -  
```
W1113 12:40:47.394740       1 warnings.go:70] metallb.io v1beta1 AddressPool is deprecated, consider using IPAddressPool
W1113 12:47:29.392394       1 warnings.go:70] metallb.io v1beta1 AddressPool is deprecated, consider using IPAddressPool
{"level":"info","ts":"2023-11-13T12:53:49Z","logger":"cert-rotation","msg":"no cert refresh needed"}
{"level":"info","ts":"2023-11-13T12:53:49Z","logger":"cert-rotation","msg":"Ensuring CA cert","name":"metallb-webhook-configuration","gvk":"admissionregistration.k8s.io/v1, Kind=ValidatingWebhookConfiguration","name":"metallb-webhook-configuration","gvk":"admissionregistration.k8s.io/v1, Kind=ValidatingWebhookConfiguration"}
{"level":"info","ts":"2023-11-13T12:53:50Z","logger":"cert-rotation","msg":"Ensuring CA cert","name":"addresspools.metallb.io","gvk":"apiextensions.k8s.io/v1, Kind=CustomResourceDefinition","name":"addresspools.metallb.io","gvk":"apiextensions.k8s.io/v1, Kind=CustomResourceDefinition"}
{"level":"info","ts":"2023-11-13T12:53:50Z","logger":"cert-rotation","msg":"Ensuring CA cert","name":"bgppeers.metallb.io","gvk":"apiextensions.k8s.io/v1, Kind=CustomResourceDefinition","name":"bgppeers.metallb.io","gvk":"apiextensions.k8s.io/v1, Kind=CustomResourceDefinition"}
W1113 12:56:59.394627       1 warnings.go:70] metallb.io v1beta1 AddressPool is deprecated, consider using IPAddressPool
W1113 13:02:57.399426       1 warnings.go:70] metallb.io v1beta1 AddressPool is deprecated, consider using IPAddressPool
[...]

PR #2281, opened 2024-02-12
Logging: propagate the log level to the controller-runtime logs

fedepaol:
Until now, controller-runtime logs were running with the default. Here we propagate the loglevel parameter to those logs too.

Note that this might cause less logs than we are used to, when triaging (but it should honor the desire of the user).

Fixes https://github.com/metallb/metallb/issues/2161


oribon (approved):
lgtm

PR #2283, opened 2024-02-13
Dev-env -p bgp: pass a temp dir to frr

fedepaol:
We create a temporary folder each time to avoid asking root permissions to the user.


oribon (approved):
lgtm

PR #2286, opened 2024-02-14
Charts: fix the metrics-certs volume of the speaker

fedepaol:
The creation of the metrics-certs volume is conditioned by FRR mode, while it should only be conditioned by the presence of the name of the secret in the values.


liornoy:
lgtm, only the pr link is missing in the release notes

fedepaol:
> lgtm, only the pr link is missing in the release notes

changed, thanks!

oribon (approved):

PR #2292, opened 2024-02-16
Communities / LocalPref routemaps: create one route map per neigh

fedepaol:
The current implementation creates one route map per localpref / community, duplicating the same route-map for each advertisement (while what we really need is one route map, many prefix-lists entries, one per service).

Here we collect all the communities / localpref / large communities per neighbor, and we use it to generate a route-map per neighbor per community / localpref only once.


oribon (approved):
lgtm

PR #2297, opened 2024-02-21
Release notes: use the k8s release tool

fedepaol:
Trying to migrate the current way to enforce the release notes:

- remove the mandatory need to change the rel notes
- provide a pr template aligned to k8s, with a field for rel notes
- provide a gh action to assign the right label to the pr, that can be used by the tool

When this is established, it will be possible to use the tool to generate the release notes automatically.
This goes under the assumption that no cherry-pick is done without a corresponding PR.

/kind cleanup


oribon (approved):
lgtm

PR #2298, opened 2024-02-25
Charts: don't deploy metallb-excludel2 cm if the speaker is disabled

mustdiechik:

/kind cleanup

- no need to install "metallb-excludel2" cm, when speaker component is disabled
- added common labels for "metallb-excludel2" cm the same way as for another manifests

NONE

**Release note**:

```release-note
The metallb-excludel2 configmap is not deployed from the Helm chart if the speaker is disabled.
```

### DIFF
--- a/charts/metallb/templates/exclude-l2-config.yaml
+++ b/charts/metallb/templates/exclude-l2-config.yaml
@@ -1,9 +1,11 @@
-{{- if .Values.speaker.excludeInterfaces.enabled }}
+{{- if and .Values.speaker.enabled .Values.speaker.excludeInterfaces.enabled }}
 apiVersion: v1
 kind: ConfigMap
 metadata:
   name: metallb-excludel2
   namespace: {{ .Release.Namespace | quote }}
+  labels:
+    {{- include "metallb.labels" . | nindent 4 }}
 data:
   excludel2.yaml: |
     announcedInterfacesToExclude:


fedepaol:
Thanks for the contribution! I see you deleted the PR template, which we are planning to use for crafting the release notes.
Would you mind restoring it and filling the right bits?
Also, would you mind signing the commit?
Also also, can you change the commit subject to what we are doing this commit (something like - don't deploy the exclude l2 config if the speaker is disabled) ?

Thanks again

charts/metallb/templates/exclude-l2-config.yaml
```diff
@@ -1,9 +1,12 @@
+{{- if .Values.speaker.enabled }}
```
cyclinder, line 1: we can use the `and` operator here.
```suggestion
{{- if .Values.speaker.enabled and .Values.speaker.excludeInterfaces.enabled }}
```

mustdiechik: fixed

charts/metallb/templates/exclude-l2-config.yaml
```diff
@@ -20,4 +22,4 @@ data:
     - ^kube-ipvs.*
     - ^cni.*
     - ^nodelocaldns.*
-{{- end }}
\ No newline at end of file
+{{- end }}
```
cyclinder, line 25: We should revert this change, we need to end with an extra blank line.

cyclinder:
Thanks for the contribution. And also, Please follow the [comment](https://github.com/metallb/metallb/pull/2298#issuecomment-1963563530) from @fedepaol, Could you mind updating the PR title? so we can easily see what are the PR doing.

fedepaol:
Thanks! Sending to merge

fedepaol:
Also, thanks @cyclinder for the review!

PR #2306, opened 2024-02-29
Add option for disabling MP BGP 

### LINKED ISSUE #2106: Make MP BGP configurable
I recently encountered a similar issue.  I hope for more flexibility in configuring the address family for neighbors.

May I inquire if there have been any recent developments regarding issue #1757?

> So far, we are using 0.12 version, I tested 0.13.7, however, it does not work as our BGP routers are split so that IPv4 handles IPv4 only connections and the same for IPv6, and MP sessions are rejected.
It would be good to make MP BGP configurable.
_Originally posted by @xhejtman in https://github.com/metallb/metallb/issues/1736#issuecomment-1357304394_

AlinaSecret:
Fixed #2106

/kind feature

```release-note
Add a field to the BGPPeer CRD to disable MP BGP for the given peer
```


e2etest/bgptests/bgp.go
```diff
@@ -1379,6 +1379,43 @@ var _ = ginkgo.Describe("BGP", func() {
 			}
 
 		})
+		ginkgo.It("FRR-MODE BGP Peer disable MP BGP", func() {
```
AlinaSecret, line 1382: I know that this test looks alike to the BGP Peer connect time test, i think in general its better to do a similar approach to the one in [frrk8s](https://github.com/metallb/frr-k8s/blob/89811e5b8238b2e0c83ee8d9382570efabc4c795/e2etests/tests/session.go#L65) where all session parameters are checked in the same test.
But because this test should not run for now in frrk8s mode i am leaving it in separate function to make the skipping easy.
i think we can combine them in the PR that adds the disable mp bgp in frrk8s mode (of course after we add the field to frrk8s) .
in general don't like code duplication but i think this option is temporary and after the tests will be combined will be a good option.

of course there are other options (creating function used by both tests or combining now and skip both of the tests in frrrk8s mode or different skipping method?) 
what do you think?

internal/bgp/frr/templates/neighboripfamily.tmpl
```diff
@@ -1,13 +1,17 @@
 {{- define "neighborenableipfamily"}}
 {{/* no bgp default ipv4-unicast prevents peering if no address families are defined. We declare an ipv4 one for the peer to make the pairing happen */}}
+{{- if or (not .DisableMP) (and .DisableMP (eq .IPFamily "ipv4"))}}
```
fedepaol, line 3: This is adds friction to readability. Can we hide it behind a function that we can implement in https://github.com/metallb/metallb/blob/cb9c6eb233b582aec7ae6fe93b73889ececf7cd3/internal/bgp/frr/config.go#L125
?

Something like `activateNeighborIPFamily`

e2etest/bgptests/bgp.go
fedepaol: It's totally fair to leave it as it is. We don't want to leave intermediate states where we are not testing the feature, and at the same time we can rework it later. Just file an issue so we won't forgot

internal/bgp/frr/templates/neighboripfamily.tmpl
AlinaSecret: added the function under the name: 
"activateNeighborFor": func(ipFamily string, neighbourFamily ipfamily.Family, disableMP bool) bool
making the ussage:
activateNeighborFor "ipv4" .IPFamily .DisableMP

in general i like names in this style - that are based on the whole signature of the function,
i think the usage is also elegant  (activateNeighborFor "ipv4" ...) 

i like less activateNeighborIPFamily because that the signature and usage have twice the ipfamily in them and i think its less fluent.

not sure what is your opinion on names if this kind, but decided on trying this option and have no problem also going with activateNeighborIPFamily

fedepaol: My idea was to pass the neighbor as a whole, so the calling site would look like `activateNeighborIPFamily . "ipv4"` but this works too, I don't have a strong opinion

PR #2308, opened 2024-03-04
PR classify: fail if there is no text for rel notes

fedepaol:
We run a simple check against the release-note entry of the PR body, to notify the users that that section is required for release notes.

>
> /kind bug
/kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

This can't be tested as part of the PR itself, however I tested it here https://github.com/fedepaol/relnotestexperients/pull/19

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/classify.yaml
+++ b/.github/workflows/classify.yaml
@@ -31,8 +31,6 @@ jobs:
           text: ${{ github.event.pull_request.body }}
           regex: '(?<!> )\/kind (\w+)'
 
-
-
       - name: Check
         run: |
           if [[ ! "${{ steps.regex-match.outputs.group1 }}" =~ ^(api-change|bug|cleanup|deprecation|design|documentation|failing|feature|flake|regression)$ ]]; then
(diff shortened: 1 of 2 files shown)


.github/workflows/classify.yaml
```diff
@@ -53,3 +51,16 @@ jobs:
       - uses: actions-ecosystem/action-add-labels@v1
         with:
           labels: kind/${{ steps.regex-match.outputs.group1 }}
+
+      - name: Check release notes were not deleted
```
oribon, line 55: nit: isn't this more checking that they are present?

.github/workflows/classify.yaml
```diff
@@ -53,3 +51,16 @@ jobs:
...
+        uses: actions/github-script@v5
+        with:
+          script: |
+            const pr = await github.rest.pulls.get({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              pull_number: context.issue.number
+            });
+            if (!pr.data.body.includes("```release-note")) {
+              throw new Error("Please don't cancel the ```release-note ``` section as it is used to build the release notes");
```
oribon, line 65: nit: if the last question is valid, this is more "pr has to have a release-note section".

.github/workflows/classify.yaml
oribon, line 1: just to be sure, this is to block something from being merged without a rel-note, but doesn't run on pr body edit?

fedepaol: it runs both on edit and on add. Currenlty it's not blocking, but it will show up as red

.github/workflows/classify.yaml
fedepaol: This is philosophical :P 
The PR template now has the section, so the only reason for the section not be there is if the contributor deleted it

oribon (approved):
lgtm

PR #2315, opened 2024-03-07
Speaker Chart: listen to localhost if securemetricsport is present

fedepaol:

>
> /kind bug

/kind cleanup

> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

When we instantiate Values.prometheus.secureMetricsPort, we add a sidecar kuberbac proxy with the metrics port as upstream.

Because of that, it's fair not to expose the unencrypted metrics port from the node itself, as it's exposed to the world.

**Release note**:

```release-note
Expose the speaker metrics port only to localhost if kuberbacproxy is added, so metrics are exported as encrypted only.
```


charts/metallb/templates/speaker.yaml
```diff
@@ -447,7 +453,7 @@ spec:
         args:
           - --logtostderr
           - --secure-listen-address=:{{ .Values.prometheus.secureMetricsPort }}
-          - --upstream=http://$(METALLB_HOST):{{ .Values.prometheus.metricsPort }}/
+          - --upstream=http://127.0.0.1:{{ .Values.prometheus.metricsPort }}/
```
cyclinder, line 456: Does it work in a ipv6-only cluster?

fedepaol: this is a very valid point :sweat_smile:

charts/metallb/templates/speaker.yaml
```diff
@@ -482,19 +495,14 @@ spec:
           - --logtostderr
           - --secure-listen-address=:{{ .Values.speaker.frr.secureMetricsPort }}
           - --tls-cipher-suites=TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,TLS_RSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_CBC_SHA256,TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256
-          - --upstream=http://$(METALLB_HOST):{{ .Values.speaker.frr.metricsPort }}/
+          - --upstream=http://127.0.0.1:{{ .Values.speaker.frr.metricsPort }}/
```
cyclinder, line 498: It seems you also missing it here :)

fedepaol: Fixed, good catch! Let's hope it won't implode again :P

fedepaol:
@cyclinder mind having a look again?

charts/metallb/templates/speaker.yaml
```diff
@@ -482,19 +495,14 @@ spec:
...
+          - --upstream=http://localhost:{{ .Values.speaker.frr.metricsPort }}/
         {{- if .Values.prometheus.speakerMetricsTLSSecret }}
           - --tls-private-key-file=/etc/metrics/tls.key
           - --tls-cert-file=/etc/metrics/tls.crt
         {{- end }}
         ports:
           - containerPort: {{ .Values.speaker.frr.secureMetricsPort }}
             name: frrmetricshttps
-        env:
-          - name: METALLB_HOST
```
cyclinder, line 494: We removed this ENV, but we are still using it. right?

frr-tools/metrics/exporter.go
```diff
@@ -23,6 +23,7 @@ import (
 )
 
 var (
+	metricsHost = flag.String("host", os.Getenv("METALLB_HOST"), "metrics HTTP host address")
```
cyclinder, line 26: We removed this ENV, but we are still using it. right?

fedepaol: I think you are right. That's for when we still want to expose the metrics, but not securely. In that case it makes sense for the host to be the ip of a node.
Good catch again!

cyclinder: I think if we remove the ENV in the `speaker.yaml`, we maybe should give it default value to "localhost". Use this instead of `os.Getenv("METALLB_HOST")`, WDYT?

fedepaol: There are two ways to deploy this machinery. One is with kuberbac proxy embedded, which exposes metrics securely. In that case, we want to listen on localhost because we don't want to expose the metrics port in the wild (because the speaker is hostnetworked).
If, on the other hand, a user doesn't need kuberbac proxy, it makes sense to listen to the node ip, because then prometheus must be able to scrape directly the metrics port (which would not be possible if we listen to localhost).

This is to say that without kuberbac proxy we do need the env variable to pass in the node ip.

cyclinder: Got it, this makes sense.

cyclinder (approved):
LGTM

PR #2317, opened 2024-03-12
Feat: e2e test add JUnit report summary in GitHub Actions

### LINKED ISSUE #2313: CI: generate the test report as junit and feed gh action with it

I am pretty sure GH actions have a nice way to format junit reports. We should explore if that is possible and produce it out of our e2e tests so triaging CI failures does not require digging into logs.

The junit is produced, the output of CI is clear about which tests failed

### Additional context

_No response_

tico88612:
**Is this a BUG FIX or a FEATURE ?**: /kind feature

- Resolve #2313 

- ~~Demo: https://github.com/metallb/metallb/actions/runs/8251026106~~
- Annotation only (latest version): https://github.com/metallb/metallb/actions/runs/8252792566

**Release note**:

```release-note
NONE
```


.github/workflows/ci.yaml
```diff
@@ -250,6 +250,13 @@ jobs:
         with:
           artifact-name: kind-logs-${{ matrix.ip-family }}-${{ matrix.bgp-type }}-${{ matrix.deployment}}
 
+      - name: Publish JUnit Report
+        uses: mikepenz/action-junit-report@v4
+        if: success() || failure()
```
fedepaol, line 255: do we need this if?

tico88612: This job will run whether the E2E test succeeds or fails.

fedepaol: yep but, since we publish in both cases, maybe the if is not needed at all?

tico88612: I'm not sure. I'd be worried that the job would not be executed because the previous job failed.

tico88612: @fedepaol I'm sure it's needed.
Test result: If this line is not added, it will be skipped.
image

This is my test code:
```
jobs:
  # This workflow contains a single job called "build"
  build:
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v3
      - name: Run a one-line script
        run: echo Hello, world!
      - name: Try to fail
        run: exit 1
      - name: Print message if we don't fail
        run: echo Should not get here
      - name: Run a multi-line script
        if: success() || failure()
        run: |
          echo Add other actions to build,
          echo test, and deploy your project.
```

fedepaol: thanks for checking!

fedepaol:
LGTM, sending to merge queue

PR #2330, opened 2024-03-21
Remove the dependencies from the kubernetes test framework

fedepaol:

>
> /kind bug
/kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

We really leverage a few features of the test framework that can be copied locally. The advantage of not having the framework as a dependency is that we won't need to pin the kubernetes dependencies each time, effectively making it easier to bump the dependencies.

I tried to make the review easier by moving a good amount of trivial changes in the first two commits. Hope this helps.

**Release note**:

```release-note
NONE
```


e2etest/pkg/k8sclient/client.go
```diff
@@ -0,0 +1,40 @@
...
+	}
+	var res *clientset.Clientset
+	res, err = clientset.NewForConfig(config)
+	if err != nil {
+		panic(err)
+	}
+	return res
+}
+
+func RestConfig() *restclient.Config {
```
cyclinder, line 30: Maybe we can use `controller-runtime.GetConfigOrDie()`

import (
    ctrl "sigs.k8s.io/controller-runtime"
)

ctrl.GetConfigOrDie()

fedepaol: definetely!

oribon:
lgtm

PR #2346, opened 2024-03-28
Use docker buildx version instead of ls for buildx check.

disco-stu:
In case one of the configured buildx contexts are unavailable (e.g. remote host not reachable), docker buildx ls will hang for several seconds.

/kind cleanup

The container build task _docker_build_cmd checks for buildx support. It uses `docker buildx ls`. However, if docker was configured with multiple contexts and some of those are not reachable for some reason the command waits for timeouts.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -77,7 +77,7 @@ def _docker_build_cmd():
     if cmd:
         out = cmd
     else:
-        out = run("docker buildx ls >/dev/null"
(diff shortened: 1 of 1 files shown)


cyclinder (approved):
Thanks for the fix!

fedepaol:
Thanks! Sent to merge queue

PR #2349, opened 2024-03-30
inv dev-env  fails with --protocol=layer2 because ip address error

lwabish:

/kind bug

In `tasks.py`, function `get_available_ips` returns string, [0] is redundant and leads to format error when generating `dev-env/layer2/config.yaml`

Meanwhile, added a `kubectl wait` to apply layer2 config after controller pod is ready or the webhook would got a `connection refused` error

**Release note**:

```release-note
NONE
```


fedepaol:
Thanks!

PR #2351, opened 2024-04-01
Fix #2311: change layer2status object namespace and add owner reference

### LINKED ISSUE #2311: ServiceL2Status follow-up: create resources in main namespace with speaker pod as OwnerRef

https://github.com/metallb/metallb/issues/2158 is merged, but there's a corner-case we don't handle where if a speaker is deleted permanently (either the node is gone, or the user doesn't want it to act as a speaker) the l2 statuses that it created would be left dangling.

Since [cross-namespace owner refs are not allowed](https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/), we think a reasonable approach would be creating the statuses in the same namespace as the speakers. When we do that, we can no longer use `servicename-node` as the resource's name, because multiple services with the same name can arrive from different namespaces.
So the idea is to have the statuses created with a generated name such as `GenerateName: <node-name>-` on the same namespace as the speaker, with the speaker pod being an ownerRef of the resource. Once a speaker is restarted / gone k8s will handle the deletion for us.

### Additional context

_No response_

lwabish:

>
 /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

fix #2311 

**Release note**:

```release-note
Enable Layer2Service status generation
```


lwabish:
Unit test and e2e test  is coming soon. Please have a look at the implement first, thanks!

oribon (commented):
thanks a lot for pushing this forward! 
Without deep-diving into the splitting logic here, this seems a bit cumbersome. On the issue I mentioned that a reasonable approach might be naming the statuses `GenerateName: <node-name>-`, and from there we can rely on the labels of the resource (to understand to which service it belongs). then we can add the service's name/namespace to the kubectl output and it should be readable. wdyt?

lwabish:
> thanks a lot for pushing this forward! Without deep-diving into the splitting logic here, this seems a bit cumbersome. On the issue I mentioned that a reasonable approach might be naming the statuses `GenerateName: <node-name>-`, and from there we can rely on the labels of the resource (to understand to which service it belongs). then we can add the service's name/namespace to the kubectl output and it should be readable. wdyt?

I think most of the complexity came from how to determine the actual namespace of the service.

As we designed before, the reconciler is receiving name/namespace from both the channel and the cluster cr changes, after moving the status objects to speaker namespace, it can't determine the service namespace directly from `req.Namespace` because `req.Namespace` is no longer always the namespace of the service.

Also, if the reconciler is triggered by the channel event, in which case the cr has not been created yet, getting information from labels seems not possible.

So In my implement, I used `svcName-svcNamespace-nodeName` as the cr name. In this case, I can split this string to get every part I need(although the process behind this is a little cumbersome because I have to solve the problem that svcName and svcNamespace may contain the splitter `-`)

Maybe you can give me an example of the `GenerateName: <node-name>-` to help me understand the detail of your previous idea?

oribon:
oh I see now what's the problem I haven't thought about it all the way through. It might be that generating a name will also be not so easy. I'll think about it and try to go over this tomorrow 😅 thanks!

oribon:
I thought about this for a while and might have a reasonable alternative that should work in theory.
Taking what you said about understanding stuff from the resource's name being problematic I suggest we try a slightly different approach:
1. Instead of using a predefined resource name, we opt to using some generated name (such as `GenerateName: <svc-name>- / l2- / etc`.)
2. We make sure that reconcile requests are raised for servicename+namespace instead of for the l2status resource name. That will require slightly modifying the current eventHandler (EnqueueRequestsFromMapFunc) you wrote, and replacing the current `For(&v1beta1.ServiceL2Status{})` with another eventHandler approach `Named("servicel2status").Watches(&v1beta1.ServiceL2Status{}, handler.EnqueueRequestsFromMapFunc(...)` . In the new eventHandler we can look at the object's labels to understand which service name + namespace is relevant for the resource (this probably means we add a service namespace label as well) and return a `reconcile.Request` accordingly.
3. Modify the existing predicate as needed (e.g reading labels instead of name)

What I like about this resource naming change is that we'll get a more compact result and won't have any logic for the name itself (or promises). This also avoids some corner cases such as trying to create a resource with a name that is too long.

With these changes, the Reconciliation won't change too much: 
fetch status for service -> list existing resource (by the relevant node+service+namespace labels) and create/update/delete as necessary.

wdyt?

lwabish:
> I thought about this for a while and might have a reasonable alternative that should work in theory. Taking what you said about understanding stuff from the resource's name being problematic I suggest we try a slightly different approach:
> 
> 1. Instead of using a predefined resource name, we opt to using some generated name (such as `GenerateName: <svc-name>- / l2- / etc`.)
> 2. We make sure that reconcile requests are raised for servicename+namespace instead of for the l2status resource name. That will require slightly modifying the current eventHandler (EnqueueRequestsFromMapFunc) you wrote, and replacing the current `For(&v1beta1.ServiceL2Status{})` with another eventHandler approach `Named("servicel2status").Watches(&v1beta1.ServiceL2Status{}, handler.EnqueueRequestsFromMapFunc(...)` . In the new eventHandler we can look at the object's labels to understand which service name + namespace is relevant for the resource (this probably means we add a service namespace label as well) and return a `reconcile.Request` accordingly.
> 3. Modify the existing predicate as needed (e.g reading labels instead of name)
> 
> What I like about this resource naming change is that we'll get a more compact result and won't have any logic for the name itself (or promises). This also avoids some corner cases such as trying to create a resource with a name that is too long.
> 
> With these changes, the Reconciliation won't change too much: fetch status for service -> list existing resource (by the relevant node+service+namespace labels) and create/update/delete as necessary.
> 
> wdyt?

Sounds good.  Let me implement this and see if there is any further promblem.

lwabish:
@oribon Please take a look at the latest implement

internal/k8s/controllers/reconciliation_test.go
```diff
@@ -465,21 +484,39 @@ var _ = Describe("FRRK8S Controller", func() {
 var _ = Describe("Layer2 Status Controller", func() {
 	Context("SetupWithManager", func() {
 		It("Should Reconcile correctly", func() {
-
-			statusObjName := fmt.Sprintf("%s-%s", testServiceName, layer2StatusReconciler.NodeName)
+			statusObjFetcherFunc := func() ([]v1beta1.ServiceL2Status, error) {
+				statusList := v1beta1.ServiceL2StatusList{}
+				err := k8sClient.List(context.TODO(), &statusList,
+					client.InNamespace(speakerNamespace),
```
oribon, line 490: can you make this in all namespaces (just so we're sure we aren't creating unnecessary ones)

internal/k8s/controllers/reconciliation_test.go
```diff
@@ -465,21 +484,39 @@ var _ = Describe("FRRK8S Controller", func() {
...
 			layer2StatusUpdateChan <- NewL2StatusEvent(testNamespace, testServiceName)
 			Eventually(func() string {
-				toCheck := v1beta1.ServiceL2Status{}
-				err := k8sClient.Get(context.TODO(), client.ObjectKey{Name: statusObjName, Namespace: testNamespace}, &toCheck)
+				statuses, err := statusObjFetcherFunc()
 				if err != nil {
+					return err.Error()
+				}
+				if len(statuses) != 1 {
+					return ""
```
oribon, line 514: nit: have an fmt.Errorf(...) explaining what happened

internal/k8s/controllers/reconciliation_test.go
```diff
@@ -465,21 +484,39 @@ var _ = Describe("FRRK8S Controller", func() {
...
+				}
+				if len(statuses[0].OwnerReferences) != 1 {
 					return ""
```
oribon, line 517: same, also can you verify the owner ref is the pod?

internal/k8s/controllers/reconciliation_test.go
```diff
@@ -489,22 +526,29 @@ var _ = Describe("Layer2 Status Controller", func() {
...
 			Eventually(func() string {
-				toCheck := v1beta1.ServiceL2Status{}
-				err := k8sClient.Get(context.TODO(), client.ObjectKey{Name: statusObjName, Namespace: testNamespace}, &toCheck)
-				if err != nil || len(toCheck.Status.Interfaces) != 1 {
+				statuses, err := statusObjFetcherFunc()
+				if err != nil {
+					return err.Error()
+				}
+				if len(statuses) != 1 {
+					return ""
```
oribon, line 534: same

internal/k8s/controllers/reconciliation_test.go
```diff
@@ -489,22 +526,29 @@ var _ = Describe("Layer2 Status Controller", func() {
...
+				}
+				if len(statuses[0].Status.Interfaces) == 0 {
 					return ""
```
oribon, line 537: same + missing owner ref validation

oribon (commented):
thanks a lot for this! overall looks good, still need to go over the e2es.

is there a reason you went with the "%s-%s-%s" format instead of a generated name for the resource? I'm concerned the names will be too long (or just messy in a `-o wide`), wonder if you hit something that you didn't like

lwabish:
> thanks a lot for this! overall looks good, still need to go over the e2es.
> 
> is there a reason you went with the "%s-%s-%s" format instead of a generated name for the resource? I'm concerned the names will be too long (or just messy in a `-o wide`), wonder if you hit something that you didn't like

Considering the name of the cr, I have not paid much attention to it so I just chose this because it first came out to my mind. We can decide the final name for crs later. Current logic has removed the dependency to cr names so I think this is a reletively independent part

e2etest/go.mod
```diff
@@ -2,6 +2,11 @@ module go.universe.tf/e2etest
 
 go 1.21
 
+replace (
+	k8s.io/kubectl => k8s.io/kubectl v0.29.3
+	k8s.io/kubelet => k8s.io/kubelet v0.29.3
+)
+
```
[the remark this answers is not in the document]
lwabish: Something similar to this has occured last time: https://github.com/metallb/metallb/pull/2198#issuecomment-1865791977

Further information is,while I am using Jetbrain IDE developing, it runs `go list -m -json all` to obtain all dependencies at project loading, I guess. 

Without the replaces, if the project requires some kubernetes code whose version is set to v0.0.0,  the `go list` command would fail.

So this is something like a work around I think, but not sure if this is the best way.Also,  I am a little confused why this problem is not triggerd in github ci env

internal/k8s/controllers/layer2_status_controller.go
```diff
@@ -85,8 +90,16 @@ func (r *Layer2StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-		state.Labels = map[string]string{LabelAnnounceNode: r.NodeName, LabelServiceName: svcName}
+		state.Labels = map[string]string{
+			LabelAnnounceNode:     r.NodeName,
+			LabelServiceName:      req.Name,
+			LabelServiceNamespace: req.Namespace,
+		}
 		state.Status = desiredStatus
+		err = controllerutil.SetOwnerReference(r.SpeakerPod, state, r.Scheme())
+		if err != nil {
+			level.Warn(r.Logger).Log("error setting owner ref", err)
```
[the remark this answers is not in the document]
lwabish: I was considering maintaining owner reference as an extra function in addition to the core function, rather than a part of the core function.
So I didn't do a fast fail when facing errors about owner reference.

internal/k8s/controllers/layer2_status_controller.go
```diff
@@ -99,32 +112,44 @@ func (r *Layer2StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 			func(ctx context.Context, object client.Object) []reconcile.Request {
-				evt, ok := object.(*l2StatusEvent)
-				if !ok {
-					level.Error(r.Logger).Log("controller", "Layer2StatusReconciler", "received an object that is not a l2StatusEvent from channel", "object", object)
-					return []reconcile.Request{}
-				}
-				level.Debug(r.Logger).Log("controller", "Layer2StatusReconciler", "enqueueing", "object", evt)
-				// unify req.Name to svcName-nodeName
+				level.Debug(r.Logger).Log("controller", "Layer2StatusReconciler", "enqueueing", "object", object)
+				label := object.GetLabels()
```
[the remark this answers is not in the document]
lwabish: You mean we don't have to filter by the labels'(svcName, svcNamespace) existence?
@oribon

internal/k8s/controllers/reconciliation_test.go
```diff
@@ -122,6 +123,32 @@ var _ = BeforeSuite(func() {
...
+	Expect(err).ToNot(HaveOccurred())
+
+	speakerPod = &corev1.Pod{
+		ObjectMeta: metav1.ObjectMeta{Name: "speaker", Namespace: speakerNamespace},
+		Spec: corev1.PodSpec{Containers: []corev1.Container{
+			{Name: "speaker", Image: "speaker"},
+		}},
+	}
+	err = k8sClient.Create(ctx, speakerPod)
+	Expect(err).ToNot(HaveOccurred())
```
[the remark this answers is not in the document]
lwabish: Tried comment this and the reconciler will not be able to set owner reference

```
"ServiceL2Status.metallb.io \"test-service-test-controller-testnode\" is invalid: metadata.ownerReferences.uid: Invalid value: \"\": uid must not be empty"}
```

e2etest/go.mod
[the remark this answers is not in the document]
oribon: yeah it's a bit odd, can you try removing it from the commit and see what happens in ci?

e2etest/l2tests/l2.go
```diff
@@ -153,15 +153,22 @@ var _ = ginkgo.Describe("L2", func() {
...
-			Consistently(refreshL2StatusNodeFunc, 5*time.Second).Should(Equal(l2Status.Status.Node))
+			}, time.Minute, time.Second).Should(Equal(l2Status.Status.Node))
+			Consistently(func() string {
+				var ss []metallbv1beta1.ServiceL2Status
+				if ss, err = status.GetSvcL2Status(ConfigUpdater.Client(), svc); err != nil || len(ss) != 1 {
+					return ""
+				}
+				return ss[0].Status.Node
+			}, 5*time.Second).Should(Equal(speakingNode))
 		})
```
oribon, line 172: can we also add validation that the status is deleted after the service is deleted? (I think it's not there, ignore this if I'm wrong)

charts/metallb/values.yaml
```diff
@@ -270,6 +270,7 @@ speaker:
     enabled: true
   # ignore the exclude-from-external-loadbalancer label
   ignoreExcludeLB: false
+  layer2Status: false
```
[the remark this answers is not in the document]
oribon: I actually like the idea of keeping the option to disable if someone really wants to (for the sake of "performance impact" or w.e) while defaulting to enabled, wdyt?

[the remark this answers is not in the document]
fedepaol: I'd rather have it and that's it. If we want to leave the code that disables it and the parameter, fine (even though it comes with a complexity cost, even if little), but I would hide the parameter from helm until we realize this doesn't scale well (and even if it does, maybe it will be better to optimize it rather than allowing users to disable).

e2etest/go.mod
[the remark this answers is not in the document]
lwabish: > yeah it's a bit odd, can you try removing it from the commit and see what happens in ci?

Ci will work fine because this actuallly doesn't affect `go mod tidy`. Maybe I can put this in a temporary changelist to avoid IDE error.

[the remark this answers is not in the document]
fedepaol: What if you go get those deps instead of replacing? Replacing pins the deps and it will require manual updates (plus, we made a considerable effort to avoid doing that), with `require` we can rely on dependabot.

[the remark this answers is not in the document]
lwabish: > What if you go get those deps instead of replacing? Replacing pins the deps and it will require manual updates (plus, we made a considerable effort to avoid doing that), with `require` we can rely on dependabot.

Got it. I tried `go get k8s.io/kubectl` and `go get k8s.io/kubelet` and this did solve the problem but meanwhile they were added to go.mod. Is it ok to submit this change to go.mod?

```
require(
...
	k8s.io/kubectl v0.29.4 // indirect
	k8s.io/kubelet v0.29.4 // indirect
...
)
```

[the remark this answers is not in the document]
fedepaol: absolutely!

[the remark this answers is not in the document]
fedepaol: @oribon I wonder if we should enforce `go list -m -json all` to work on CI, as we want to have a smooth dev experience (not only to those using vim :trollface: )

lwabish:
@oribon pls take a look agian

e2etest/l2tests/interface_selector.go
```diff
@@ -107,36 +107,26 @@ var _ = ginkgo.Describe("L2-interface selector", func() {
...
-						return fmt.Errorf("")
-					}
-					l2Status = ss[0]
+				var ss []metallbv1beta1.ServiceL2Status
+				if ss, err = status.GetSvcL2Status(ConfigUpdater.Client(), svc); err == nil {
+					l2Statuses = ss
 				}
 				return err
 			}, 2*time.Minute, 1*time.Second).ShouldNot(HaveOccurred())
+			Expect(l2Statuses).To(HaveLen(1))
```
oribon, line 118: why this change? I think keeping it in the eventually is clearer (returning an error with what we got if it's not 1)

e2etest/l2tests/l2.go
```diff
@@ -153,15 +147,33 @@ var _ = ginkgo.Describe("L2", func() {
...
+			service.Delete(cs, svc)
+			ginkgo.By("validating the status object is delete after the service is deleted")
+			Eventually(func() error {
+				var ss []metallbv1beta1.ServiceL2Status
+				if ss, err = status.GetSvcL2Status(ConfigUpdater.Client(), svc); err == nil {
+					l2Statuses = ss
+				}
+				return err
+			}, 2*time.Minute, 1*time.Second).ShouldNot(HaveOccurred())
+			Expect(l2Statuses).To(HaveLen(0))
```
oribon, line 176: same (have it in eventually)

internal/k8s/controllers/layer2_status_controller.go
```diff
@@ -99,32 +112,30 @@ func (r *Layer2StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	level.Debug(r.Logger).Log("controller", "Layer2StatusReconciler", "updated state", dumpResource(state))
-
 	return ctrl.Result{}, nil
 }
 
 func (r *Layer2StatusReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	p := predicate.NewPredicateFuncs(func(object client.Object) bool {
 		if s, ok := object.(*v1beta1.ServiceL2Status); ok {
-			return strings.HasSuffix(s.Name, r.NodeName)
+			return r.NodeName == s.GetLabels()[LabelAnnounceNode]
```
oribon, line 121: nit: add a verification that the labels aren't nil

internal/k8s/controllers/layer2_status_controller.go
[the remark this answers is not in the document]
oribon: I think this is still the same since last time

internal/k8s/controllers/reconciliation_test.go
```diff
@@ -122,6 +124,35 @@ var _ = BeforeSuite(func() {
 	})
 	Expect(err).ToNot(HaveOccurred())
 
+	ctx, cancel = context.WithCancel(context.TODO())
+	// test service is located in testNamespace
+	// speaker pod is deployed in speakerNamespace
+	// the layer2 status crs are maintained in speakerNamespace
+	err = func(names ...string) error {
```
oribon, line 131: nit: having this as a loop over []string{testNamespace, speakerNamespace} is more readable imo

oribon (commented):
Left a few nits, overall this looks good from my side.
Regarding the object's name I'm ok with the current approach but I think we should opt for some generated name "l2- / <NODE-NAME>- / etc." as I believe a `node-service-namespace` format is a bit cumbersome. If we go that route we should also make sure the service's name/namespace is visible when running a `kubectl get -o wide`.

Also, can you organize the commits as we did last time?

I'll be away for the next ~2.5 weeks, thanks a lot for your contributions with this feature we really appreciate it!

/cc @fedepaol

fedepaol:
> I'll be away for the next ~2.5 weeks, thanks a lot for your contributions with this feature we really appreciate it!
> 
> /cc @fedepaol

@lwabish ping me when you adjust the commits and fix the last issues and I'll move forward from here, you are in good hands!

e2etest/go.mod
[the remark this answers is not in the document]
lwabish: > @oribon I wonder if we should enforce `go list -m -json all` to work on CI, as we want to have a smooth dev experience (not only to those using vim :trollface: )

Speaking of this, you can take a look at #2349 , which is a bug related to dev experience

internal/k8s/controllers/layer2_status_controller.go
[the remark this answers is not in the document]
lwabish: > I think this is still the same since last time

Oh, I misunderstood you and thought you were talking about the predicate part.

This part of code(in `Watches(&v1beta1.ServiceL2Status{}, ...)`) is not designed for validation, but for building reconciler request from cr label insdead of cr metadata, which is what we discussed before at the very beginning of this pr to solve some tricky problem.

So I think we should stick to the current `Watches(&v1beta1.ServiceL2Status{}, ...)` part and revert some of my previous commits which deleted some validation of label in predicate part due to my misunderstanding.

wdyt @fedepaol

lwabish:
@fedepaol 
I have rebased the commits. Now let me summarize what is left to be done from my view. 

1. Make sure the `SetupWithManager` function of `Layer2StatusReconciler`  is implemented as we expected. 
    Refer to this review and our discussion https://github.com/metallb/metallb/pull/2351#discussion_r1554887923
2. Do some adjustment to layer2 stauts object names based on generated name, which I will finish at the last step.

e2etest/go.mod
fedepaol, line 1: please adjust the commit message with the same what - how format.

charts/metallb/templates/speaker.yaml
fedepaol, line 1: please adjust the commit message with the same what - how format.

In this case:
what - expose the speaker's name as env variable / add permissions
why: because we need to fetch the pod object to add the owner ref.

config/manifests/metallb-frr-k8s-prometheus.yaml
```diff
@@ -1846,6 +1846,7 @@ rules:
   - pods
   verbs:
   - list
+  - get
```
fedepaol, line 1849: I am tempted to say this is not necessary as helm is passing but we didn't add the corresponding permission.
Also, we should've added it here https://github.com/metallb/metallb/blob/fe9434970028279a062a017c8d27c3867d23a18e/config/rbac/role.yaml#L152 and let the generation part do the rest. I am also surprised the CI didn't catch this :worried:

config/manifests/metallb-frr-k8s-prometheus.yaml
```diff
@@ -3055,6 +3056,10 @@ spec:
           valueFrom:
             fieldRef:
               fieldPath: spec.nodeName
+        - name: METALLB_POD_NAME
```
fedepaol, line 3059: same, this should go under metallb/config/controllers/speaker.yaml

config/manifests/metallb-frr-k8s-prometheus.yaml
fedepaol: ah this drove me mad (I even filed a wip pr to validate). You committed the changes in the kustomize bases in the first commit. I think it makes sense to move them to this commit too.

lwabish: The rbac permission of getting for pods was added by me manually at first.
After this modification, ci would check if manifests are  in sync by running `inv generatemanifests`. So I have to commit what `inv generatemanifests` modified (manifests in `config/manifests/`).
@fedepaol

fedepaol: yes that makes sense. I am just arguing about having all of them in the same commit.
Also, you didn't add the rbac to the helm chart (but the helm lane is passing), so I'd check if that capability is really required (and if so, let's add it to the chart as well)

lwabish: Got you. I was mainly focused on the reconciler but did not pay much attention to the manifests and helm chart(thought these parts are covered by the `inv generatemanifests`).I'll try optimizing these parts.

fedepaol: I'd give it a try without the rbac to see if it's really required.

e2etest/go.mod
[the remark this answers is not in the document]
fedepaol: this has still to be replaced with require, right?

e2etest/go.mod
```diff
@@ -14,10 +19,10 @@ require (
 	github.com/prometheus/client_golang v1.19.0 // indirect
 	go.uber.org/zap v1.26.0 // indirect
 	golang.org/x/sys v0.18.0 // indirect
-	k8s.io/api v0.29.3
+	k8s.io/api v0.29.4
```
fedepaol, line 22: is there a reason for these bumps?

e2etest/l2tests/l2.go
```diff
@@ -128,40 +128,56 @@ var _ = ginkgo.Describe("L2", func() {
...
+				speakingNode = node.Name
 				return node.Name
-			}
-			Eventually(refreshL2StatusNodeFunc, time.Minute, time.Second).Should(Equal(l2Status.Status.Node))
-			Consistently(refreshL2StatusNodeFunc, 5*time.Second).Should(Equal(l2Status.Status.Node))
+			}, time.Minute, time.Second).Should(Equal(l2Status.Status.Node))
+			Consistently(func() string {
+				var ss []metallbv1beta1.ServiceL2Status
+				if ss, err = status.L2ForService(ConfigUpdater.Client(), svc); err != nil || len(ss) != 1 {
+					return ""
```
fedepaol, line 164: we are going to have a bad time understanding why this failed. Let's be more verbose and split the error case from the len case, like we do below

e2etest/l2tests/interface_selector.go
```diff
@@ -107,36 +107,27 @@ var _ = ginkgo.Describe("L2-interface selector", func() {
...
-			}, 5*time.Second).ShouldNot(HaveOccurred())
+			}, 2*time.Minute, 1*time.Second).ShouldNot(HaveOccurred())
 
-			Consistently(func() error {
-				l2Status, err = status.GetL2Status(ConfigUpdater.Client(), svc, l2Status.Status.Node)
-				if err != nil || l2Status.Status.Interfaces[0].Name != NodeNics[0] {
-					return fmt.Errorf("")
+			Consistently(func() string {
+				var ss []metallbv1beta1.ServiceL2Status
+				if ss, err = status.L2ForService(ConfigUpdater.Client(), svc); err != nil || len(ss) != 1 {
```
fedepaol, line 123: let's be more descriptive when this fails, like we do above (I know it was already sub-optimal :-) )

e2etest/go.mod
lwabish: These are submitted with this: https://github.com/metallb/metallb/pull/2351#discussion_r1568464636

fedepaol:
@lwabish as always, no pressure, just checking you are not waiting something from us.
I think there are a few unsolved discussions.

lwabish:
> @lwabish as always, no pressure, just checking you are not waiting something from us.
> 
> I think there are a few unsolved discussions.

Oh,yes. I have been working on something about job switching and almost done.

I should be able to continue in the coming days this week🤓

e2etest/go.mod
lwabish: @fedepaol So Should I downgrade this dependency version?

e2etest/go.mod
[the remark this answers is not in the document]
lwabish: > this has still to be replaced with require, right?

Seem right. Without replaces `go list` still reports errors
```
❯ cd e2etest
❯ go list -m -json all
go: k8s.io/kubectl@v0.0.0: invalid version: unknown revision v0.0.0
go: k8s.io/kubelet@v0.0.0: invalid version: unknown revision v0.0.0
```

lwabish:
@fedepaol Please take a look at the current implement. 😄

fedepaol:
> @fedepaol Please take a look at the current implement. 😄

will do by eod, thanks!

e2etest/go.mod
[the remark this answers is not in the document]
fedepaol: even with require, you mean?

internal/k8s/controllers/layer2_status_controller.go
```diff
@@ -61,19 +64,24 @@ func (r *Layer2StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	level.Info(r.Logger).Log("controller", "Layer2StatusReconciler", "start reconcile", req.NamespacedName.String())
 	defer level.Info(r.Logger).Log("controller", "Layer2StatusReconciler", "end reconcile", req.NamespacedName.String())
 
-	svcName := strings.TrimSuffix(req.Name, fmt.Sprintf("-%s", r.NodeName))
-	ipAdvS := r.StatusFetcher(types.NamespacedName{Name: svcName, Namespace: req.Namespace})
+	serviceName, serviceNamespace := req.Name, req.Namespace
+	statusName, statusNamespace := fmt.Sprintf("%s-%s-%s", req.Name, req.Namespace, r.NodeName), r.Namespace
```
fedepaol, line 68: nit: use serviceName as the argument in the formatting string will make it clearer how the statusName is built with.
Also, @oribon had concerns about generating such name in this way with the risk of going over the maximum size. The alternative would be to use a generated name, but doing so would make the status - service / node binding implemented only with something immutable like labels.

If we do so, we'd risk to have somebody changing the label and leaving a rogue status around. A possible alternative would be:
- generated name
- node and service name as labels (for better discoverability)
- node and service name explicit in the status part (to provide immutability of the information)

If we do that, we'll probably have to add an index as well so we can retrieve the entry easily (we won't be able to do that by name anymore).

Thoughts?

Based on the outcome of this discussion, we'll have to choose what to do with the predicate.

e2etest/pkg/status/l2status.go
```diff
@@ -4,37 +4,29 @@ package status
...
-		s, err := GetL2Status(cs, svc, node.Name)
-		if err != nil && errors.IsNotFound(err) {
-			continue
-		}
-		if err != nil {
-			return nil, err
-		}
-		l2Statuses = append(l2Statuses, s)
+func L2ForService(cs client.Client, svc *v1.Service) ([]v1beta1.ServiceL2Status, error) {
+	statusList := v1beta1.ServiceL2StatusList{}
```
fedepaol, line 20: if we return error here if we get more than one instance, the calling sites will be cleaner (also, it feels weird to read "status.ForService" and then see that we are receiving more than one status.

e2etest/go.mod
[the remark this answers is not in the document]
lwabish: > even with require, you mean?

yes

internal/k8s/controllers/layer2_status_controller.go
lwabish: What do you mean by adding an index? Can you give me an quick example for better understanding?

fedepaol: Something along the line of what we do here https://github.com/metallb/metallb/blob/main/internal/k8s/k8s.go#L246 where we set a key to retrieve the ep slice based on the service name

e2etest/pkg/status/l2status.go
```diff
@@ -5,36 +5,40 @@ package status
...
-	for _, node := range nodes.Items {
-		s, err := GetL2Status(cs, svc, node.Name)
-		if err != nil && errors.IsNotFound(err) {
-			continue
-		}
-		if err != nil {
-			return nil, err
-		}
-		l2Statuses = append(l2Statuses, s)
+func L2sForService(cs client.Client, svc *v1.Service) ([]v1beta1.ServiceL2Status, error) {
```
oribon, line 31: nit: this is a bit confusing, we can make this private or move the logic to the `L2ForService`, and create another function (e.g `L2StatusExistsFor`) for the other place you determine if a status exists or not.

internal/k8s/controllers/layer2_status_controller.go
oribon: +1 for that idea!

e2etest/pkg/status/l2status.go
fedepaol: We can just use L2ForService as in every place we use L2sForServices we error if we get more than one entry

lwabish: There is one single case we need to validate there's 0 status obj and we can't ensure this from the error returned by L2ForService. So I had to expose L2sForService.

internal/k8s/controllers/layer2_status_controller.go
lwabish: Let me try to tidy this up.
With completely generated names for crs, along with index which maintains the node->svc relationship, we would    use something like this in predicate to determine if a cr belongs to a specific node(speaker), am I right ?
https://github.com/metallb/metallb/blob/83b5dd4739e32d22348524fbf7ffedc4f1a7ad1f/internal/k8s/controllers/endpoints.go#L16-L18

lwabish: Thought some details and found there may be a tiny problem with the nodeName.
When the first time a status cr is created, there is no status sub-resource. The status sub-resource is patched after a re-enqueue.
So It seems not possible to implement our predicate based on something like status.nodeName(metadata.name/metadata.labels seems more applicable).

lwabish: Will submit this change this weekend 😄

lwabish:
I have change the cr names to fully random string.

internal/k8s/controllers/layer2_status_controller.go
```diff
@@ -61,32 +67,54 @@ func (r *Layer2StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
-		Namespace: req.Namespace,
-	}}
-	err := r.Client.Get(ctx, types.NamespacedName{Name: req.Name, Namespace: req.Namespace}, state)
-	if err != nil && !errors.IsNotFound(err) {
-		return ctrl.Result{}, err
+	var state *v1beta1.ServiceL2Status
+	if len(stateList.Items) == 0 {
+		state = &v1beta1.ServiceL2Status{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      statusName,
```
oribon, line 93: We don't really need to generate the string randomly ourselves, we can just use here `GenerateName: "l2-"` instead of `Name` :)

oribon (commented):
@lwabish thanks a lot for this, haven't gone through the new indexing change - mind fixing the conflicts so we can see the e2e running first?

lwabish:
> @lwabish thanks a lot for this, haven't gone through the new indexing change - mind fixing the conflicts so we can see the e2e running first?

I wonder why conflict occured in `e2etest/go.work.sum` and fixing this is confusing 😢

internal/k8s/controllers/layer2_status_controller.go
lwabish: Oh, forgot that

lwabish:
@oribon ready for your review

api/v1beta1/servicel2status_types.go
```diff
@@ -26,6 +26,10 @@ import (
 type MetalLBServiceL2Status struct {
 	// Node indicates the node that receives the directed traffic
 	Node string `json:"node,omitempty"`
+	// ServiceName indicates the service this status represents
+	ServiceName string `json:"serviceName,omitempty"`
+	// ServiceNamespace indicates the namespace of the service
+	ServiceNamespace string `json:"serviceNamespace,omitempty"`
```
oribon, line 32: let's add a printcolumn for these like you did for `.status.node`

oribon, line 32: let's make these fields immutable:
```go
	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
	Node string `json:"node,omitempty"`
	// ServiceName indicates the service this status represents
	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
	ServiceName string `json:"serviceName,omitempty"`
	// ServiceNamespace indicates the namespace of the service
	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="Value is immutable"
	ServiceNamespace string `json:"serviceNamespace,omitempty"`
```
verified on a local cluster that it actually works when using:
```
k edit -n metallb-system servicel2status ".." --subresource=status
```

internal/k8s/controllers/layer2_status_controller.go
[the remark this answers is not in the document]
oribon: sorry about the earlier comment, wdyt about having the predicate act on the status fields after we make them immutable? cc @fedepaol

oribon (commented):
awesome work! 
left a few small comments

internal/k8s/controllers/layer2_status_controller.go
```diff
@@ -61,32 +66,53 @@ func (r *Layer2StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	defer level.Info(r.Logger).Log("controller", "Layer2StatusReconciler", "end reconcile", req.NamespacedName.String())
 
-	svcName := strings.TrimSuffix(req.Name, fmt.Sprintf("-%s", r.NodeName))
-	ipAdvS := r.StatusFetcher(types.NamespacedName{Name: svcName, Namespace: req.Namespace})
+	serviceName, serviceNamespace := req.Name, req.Namespace
+
+	ipAdvS := r.StatusFetcher(types.NamespacedName{Name: serviceName, Namespace: serviceNamespace})
+	stateList := &v1beta1.ServiceL2StatusList{}
+	if err := r.Client.List(ctx, stateList, client.MatchingFields{
+		serviceIndexName: types.NamespacedName{Name: serviceName, Namespace: serviceNamespace}.String(),
```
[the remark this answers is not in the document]
lwabish: This is what I thought about this problem: https://github.com/metallb/metallb/pull/2351#discussion_r1616510392.
The duty of avoiding multiple speaker problem still belongs to predicate.

e2etest/pkg/status/l2status.go
```diff
@@ -5,36 +5,38 @@ package status
...
-		l2Statuses = append(l2Statuses, s)
+	statuses := statusList.Items
+	if len(statuses) == 0 {
+		return nil, errors.NewNotFound(schema.ParseGroupResource("serviceL2Status.metallb.io"), fmt.Sprintf("%s/%s", svc.Name, svc.Namespace))
+	}
+	if len(statuses) > 1 {
+		return nil, fmt.Errorf("got more than 1 serviceL2Status object: %d", len(statuses))
 	}
-	return l2Statuses, nil
+	return &(statuses[0]), nil
```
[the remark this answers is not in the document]
fedepaol: Not sure about this. StatusList is a collection, here we only want the first element

internal/k8s/controllers/layer2_status_controller.go
```diff
@@ -61,32 +66,53 @@ func (r *Layer2StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	var result controllerutil.OperationResult
 	result, err = controllerutil.CreateOrPatch(ctx, r.Client, state, func() error {
-		state.Labels = map[string]string{LabelAnnounceNode: r.NodeName, LabelServiceName: svcName}
+		state.Labels = map[string]string{
+			LabelAnnounceNode:     r.NodeName,
+			LabelServiceName:      serviceName,
+			LabelServiceNamespace: serviceNamespace,
+		}
 		state.Status = desiredStatus
+		err = controllerutil.SetOwnerReference(r.SpeakerPod, state, r.Scheme())
```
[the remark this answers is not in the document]
fedepaol: return err?
I like this more, it feels more idiomatic (and consistent with the rest)

e2etest/l2tests/l2.go
```diff
@@ -128,40 +128,45 @@ var _ = ginkgo.Describe("L2", func() {
...
+				}
+				return s.Status.Node
+			}, 5*time.Second).Should(Equal(speakingNode))
+
+			service.Delete(cs, svc)
+			ginkgo.By("validating the status object is deleted after the service is deleted")
+			Eventually(func() bool {
+				_, err = status.L2ForService(ConfigUpdater.Client(), svc)
+				return pkgerr.IsNotFound(err)
+			}, 2*time.Minute, 1*time.Second).Should(Equal(true))
```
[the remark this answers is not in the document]
fedepaol: I'd say both have value (and a different one). The by tells us in which phase of the test we are, the assertion reason gives us more information about why the test failed.

internal/k8s/controllers/layer2_status_controller.go
```diff
@@ -99,40 +125,89 @@ func (r *Layer2StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+				level.Error(r.Logger).Log("controller", "Layer2StatusReconciler", "missing meta", "object", object)
+				return false
+			}
+			var node string
+			if node, ok = label[LabelAnnounceNode]; !ok {
+				level.Error(r.Logger).Log("controller", "Layer2StatusReconciler", "missing meta", "object", object)
+				return false
+			}
+			// only trigger the reconciler if the service is announced by this node
+			if node != r.NodeName {
```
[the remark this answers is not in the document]
karampok: There is no
```
level.Error(r.Logger).Log("controller", "Layer2StatusReconciler", "missing meta", "object", object)
```
before thel last `return false`, while there is in all previous `return false`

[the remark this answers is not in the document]
fedepaol: That's because this is the happy path (the node can or can not be the current node) while the others are exceptional conditions that are worth labeling.

internal/k8s/controllers/layer2_status_controller.go
```diff
@@ -61,32 +67,59 @@ func (r *Layer2StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
+	}
+
+	var errs []error
 	if len(ipAdvS) == 0 {
-		err := r.Client.Delete(ctx, &v1beta1.ServiceL2Status{ObjectMeta: metav1.ObjectMeta{Name: req.Name, Namespace: req.Namespace}})
-		return ctrl.Result{}, client.IgnoreNotFound(err)
+		for key := range serviceL2statuses.Items {
+			if err := r.Client.Delete(ctx, &serviceL2statuses.Items[key]); err != nil && !errors.IsNotFound(err) {
+				errs = append(errs, err)
+			}
```
oribon, line 85: I understand this should probably be handled by the predicate, but let's be on the safe side and validate that the status first belongs to this node (`.Status.Node`) before deleting it.

internal/k8s/controllers/layer2_status_controller.go
```diff
@@ -99,40 +132,89 @@ func (r *Layer2StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
...
 	return ctrl.Result{}, nil
 }
 
 func (r *Layer2StatusReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	p := predicate.NewPredicateFuncs(func(object client.Object) bool {
 		if s, ok := object.(*v1beta1.ServiceL2Status); ok {
-			return strings.HasSuffix(s.Name, r.NodeName)
+			// only objects with complete labels that can illustrate which service it is related to
+			// can trigger the reconciler
+			label := s.GetLabels()
```
oribon, line 143: now that the status part is immutable we can remove anything related to labels in the predicate :)

oribon (approved):
last few small comments, other than that looks good from my side!
thanks a lot for helping with this (and for your patience)!!

internal/k8s/controllers/layer2_status_controller.go
lwabish: Please take a look at this, am I right about the predicates?https://github.com/metallb/metallb/pull/2351#discussion_r1616510392

e2etest/l2tests/l2.go
```diff
@@ -128,40 +128,45 @@ var _ = ginkgo.Describe("L2", func() {
...
 			allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
 			Expect(err).NotTo(HaveOccurred())
 
-			var l2Statuses []*metallbv1beta1.ServiceL2Status
+			var l2Status *metallbv1beta1.ServiceL2Status
 			Eventually(func() error {
-				var ss []*metallbv1beta1.ServiceL2Status
-				if ss, err = status.GetSvcPossibleL2Status(ConfigUpdater.Client(), svc, allNodes); err == nil {
-					l2Statuses = ss
+				if l2Status, err = status.L2ForService(ConfigUpdater.Client(), svc); err != nil {
```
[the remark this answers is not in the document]
lwabish: this is for better readability?

internal/k8s/controllers/layer2_status_controller.go
oribon: oh sorry I missed that, I guess you're right since `CreateOrPatch` updates the status subresource only when the object is there . I'm fine with leaving it on the labels then (at least for now)

e2etest/l2tests/l2.go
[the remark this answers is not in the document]
karampok: this is for defining better the variable scope, kind of https://github.com/uber-go/guide/blob/master/style.md#reduce-scope-of-variables. In general (I think at least) variables should be scoped within if scope. And for consistency.

lwabish:
@fedepaol Hi, I have rebased the commits.Please take a look at the latest code

fedepaol:
Second that! Thanks for the patience of following our sometimes not aligned comments, and for filling something that was greatly missing in MetalLB :heart:

lwabish:
My pleasure to contribute and thanks you guys!

Issue #1307, opened 2022-04-05
tasks.py - deploy all the needed dependencies


fedepaol:
Now that we have extra deps, such as controller-gen and kustomize, we should be able to deploy them automatically from tasks.py

fedepaol:
sorry if I reply so late. The idea here is to have a simple command in tasks.py that deploys the dependencies required such as controller-gen

26tanishabanik:
@fedepaol , I think commands for installing controller-gen is added, is that right?

2SpaceMasterRace:
Hey @fedepaol I should be able to work on this

Is it cool if I get assigned to this issue?

fulviodenza:
Hi @fedepaol I was starting looking into this and found out that `controller-gen` has already been implemented. While Kustomize is not, I can take the issue if Kustomize dep is still valid.

fedepaol:
@fulviodenza absolutely, and thanks!

fedepaol:
Closing, kustomize is not really needed as we moved to `kubectl kustomizze` and I added gingko which was the last standing one

PR #2366, opened 2024-04-15
Add Kustomize dependency

fulviodenza:

/kind dependencies

This PR adds the dependency for Kustomize as requested in #1307 in the python script to fetch dependencies, the controller-gen was not required anymore.
We can use this PR also to update the tasks.py file in case there are some dependencies to update. Further we should verify the kustomize version.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -333,7 +333,7 @@ def generate_manifest(ctx, crd_options="crd:crdVersions=v1", bgp_type="native", 
         layer = bgp_type
         if with_prometheus:
             layer = "prometheus-" + layer
-        res = run("{} kustomize config/{} > {}".format(kubectl_path, layer, output))
+        res = run("{} build config/{} > {}".format(kustomize_path, layer, output))
         if not res.ok:
             raise Exit(message="Failed to kustomize manifests")
 
--- a/tasks.py
+++ b/tasks.py
@@ -1072,7 +1072,7 @@ def fetch_controller_gen():
 @cache
 def fetch_kustomize():
     kustomize_zipfile = f"kustomize_{kustomize_version}_$(go env GOOS)_$(go env GOARCH).tar.gz"
-    curl_command = f"curl -L -o {kustomize_path}/{kustomize_zipfile} https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/{kustomize_version}/{kustomize_file}"
+    curl_command = f"curl -L -o {kustomize_path}/{kustomize_zipfile} https://github.com/kubernetes-sigs/kustomize/releases/download/kustomize/{kustomize_version}/{kustomize_zipfile}"
     tar_command = f"tar xzf {kustomize_path}/{kustomize_zipfile} && -C {build_path}/bin && rm -rf {kustomize_path}"
     get_version_command = f"{kustomize_path} version"
     fetch_dependency(kustomize_path, kustomize_version, curl_command+tar_command, get_version_command, "Version:")
--- a/tasks.py
+++ b/tasks.py
@@ -32,10 +32,12 @@ all_architectures = set(["amd64",
 default_network = "kind"
 extra_network = "network2"
 controller_gen_version = "v0.14.0"
+kubectl_version = "v1.27.0"
+kustomize_version = "v5.4.0"
 build_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "build")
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -32,10 +32,12 @@
 default_network = "kind"
 extra_network = "network2"
 controller_gen_version = "v0.14.0"
+kubectl_version = "v1.27.0"
+kustomize_version = "v5.4.0"
```
fulviodenza, line 36: We should verify that this is actually the version needed, I could not find which exact version metallb uses

fedepaol: I guess the latest should work. What's important is pinning so we'll have reproducible behaviour.

tasks.py
```diff
@@ -32,10 +32,12 @@
...
 build_path = os.path.join(os.path.dirname(os.path.abspath(__file__)), "build")
 kubectl_path = os.path.join(build_path, "kubectl")
 controller_gen_path = os.path.join(build_path, "bin", "controller-gen")
-kubectl_version = "v1.27.0"
+kustomize_path = os.path.join(build_path, "bin", "kustomize")
```
fedepaol, line 40: can you also use this in place where we currently call kustomize straight? IIUC should only be here https://github.com/metallb/metallb/blob/28b13e36bafc10f03557b76be83f6a8941136566/tasks.py#L324 (also, call fetch_kustomize on top of the function)

fedepaol (changes requested):

fedepaol:
Sorry for the delay. Left a couple of comments. Can you also sign the commit?

tasks.py
fulviodenza: Hi, I've added the fetch_kustomize on top of the function as requested, 
For what concerns the comment
> can you also use this in place where we currently call kustomize straight
I may be misunderstanding, but as it is now I should be already using that `kustomize_path`  when we call kustomize. Did you find any place where I'm missing it?

tasks.py
fulviodenza: sounds good, is there some specific documentation page i should update?

tasks.py
fedepaol: ah, apologies. I did a quick search and found https://github.com/metallb/metallb/blob/28b13e36bafc10f03557b76be83f6a8941136566/tasks.py#L335 where we use kubectl kustomize. Ignore my comment

fulviodenza: Actually I'm using it in my `fetch_kustomize` function

fedepaol: yep but the point of `fetch_kustomize` is to fetch it to use it :-)
If we use `kubectl kustomize` there is no point in even fetching it

fedepaol: so, let's use kustomize in https://github.com/metallb/metallb/blob/28b13e36bafc10f03557b76be83f6a8941136566/tasks.py#L335 
instead of kubectl kustomize.

Does it make sense?

fulviodenza:
I will fix the commit sign at the end so that I don't have to do it at any rebase

fedepaol:
Tried locally, getting the following error:

```
inv generatemanifests
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0Warning: Failed to open the file
Warning: /home/fpaoline/devel/upstream/metallb/build/bin/kustomize/kustomize_v5
Warning: .4.0_linux_amd64.tar.gz: No such file or directory
100     9  100     9    0     0     31      0 --:--:-- --:--:-- --:--:--    31
curl: (23) Failure writing output to destination
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
  0     0    0     0    0     0      0      0 --:--:--  0:00:01 --:--:--     0curl: (6) Could not resolve host: xzf
curl: (3) URL rejected: No host part in the URL
```

fulviodenza:
Can i ask you how you are testing it locally? @fedepaol

fedepaol:
> Can i ask you how you are testing it locally? @fedepaol

I just run `inv generatemanifests`. 
CI failed with the same error: https://github.com/metallb/metallb/actions/runs/8815367797/job/24363517861?pr=2366

PR #2384, opened 2024-04-22
Improve error message when keepalivetime is invalid

karampok:
/kind cleanup
/kind documentation

The current return error on the validation webhook looks like
```
Error from server (Forbidden): error when creating "peers.yaml": admission webhook "bgppeersvalidationwebhook.metallb.io" denied the request: parsing peer peer-with-kt-greater-than-ht: invalid keepaliveTime {"1m0s"}
```
which does not clue the real reason why that fails validation which is that keepalive time is greater than holdtime.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -394,7 +394,8 @@ func peerFromCR(p metallbv1beta2.BGPPeer, passwordSecrets map[string]corev1.Secr
 
 	// keepalive must be lower than holdtime
 	if keepaliveTime > holdTime {
-		return nil, fmt.Errorf("invalid keepaliveTime %q", p.Spec.KeepaliveTime)
+		return nil, fmt.Errorf("invalid keepaliveTime %q must be smaller than holdtime %q",
+			keepaliveTime, holdTime)
 	}
 
 	// Ideally we would set a default RouterID here, instead of having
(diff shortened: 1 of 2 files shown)


internal/config/config.go
```diff
@@ -394,7 +394,8 @@ func peerFromCR(p metallbv1beta2.BGPPeer, passwordSecrets map[string]corev1.Secr
 
 	// keepalive must be lower than holdtime
 	if keepaliveTime > holdTime {
-		return nil, fmt.Errorf("invalid keepaliveTime %q", p.Spec.KeepaliveTime)
+		return nil, fmt.Errorf("invalid keepaliveTime %q must be smaller than holdtime %q",
```
fedepaol, line 397: the commit says "add unit test", but the core part is the changed message.

Let's reword the commit, with the unit test being a part of what the commit does.

karampok:
Is there an e2e test which given a bgp peer CR, it checks the actual FRR config that has some property? (in this case the actual values of keepalivetime/holdtime)

fedepaol:
> Is there an e2e test which given a bgp peer CR, it checks the actual FRR config that has some property? (in this case the actual values of keepalivetime/holdtime)

https://github.com/metallb/metallb/blob/fc0636ac1cf423842f9c4f48c31d2b051281db83/e2etest/bgptests/bgp.go#L1259

karampok:
thanks, https://github.com/metallb/metallb/blob/main/e2etest/bgptests/bgp.go#L1334 seems it is well covered there

Issue #2368, opened 2024-04-15
Add graceful-restart support


fedepaol:
### Is your feature request related to a problem?

Whenever a version of metallb is updated, the speakers are rolled one at the time. This is causing the behaviour described here https://metallb.universe.tf/concepts/bgp/#limitations to be amplified for each node of the cluster.

Graceful restart should mitigate this problem.

### Describe the solution you'd like

Either a per - peer graceful restart option, or a global one. To be discussed.
We should validate that all the proper signals are sent to the frr container and thus to the bgpd daemon (and e2e this), and cover the same feature in frr-k8s too.

### I've read and agree with the following

- [X] I've checked all open and closed issues and my request is not there.
- [X] I've checked all open and closed pull requests and my request is not there.

karampok:
Hi @fedepaol I can work on that. Thanks

karampok:
Hi @aelbarkani, gracefull restart is helping to avoid happening any BGP change  when control plane FRR process is restarted (e.g. update image of the container), while the dataplane continues to work.

AFAICT graceful shutdown is when you want to stop for long the peering, and actually stop the traffic. But could you  expand? thanks

aelbarkani:
@karampok the limitation I had in mind is when you remove a node from the cluster for maintenance or just for down scaling. Shouldn't be there a sort of signaling to the routers saying that the current node is going to shutdown so the routers can stop routing the traffic to it gracefully ? I suppose BFD protocol mitigates the problem partially by detecting the link failure very fast, but maybe it's a better solution to do it gracefully. Also, wouldn't this mechanism avoid the limitation described in the docs https://metallb.universe.tf/concepts/bgp/#limitations ?
`The consequence of this is that any time the IP→Node mapping changes for your service, you should expect to see a one-time hit where most active connections to the service break. There’s no ongoing packet loss or blackholing, just a one-time clean break.`.

By the way, I'm no network expert, so don't hesitate to challenge what I'm saying :).

karampok:
[...]

PR #2386, opened 2024-04-26
Add graceful-restart support

karampok:
/kind feature

Issue covers #2368 

**Release note**:

```release-note
TODO
```


karampok:
Note to discuss:
```
Graceful restart configuration changed, reset this peer to take effect
% The Graceful Restart command used is not valid at this moment.
line 76: Failure to communicate[13] to bgpd, line:  neighbor 11.11.11.1 graceful-restart

[311|bfdd] done
[294|zebra] done
[300|bgpd] Configuration file[/etc/frr/frr.conf] processing failure: 13
2024-04-26 13:18:54,515 WARNING: frr-reload.py failed due to
vtysh (exec file) exited with status 13
Failed to fully apply configuration file 1 seconds
Caught SIGHUP and acquired lock! Reloading FRR..
Checking the configuration file syntax
```
it seems it fails, and retries to apply that the speaker/frr-reloader.

karampok:
The above might be related to https://github.com/FRRouting/frr/issues/8403

karampok:
In my test: I achieved zero downtime if I have graceful restart either in global `bgp graceful restarst` or in neighbor mode `neighbor x.y.z.k graceful restart` and having that in  `bgp graceful-restart preserve-fw-state`. 

(Test= in runtime I delete the speaker and see if the client loses connectivity)

The second command that enables the F -bit is only available in global mode, not available per neighbor. Therefore a solution which we add graceful-restart in neighbor level and f-bit in global (because is not safe to just have that there by default) seem not optimal.

@fedepaol How much do you object to have it global (one for enable gracefull restart/one to on/off f-bit), and also have the option to opt out per neighbor? Any other ideas?

karampok:
https://github.com/FRRouting/frr/issues/15880 created this issue upstream, even though we are not really blocked on that (well we will get an error but doe not look like it has an impact)

fedepaol:
> The second command that enables the F -bit is only available in global mode, not available per neighbor. Therefore a solution which we add graceful-restart in neighbor level and f-bit in global (because is not safe to just have that there by default) seem not optimal.
> 
> @fedepaol How much do you object to have it global (one for enable gracefull restart/one to on/off f-bit), and also have the option to opt out per neighbor? Any other ideas?

Won't the global one affect only peers with gr enabled? If so, doesn't sound too terrible to always enable it.

karampok:
My preference would be to be configurable to be in the safe side. I cannot find argument against always having it (or being default on) but no idea for which case this option exists.

api/v1beta2/bgppeer_types.go
```diff
@@ -85,6 +85,12 @@ type BGPPeerSpec struct {
 	// +optional
 	BFDProfile string `json:"bfdProfile,omitempty"`
 
+	// GracefulRestart allows BGP peer to continue to forward data packets along
+	// known routes while the routing protocol information is being restored.
+	// Needed for FRR mode only.
+	// +optional
+	GracefulRestart bool `json:"gracefulRestart,omitempty"`
```
fedepaol, line 93: nit: EnableGracefulRestart

api/v1beta2/bgppeer_types.go
fedepaol, line 1: nit on the commit subject: this commit does more than adding the field to the api, we are adding the support end to end (which is totally fine, but let's change the subject to "add graceful restart support"

e2etest/bgptests/bgp.go
```diff
@@ -427,6 +427,33 @@ var _ = ginkgo.Describe("BGP", func() {
...
+		ginkgo.By("configure peer")
+
+		resources := config.Resources{
+			Peers: metallb.WithGracefulRestart(metallb.PeersForContainers(FRRContainers, ipFamily)),
+		}
+
+		err := ConfigUpdater.Update(resources)
+		Expect(err).NotTo(HaveOccurred())
+
+		for _, c := range FRRContainers {
```
fedepaol, line 440: I think it'd make sense to have a real e2e test where we shut down a speaker and we check the traffic is still flowing.

First thing that comes to my mind is, focus on a node, exclude it from the daemonset selector, have a pod running only there with etp=local, ensure the traffic flows when the speaker is down

fedepaol: It's probably easier if we raise the graceful restart timers of the peered neighbors, so we'll have control

api/v1beta2/bgppeer_types.go
karampok: I can change that, sure

api/v1beta2/bgppeer_types.go
karampok: I have three commits and the e2e is added on its own commit, no?

e2etest/bgptests/bgp.go
fedepaol: Let's start with the default and see if it's enough then

api/v1beta2/bgppeer_types.go
karampok: done

api/v1beta2/bgppeer_types.go
karampok: done

api/v1beta2/bgppeer_types.go
```diff
@@ -85,6 +85,13 @@ type BGPPeerSpec struct {
 	// +optional
 	BFDProfile string `json:"bfdProfile,omitempty"`
 
+	// EnableGracefulRestart allows BGP peer to continue to forward data packets along
+	// known routes while the routing protocol information is being restored.
+	// Needed for FRR mode only.
+	// +optional
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="EnableGracefulRestart cannot be changed after creation, because GR requires rest BGP session"
```
oribon, line 92: nit: "Needed" should be more about "Supported" here?
also I'd put in the description the point about "this field is immutable because it requires restart of the BGP session" and keep the validation message "EnableGracefulRestart cannot be changed after creation"

internal/config/validation.go
```diff
@@ -41,6 +41,9 @@ func DiscardFRROnly(c ClusterResources) error {
 		if p.Spec.ConnectTime != nil {
 			return fmt.Errorf("peer %s has connect time set on native bgp mode", p.Spec.Address)
 		}
+		if p.Spec.EnableGracefulRestart {
+			return fmt.Errorf("peer %s has enabled GR (GracefulRestart) flag set on native bgp mode", p.Spec.Address)
```
oribon, line 45:
```suggestion
			return fmt.Errorf("peer %s has GracefulRestart flag set on native bgp mode", p.Spec.Address)
```

api/v1beta2/bgppeer_types.go
karampok: sounds good, done

e2etest/bgptests/bgp.go
```diff
@@ -147,6 +165,65 @@ var _ = ginkgo.Describe("BGP", func() {
...
+							err := validateServiceNoWait(svc, allNodes.Items, cc)
+							if err != nil && !errors.Is(err, ErrStaleRoute) {
+								return err
+							}
+						}
+					}
+				})
+			}
+
+			err = metallb.RestartSpeakerPods(cancel, cs)
```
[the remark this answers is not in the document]
karampok: What do you mean "are restarted". Are you suggesting  blocking in a function which will check speakers are restarted and then continue to an eventuallly gingko body that loops over all external frr docker containers doing in serial (validateServiceNoWait)?

e2etest/bgptests/bgp.go
```diff
@@ -98,9 +101,25 @@ var _ = ginkgo.Describe("BGP", func() {
 	})
 
 	ginkgo.BeforeEach(func() {
+
+		clientconfig := k8sclient.RestConfig()
+		var err error
```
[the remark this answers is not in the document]
karampok: I had tod that because  otherwise it fails like https://github.com/metallb/metallb/actions/runs/9267228538/job/25493436433 (a wip commit that I reverse)

e2etest/bgptests/bgp.go
```diff
@@ -147,6 +165,65 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			testservice.ValidateDesiredLB(svc)
+
+			ctx, cancel := context.WithTimeout(context.Background(), 120*time.Second)
+			defer cancel()
+			eg, ctx := errgroup.WithContext(ctx)
+			for _, c := range FRRContainers {
+				validateService(svc, allNodes.Items, c)
+				cc := c // https://go.dev/doc/faq#closures_and_goroutines
```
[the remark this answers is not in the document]
karampok: done (probably not needed anymore though or will be done)

e2etest/pkg/metallb/metallb.go
```diff
@@ -32,6 +32,38 @@ func init() {
...
+
+	f := func(context.Context) (bool, error) {
+		pods, err := SpeakerPods(cs)
+		if err != nil {
+			return false, err
+		}
+
+		ret := true
+		for _, p := range pods {
+			ret = ret && k8s.PodIsReady(p)
```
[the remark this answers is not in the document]
karampok: done

e2etest/pkg/metallb/metallb.go
```diff
@@ -32,6 +32,38 @@ func init() {
...
+	}
+	for _, p := range pods {
+		err := cs.CoreV1().Pods(Namespace).Delete(context.Background(), p.Name, metav1.DeleteOptions{})
+		if err != nil {
+			return err
+		}
+	}
+
+	f := func(context.Context) (bool, error) {
+		pods, err := SpeakerPods(cs)
```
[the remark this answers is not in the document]
karampok: TODO(i time.Sleep(5) for the time being)

e2etest/pkg/metallb/metallb.go
```diff
@@ -32,6 +32,38 @@ func init() {
 	}
 }
 
+func RestartSpeakerPods(cancel context.CancelFunc, cs clientset.Interface) error {
```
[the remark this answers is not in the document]
karampok: made it non blocksing as we discussed

e2etest/pkg/metallb/metallb.go
```diff
@@ -32,6 +32,38 @@ func init() {
...
+		return err
+	}
+	for _, p := range pods {
+		err := cs.CoreV1().Pods(Namespace).Delete(context.Background(), p.Name, metav1.DeleteOptions{})
+		if err != nil {
+			return err
+		}
+	}
+
+	f := func(context.Context) (bool, error) {
```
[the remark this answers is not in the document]
karampok: done, but is there a drawback to have it assigned to variable?

e2etest/bgptests/bgp.go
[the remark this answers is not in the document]
karampok: refactored

e2etest/bgptests/bgp.go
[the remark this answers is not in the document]
karampok: There is some state that changes because I delete the speaker containers, so it must that logic go per test not per suite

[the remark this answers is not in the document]
fedepaol: ah, I saw the reason. The FRR provider holds a reference to the speaker / frr-k8s pods.

This is a bit confusing though. FRRProvider is global, we initialize it two different times and we use it also in other files. I have the feeling this might bit us in the future (for example if we add a new context in a new file). 
How about giving the FRRProvider a Refresh() method that updates the speaker? So we re-load the proper speakers after any test that restarts the speakers?

Either that, or we fetch the speakers every time we call Execute. Shouldn't be terrible either as it fetches them from the local cache.

e2etest/bgptests/bgp.go
```diff
@@ -147,6 +164,134 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			//				}
+			//			}
+			//		}
+			//	}
+			// }()
+
+			Expect(metallb.RestartSpeakerPodsNoWait(cs)).NotTo(HaveOccurred())
+
+			Eventually(func() error {
+				c := func() error {
```
fedepaol, line 251: why do you need to assign to a variable? Can't you just iterate over the containers and assert from there?

e2etest/bgptests/bgp.go
```diff
@@ -147,6 +164,134 @@ var _ = ginkgo.Describe("BGP", func() {
...
+						err := validateServiceNoWait(svc, allNodes.Items, c)
+						if err != nil {
+							return err
+						}
+					}
+					return nil
+				}
+
+				if err := c(); err != nil {
+					if !errors.Is(err, ErrStaleRoute) {
```
fedepaol, line 262: why is staleroute skipped?

e2etest/bgptests/bgp.go
```diff
@@ -147,6 +164,134 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			//			if err := checkSVC(2 * time.Second); err != nil {
+			//				if !errors.Is(err, ErrStaleRoute) {
+			//					monitoringError = errors.Join(monitoringError, err)
+			//				}
+			//			}
+			//		}
+			//	}
+			// }()
+
+			Expect(metallb.RestartSpeakerPodsNoWait(cs)).NotTo(HaveOccurred())
```
fedepaol, line 248: I'd stick with the err := and the assert on the next line pattern.

e2etest/bgptests/bgp.go
karampok: I find easier to read but not strong opinion, I will change it

e2etest/pkg/metallb/metallb.go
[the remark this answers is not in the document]
karampok: done

e2etest/bgptests/bgp.go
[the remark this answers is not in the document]
karampok: I could go either, maybe the second seems simpler. Do you mean changing here https://github.com/metallb/metallb/blob/95274d94dffe0baed35330ebbbbe3f82282a19d1/e2etest/pkg/frr/provider/provider.go#L49 ? Can you expand?

e2etest/bgptests/bgp.go
karampok: done

e2etest/bgptests/bgp.go
karampok: done

e2etest/bgptests/bgp.go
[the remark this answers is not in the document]
karampok: done

karampok:
@fedepaol I have completed the last commit with adding the e2e test (plus the reverse to see a failure without graceful restart). I think I addressed all your comments. Let me know if I miss something. Thanks

e2etest/pkg/metallb/metallb.go
```diff
@@ -32,6 +33,35 @@ func init() {
 	}
 }
 
+func RestartSpeakerPodsNoWait(cs clientset.Interface) error {
```
fedepaol, line 36: Let's rename this, as it waits :)

e2etest/bgptests/bgp.go
```diff
@@ -147,6 +147,115 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			}, 2*time.Minute, time.Second).ShouldNot(HaveOccurred(), "no downtime until speakers are ready")
+
+			for _, c := range FRRContainers {
+				validateService(svc, allNodes.Items, c)
+			}
+		}
+
+		ginkgo.Context("and when GR enabled", func() {
+			ginkgo.BeforeEach(func() {
+				grEnabled = true
```
fedepaol, line 227: let's have this a member of the assertDuringSpeakerRestart function (shouldFail, for example)

e2etest/bgptests/bgp.go
```diff
@@ -147,6 +147,115 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			}
+			err = metallb.RestartSpeakerPodsNoWait(cs)
+			Expect(err).NotTo(HaveOccurred())
+
+			if !grEnabled {
+				Eventually(func() error {
+					for _, c := range FRRContainers {
+						err := validateServiceNoWait(svc, allNodes.Items, c)
+						if err != nil {
+							if errors.Is(err, ErrStaleRoute) {
```
fedepaol, line 183: let's avoid extra nesting and assert stale route first

e2etest/bgptests/bgp.go
```diff
@@ -147,6 +147,115 @@ var _ = ginkgo.Describe("BGP", func() {
...
+					return nil
+				}, 2*time.Minute, time.Second).Should(HaveOccurred(), "a downtime should be observed")
+				return
+			}
+
+			Eventually(func() error {
+				for _, c := range FRRContainers {
+					err := validateServiceNoWait(svc, allNodes.Items, c)
+					if err != nil {
+						if errors.Is(err, ErrStaleRoute) {
```
fedepaol, line 199: same

e2etest/bgptests/bgp.go
```diff
@@ -147,6 +147,115 @@ var _ = ginkgo.Describe("BGP", func() {
...
+							continue // when GR, is normal to observe stale routes
+						}
+						Expect(err).NotTo(HaveOccurred(), "downtime was observed")
```
fedepaol, line 202: this can be outside of the if

e2etest/bgptests/bgp.go
karampok: done

e2etest/bgptests/bgp.go
karampok: not done, can you expand what you mean?

e2etest/pkg/metallb/metallb.go
karampok: done

e2etest/bgptests/bgp.go
karampok: done

e2etest/bgptests/bgp.go
fedepaol: instead of having a global that share some state, I'd make assertDuringSpeakerRestart pure with that regard and allow it to accept an extra parameter which is that bool. The same bool can then be a parameter of the table tests (bonus point, use a constant with the proper name instead of a bool).

karampok: Currently I have two tables, exactly the same with the diff the beforeeach section. I think the way that the Describe scope have a "global" state, and that in the before each you define it (as you do with the frr config. gr to true) is more readable to understand the test behavior and fits better ginkgo approach. The description looks like 

```
BGP GracefulRestart, when speakers restart and when GR enabled dataplane should keep working (entry x,y,z)
BGP GracefulRestart, when speakers restart and when GR disabled dataplane should have a downtime (entry x,y,z)
```

If I add as extra to the function, then you suggest to actually merge the test into single context/table (right?) and have the  bool name hint  what we test inside the entry  args instead of the ginkgo description. tbh I would prefer to keep it that way but I can change it if you want.

karampok:
Another version pushed that does make the function not depend on external variable

e2etest/bgptests/bgp.go
```diff
@@ -147,6 +147,117 @@ var _ = ginkgo.Describe("BGP", func() {
...
+
+			for _, c := range FRRContainers {
+				validateService(svc, allNodes.Items, c)
+			}
+		}
+
+		ginkgo.Context("and when GR enabled", func() {
+
+			assertDuringSpeakerRestartWithGR := func(pairingIPFamily ipfamily.Family, poolAddresses []string, tweak testservice.Tweak) {
+				assertDuringSpeakerRestart(true, pairingIPFamily, poolAddresses, tweak)
```
fedepaol, line 224: nit: let's define a constant "EnableGracefulRestart = true" and let's use that instead of true / false (also when checking)

karampok: done

e2etest/pkg/wget/wget.go
```diff
@@ -25,7 +25,7 @@ func Do(address string, exc executor.Executor) error {
 
 	// Retry loop to handle wget NetworkFailure errors
 	for {
-		out, err = exc.Exec("wget", "-O-", "-q", address, "-T", "60")
+		out, err = exc.Exec("wget", "-O-", "-q", address, "-T", "5")
```
fedepaol, line 28: is this change related?

e2etest/pkg/frr/provider/provider.go
```diff
@@ -43,22 +47,28 @@ func NewFRRMode(r *rest.Config) (Provider, error) {
...
 
-	return frrModeProvider{speakers: speakers}, nil
-}
-
-func (f frrModeProvider) FRRExecutorFor(ns, name string) (executor.Executor, error) {
-	_, ok := f.speakers[name]
+	_, ok := speakers[name]
 	if !ok {
-		return nil, fmt.Errorf("speakers %s/%s not found in known speakers %v", ns, name, f.speakers)
+		return nil, fmt.Errorf("speakers %s/%s not found in known speakers %v", ns, name, speakers)
```
fedepaol, line 52: nit: speaker %s not found in..

e2etest/pkg/jigservice/jig.go
```diff
@@ -415,7 +415,7 @@ func (j *TestJig) UpdateService(ctx context.Context, update func(*v1.Service)) (
 
 // WaitForLoadBalancer waits the given service to have a LoadBalancer, or returns an error after the given timeout
 func (j *TestJig) WaitForLoadBalancer(ctx context.Context, timeout time.Duration) (*v1.Service, error) {
-	ginkgo.GinkgoWriter.Printf("Waiting up to %v for service %q to have a LoadBalancer", timeout, j.Name)
+	ginkgo.GinkgoWriter.Printf("\tWaiting up to %v for service %q to have a LoadBalancer\n", timeout, j.Name)
```
fedepaol, line 418: is this related? If not but makes sense, please add it in a separte commit

e2etest/pkg/metallb/metallb.go
```diff
@@ -32,6 +33,35 @@ func init() {
...
+			return false, err
+		}
+		for _, p := range npods {
+			if slices.Contains(oldNames, p.Name) {
+				return false, nil
+			}
+		}
+		return true, nil
+	})
+
```
fedepaol, line 62: nit: unnecessary line

fedepaol:
few nits, almost there!

karampok (commented):
What about now :) let me know if you want to rebase. Thanks

fedepaol:
> What about now :) let me know if you want to rebase. Thanks

rebase is not necessary as we use merge queues.
LGTM, sending to merge queue, thanks!

PR #2408, opened 2024-05-21
E2E: use the local kubectl instead of the one on the host

fedepaol:

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design
/kind flake
> /kind failing
> /kind documentation
> /kind regression

Using the kubectl version on the host instead of a pinned one may result in flakiness / sudden unexpected behaviour.
Specifically, now we have a set of flakes like the one found in https://github.com/metallb/metallb/pull/2406

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etest/pkg/executor/executor.go
+++ b/e2etest/pkg/executor/executor.go
@@ -3,9 +3,12 @@
 package executor
 
 import (
+	"errors"
 	"os"
 	"os/exec"
 )
+
+var Kubectl string
 
 type Executor interface {
 	Exec(cmd string, args ...string) (string, error)
@@ -58,7 +61,10 @@ func ForPod(namespace, name, container string) Executor {
 }
 
 func (p *podExecutor) Exec(cmd string, args ...string) (string, error) {
+	if Kubectl == "" {
+		return "", errors.New("the kubectl parameter is not set")
+	}
 	fullargs := append([]string{"exec", p.name, "-n", p.namespace, "-c", p.container, "--", cmd}, args...)
-	out, err := exec.Command("kubectl", fullargs...).CombinedOutput()
+	out, err := exec.Command(Kubectl, fullargs...).CombinedOutput()
 	return string(out), err
 }
--- a/tasks.py
+++ b/tasks.py
@@ -930,10 +930,10 @@ def e2etest(ctx, name="kind", export=None, kubeconfig=None, system_namespaces="k
     if external_frr_image != "":
         external_frr_image = "--frr-image=" + (external_frr_image)
     testrun = run("cd `git rev-parse --show-toplevel`/e2etest &&"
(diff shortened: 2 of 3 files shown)


e2etest/pkg/executor/executor.go
```diff
@@ -58,7 +61,10 @@ func ForPod(namespace, name, container string) Executor {
 }
 
 func (p *podExecutor) Exec(cmd string, args ...string) (string, error) {
+	if Kubectl == "" {
+		return "", errors.New("the kubectl parameter is not set")
+	}
 	fullargs := append([]string{"exec", p.name, "-n", p.namespace, "-c", p.container, "--", cmd}, args...)
-	out, err := exec.Command("kubectl", fullargs...).CombinedOutput()
+	out, err := exec.Command(Kubectl, fullargs...).CombinedOutput()
```
karampok, line 68: no changes asked, just question: having   exec in k8s's pod by using go client would that also also an option?

tasks.py
```diff
@@ -930,10 +930,10 @@ def e2etest(ctx, name="kind", export=None, kubeconfig=None, system_namespaces="k
     if external_frr_image != "":
         external_frr_image = "--frr-image=" + (external_frr_image)
     testrun = run("cd `git rev-parse --show-toplevel`/e2etest &&"
-                  "KUBECONFIG={} ginkgo {} --junit-report={} --timeout=3h {} {} -- --kubeconfig={} --service-pod-port={} -ipv4-service-range={} -ipv6-service-range={} {} --report-path {} {} -node-nics {} -local-nics {} {} -bgp-mode={}  -with-vrf={} {} --host-bgp-mode={}".format(
+                  "KUBECONFIG={} ginkgo {} --junit-report={} --timeout=3h {} {} -- --kubeconfig={} --service-pod-port={} -ipv4-service-range={} -ipv6-service-range={} {} --report-path {} {} -node-nics {} -local-nics {} {} -bgp-mode={} -with-vrf={} {} --host-bgp-mode={} --kubectl={}".format(
         kubeconfig, ginkgo_params, junit_report, ginkgo_focus, ginkgo_skip, kubeconfig, service_pod_port, ipv4_service_range,
         ipv6_service_range, opt_skip_docker, report_path, prometheus_namespace, node_nics, local_nics,
-        external_containers, bgp_mode, with_vrf, external_frr_image, host_bgp_mode), warn="True")
+        external_containers, bgp_mode, with_vrf, external_frr_image, host_bgp_mode, kubectl_path), warn="True")
```
karampok, line 936: where do we actually put a a value here?

e2etest/pkg/executor/executor.go
fedepaol: yep, but the previous implementation was using kubectl, and that avoids bringing in extra dependencies

tasks.py
karampok: nevermind we already had that path in the file

fedepaol: not sure I understood what do you mean. `kubectl_path` is a variable that holds the path. It's gonna land in the --kubectl parameter above.

PR #2418, opened 2024-06-03
Use distroless image to build controller and speaker containers

minmzzhang:
Inspired by https://github.com/metallb/metallb/pull/2117
fixes https://github.com/metallb/metallb/issues/2110

A big thank you to @oribon and @georgettica for the detail discussion in #2117 which is **the** design document I followed to this solution. Also appreciate @oribon 's help during design implementation, that I was able to run local integration and e2e tests! 

/kind feature

gcr.io/distroless/static base image is used to replace the original docker.io/alpine image, for a distroless solution that controller and speaker images will be built from scratch. Since distroless base image has no shell entry, there are two major impacts with this solution:

1. Some of the frr init containers which need to move files around will need a local copy tool. 
2. Some of the e2e tests which need to execute certain commands (e.g. "ifconfig") from inside the speaker will need to have an ephemeral container attached to the pod to complete the task. 

In this PR, a local copy tool (cp-tool) was introduced to cope with the first concern, and a podDebugExecutor was introduced under e2etest to attach a debug ephemeral container when there is a need to execute additional commands.

**Release note**:

```release-note
The speaker and the controller containers are now based on a distroless image, reducing the attack surface
```

### DIFF
--- a/e2etest/l2tests/interface_selector.go
+++ b/e2etest/l2tests/interface_selector.go
@@ -25,6 +25,8 @@ import (
 	"go.universe.tf/e2etest/pkg/status"
 )
 
+const agnostImage = "registry.k8s.io/e2e-test-images/agnhost:2.45"
+
 var (
 	NodeNics  []string
 	LocalNics []string
@@ -170,7 +172,7 @@ var _ = ginkgo.Describe("L2-interface selector", func() {
 			}, 1*time.Minute, 1*time.Second).Should(Not(HaveOccurred()))
 			speakerPod, err := metallb.SpeakerPodInNode(cs, svcNode.Name)
 			Expect(err).NotTo(HaveOccurred())
-			selectorMac, err := mac.GetIfaceMac(NodeNics[0], executor.ForPod(speakerPod.Namespace, speakerPod.Name, "speaker"))
+			selectorMac, err := mac.GetIfaceMac(NodeNics[0], executor.ForPodDebug(speakerPod.Namespace, speakerPod.Name, "speaker", agnostImage))
 			Expect(err).NotTo(HaveOccurred())
 
 			ingressIP := jigservice.GetIngressPoint(&svc.Status.LoadBalancer.Ingress[0])
--- a/e2etest/pkg/executor/executor.go
+++ b/e2etest/pkg/executor/executor.go
@@ -6,6 +6,8 @@ import (
 	"errors"
 	"os"
 	"os/exec"
+
+	utilrand "k8s.io/apimachinery/pkg/util/rand"
 )
 
 var Kubectl string
@@ -65,6 +67,37 @@ func (p *podExecutor) Exec(cmd string, args ...string) (string, error) {
 		return "", errors.New("the kubectl parameter is not set")
 	}
 	fullargs := append([]string{"exec", p.name, "-n", p.namespace, "-c", p.container, "--", cmd}, args...)
+	out, err := exec.Command(Kubectl, fullargs...).CombinedOutput()
+	return string(out), err
+}
+
+// add ephemeral container to deal with distroless image
+type podDebugExecutor struct {
+	namespace string
+	name      string
+	container string
+	image     string
+}
+
+func ForPodDebug(namespace, name, container, image string) Executor {
+	return &podDebugExecutor{
+		namespace: namespace,
+		name:      name,
+		container: container,
+		image:     image,
+	}
+}
+
+func (pd *podDebugExecutor) Exec(cmd string, args ...string) (string, error) {
+	if Kubectl == "" {
+		return "", errors.New("the kubectl parameter is not set")
+	}
+
+	imageArg := "--image=" + pd.image
+	targetArg := "--target=" + pd.container
+	debuggerArg := pd.container + "-debugger-" + utilrand.String(5)
+
+	fullargs := append([]string{"debug", "-it", "-n", pd.namespace, "-c", debuggerArg, targetArg, imageArg, pd.name, "--", cmd}, args...)
 	out, err := exec.Command(Kubectl, fullargs...).CombinedOutput()
 	return string(out), err
 }
--- a/frr-tools/cp-tool/cp-tool.go
+++ b/frr-tools/cp-tool/cp-tool.go
@@ -1,0 +1,63 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package main
+
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+)
+
+func CopyFile(src, dst string) error {
+	srcFile := filepath.Clean(src)
+	destFile := filepath.Clean(dst)
+	source, err := os.Open(srcFile)
+	if err != nil {
+		err = fmt.Errorf("failed to open source file %s: %w", srcFile, err)
+		return err
+	}
+	defer source.Close()
+
+	srcStat, err := os.Stat(srcFile)
+	if err != nil {
+		err = fmt.Errorf("failed to check stats for %s: %w", srcFile, err)
+		return err
+	}
+
+	destination, err := os.Create(destFile)
+	if err != nil {
+		err = fmt.Errorf("failed to create destination file: %s: %w", destFile, err)
+		return err
+	}
+	defer destination.Close()
+
+	_, err = io.Copy(destination, source)
+	if err != nil {
+		err = fmt.Errorf("failed to copy %s to %s: %w", srcFile, destFile, err)
+		return err
+	}
+
+	err = os.Chmod(destFile, srcStat.Mode())
+	if err != nil {
+		err = fmt.Errorf("failed to apply permission on %s", destFile)
+		return err
+	}
+
+	return nil
+}
+
+func main() {
+	if len(os.Args) != 3 {
+		fmt.Printf("Please provide two command line arguments.")
+		return
+	}
+
+	sourceFile, destinationFile := os.Args[1], os.Args[2]
+
+	err := CopyFile(sourceFile, destinationFile)
+	if err != nil {
+		fmt.Println(err)
+		return
+	}
+}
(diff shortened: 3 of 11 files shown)


e2etest/pkg/executor/executor.go
```diff
@@ -68,3 +68,25 @@ func (p *podExecutor) Exec(cmd string, args ...string) (string, error) {
...
+		namespace: namespace,
+		name:      name,
+	}
+}
+
+func (pd *podDebugExecutor) Exec(cmd string, args ...string) (string, error) {
+	if Kubectl == "" {
+		return "", errors.New("the kubectl parameter is not set")
+	}
+	fullargs := append([]string{"debug", "-it", "-n", pd.namespace, pd.name, "-c", "pod-executor-debugger", "--image=busybox", "--", cmd}, args...)
```
oribon, line 89: I think we should randomize the container name here `pod-executor-debugger-..`, the reason being that if you run the same debug container multiple times it only remembers the first output (the first debug command that caused the container to be terminated) which could cause confusion if the test is ran multiple times on the same cluster

e2etest/pkg/executor/executor.go
```diff
@@ -68,3 +68,25 @@ func (p *podExecutor) Exec(cmd string, args ...string) (string, error) {
 	out, err := exec.Command(Kubectl, fullargs...).CombinedOutput()
 	return string(out), err
 }
+
+// add ephemeral container to deal with distroless image
+type podDebugExecutor struct {
+	namespace string
+	name      string
```
oribon, line 75: wdyt about addibg an image for the container here (instead of defaulting to busybox)? this will make the call for `ForPodDebug` more resemblant to the usual `kubectl debug` where the caller specifies the image.

frr-tools/cp-tool/cp-tool.go
```diff
@@ -0,0 +1,37 @@
...
+	if len(os.Args) != 3 {
+		level.Error(logger).Log("Please provide two command line arguments.")
+		return
+	}
+
+	sourceFile, destinationFile := os.Args[1], os.Args[2]
+	sourceFile = filepath.Clean(sourceFile)
+	destinationFile = filepath.Clean(destinationFile)
+
+	err = cp.Copy(sourceFile, destinationFile)
```
oribon, line 32: did you try something that doesn't require importing another dependency? I wonder if we can just do this with some `io.Copy` (or anything else from the standard library) that wouldn't be too messy since our usecase is pretty simple

frr-tools/cp-tool/cp-tool.go
```diff
@@ -0,0 +1,37 @@
...
+	"path/filepath"
+
+	cp "github.com/otiai10/copy"
+
+	"github.com/go-kit/log/level"
+	"go.universe.tf/metallb/internal/logging"
+)
+
+func main() {
+	logger, err := logging.Init("error")
```
oribon, line 17: we don't really need a logger here, let's just panic / fmt.Printf on the few occasions where needed

oribon (commented):
awesome work! thanks a lot
can you also squash your commits into one with the gist of the change?(https://metallb.io/community/#commit-messages)

e2etest/pkg/executor/executor.go
minmzzhang: added the randomized string for debugger.

e2etest/pkg/executor/executor.go
minmzzhang: A randomized string (length=5) was added to the pod executor debugger.

e2etest/pkg/executor/executor.go
minmzzhang: image name is configurable now.

frr-tools/cp-tool/cp-tool.go
minmzzhang: Originally there was file permission issue for the local copy tool when using io.Copy, so I switched to a community package. With the new submission, I've added the preserve permission functionality, so it should be clean now.

frr-tools/cp-tool/cp-tool.go
minmzzhang: logger removed.

frr-tools/cp-tool/cp-tool.go
```diff
@@ -0,0 +1,77 @@
...
+
+func main() {
+	if len(os.Args) != 3 {
+		fmt.Printf("Please provide two command line arguments.")
+		return
+	}
+
+	sourceFile, destinationFile := os.Args[1], os.Args[2]
+	sourceFile = filepath.Clean(sourceFile)
+	destinationFile = filepath.Clean(destinationFile)
```
oribon, line 70: small nit: this is probably redundant with the one happening inside the function

frr-tools/cp-tool/cp-tool.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"fmt"
+	"io"
+	"os"
+	"path/filepath"
+)
+
+func CopyFile(srcFile string, destFile string) error {
+	srcFile = filepath.Clean(srcFile)
+	destFile = filepath.Clean(destFile)
```
oribon, line 14: nit: rename the input to src,dst (so the new vars are declared with new names)

frr-tools/cp-tool/cp-tool.go
```diff
@@ -0,0 +1,77 @@
...
+	defer src.Close()
+
+	srcStat, err := os.Stat(srcFile)
+	if err != nil {
+		err = fmt.Errorf("failed to check stats for %s: %w", srcFile, err)
+		return err
+	}
+
+	srcPerm := srcStat.Mode().Perm()
+	fmt.Printf("File permission before copying %v \n", srcPerm)
```
oribon, line 29: nit: I think we can remove it (unless you think this adds value)

frr-tools/cp-tool/cp-tool.go
```diff
@@ -0,0 +1,77 @@
...
+	}
+
+	dstStat, err := os.Stat(destFile)
+	if err != nil {
+		err = fmt.Errorf("failed to check stats for %s: %w", destFile, err)
+		return err
+	}
+
+	dstPerm := dstStat.Mode().Perm()
+	fmt.Printf("File permission After copying %v \n", dstPerm)
```
oribon, line 57: same?

oribon (approved):
left a few small comments, looks really good!
also, please change the commit message to be more about "what is the change" and not "how we changed" :) 
approved, cc @fedepaol to also give this a look

frr-tools/cp-tool/cp-tool.go
minmzzhang: removed the redundant part.

frr-tools/cp-tool/cp-tool.go
minmzzhang: renamed the inputs

e2etest/pkg/executor/executor.go
```diff
@@ -68,3 +70,34 @@ func (p *podExecutor) Exec(cmd string, args ...string) (string, error) {
...
+		image:     image,
+	}
+}
+
+func (pd *podDebugExecutor) Exec(cmd string, args ...string) (string, error) {
+	if Kubectl == "" {
+		return "", errors.New("the kubectl parameter is not set")
+	}
+
+	imagearg := "--image=" + pd.image
```
fedepaol, line 96: nit: imageArg, targetArg, debuggerArg

e2etest/l2tests/interface_selector.go
```diff
@@ -170,7 +170,7 @@ var _ = ginkgo.Describe("L2-interface selector", func() {
 			}, 1*time.Minute, 1*time.Second).Should(Not(HaveOccurred()))
 			speakerPod, err := metallb.SpeakerPodInNode(cs, svcNode.Name)
 			Expect(err).NotTo(HaveOccurred())
-			selectorMac, err := mac.GetIfaceMac(NodeNics[0], executor.ForPod(speakerPod.Namespace, speakerPod.Name, "speaker"))
+			selectorMac, err := mac.GetIfaceMac(NodeNics[0], executor.ForPodDebug(speakerPod.Namespace, speakerPod.Name, "speaker", "busybox"))
```
fedepaol, line 173: the only concern I have here is relying on another image (from dockerhub, also), which may result in hitting the pull limits. How about using the agnhost image we are already using [here](https://github.com/metallb/metallb/blob/main/e2etest/pkg/jigservice/jig.go#L42) (assuming it already has all we need to debug)?

website/content/troubleshooting/_index.md
```diff
@@ -288,3 +288,11 @@ Additionally, the status of the service and of the endpoints must be provided:
 kubectl get endpointslices <my-service> -o yaml
 kubectl get svc <my_service> -o yaml
 ```
+
+### How to debug a container with distroless image
```
fedepaol, line 292:
```suggestion
### How to debug the speaker and the controller containers
```

website/content/troubleshooting/_index.md
```diff
@@ -288,3 +288,11 @@ Additionally, the status of the service and of the endpoints must be provided:
...
+
+With the distroless solution, if someone needs to debug inside speaker container, an ephemeral container should be used:
```
fedepaol, line 294:
```suggestion
Due to the fact that both the speaker and the controller containers are based on a distroless image, an ephemeral container should be used to debug inside them: 
```

fedepaol:
Also, I'd reword the release notes section, saying that now the speaker and the controller are based on distroless.

e2etest/pkg/executor/executor.go
minmzzhang: changed

e2etest/l2tests/interface_selector.go
minmzzhang: good idea! changed.

website/content/troubleshooting/_index.md
minmzzhang: modified.

website/content/troubleshooting/_index.md
minmzzhang: changed..

fedepaol:
LGTM

PR #2420, opened 2024-06-04
Charts: don't log history in liveness probe

freedge:
frr-metrics liveness probe runs a show daemons command repeatedly and append that at the end of a file. When the file is longer than 1k lines the history gets truncated which involves a few system calls like renames. When the system is under stress this rename call can take a few seconds, exceeding the timeout of the liveness probe.

When this happens repeatedly frr process gets restarted.

This was noticed as part of the investigation in
https://issues.redhat.com/browse/OCPBUGS-34761?focusedId=24850879&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-24850879

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

fix occasional frr crash

replaces https://github.com/metallb/metallb-operator/pull/473
would love to have that in OCP 4.14

**Release note**:

```release-note
NONE
```

### DIFF
--- a/charts/metallb/templates/speaker.yaml
+++ b/charts/metallb/templates/speaker.yaml
@@ -443,6 +443,9 @@ spec:
           {{- if .Values.prometheus.secureMetricsPort }}
           - --host=localhost
           {{- end }}
+        env:
+          - name: VTYSH_HISTFILE
+            value: /dev/null
         ports:
           - containerPort: {{ .Values.speaker.frr.metricsPort }}
             name: monitoring
--- a/config/frr/speaker-patch.yaml
+++ b/config/frr/speaker-patch.yaml
@@ -105,6 +105,9 @@ spec:
           command: ["/etc/frr_metrics/frr-metrics"]
           args:
             - --metrics-port=7473
+          env:
+            - name: VTYSH_HISTFILE
+              value: /dev/null
           ports:
             - containerPort: 7473
               name: monitoring
--- a/config/manifests/metallb-frr-prometheus.yaml
+++ b/config/manifests/metallb-frr-prometheus.yaml
@@ -2054,6 +2054,9 @@ spec:
         - --metrics-port=7473
         command:
         - /etc/frr_metrics/frr-metrics
+        env:
+        - name: VTYSH_HISTFILE
+          value: /dev/null
         image: quay.io/frrouting/frr:9.0.2
         name: frr-metrics
         ports:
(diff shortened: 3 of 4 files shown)


fedepaol:
Hi @freedge , I tried this locally and I can see the log being filled, even with the env being set.

The only way I managed to disable it was via the -H parameter here https://github.com/metallb/metallb/blob/main/frr-tools/metrics/vtysh/vtysh.go#L14

freedge:
I tried it too, manually on an openshift installation, providing
```
        env:
        - name: VTYSH_HISTFILE
          value: /dev/null
```
in the daemonset for the frr-metrics container, and it works

fedepaol:
> I tried it too, manually on an openshift installation, providing
> 
> ```
>         env:
>         - name: VTYSH_HISTFILE
>           value: /dev/null
> ```
> 
> in the daemonset for the frr-metrics container, and it works

weird, let me try it again then :-) 
It could also be related to the different frr version we have upstream

freedge:
the man page says
https://man.cx/vtysh(1)

```
VTYSH_HISTFILE

Override the history file for vtysh commands. Logging can be turned off using VTYSH_HISTFILE=/dev/null vtysh. Environment is preferred way to override the history file path over command line argument (-H/--histfile).
```

I tried with frr-8.3.1-5.el9_2.4.x86_64 and locally I run frr-8.5.3-4.el9.x86_64

fedepaol:
Yep, I saw the man page.
This is what I have locally:

```
k exec -n metallb-system speaker-h4tgn -c frr-metrics env | grep VTYSH
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
VTYSH_HISTFILE=/dev/null
```

and 

```
 k exec -n metallb-system speaker-h4tgn -c frr-metrics cat /root/.history_frr | wc -l
kubectl exec [POD] [COMMAND] is DEPRECATED and will be removed in a future version. Use kubectl exec [POD] -- [COMMAND] instead.
207
```

Will try on OCP

freedge:
I'm not familiar enough, should
 config/manifests/metallb-frr.yaml
be changed as well?

fedepaol:
> ~I'm not familiar enough, should config/manifests/metallb-frr.yaml be changed as well?~ nevermind `invoke generatemanifests` takes care of this

yep exactly. And we have CI all over the place to ensure the generated bits won't drift from the base :sunglasses:

fedepaol:
@freedge mind generating the manifests with `inv generatemanifests` ? I see CI is failing here

fedepaol:
Also, please change the commit body as now it's not charts only anymore

fedepaol:
sending to merge queue, LGTM

PR #2437, opened 2024-06-20
update the location of the metrics exporter

emmanuel-ferdman:

/kind documentation

PR #1732 moved the location of the metrics exporter. This PR adjusts sources to changes.

**Release note**:

```release-note
update the location of the metrics exporter
```


fedepaol:
Sorry for the delay.
I'd keep the doc as it is. It was listing the location at the time of the proposal, don't think we need to update it.

PR #2442, opened 2024-06-26
bug: fix ip pool total count caculate failed

changluyi:
BUG
>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

when ippool address contains a x.x.x.0/32  or  x.x.x.255/32 IP  , the metric of metallb_allocator_addresses_total will be -1

because the firstIP is equal lastIP, it should only count once, but actually twice.

**Release note**:

```release-note

```


fedepaol:
Thanks @changluyi , mind extending the commit reason on why / what this commit is fixing so it will be clear in the future?

PR #2444, opened 2024-07-02
Support running frr-k8s in a separate namespace

fedepaol:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
Add helm / configuration parameters to allow MetalLB to consume an frr-k8s instance running in a separate namespace.
```


speaker/bgp_controller.go
```diff
@@ -265,6 +275,30 @@ func (c *bgpController) syncPeers(l log.Logger) error {
...
+		return plainTextPassword, v1.SecretReference{}
+	case bgpFrrK8s:
+		// in case of passthrough, we don't propagate the converted password,
+		// so it's either the password as plain text or the secret
+		if secret == SecretPassThrough {
+			return cfg.Password, cfg.PasswordRef
+		}
+		return plainTextPassword, v1.SecretReference{}
+	}
+	return "", v1.SecretReference{}
```
oribon, line 299: imo it feels a bit odd to have this caring about the bgp implementation.
without the new `SecretPassword` the `Password` field acted as "plain text password" (either if it came from the crd directly or a secret)
would it work if we had this function being:
`if passthrough` => `return cfg.Password, cfg.PasswordRef`
`if convert` => `return cfg.Password, v1.SecretReference{}`
if so I think it is cleaner (letting the bgp implementation to decide)
I think this makes sense as the frr-k8s controller itself has the logic of when should it use the ref vs plaintext:
```go
			password := ""
			if reflect.DeepEqual(s.PasswordRef, corev1.SecretReference{}) {
				password = s.Password
			}
```
wdyt?

internal/config/config.go
```diff
@@ -439,9 +441,9 @@ func peerFromCR(p metallbv1beta2.BGPPeer, passwordSecrets map[string]corev1.Secr
 		return nil, fmt.Errorf("can not have both password and secret ref set in peer config %q/%q", p.Namespace, p.Name)
 	}
 
-	password := p.Spec.Password
+	secretPassword := ""
 	if p.Spec.PasswordSecret.Name != "" {
-		password, err = passwordFromSecretForPeer(p, passwordSecrets)
+		secretPassword, err = passwordFromSecretForPeer(p, passwordSecrets)
```
oribon, line 446: w/ the other comment, I think it's cleaner to keep the "Password" field as "plain text password", not having the distinction on where it came from.

.github/workflows/ci.yaml
```diff
@@ -205,13 +212,30 @@ jobs:
...
+          repository: metallb/frr-k8s
+          path: frr-k8s
+          ref: v0.0.11
+
+      - name: Deploy FRR-K8s
+        if: matrix.bgp-type == 'frr-k8s-external'
+        run: |
+          cd ${GITHUB_WORKSPACE}/frr-k8s
+          make docker-build
+          make deploy
```
oribon, line 228: can we deploy the all-in-one for that version instead of checkout+build+.. ?

speaker/bgp_controller.go
fedepaol: The choice here was to be explicit from above (ie when I call the bgp implementation) or doing some extra logic under the api surface.
I could've propagated the passthrough information, but then we would've been passing two password information without knowing (or, relying on) some hidden behavior down the carpet.

Propagating that to frr-k8s itself is even worse imo, because again, we fill two fields relying on the behavior implemented in a different process.

Here each implementation is doing what we are telling it to do from above, without doing extra hidden things like using only the password or the secret.

.github/workflows/ci.yaml
fedepaol: ah yes, that's much better. I stuck to what I had in my local dev env but this is way more straightforward!

fedepaol: I guess I can do everything from dev-env then

speaker/bgp_controller.go
oribon: I understand the rationale (not sure I like it better, but this is nothing critical). If we're going that way we can remove that code snippet I attached?
`Propagating that to frr-k8s itself is even worse imo`
I agree, I was talking about the metallb frr-k8s-controller (where the code snippet is from) 😅

fedepaol: ah lol! Sorry, I thought that was frr-k8s.

Yep, we can remove / log / panic if both are set down there.

oribon (approved):
lgtm (once ci stabilizes 😄)

PR #2454, opened 2024-07-12
Prevent session flapping when BFD is enabled and services are added / removed

fedepaol:

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

  The generated configuration file has the format

  `neighbor xxxx bfd profile zzz`

  which works, but causes session flapping whenever the configuration is
  reloaded, as the converted configuration contains two entries:

```
  neighbor xxxx bfd
  neighbor xxxx bfd profile zzz
```

  Reloading the configuration causes the first line to be removed:
```
  Lines To Delete
  ===============
  ip forwarding
  ipv6 forwarding
  router bgp 64520
   no neighbor 10.74.239.52 bfd
```
  which causes the session to flap. Here we generate also the

  `neighbor xxxx bfd`

  entry to avoid flaps.

**Release note**:

```release-note
Fix session flapping when adding / removing services and BFD is enabled.
```

### DIFF
--- a/e2etest/bgptests/bgp.go
+++ b/e2etest/bgptests/bgp.go
@@ -662,6 +662,33 @@ var _ = ginkgo.Describe("BGP", func() {
 				return nil
 			}, 4*time.Minute, 1*time.Second).ShouldNot(HaveOccurred())
 
+			ginkgo.By("checking the sessions don't flap when changing the configuration")
+
+			previousNeighbors := map[string]frr.NeighborsMap{}
+			for _, c := range FRRContainers {
+				neighbors, err := frr.NeighborsInfo(c)
+				Expect(err).NotTo(HaveOccurred())
+				previousNeighbors[c.Name] = neighbors
+			}
+			ginkgo.By("creating another the service")
+			svc1, _ := testservice.CreateWithBackend(cs, testNamespace, "external-local-lb1", tweak)
+			defer testservice.Delete(cs, svc1)
+
+			Consistently(func() error {
+				for _, c := range FRRContainers {
+					neighbors, err := frr.NeighborsInfo(c)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(neighbors).To(HaveLen(len(previousNeighbors[c.Name])))
+
+					for _, n := range neighbors {
+						previousDropped := previousNeighbors[c.Name][n.IP.String()].ConnectionsDropped
+						if n.ConnectionsDropped > previousDropped {
+							return fmt.Errorf("increased connections dropped from %s to %s, previous: %d current %d", c.Name, n.IP.String(), previousDropped, n.ConnectionsDropped)
+						}
+					}
+				}
+				return nil
+			}, 10*time.Second, 1*time.Second).ShouldNot(HaveOccurred())
 		},
 			ginkgo.Entry("IPV4 - default",
 				metallbv1beta1.BFDProfile{
--- a/e2etest/pkg/frr/bgp.go
+++ b/e2etest/pkg/frr/bgp.go
@@ -32,18 +32,24 @@ func NeighborInfo(neighborName string, exec executor.Executor) (*Neighbor, error
 	return neighbor, nil
 }
 
+type NeighborsMap map[string]*Neighbor
+
 // NeighborsInfo returns informations for the all the neighbors in the given
 // executor.
-func NeighborsInfo(exec executor.Executor) ([]*Neighbor, error) {
-	res, err := exec.Exec("vtysh", "-c", "show bgp neighbor json")
+func NeighborsInfo(exec executor.Executor) (NeighborsMap, error) {
+	jsonRes, err := exec.Exec("vtysh", "-c", "show bgp neighbor json")
 	if err != nil {
 		return nil, errors.Join(err, errors.New("Failed to query neighbours"))
 	}
-	neighbors, err := ParseNeighbours(res)
+	neighbors, err := ParseNeighbours(jsonRes)
 	if err != nil {
-		return nil, errors.Join(err, fmt.Errorf("Failed to parse neighbours %s", res))
+		return nil, errors.Join(err, fmt.Errorf("Failed to parse neighbours %s", jsonRes))
 	}
-	return neighbors, nil
+	res := map[string]*Neighbor{}
+	for _, n := range neighbors {
+		res[n.IP.String()] = n
+	}
+	return res, nil
 }
 
 // Routes returns informations about routes in the given executor
--- a/e2etest/pkg/frr/parse_test.go
+++ b/e2etest/pkg/frr/parse_test.go
@@ -88,8 +88,7 @@ func TestNeighbour(t *testing.T) {
           "sentPrefixCounter":%d
         }
       },
-      "connectionsEstablished":0,
-      "connectionsDropped":0,
+      "connectionsDropped":2,
       "lastResetTimerMsecs":253000,
       "lastResetDueTo":"Waiting for peer OPEN",
       "lastResetCode":32,
@@ -185,6 +184,9 @@ func TestNeighbour(t *testing.T) {
 			}
 			if n.ConfiguredKeepAliveTime != 30000 {
 				t.Fatal("unexpected configured keepalivetime time", n.ConfiguredKeepAliveTime)
+			}
+			if n.ConnectionsDropped != 2 {
+				t.Fatal("unexpected connections dropped", n.ConnectionsDropped)
 			}
 		})
 	}
--- a/e2etest/pkg/frr/parse.go
+++ b/e2etest/pkg/frr/parse.go
@@ -27,6 +27,7 @@ type Neighbor struct {
 	ConfiguredKeepAliveTime int
 	ConfiguredConnectTime   int
 	AddressFamilies         []string
+	ConnectionsDropped      int
 }
 
 type Route struct {
@@ -55,6 +56,7 @@ type FRRNeighbor struct {
 	AddressFamilyInfo            map[string]struct {
 		SentPrefixCounter int `json:"sentPrefixCounter"`
 	} `json:"addressFamilyInfo"`
+	ConnectionsDropped int `json:"connectionsDropped"`
 }
 
 type GracefulRestartInfo struct {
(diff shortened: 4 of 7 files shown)


e2etest/bgptests/bgp.go
```diff
@@ -662,6 +662,33 @@ var _ = ginkgo.Describe("BGP", func() {
...
 
+			ginkgo.By("checking the sessions don't flap when changing the configuration")
+
+			previousNeighbors := map[string]frr.NeighborsMap{}
+			for _, c := range FRRContainers {
+				neighbors, err := frr.NeighborsInfoMap(c)
+				Expect(err).NotTo(HaveOccurred())
+				previousNeighbors[c.Name] = neighbors
+			}
+			ginkgo.By("creating another the service")
```
karampok, line 673: `creating and delete another service` ?

e2etest/bgptests/bgp.go
```diff
@@ -662,6 +662,33 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			svc1, _ := testservice.CreateWithBackend(cs, testNamespace, "external-local-lb1", tweak)
+			testservice.Delete(cs, svc1)
```
karampok, line 675: is the Delete really need to observe the bug?

e2etest/bgptests/bgp.go
```diff
@@ -662,6 +662,33 @@ var _ = ginkgo.Describe("BGP", func() {
...
+
+					for _, n := range neighbors {
+						previousDropped := previousNeighbors[c.Name][n.IP.String()].ConnectionsDropped
+						if n.ConnectionsDropped > previousDropped {
+							return fmt.Errorf("increased connections dropped from %s to %s, previous: %d current %d", c.Name, n.IP.String(), previousDropped, n.ConnectionsDropped)
+						}
+					}
+				}
+				return nil
+			}, 5*time.Second, 1*time.Second).ShouldNot(HaveOccurred())
```
karampok, line 691: why 5 and not 10 or 20 seconds?

e2etest/bgptests/bgp.go
```diff
@@ -662,6 +662,33 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			}
+			ginkgo.By("creating another the service")
+			svc1, _ := testservice.CreateWithBackend(cs, testNamespace, "external-local-lb1", tweak)
+			testservice.Delete(cs, svc1)
+
+			Consistently(func() error {
+				for _, c := range FRRContainers {
+					neighbors, err := frr.NeighborsInfo(c)
+					Expect(err).NotTo(HaveOccurred())
+					Expect(neighbors).To(HaveLen(len(previousNeighbors[c.Name])))
```
karampok, line 681: key in map if not exists (not I can see that happen) panics

e2etest/bgptests/bgp.go
```diff
@@ -662,6 +662,33 @@ var _ = ginkgo.Describe("BGP", func() {
...
+
+					for _, n := range neighbors {
+						previousDropped := previousNeighbors[c.Name][n.IP.String()].ConnectionsDropped
+						if n.ConnectionsDropped > previousDropped {
```
karampok, line 685: Can you add in the commit one/two sentence doc about the timer? or link

e2etest/bgptests/bgp.go
fedepaol: leftover, I'll just defer the delete as we do with the others

e2etest/bgptests/bgp.go
fedepaol: nah, I meant it to be the usual deferred deletion

e2etest/bgptests/bgp.go
fedepaol: I tested locally with a power machine and 5 seconds was still making the bug appear.
It's a `Consistently`, so the longer the time, the longer the execution time so we should be cautious when using them. If we starts seeing it flaking, we'll make it longer.

e2etest/bgptests/bgp.go
fedepaol: what timer?

e2etest/bgptests/bgp.go
fedepaol: yep but in this case I think it's fair because:
- it's a test, we'll see where it fails and eventually adjust it
- we are filling the map with the same set of keys we are using here, just a few lines above, so it shouldn't happen

All in all, the extra check is not worth the break in following the code imo

e2etest/bgptests/bgp.go
karampok: hiding the error might be the case. I mean if you create the svc but process takes more than 5 seconds to apply the config, then your test will be green but behavior  can regress.

e2etest/bgptests/bgp.go
karampok: in 10 years from now it could be many lines diff :)
we can let it or you can use the https://onsi.github.io/gomega/#havekeykey-interface is single line

e2etest/bgptests/bgp.go
karampok: sorry I corrected timer, what is the doc for this ConnectionDropped

e2etest/bgptests/bgp.go
karampok: will let is as is or use a defer?

e2etest/bgptests/bgp.go
fedepaol: yep, but the same can be said whatever time we put in. I can extend to 10, but I don't think it's worth doing more than that.

e2etest/bgptests/bgp.go
fedepaol: > in 10 years from now it could be many lines diff :) we can let it or you can use the https://onsi.github.io/gomega/#havekeykey-interface is single line

Is there any added value if we add the assertion vs having it explode? The test will fail anyway (even more elegantly) and it will still tell us the missing key.

e2etest/bgptests/bgp.go
fedepaol: it's a property of the output of `bgp show neighbor`. I can add a comment saying connectiondropped is the number of dropped connections, but feels a bit redundant.

fedepaol:
interestingly, the frr-k8s lanes still failed (which makes sense, because frr-k8s is still bugged), but the external did not.
Also, weird that the operator lane didn't fail either.

e2etest/bgptests/bgp.go
karampok: we can let  5sec as long as you still think is good

karampok:
> interestingly, the frr-k8s lanes still failed (which makes sense, because frr-k8s is still bugged), but the external did not. Also, weird that the operator lane didn't fail either.

maybe you could comment out the implementation, let only the test and run the ci couple of times?

e2etest/bgptests/bgp.go
fedepaol: The test is failing on frr-k8s jobs, meaning that it is "at least" borderline to be good. Let's put 10 to be on the safe side. Once we see our tests are taking too long, we'll do a round to understand why.

fedepaol:
> > interestingly, the frr-k8s lanes still failed (which makes sense, because frr-k8s is still bugged), but the external did not. Also, weird that the operator lane didn't fail either.
> 
> maybe you could comment out the implementation, let only the test and run the ci couple of times?

Yup, but the fact that some frr-k8s lanes are failing should mean that the test is testing.
Might be related to the timing, let's increase to 10 hoping to make it fail all the frr-k8s / external lanes.

Last odd is the operator lane failing, as it should run with the image being built here. checking the logs before re-pushing

fedepaol:
The reason why the operator lane is failing is because it's running with frr 8.4.2:

```
                    {
                        "name": "FRR_IMAGE",
                        "value": "quay.io/frrouting/frr:8.4.2"
                    },
```
Which has another bug related to multihop (https://github.com/FRRouting/frr/issues/14242):

```
Lines To Delete
===============
no log file /etc/frr/frr.log
ipv6 forwarding
no debug bgp updates in
no debug bgp updates out
router bgp 64512
 no neighbor 172.30.0.2 ebgp-multihop 255  <- causing the flap

```

we need to update the operator too.

e2etest/bgptests/bgp.go
fedepaol: oops :-)
Fixed

e2etest/pkg/frr/bgp.go
```diff
@@ -32,6 +32,22 @@ func NeighborInfo(neighborName string, exec executor.Executor) (*Neighbor, error
 	return neighbor, nil
 }
 
+type NeighborsMap map[string]*Neighbor
+
+// NeighborsInfo returns informations for the all the neighbors in the given
+// executor.
+func NeighborsInfoMap(exec executor.Executor) (NeighborsMap, error) {
+	neighbors, err := NeighborsInfo(exec)
```
karampok, line 40: just a thought, we could change this function to always return a map (but is not worth the refactoring)

karampok (approved):

e2etest/pkg/frr/bgp.go
fedepaol: Let me check...

fedepaol: done

e2etest/pkg/frr/parse_test.go
```diff
@@ -186,6 +185,9 @@ func TestNeighbour(t *testing.T) {
 			if n.ConfiguredKeepAliveTime != 30000 {
 				t.Fatal("unexpected configured keepalivetime time", n.ConfiguredKeepAliveTime)
 			}
+			if n.ConnectionsDropped != 2 {
+				t.Fatal("unexpected configured keepalivetime time", n.ConfiguredKeepAliveTime)
```
karampok, line 189: message is wrong, no?

fedepaol: definetely!

fedepaol: fixed

karampok (approved):

PR #2470, opened 2024-07-24
Stop advertisement  when speaker on  a cordoned node

karampok:
/kind feature

When we have peering with  graceful restar, and when we have a svc of policy Cluster,  and a node becomes cordoned, that node continues advertizes prefixes. If that node reboots (which is probably scenario after drain a node), the peer will keep the routes for the full graceful restart timer. That gives downtime.

**Release note**:

```release-note
Clear advertisement when speaker runs in a cordoned (unschedulable) node
```


speaker/bgp_controller.go
```diff
@@ -394,13 +417,29 @@ func (c *bgpController) SetNode(l log.Logger, node *v1.Node) error {
 	if c.myNode != node.Name {
 		return nil
 	}
+	level.Debug(l).Log("event", "nodeChange", "msg", "start")
+	defer level.Debug(l).Log("event", "nodeChange", "msg", "end")
+
+	if v := node.Spec.Unschedulable; v != c.unschedulable {
```
fedepaol, line 423: we shouldn't make logic on the previous value as it's going to make the controller weak. Let' shook in the same place we have the `k8snodes.IsNetworkUnavailable` checks so the controller is level driven and not edge driven.
This will also make the clearAds not necessary.

speaker/bgp_controller.go
```diff
@@ -169,6 +170,13 @@ func (c *bgpController) ShouldAnnounce(l log.Logger, name string, _ []net.IP, po
 		return "nodeNetworkUnavailable"
 	}
 
+	if v, ok := nodes[c.myNode]; ok {
+		if v.Spec.Unschedulable {
```
fedepaol, line 174: I'd extend the IsNetworkUnavailable function to some more generic name and embed any "this node is not good" logic into that.

speaker/bgp_controller.go
karampok: Can you expand more on this comment? I think I miss some logic:
the `IsNetworkUnavailable` is being within the `c.handleService` which is called by ctrl.SetBalancer`
So if understand the top down logic

`client -> k8s.Listener-> serviceChanged -> call ctrl.SetBalancer `

which executed when service changes.

How that executed when there is a node change

those are the logs when running the NetworkUnavailable
```
{"caller":"bgp_controller.go:420","event":"nodeChange","level":"debug","msg":"start","ts":"2024-07-24T13:17:27Z"}
{"caller":"bgp_controller.go:442","event":"nodeLabelsNoChange","level":"debug","msg":"Node labels same, do thing","ts":"2024-07-24T13:17:27Z"}
{"caller":"bgp_controller.go:443","event":"nodeChange","level":"debug","msg":"end","ts":"2024-07-24T13:17:27Z"}
```
thanks

fedepaol: When a node status changes in a way that might make all the services not available, we reprocess the services and some of them can be revoked.

In general, if some logic is implemented in some way, it's reasonable to not add a different way to implement something similar. There are already a set of conditions that might make a node not able to announce the service.

internal/k8s/nodes/nodes.go
```diff
@@ -3,9 +3,35 @@
...
+
 	corev1 "k8s.io/api/core/v1"
 )
 
+func IsNodeAvailable(n *corev1.Node) (bool, string) {
+	ret := true
+	reason := []string{}
+
+	if IsNetworkUnavailable(n) {
+		ret = ret && false
```
fedepaol, line 16: We can exit early with the first offending reason. Also, I am tempted to say that (bool, string) is worse than returning a speaking error.

internal/k8s/nodes/nodes.go
```diff
@@ -3,9 +3,35 @@
...
+
+	if IsNodeUnschedulable(n) {
+		ret = ret && false
+		reason = append(reason, "nodeUnschedulable")
+	}
+	return ret, strings.Join(reason, ",")
+}
+
+func IsNodeUnschedulable(n *corev1.Node) bool {
+	if n == nil {
```
fedepaol, line 28: Is this an expected scenario? If not, I'd have it blow up instead of going defensive.

speaker/main.go
```diff
@@ -570,13 +571,17 @@ func (c *controller) SetNode(l log.Logger, node *v1.Node) controllers.SyncState
...
 
-	if conditionChanged || labelNodeExcludeBalancersChanged {
+	if conditionChanged || labelNodeExcludeBalancersChanged || schedulableConditionChanged {
 		return controllers.SyncStateReprocessAll
 	}
 
 	return controllers.SyncStateSuccess
 }
 
+func isNodeSchedulableChanged(nodeName string, oldNodes map[string]*v1.Node, newNode *v1.Node) bool {
```
fedepaol, line 581: let's have a `isNodeAvailableChanged`  function here, so it's going to be easier in case a new condition is added.

internal/k8s/controllers/node_controller.go
```diff
@@ -82,6 +82,10 @@ func (r *NodeReconciler) SetupWithManager(mgr ctrl.Manager) error {
 				level.Error(r.Logger).Log("controller", "NodeReconciler", "error", "old object is not node", "name", oldNodeObj.GetName())
 				return true
 			}
+			if oldNodeObj.Spec.Unschedulable != newNodeObj.Spec.Unschedulable {
```
fedepaol, line 85: We never return false if they are equal. This must be appended to the below if (or we must flip it to a sequence and change the default, but in that case I'd advise to unit test this)

e2etest/bgptests/cordon_node.go
```diff
@@ -0,0 +1,147 @@
...
+	ginkgo.Describe("When create an IPV4 service and then cordon node", func() {
+		var (
+			svc                *corev1.Service
+			allNodes           *v1.NodeList
+			schedulableNodeIPs []net.IP
+		)
+
+		ginkgo.BeforeEach(func() {
+			if BGPMode == "frr" {
+				for _, c := range FRRContainers {
```
fedepaol, line 84: If it relates to graceful restart, we should make sure that runs only with FRR enabled by adding FRR in the name of the test so it can be skipped from outside.

e2etest/bgptests/cordon_node.go
```diff
@@ -0,0 +1,147 @@
...
+				validateService(svc, allNodes.Items, c)
+			}
+
+			testservice.Delete(cs, svc)
+		})
+
+		//			ginkgo.It("speaker on cordoned node should have empty frr config", func() {})
+		ginkgo.It("bgp peers should not include cordoned node IP as next hop", func() {
+			Eventually(func() error {
+				for _, c := range FRRContainers {
```
fedepaol, line 128: I think you can use `checkServiceNotOnNodes` here to ensure that the cordoned node is not announcing.
Also, I'd try to uncordon it to see if the service is restored.

e2etest/bgptests/cordon_node.go
```diff
@@ -0,0 +1,147 @@
+/*
```
fedepaol, line 1: Just add the spdx license

e2etest/bgptests/cordon_node.go
```diff
@@ -0,0 +1,147 @@
...
+	ginkgo.AfterEach(func() {
+		if ginkgo.CurrentSpecReport().Failed() {
+			dumpBGPInfo(ReportPath, ginkgo.CurrentSpecReport().LeafNodeText, cs, testNamespace)
+			k8s.DumpInfo(Reporter, ginkgo.CurrentSpecReport().LeafNodeText)
+		}
+		err := k8s.DeleteNamespace(cs, testNamespace)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	ginkgo.Describe("When create an IPV4 service and then cordon node", func() {
```
fedepaol, line 75: We'll need a similar test for L2

e2etest/bgptests/cordon_node.go
karampok: since we there  is no peer.Close, I would say graceful restart is orthogonal.  I remove it

e2etest/bgptests/cordon_node.go
karampok: Using that function  is a very good idea and works with VRFs which is tricky because you have to take IPs from the containers. Thanks

>  I'd try to uncordon it to see if the service is restored.
I do test it in the after each
```
		ginkgo.AfterEach(func() {

			err := k8s.UnCordonNode(cs, &allNodes.Items[0])

				validateService(svc, allNodes.Items, c)
			}

			testservice.Delete(cs, svc)
		})


```
but I would not object that this might be a test case on its own

e2etest/bgptests/cordon_node.go
karampok: done

internal/k8s/nodes/nodes.go
karampok: unit test are panicking 

```
 go test ./speaker/...
--- FAIL: TestBGPSpeakerEPSlices (0.00s)
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x198 pc=0x14901e2]

goroutine 67 [running]:
testing.tRunner.func1.2({0x16a7fe0, 0x2748e40})
        /home/kka/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-amd64/src/testing/testing.go:1631 +0x24a
testing.tRunner.func1()
        /home/kka/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-amd64/src/testing/testing.go:1634 +0x377
panic({0x16a7fe0?, 0x2748e40?})
        /home/kka/go/pkg/mod/golang.org/toolchain@v0.0.1-go1.22.4.linux-amd64/src/runtime/panic.go:770 +0x132
go.universe.tf/metallb/internal/k8s/nodes.conditionStatus(...)
```
therefore the pattern that also exists in 
https://github.com/metallb/metallb/blob/main/internal/k8s/nodes/nodes.go#L16-L18
and 
https://github.com/metallb/metallb/blob/main/internal/k8s/nodes/nodes.go#L16-L18

internal/k8s/controllers/node_controller.go
karampok: I refactored that part as in this https://github.com/metallb/metallb/pull/2470/commits/64be8c947e8b1051b9d236de8c489ebf9d0d6113

The implementation is a bit more modular and it makes it clear if we want to extend it


Added unit-test also, is there any test case I miss?

speaker/bgp_controller.go
karampok: true, I saw the logic, thanks

speaker/main.go
karampok: done

internal/k8s/nodes/nodes.go
karampok: speaking error then, not great because this is not really an error but I agree (bool,string) or (string) and compare to empty is worse.

e2etest/bgptests/cordon_node.go
karampok: added

e2etest/bgptests/cordon_node.go
fedepaol: fair enough then

e2etest/bgptests/cordon_node.go
fedepaol, line 1: nit on the commit message: double signed-off

e2etest/go.mod
```diff
@@ -16,10 +16,10 @@ require (
 	github.com/prometheus/client_golang v1.19.0 // indirect
 	go.uber.org/zap v1.26.0 // indirect
 	golang.org/x/sys v0.18.0 // indirect
-	k8s.io/api v0.30.1
+	k8s.io/api v0.30.3
```
fedepaol, line 19: why did we need this bump (sorry if I asked already)?

e2etest/pkg/k8s/nodes.go
```diff
@@ -151,3 +153,62 @@ func conditionStatus(n *corev1.Node, ct corev1.NodeConditionType) corev1.Conditi
 
 	return corev1.ConditionUnknown
 }
+
+func CordonNode(cs kubernetes.Interface, node *corev1.Node) error {
+	if cs == nil {
```
fedepaol, line 158: I think it's fine to explode here. After all this is a test, we'll have a report saying where it exploded and we have a bit less clutter. Same for node below.

e2etest/bgptests/cordon_node.go
```diff
@@ -0,0 +1,105 @@
...
+			allNodes, err = cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			for _, c := range FRRContainers {
+				validateService(svc, allNodes.Items, c)
+			}
+
+			err = k8s.CordonNode(cs, &allNodes.Items[0])
+			Expect(err).NotTo(HaveOccurred(), "k8s api call to cordon node failed")
+			Eventually(func() bool {
```
fedepaol, line 75: would it make sense to have this part of cordon node (ie cordon node is synchronous and waits for the node to be cordoned) ?

e2etest/bgptests/cordon_node.go
```diff
@@ -0,0 +1,105 @@
...
+				ret, err := k8s.IsNodeCordoned(cs, &allNodes.Items[0])
+				Expect(err).NotTo(HaveOccurred())
+				return ret
+			}, 30*time.Second, time.Second).Should(BeTrue(), "node.spec.Unschedulable not true")
+
+		})
+
+		ginkgo.AfterEach(func() {
+
+			err := k8s.UnCordonNode(cs, &allNodes.Items[0])
```
fedepaol, line 85: here we don't wait for the node to be uncordoned, we might leak state across tests.

e2etest/l2tests/cordon_node.go
fedepaol, line 1: we should reorganize the commits. I think it make sense to have one commit for "handling hte cordoned node" scenario, both l2 and bgp, and another for e2e tests.

e2etest/l2tests/cordon_node.go
```diff
@@ -0,0 +1,132 @@
...
+	var (
+		cs            clientset.Interface
+		testNamespace string
+	)
+
+	ginkgo.BeforeEach(func() {
+		err := ConfigUpdater.Clean()
+		Expect(err).NotTo(HaveOccurred(), "cleaning k8s api CRs failed")
+		cs = k8sclient.New()
+		testNamespace, err = k8s.CreateTestNamespace(cs, "bgp")
```
fedepaol, line 39: nit: this is l2

e2etest/l2tests/cordon_node.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+		err := k8s.DeleteNamespace(cs, testNamespace)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	ginkgo.Describe("When create an IPV4 service and then cordon node", func() {
+		var (
+			svc       *corev1.Service
+			allNodes  *corev1.NodeList
+			nodeToSet *corev1.Node
```
fedepaol, line 75: let's call this `nodeToCordon`

e2etest/l2tests/node_service.go
```diff
@@ -15,22 +15,22 @@ import (
 	corev1 "k8s.io/api/core/v1"
 )
 
-func nodeForService(svc *corev1.Service, nodes []corev1.Node) (string, error) {
+func nodeForService(svc *corev1.Service, nodes []corev1.Node) (*corev1.Node, error) {
```
fedepaol, line 18: can you explain why do we need this refactor (possibly in the commit message, and maybe in a separate commit if it's not completely needed for this new test)?

internal/k8s/controllers/node_controller_test.go
```diff
@@ -121,3 +123,109 @@ func TestNodeController(t *testing.T) {
...
+			},
+			expected: false,
+		},
+	}
+
+	for name, test := range tests {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+			t.Log(name)
+			if got, expected := p.Update(test.event), test.expected; got != expected {
```
fedepaol, line 226: let's also add Create and delete to verify they always pass

internal/k8s/controllers/node_controller_test.go
```diff
@@ -121,3 +123,109 @@ func TestNodeController(t *testing.T) {
...
+			},
+			expected: false,
+		},
+		"label change": {
+			event: event.UpdateEvent{
+				ObjectOld: &corev1.Node{},
+				ObjectNew: &corev1.Node{ObjectMeta: metav1.ObjectMeta{Labels: map[string]string{"x": "y"}}}},
+			expected: true,
+		},
+		"spec schedulable change": {
```
fedepaol, line 177: let's add one where both schedulable and label change

internal/k8s/nodes/nodes.go
fedepaol: fair enough!

speaker/main.go
```diff
@@ -570,19 +569,22 @@ func (c *controller) SetNode(l log.Logger, node *v1.Node) controllers.SyncState
...
-}
+func isNodeAvailableChanged(oldNodes map[string]*v1.Node, newNode *v1.Node) bool {
+	oldNode, exists := oldNodes[newNode.Name]
+	if !exists {
+		return false
+	}
 
-func isLabelNodeExcludeBalancersChanged(nodeName string, oldNodes map[string]*v1.Node, newNode *v1.Node) bool {
-	return k8snodes.IsNodeExcludedFromBalancers(oldNodes[nodeName]) != k8snodes.IsNodeExcludedFromBalancers(newNode)
+	return k8snodes.IsNodeUnschedulable(oldNode) != k8snodes.IsNodeUnschedulable(newNode) ||
```
fedepaol, line 585: let's avoid ors and have multiple if xxx -> return true

e2etest/go.mod
karampok: probably not, I revert and I added only the changes that this command
```
go get k8s.io/kubectl/pkg/drain@v0.30.1
```
gave me .

e2etest/bgptests/cordon_node.go
karampok: done

internal/k8s/controllers/node_controller_test.go
karampok: done

karampok (commented):
Thanks for the feedback

fedepaol:
lgtm, sent to merge queue

remram44:
This brought all networking down into my cluster. I set my own tolerations for the speaker, I definitely didn't expect that the speaker will correctly run on a node **but refuse to advertise prefixes because of taints on the node**. This is extremely surprising and feels very wrong.

At the very least please put a "breaking change" warning when you change the behavior of advertisement in a manner where all advertisements might get disabled.

karampok:
Hi @remram44, sorry for the inconvenience. Could you please explain more your use case, why do you need the cordoned to be part of the networking  dataplane?  Ideally if you can create a new issue so we can follow it up. Thx

kfox1111:
This seems like it trades one bug for another?

If you have only one pod in a load balancer, and you cordon the node to prevent new workload from hitting the node, with no intention of draining, it would leave the load balancer dead to new connections.

Seems like it should be checking something like https://github.com/kubernetes/enhancements/issues/4212 once it is a thing.

remram44:
First of all, no, because the load balancer doesn't necessarily only routes to local workloads. This is only the case if `externalTrafficPolicy=Local` which is not the default.

Second of all, your workloads might have a toleration for `unschedulable=reserved-for-foobar`. In fact if you are deploying a workload to your ingress nodes, you do it with a DaemonSet, which tolerates `unschedulable` by default.

`unschedulable` is *not* a "rebooting soon" signal. In fact the [graceful node shutdown](https://kubernetes.io/docs/concepts/cluster-administration/node-shutdown/#graceful-node-shutdown) documentation explicitly indicates that `not-ready` is the taint that will be set by a node when shutting down.

You have grabbed a random taint that is not meant for this and assumed that everyone will use it for nothing but shutdowns. And you have done so in a non-configurable manner, in a patch release. Yesterday wasn't fun.

fedepaol:
> You have grabbed a random taint that is not meant for this and assumed that everyone will use it for nothing but shutdowns. And you have done so in a non-configurable manner, in a patch release. Yesterday wasn't fun.

Hey, sorry we caused disruption on your environment. This was done in good faith with the idea to make MetalLB better, clearly not to cause any problem. We probably oversaw this behavior as it always happen when you jump from one thing to the other. We'll try to amend, although I am not sure we'll be able to before the break.

As @karampok said, it would be great if you can file an issue so other users will find it in case they hit the same problem.

fedepaol:
cc @oribon

DerRockWolf:
Did somebody already open an issue regarding the above mentioned problem? (https://github.com/metallb/metallb/pull/2470#issuecomment-2555779581)

karampok:
https://github.com/metallb/metallb/pull/2715 that has been reverted.

fedepaol:
> > #2715 that has been reverted.
> 
> Is there an ETA when the revert will be released?

we are targeting this week for releasing a new versoin

PR #1417, opened 2022-06-07
Add support for LoadBalancerClass

fedepaol:
Loadbalancer class enable different loadbalancers to co-exist. We add a parameter to the speaker and the controller to process only the services consistent with the loadbalancer class provided to metallb.

Fixes https://github.com/metallb/metallb/issues/685 and https://github.com/metallb/metallb/issues/996

PR #2474, opened 2024-07-29
Add option to keep watching services without a loadbalancer class

tommasopozzetti:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

PR https://github.com/metallb/metallb/pull/1417 introduced support for LoadBalancer Class. However, with the current design, metallb can either look at every LoadBalancer service in the cluster (irrespectively of their LoadBalancer Class) or only the ones that specify the right LoadBalancer class. There is no way to have metallb "ignore" services that specify a given LoadBalancer Class that is not the one managed by metallb while also remaining the "default" load balancer implementation for services not specifying a LoadBalancer class at all.

This feature is very useful in big clusters with thousands of services where only a couple might need to be dealt with differently through a different load balancer implementation. Today there is no way to achieve this without refactoring every other service to also specify the load balancer class.

This PR introduces an extra flag `watch-lb-without-class` (being meaningful only if the `lb-class` flag is also passed and defaulting to `false` for backwards compatibility) which instructs metallb running in `lb-class` mode to still reconcile and manage also services which do not specify a load balancer class at all.

This is the same design that projects like ingress-nginx use for the same purpose (with ingress classes): https://github.com/kubernetes/ingress-nginx/blob/helm-chart-4.11.1/charts/ingress-nginx/values.yaml#L81

This PR should be fully backwards compatible such that users that do not wish to use this feature do not need to change anything at all in their current configuration and will observe the same behavior as they used to prior to this change.

**Release note**:

```release-note
A new flag `watch-lb-without-class` has been added. When used in conjunction with `lb-class` it instructs metallb to reconcile and manage also load balancer services which do not specify a load balancer class.
```

### DIFF
--- a/charts/metallb/README.md
+++ b/charts/metallb/README.md
@@ -66,6 +66,7 @@ Kubernetes: `>= 1.19.0-0`
 | fullnameOverride | string | `""` |  |
 | imagePullSecrets | list | `[]` |  |
 | loadBalancerClass | string | `""` |  |
+| watchLoadBalancerWithoutClass | bool | `false` |  |
 | nameOverride | string | `""` |  |
 | prometheus.controllerMetricsTLSSecret | string | `""` |  |
 | prometheus.metricsPort | int | `7472` |  |
--- a/charts/metallb/templates/controller.yaml
+++ b/charts/metallb/templates/controller.yaml
@@ -68,6 +68,9 @@ spec:
         {{- if .Values.loadBalancerClass }}
         - --lb-class={{ .Values.loadBalancerClass }}
         {{- end }}
+        {{- if .Values.watchLoadBalancerWithoutClass }}
+        - --watch-lb-without-class
+        {{- end }}
         {{- if .Values.controller.webhookMode }}
         - --webhook-mode={{ .Values.controller.webhookMode }}
         {{- end }}
--- a/charts/metallb/templates/speaker.yaml
+++ b/charts/metallb/templates/speaker.yaml
@@ -246,6 +246,9 @@ spec:
         {{- if .Values.loadBalancerClass }}
         - --lb-class={{ .Values.loadBalancerClass }}
         {{- end }}
+        {{- if .Values.watchLoadBalancerWithoutClass }}
+        - --watch-lb-without-class
+        {{- end }}
         {{- if .Values.speaker.wanConfig }}
         - --ml-wan-config
         {{- end }}
--- a/charts/metallb/values.schema.json
+++ b/charts/metallb/values.schema.json
@@ -162,6 +162,9 @@
     "loadBalancerClass": {
       "type":"string"
     },
+    "watchLoadBalancerWithoutClass": {
+      "type": "boolean"
+    },
     "rbac": {
       "description": "RBAC configuration",
       "type": "object",
--- a/charts/metallb/values.yaml
+++ b/charts/metallb/values.yaml
@@ -6,6 +6,7 @@ imagePullSecrets: []
 nameOverride: ""
 fullnameOverride: ""
 loadBalancerClass: ""
+watchLoadBalancerWithoutClass: false
 
 # To configure MetalLB, you must specify ONE of the following two
 # options.
--- a/controller/main.go
+++ b/controller/main.go
@@ -138,21 +138,22 @@ func (c *controller) SetPools(l log.Logger, pools *config.Pools) controllers.Syn
 
 func main() {
 	var (
-		port                = flag.Int("port", 7472, "HTTP listening port for Prometheus metrics")
-		namespace           = flag.String("namespace", os.Getenv("METALLB_NAMESPACE"), "config / memberlist secret namespace")
-		mlSecret            = flag.String("ml-secret-name", os.Getenv("METALLB_ML_SECRET_NAME"), "name of the memberlist secret to create")
-		deployName          = flag.String("deployment", os.Getenv("METALLB_DEPLOYMENT"), "name of the MetalLB controller Deployment")
-		logLevel            = flag.String("log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", logging.Levels.String()))
-		enablePprof         = flag.Bool("enable-pprof", false, "Enable pprof profiling")
-		disableCertRotation = flag.Bool("disable-cert-rotation", false, "disable automatic generation and rotation of webhook TLS certificates/keys")
-		certDir             = flag.String("cert-dir", "/tmp/k8s-webhook-server/serving-certs", "The directory where certs are stored")
-		certServiceName     = flag.String("cert-service-name", "metallb-webhook-service", "The service name used to generate the TLS cert's hostname")
-		loadBalancerClass   = flag.String("lb-class", "", "load balancer class. When enabled, metallb will handle only services whose spec.loadBalancerClass matches the given lb class")
-		webhookMode         = flag.String("webhook-mode", "enabled", "webhook mode: can be enabled, disabled or only webhook if we want the controller to act as webhook endpoint only")
-		webhookSecretName   = flag.String("webhook-secret", "metallb-webhook-cert", "webhook secret: the name of webhook secret, default is metallb-webhook-cert")
-		webhookHTTP2        = flag.Bool("webhook-http2", false, "enables http2 for the webhook endpoint")
-		tlsMinVersion       = flag.String("tls-min-version", "", "Minimum TLS version supported for the webhook server, Possible values: "+strings.Join(cliflag.TLSPossibleVersions(), ", "))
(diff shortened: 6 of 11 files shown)


fedepaol:
Hey, thanks for the PR and sorry for the delay, I was away.

> There is no way to have metallb "ignore" services that specify a given LoadBalancer Class that is not the one managed by metallb while also remaining the "default" load balancer implementation for services not specifying a LoadBalancer class at all.

The way to handle this is, have metallb be the default lb and specify the lb class only to the few services that need to be handled specially.

But if I understood your scenario, it's about having metallb be the default loadbalancer plus handling a specific lb class (while having other lb classes handled by other lb implementations), which feels very specific to your use case where you have some services specifying a lb class that does not make them special.

I think that the current model adheres to what is the k8s definition for lb class, so I'd tend to lean to the current implementation, but I am open to discuss if I missed something.

tommasopozzetti:
Hey @fedepaol thanks for taking a look!

> The way to handle this is, have metallb be the default lb and specify the lb class only to the few services that need to be handled specially.

This is precisely what this PR aims to enable! Unless I am missing something (and please feel free to correct me if so), this is impossible today without this PR! If metallb is the default load balancer (i.e. running without the `lb-class` argument), it won't care about the LoadBalancer class of any service, whether it is specified or not. So if I only specify the LoadBalancer class of the ones I want to be handled by a different implementation, metallb will still reconcile those and fight with the other implementation.

By adding this functionality, metallb fully conforms to LoadBalancer class concept of k8s when run with the `lb-class` argument while also allowing to be set as the default reconciler for any service not specifying explicitly a load balancer class.

Let me know if this makes sense or if I am missing something!
Thank you!

fedepaol:
> it won't care about the LoadBalancer class of any service, whether it is specified or not. So if I only specify the LoadBalancer class of the ones I want to be handled by a different implementation, metallb will still reconcile those and fight with the other implementation.

This is not correct. If metallb is the default LB (ie the loadbalancer class parameter is not specified), it will ignore all the services coming with the LB Class set. We also have an e2e ensuring the behavior: https://github.com/metallb/metallb/blob/main/e2etest/l2tests/loadbalancer_class.go

The if that ensures it is the third https://github.com/metallb/metallb/blob/c76e72d1cf3f675d6bd83caa17f95df803fa6e33/internal/k8s/controllers/service_controller.go#L172

Your patch allows metallb to handle a lb class AND to be the default one.

tommasopozzetti:
@fedepaol I did some more testing, I could have sworn I checked and it didn't work but I'm clearly mistaken. Specifying the class for the "special" ones only works for me.
We can close this if there is no interest in the functionality of allowing both a specific class + default behavior
Thank you!

fedepaol:
> @fedepaol I did some more testing, I could have sworn I checked and it didn't work but I'm clearly mistaken. Specifying the class for the "special" ones only works for me. We can close this if there is no interest in the functionality of allowing both a specific class + default behavior Thank you!

Glad to hear that! Yep, I'd close this as I don't think it's worth the extra complexity.

tmmorin:
Hello @fedepaol 

As you rightly explained above, it is true that in "cruise mode", MetalLB as-is can co-exist with another LB implementation.

However, I think that this PR has value **to handle transition scenarios** where MetalLB initially would not have `--lb-class` and would need to transition to using `--lb-class=..`.

Today MetalLB offers only two behaviors:
- (1) without `--lb-class`: metallb runs as "lbclass-blind" (process services without an LB class, ignores services which specify spec.lbClass)
- or (2) with `--lb-class`: metallb runs "lbclass-only", and does **not** process services which do not have spec.lbClass
- (there is no (2bis): with `--lb-class` but still processing services which do not have spec.lbClass)

Having only (1) and (2) makes transition painful. Indeed, with (1) as a starting point, when you switch to `--lb-class=metallb`, you can't avoid having to delete/recreate all services to set their spec.loadbalancerClass (delete/recreate is required because spec.loadbalancerClass is read-only); each Service will be disrupted until it's updated ; changing the service definition may involve coordinating with the teams that deploy applications on your cluster, so this is quite a pain.

If you have "(2bis)", then the transition can be rather smooth: since no disruption happens on the service, the transition can spread over a time as long as needed. Each service will see a small disruption, but each of those is as short as delete/recreate, and can be handled by application teams.

Now, one question could be "why would I need to do a transition where I start to use --lb-class=x" ?

I would be tempted to simply say that this is generally a good idea to support any transition in deployment options. Perhaps this rationale is sufficient :)

As illustrations, here are be real-life scenarios:
* "incrementally transition from metallb to another LB implementation"
    * okay, maybe this isn't the most enticing scenario to describe to a metallb maintainer ;-), but you still might like to state that you trust sooo much metallb that you take the risk of making it easy for users to leave... ;)
* "incrementally transition from metallb-by-default to practices where an explicit lbclass is used"
* "introduction of an LB implementation that does not support loadbalancer classes"
    * (it is this last use-case that lead our team to look at this PR, we do platform engineering)
    * the scenario we have is the desire by some users to use the LB  implementation from cloud-provider -- cloud-provider LB implementations do not support lbClass (see [here](https://github.com/kubernetes/cloud-provider/blob/efee7eb51d29c985742dd272c3d8aa3c6bd90f04/controllers/service/controller.go#L864), in the code common to all cloud-provider implementation, Services with `service.Spec.LoadBalancerClass` are ignored) and at least the one we use (cloud-provider-openstack) does not do any further filtering, and will process all services having no spec.lbClass
    * ... as a result, to introduce this cloud-provider in these existing clusters which have metallb running without `--lb-class`, today our only option is to reconfigure metallb with `--lb-class`

With the above in mind, would you agree that the feature introduced by this PR is useful ?

(for reference only: kube-vip LB implementation [has](https://kube-vip.io/docs/usage/kubernetes-services/#optional-behavior) such a feature -- the reason, as far as I understand, is because kube-vip initial support for loadbalancer classes was only "(2bis)" which has other migrations problems and they wanted to also have "(2)", like metallb, but they had to make it optional to avoid making a breaking change)

fedepaol:
>       * (it is this last use-case that lead our team to look at this PR, we do platform engineering)
>       * the scenario we have is the desire by some users to use the LB  implementation from cloud-provider -- cloud-provider LB implementations do not support lbClass (see [here](https://github.com/kubernetes/cloud-provider/blob/efee7eb51d29c985742dd272c3d8aa3c6bd90f04/controllers/service/controller.go#L864), in the code common to all cloud-provider implementation, Services with `service.Spec.LoadBalancerClass` are ignored) and at least the one we use (cloud-provider-openstack) does not do any further filtering, and will process all services having no spec.lbClass
>       * ... as a result, to introduce this cloud-provider in these existing clusters which have metallb running without `--lb-class`, today our only option is to reconfigure metallb with `--lb-class`
> 
> 
> With the above in mind, would you agree that the feature introduced by this PR is useful ?
> 

Not sure I understood your use case. Here you are saying that cloud provider implementations do not support lb class, but iirc this was about having metallb caring about lbclass = empty and lbclass = metallb (or something like that).
Mind expanding the reason why you'd need this? IIUC the lbclass empty would be handled by the cloud provider's load balancer, so not sure I follow.


> (for reference only: kube-vip LB implementation [has](https://kube-vip.io/docs/usage/kubernetes-services/#optional-behavior) such a feature -- the reason, as far as I understand, is because kube-vip initial support for loadbalancer classes was only "(2bis)" which has other migrations problems and they wanted to also have "(2)", like metallb, but they had to make it optional to avoid making a breaking change)

tmmorin:
> IIUC the lbclass empty would be handled by the cloud provider's load balancer, so not sure I follow.

yes, you are correct

indeed for a setup like this, nothing is needed:
* metallb set with lbclass=metallb (handles only services with spec.lbClass=metallb)
* cloud provider handles the services which have no spec.lbClass

the scenario that calls for this feature is about _transitioning to such a setup_ 

for instance, this scenario is problematic:
* initial state: only metallb installed, without having been setup with lbclass (and no service has spec.lbClass)
* target state: the one described above (metallb set with lbclass=metallb, cloud provider handles the services which have no spec.lbClass)

This scenario is problematic because if you reconfigure metallb with lbclass=metallb then, at the moment you do it, metallb will stop handling the currently defined Services.

The only way to handle such a transition smoothly is to have an intermediate step with metallb configured to handle _both_ services  without spec.lbClass and the service with spec.lbClass=metallb.  When applying this configuration nothing will break, and you then get all the time you want to redefine the Service to let them have spec.lbClass=metallb.  And once you have covered them all, you can then enable the cloud-provider LB implementation  (and it will handle new Services on which spec.lbClass would be purposefully omitted to ensure that the cloud-provider LB implementation would manage them)

fedepaol:
Wrapping up on this as I just realized I never replied back. All in all, this feels a corner case that is not worth the (albeit tiny) complexity that is going to be added to the project.
Sorry for the delay in the reply :/

PR #2477, opened 2024-07-30
add readOnlyRootFilesystem securityContext to containers

NicoJDE:

/kind feature

This PR adds the securityContext.readOnlyRootFilesystem for the speaker Pod. 

Setting the readOnlyRootFilesystem flag to true reduces the attack surface of your containers since an attacker cannot manipulate the executables of your container, nor make any other changes to your root file system.

**Release note**:

```release-note
Sets readOnlyRootFilesystem for the Speaker Pods
```

### DIFF
--- a/charts/metallb/templates/speaker.yaml
+++ b/charts/metallb/templates/speaker.yaml
@@ -369,6 +369,7 @@ spec:
       {{- if .Values.speaker.frr.enabled }}
       - name: frr
         securityContext:
+          readOnlyRootFilesystem: true
           capabilities:
             add:
             - NET_ADMIN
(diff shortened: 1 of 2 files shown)


fedepaol:
Thanks! Can you add this change to the kustomize based configuration too (here https://github.com/metallb/metallb/blob/main/config/controllers/speaker.yaml and probably here https://github.com/metallb/metallb/blob/main/config/frr/speaker-patch.yaml ) and regenerate the manifest with inv ?

NicoJDE:
@fedepaol sry for the delay. I added the patch

PR #2488, opened 2024-08-09
Ability to install CRDs-only Metallb Chart

kiemlicz:
/kind feature

In my organization we need to install CRDs upfront during a separate process and we would like to use this Chart.  
Currently it is impossible to set the values that enable CRD only as `templates/webhooks.yaml` are installed unconditionally. Additionally `crds` Chart is not published anywhere  

Given my context of having CRDs-only installed, this PR is a slippery-slope.    
It doesn't guarantee that in future some change would introduce other resources thus breaking the assumption of having CRDs-only Chart.    
Alternative method would be for metallb organization to publish the CRDs only Chart (which seems to be already created in `charts/` directory)

**Release note**:

```release-note
Support of conditional webhooks installation
```

### DIFF
--- a/website/content/release-notes/_index.md
+++ b/website/content/release-notes/_index.md
@@ -3,6 +3,16 @@ title: Release Notes
 weight: 8
 ---
 
+## Version 0.14.8
+
+### Feature
+
+- Use frr-k8s 0.0.14. (#2467, @fedepaol)
+
+### Bug fixes
+
+- Fix the broken all-in-one/kustomize manifests from v0.14.7.
+
 ## Version 0.14.7
 
 ### Bug fixes


fedepaol:
You might want to comment out the conversion stanza for bgppeer https://github.com/metallb/metallb/blob/main/charts/metallb/charts/crds/templates/crds.yaml#L335 otherwise creation of bgppeers will be not allowed due to the service not being there.

kiemlicz:
Thank you for comment!  
This got auto-closed after commit message rewording (rebase) and force-pushing branch (commit lint fix)

[Here's updated PR](https://github.com/metallb/metallb/pull/2502)

sandeep-ongit:
+1

PR #2494, opened 2024-08-16
Define default holdtime value in CRD and extend keepalivetime validation

karampok:
/kind feature

- if no holdtime, then there is default from metallb code to 90 that is not visible to the user when fetching the bgppeer CR
- Keepalivetime value cannot be between 1/3 to 3/3 of holdtime, FRR just changes it. IMO is better if defined in the CR then this value to be landed in the frr config exactly or fail in validation

```
when the config `neighbor 192.168.111.1 timers  60 90` is entered in the frr process, the frr process alters the 60 to 30 
```

**Release note**:

```release-note
Define default holdtime value in CRD and extend keepalivetime validation to fail if larger than 1/3 of holdtime
```


karampok:
Draft because there are some DISCUSS points as comments, but needs review. Thanks

karampok:
hi @oribon could you take a look here, thanks!

internal/config/config.go
```diff
@@ -385,7 +382,13 @@ func peerFromCR(p metallbv1beta2.BGPPeer, passwordSecrets map[string]corev1.Secr
 	if ip == nil {
 		return nil, fmt.Errorf("invalid BGPPeer address %q", p.Spec.Address)
 	}
+
 	holdTime := p.Spec.HoldTime.Duration
+	// DISCUSS1: default value and validation can go away because
+	// it takes place directly in CRD, should we remove them?
+	// DISCUSS2: in RFC the value 0 is valid for holdtime and it means
+	//  BGP session will not expect any Keepalive messages, and it will not
+	//  maintain a session timeout due to inactivity. Do we want to handle it?
```
oribon, line 391: in this case (when 0 has a different meaning than unspecified) we should be using a pointer

internal/config/config.go
```diff
@@ -394,13 +397,16 @@ func peerFromCR(p metallbv1beta2.BGPPeer, passwordSecrets map[string]corev1.Secr
...
+	// BGP speaker (router) will not send Keepalive messages to its peer. Should
+	// we allow it? it might makes sense for BFD?
 	if keepaliveTime == 0 {
 		keepaliveTime = holdTime / 3
 	}
 
-	// keepalive must be lower than holdtime
-	if keepaliveTime > holdTime {
-		return nil, fmt.Errorf("invalid keepaliveTime %q must be smaller than holdtime %q",
+	// keepalive must be lower than 1/3 holdtime, otherwise FRR adapts it.
```
oribon, line 407: I think this code is shared between the modes, so relying on FRR to adapt it isn't always the case

PR #2495, opened 2024-08-16
Service annotations: change annotation prefix to metallb.io

### LINKED ISSUE #2392: Deprecate the metallb.universe.tf annotations and start handling metallb.io

metallb.universe.tf based annotations need to be deprecated and we should move to metallb.io based ones.
We should start slowly warning the users that the old ones are going to disappear soon or later, but in first instance we should handle both.

Handling both types of annotations, and I think the webhook can throw a warning to the user, so we could leverage it to notify that the old annotation is going to be deprecated.

### Additional context

_No response_

disco-stu:
This commit deprecates the old annotation prefix metallb.universe.tf in favor of metallb.io.

The introduced change is backwards compatible. E.g., if a services specifies a address-pool using the old prefix, it will still be accepted by the controller. However, if two address-pool annotations are specified the one with the newer prefix (metallb.io/address-pool) has precedence.

If a services uses the old prefix a deprecation warning is generated. The corresponding message will look similar to this one: "Warning: service object contains deprecated metallb annotation metallb.universe.tf/address-pool".

Fixed #2392

**Is this a FEATURE ?**:

/kind feature

Migrate from "metallb.universe.tf" annotation prefix to "metallb.io"

This pull request currently lacks e2e tests to check the usage of deprecated annotations. Moreover, it lacks documentation Adjustments. Therefore, this PR must be considered WIP.

**Release note**:

```release-note
Action Required: The annotation prefix "metallb.universe.tf" is deprecated by now. Please use the new prefix "metallb.io". E.g.  "metallb.io/address-pool" instead of "metallb.universe.tf/address-pool". The old prefix is still supported. However, migrating to the new prefix is highly recommend. Future versions might remove this backwards compatibility. 
```

### DIFF
--- a/controller/service.go
+++ b/controller/service.go
@@ -30,10 +30,18 @@ import (
 )
 
 const (
-	AnnotationPrefix             = "metallb.universe.tf"
+	AnnotationPrefix             = "metallb.io"
 	AnnotationAddressPool        = AnnotationPrefix + "/" + "address-pool"
 	AnnotationLoadBalancerIPs    = AnnotationPrefix + "/" + "loadBalancerIPs"
 	AnnotationIPAllocateFromPool = AnnotationPrefix + "/" + "ip-allocated-from-pool"
+	AnnotationAllowSharedIP      = AnnotationPrefix + "/" + "allow-shared-ip"
+
+	// Deprecated Annotations. Used for backward compatibility.
+	DeprecatedAnnotationPrefix             = "metallb.universe.tf"
+	DeprecatedAnnotationAddressPool        = DeprecatedAnnotationPrefix + "/" + "address-pool"
+	DeprecatedAnnotationLoadBalancerIPs    = DeprecatedAnnotationPrefix + "/" + "loadBalancerIPs"
+	DeprecatedAnnotationIPAllocateFromPool = DeprecatedAnnotationPrefix + "/" + "ip-allocated-from-pool"
+	DeprecatedAnnotationAllowSharedIP      = DeprecatedAnnotationPrefix + "/" + "allow-shared-ip"
 )
 
 var ErrConverge = fmt.Errorf("failed to converge")
@@ -104,7 +112,7 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 	if len(lbIPs) != 0 {
 		// This assign is idempotent if the config is consistent,
 		// otherwise it'll fail and tell us why.
-		if err = c.ips.Assign(key, svc, lbIPs, k8salloc.Ports(svc), k8salloc.SharingKey(svc), k8salloc.BackendKey(svc)); err != nil {
+		if err = c.ips.Assign(key, svc, lbIPs, k8salloc.Ports(svc), SharingKey(svc), k8salloc.BackendKey(svc)); err != nil {
 			level.Info(l).Log("event", "clearAssignment", "error", err, "msg", "current IP not allowed by config, clearing")
 			c.client.Infof(svc, "ClearAssignment", "current IP for %q not allowed by config, will attempt for new IP assignment: %s", key, err)
 			c.clearServiceState(key, svc)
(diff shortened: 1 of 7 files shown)


controller/service.go
```diff
@@ -30,14 +30,40 @@ import (
...
+	DeprecatedAnnotationAddressPool        = DeprecatedAnnotationPrefix + "/" + "address-pool"
+	DeprecatedAnnotationLoadBalancerIPs    = DeprecatedAnnotationPrefix + "/" + "loadBalancerIPs"
+	DeprecatedAnnotationIPAllocateFromPool = DeprecatedAnnotationPrefix + "/" + "ip-allocated-from-pool"
+	DeprecatedAnnotationAllowSharedIP      = DeprecatedAnnotationPrefix + "/" + "allow-shared-ip"
 )
 
 var ErrConverge = fmt.Errorf("failed to converge")
 
+// SharingKey extracts the sharing key for a service.
+func SharingKey(svc *v1.Service) string {
```
fedepaol, line 50: why not changing the implementation in k8salloc?

controller/service.go
```diff
@@ -30,14 +30,40 @@ import (
...
+	if _, ok := svc.Annotations[AnnotationAllowSharedIP]; ok {
+		return svc.Annotations[AnnotationAllowSharedIP]
+	}
+	return svc.Annotations[DeprecatedAnnotationAllowSharedIP]
+}
+
+func getAnnotationValue(str ...string) string {
```
fedepaol, line 57: I'd:
- move this to the bottom as this is an "utility-like" function
- rework this by providing the annotations map, the supported annotation and the deprecated annotation:

`func valueForAnnotation(annotations map[string]string, stableAnnotation, deprecatedAnnotation) string` as its implementation will be more explicit than fetching the first one and relying on the caller setting the first as the stable one

fedepaol:
Thanks a bunch for this! Left a couple of comments, plus I have the following considerations:
- we are not e2e-ing anymore the old annotations but users might be still using them. I don't see a good / smart way to test both (and it's too expensive), but it'd make sense to have a handful of tests that cover the old annotations.

- I am not entirely sold over the complexity of adding an extra webhook covering all the services. The failurePolicy is ignore, so should be good but we've been bitten already multiple times by validating webhooks, that I don't feel super comfortable in adding them to something that is cluster-wide such as services


cc @oribon for giving a look too

controller/service.go
disco-stu: My modification uses some of the constants defined in service.go (package controller). However, importing controller into k8salloc produces an import loop and I don't liked the idea of defining the annotations multiple times in different packages. Moreover, SharingKey(...) is only used in service.go.

controller/service.go
disco-stu: I changed the implementation based on your input and moved the function to the bottom of the file as you requested.

oribon:
> am not entirely sold over the complexity of adding an extra webhook covering all the services.

same, not sure it's worth the overhead for the annotation (affecting all services in the cluster). I guess a log/event saying "you're using a deprecated annotation, consider using x" would be more than enough

fedepaol:
> > > am not entirely sold over the complexity of adding an extra webhook covering all the services.
> > 
> > 
> > same, not sure it's worth the overhead for the annotation (affecting all services in the cluster). I guess a log/event saying "you're using a deprecated annotation, consider using x" would be more than enough
> 
> I actually have slight concerns too. I have implemented the warning as a webhook for two reasons:
> 
>     1. description in issue [Deprecate the metallb.universe.tf annotations and start handling metallb.io #2392](https://github.com/metallb/metallb/issues/2392)
> 
>     2. a user who creates a load balancing service does not necessarily have to have access to the logs of MetalLB
> 
> 
> My suggestion would therefore be to generate an event for the service object.

+1 for the event. No one reads logs, even if they have access to them :P

controller/service.go
```diff
@@ -98,13 +106,21 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 		}
 	}
 
+	// Check for any deprecated annotations
+	for key := range svc.Annotations {
```
fedepaol, line 110: if we put this here, we'll generate an event (for every deprecated annotation) each time we receive an update of the service or the configuration changes (this is because we call this function on every service update, or config change, to make sure the allocation is still valid), and we might risk to flood the cluster.

One possible decent tradeoff is to trigger this check and the events only when the service gets the ip. This means we won't warn users if they have an invalid configuration, but I think it's a good compromise and better than generating too many events, probably somewhere around here https://github.com/metallb/metallb/blob/cd0b6fe9ae1375e0e696c40d040cfc6f86465e42/controller/main.go#L89-L90

(please leave also a comment with the reason why we are doing it there, as it feels a bit odd to have it in the middle there).

PR #2496, opened 2024-08-19
Ability to specify extra annotations for CRDs

kiemlicz:
/kind feature

Please check [associated PR](https://github.com/metallb/metallb/pull/2488) for more context.  
I want to move CRDs installation into separate Chart (this is addressed in aforementioned PR) and mark these resources using special annotations like: [`resource-policy`](https://helm.sh/docs/howto/charts_tips_and_tricks/#tell-helm-not-to-uninstall-a-resource) - hence this PR

Added new key: `.crds.annotations` with default value of `{}`

**Release note**:

```release-note
Added ability to configure extra annotations for CRDs
```


fedepaol:
If the goal is to add `helm.sh/resource-policy: keep`, would it make sense to have a custom flag instead of a generic parameter?

charts/metallb/charts/crds/templates/crds.yaml
```diff
@@ -3,6 +3,9 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.14.0
+    {{- with .Values.annotations }}
```
fedepaol, line 6: this will fail CI because the crds are generated automatically from the ones under config. You should change inv so it injects this section

kiemlicz:
> If the goal is to add `helm.sh/resource-policy: keep`, would it make sense to have a custom flag instead of a generic parameter?

Really depends on your take on this matter and what are your guidelines, just let me know and I'll adjust.  

Personally I don't like being that specific in regards of annotations (flag that will attach that single annotation only).   
Just let the user decide and don't take his freedom.  
Maybe so other tool used for deployment along with Helm would use `helmBrother.sh/resource-policy: reallyKeep` and then Chart upgrade is needed, whereas with configurable annotation user can adjust on its own

charts/metallb/charts/crds/templates/crds.yaml
kiemlicz: I've added generation of these `{{` annotations in some as little-hacky-as-possible way 
Please re-check

config/crd/kustomization.yaml
```diff
@@ -14,3 +14,12 @@ patches:
   target:
     kind: CustomResourceDefinition
     name: bgppeers.metallb.io
+- patch: |-
+    - op: add
+      path: "/metadata/annotations/metallb-internal"
```
kiemlicz, line 19: cannot omit `metallb-internal`, possibly due to [kustomize bug](https://github.com/kubernetes-sigs/kustomize/issues/4786#issuecomment-1292356615)

kiemlicz:
The tests should pass (unit/lint), I'm unsure about e2e - they yield non-deterministic results :thinking:

kiemlicz:
not stale

kiemlicz:
Why is this kept for that long?  
If there are any comments or objections - I need to hear/see them

fedepaol:
> Why is this kept for that long? If there are any comments or objections - I need to hear/see them

Sorry, I think last time I checked CI was not entirely happy, then I got dragged by other stuff. I'll go over this soon-ish, in the meantime would you mind squashing the commits? Thanks!

config/manifests/metallb-frr-k8s-prometheus.yaml
```diff
@@ -12,6 +12,10 @@ kind: CustomResourceDefinition
 metadata:
   annotations:
     controller-gen.kubebuilder.io/version: v0.16.3
+    metallb-internal: |-
```
fedepaol, line 15: I am actually surprised this works!

kiemlicz:
Thanks, will check and update

kiemlicz:
Do you have other comments besides squashing?

Issue #2423, opened 2024-06-06
Add support for BGP on link-local addresses


doka380:
### Is your feature request related to a problem?

At the moment, it's not possible to use link-local addresses for building BGP peering, since there is no ability to specify interface where peer can be found. Value e.g. `fe80::xxxx:xxxx:xxxx:xxxx%enp1s0` for `peerAddress` in `BGPPeer` rejected with the error `Error from server (Forbidden): error when creating "metlb-conf.yaml": admission webhook "bgppeersvalidationwebhook.metallb.io" denied the request: parsing peer gk-peer-v6: invalid BGPPeer address "fe80::b8a3:17ff:fe0c:c54a%enp1s0"`

### Describe the solution you'd like

This is already supported by FRR:
```
 neighbor fe80::a876:4dff:fe77:408 remote-as 64501
 neighbor fe80::a876:4dff:fe77:408 interface enp1s0
```
and the second line of the config above (interface specification) can be generated based on the following BGPPeer specification:
```
spec:
  peerASN: 64501
  peerAddress: fe80::a876:4dff:fe77:408
  peerInterface: enp1s0
```

### Additional context

This change will allow to use any available addressing in IPv6 environment, including link-local addresses.

There was a discussion here about this - https://kubernetes.slack.com/archives/C8FSTGU56/p1717574500705649

### I've read and agree with the following

- [X] I've checked all open and closed issues and my request is not there.
- [X] I've checked all open and closed pull requests and my request is not there.

fedepaol:
Thanks for raising this! We are working on it!
At some point we'll create a design proposal.

doka380:
Just to make it clear - this is not something exotic, the same approach also implemented in another routing software - Bird - https://bird.network.cz/?get_doc&v=20&f=bird-6.html#ss6.4 -
```
interface string
     Define interface we should use for link-local BGP IPv6 sessions. Interface can also be specified
     as a part of neighbor address (e.g., neighbor fe80::1234%eth0 as 65000;).
```

karampok:
@fedepaol  you could assign that to me, it might relevant to the unnumber BGP peering (as we would add interface name in the config), thanks

PR #2508, opened 2024-08-30
Unnumbered BGP Peering design document

karampok:
/kind design

In part of this PR there is
1. A design document about unnumbered BGP peering, discussing implementation and API changes
2. A draft version of how an e2e test can run, more for discussion rather than actually merging that file in that PR.

My plan would be that first we discuss how the raw frr config would look like, then the API in frr-k8s, then in metallb.

Thanks

**Release note**:

```release-note
None
```

### DIFF
--- a/design/unnumbered-bgp.md
+++ b/design/unnumbered-bgp.md
@@ -1,0 +1,531 @@
+# MetalLB support for BGP Unnumbered
+
+## Summary
+
+BGP Unnumbered allows routers to peer with each other when direct connected
+without the need for IPs (the BGP configuration uses interface names). Once
+peering takes places, exchange IPv4 or IPv6 prefixes can take place. That
+feature can be used by MetalLB to simplify configuration both on the network
+fabric and on the cluster side.
+
+## Motivation
+
+Using BGP unnumbered peering, which dynamically discovers IPV6 neighbors,
+reduces the burden of the network administrator to configue all
+interfaces on the network fabric or on the cluster nodes to have IPv4
+addressing just for the BGP peering. By using BGP unnumbered, cluster administrator
+avoids to specify the address of each neihbor. Unnumbered BGP utilizes IPv6
+link local address to automatically decide which peer to connect to.
+
+## Limitation/Requirements
+
+- To use unnumbered BGP feature, there must be a link (point-to-point, directly
+  connected) between the interfaces two BGP speakers are using
+- IPv6 must be enabled on that interfaces, and with LLA (Link Local IPv6 Address) enabled.
+  Note: the use of IPv6 LLA does not require a full IPv6 deployment
+- IPv6 RAs (Router Advertisments) must be enabled (it should be safe when p2p and no IP address on the interface).
+
+## Goals
+
+- Make the BGP peering from Server to the ToR (TopOfRack) easier to configure
+  and have the same configuration for all servers (assuming the interfaces' names
+  are the same across the cluster nodes)
+- Remove the task of network admins to setup up addressing on those interfaces
+  if it is needed only for BGP peering
+
+## Non Goals
+
+- Make it work over no point-to-point connection
+- Define/Discuss dual ToR architecture, on top of bond interfaces or other HA strategies
+- [Dynamic BGP Peering](https://ipwithease.com/dynamic-bgp-peering/) is a
+  feature different than unnumbered BGP. Both feature reduce the configuration
+  needed for the BGP peering. The main difference is that Dynamic BGP peering
+  still needs IP addresses and does not require point-to-point links.
+
+## User Stories
+
+### Story 1
+
+As cluster administrator, I want to configure metallb without configuring IP addresses.
+
+### Story 2
+
+As network administrator, I want to avoid configuring IP addresses for
+point-to-point connection just for the BPG peering reason.
+
+### Story 3
+
+As network administrator, I want to reduce the size of the configuration in the ToR router.
+
+## The API Changes
+
+A new field will be introduced, `Interface string` which if defined then Unnumber
+BGP takes place. In that case, the address must NOT have value. The API doc will
+look like:
+
+```
+Interface, if defined, instructs Metallb to setup Unnumber BGP peering on the interface,
+which means that the peering configuration only includes the interface and does
+not include IP address. Address and Interface are mutually exclusive and one of
+them must be specified.
+```
+
+#### FRR-K8S
+
+```go
+type Neighbor struct {
+  Address string `json:"address"`
+  Interface string `json:"interface"`
+}
+```
+
+##### MetalLB
+
+```go
+type BGPPeerSpec struct {
+  Address string `json:"peerAddress"`
+  Interface string `json:"interface"`
+}
+```
+
+Example CR
+
+```yaml
+apiVersion: metallb.io/v1beta2
+kind: BGPPeer
+metadata:
+  name: peer-unnumber
+  namespace: metallb-system
+spec:
+  myASN: 64512
+  ASN: 645000
+  interface: net0
+```
+
+#### Rejected Alternative - Re-Use Address field
+
+We can re-use `Address string` to either be IP or interface, if not valid IP we
+indicate Unnumbered BGP peering. This looks the minimal API, but it will not
+have clean error path. E.g. if the user input is not a valid IP, the
+implementation will shadow all the existing errors and will fail in the
+runtime.
+
+#### Rejected Alternative - Add new Unnumber bool field
+
+We can introduce a new field `Unnumber bool`, which if true the we indicate
+Unnumber BGP Peering. If `true` the Address field holds the interface value.
+Bool option makes the configuration error-free, because the user declares
+explicitly that they want Unnumber and there is no "guessing" what metallb
+does. Unnumber case is disconnected from LLA support case.
+
+### Design details
+
+To make unnumber BGP work, we need to modify the neighborsession
+as defined in the following template [neighborsession.tmpl](internal/bgp/frr/templates/neighborsession.tmpl)
+and make the following changes
+
+1. Every time there is FRR directive that uses the address, to use the
+   interface name
+
+```
+before:  neighbor 192.168.10.10 timers 180 540
+after:  neighbor net0 timers 180 540
+```
+
+2. The `remote-as` directive must be replaced
+
+```
+before: neighbor 192.168.10.10 remote-as 65004
+after: neighbor net0 interface remote-as 65004
+```
+
+3. There is currently a single directive that does not accept interface.
+   Therefore it must be always excluded.
+
+```
+neighbor {{.neighbor.Addr}} disable-connected-check
+```
+
+### Topics
+
+#### IPv6 RAs (Route Advertisments) enabled on both ends of the link
+
+According to the design of unnumber BGP peering, we need to enable RAs. By
+observation it seems that FRR/Zebra component is auto-enabling RAs when the BGP
+unnumber session is configured. This is the [commit](https://github.com/FRRouting/frr/commit/4a04e5f7968d46e446bdb530724da18b90b7f5f8).
+Therefore the Metallb implementation does not require that RAs are explicitly
+enabled in the FRR config unless we have other reasons (e.g. have different period).
+
+#### Do I need an IPv{4,6} to enable dataplane traffic
+
+There is no need for the interface to have an IP in order the BGP
+session to be established. This is the core of unnumber BGP. Nevertheless,
+when doing BGP peering between a non-router component and a route, then an IP
+might be needed for the dataplane traffic to work.
+
+**Note:** the rest of the section is outside the scope of Metallb and is related
+to the CNI that cluster is being using. We discuss it as that might be important.
+
+##### Ingress traffic to LB IP and BGP route learning
+
+When external client (outside peer router) talks to an LB address and there is BGP route learning
+(= local k8s nodes adds routing from peer) **there is no need for an IP**
+
+```
+# packet in/out of k8s node
+net0  In  IP 200.100.100.1.40524 > 5.5.5.5.80: Flags [S], seq 1203290349,
+veth3e626b76 Out IP 200.100.100.1.40524 > 10.244.2.4.80: Flags [S], seq 1203290349,
+veth3e626b76 In  IP 10.244.2.4.80 > 200.100.100.1.40524: Flags [S.], seq 3181148864, ack 1203290350,
+net0  Out IP 5.5.5.5.80 > 200.100.100.1.40524: Flags [S.], seq 3181148864, ack 1203290350,
+
+# k8s node
+kind-worker:/# ip --br add show
+    lo               UNKNOWN        127.0.0.1/8 ::1/128
+    veth2c551257@if2 UP             10.244.2.1/32 fe80::1498:dbff:fe25:6e1b/64
+    eth0@if325       UP             172.20.20.4/24 2001:172:20:20::4/64 fe80::42:acff:fe14:1404/64
+    eth1@if330       UP             172.18.0.3/16 fc00:f853:ccd:e791::3/64 fe80::42:acff:fe12:3/64
+    net0@if3         UP             fe80::dcad:beff:feff:1161/64
+kind-worker:/# ip route
+    default via 172.20.20.1 dev eth0
+    10.244.0.0/24 via 172.20.20.3 dev eth0
+    10.244.1.0/24 via 172.20.20.2 dev eth0
+    10.244.2.3 dev veth2c551257 scope host
+    172.18.0.0/16 dev eth1 proto kernel scope link src 172.18.0.3
+    172.20.20.0/24 dev eth0 proto kernel scope link src 172.20.20.4
+    172.30.0.0/16 via 172.20.20.6 dev eth0
+    200.100.100.0/24 nhid 68 via inet6 fe80::2c5f:eff:fec4:cf7b dev net0 proto bgp metric 20
+kind-worker:/# ip route get 200.100.100.1
+    200.100.100.1 via inet6 fe80::2c5f:eff:fec4:cf7b dev net0 src 172.20.20.4 uid 0
+        cache
+```
+That works as long as the NAT/Masq rule properly sets back the source IP to be the LB
+which depends at the end of the CNI/use or not kube-proxy.
+
+##### Ingress traffic to LB IP and static routes
+
+This does NOT work without IP on the interface. Even if route is configured
+with the interface `ip route add 200.100.100.0/24 dev net0` the traffic breaks
+due to ARP.
+
+```
+net0  In  IP 200.100.100.1.43198 > 5.5.5.5.80: Flags [S], seq 3169002768,
+veth3e626b76 Out IP 200.100.100.1.43198 > 10.244.2.4.80: Flags [S], seq 3169002768,
+veth3e626b76 In  IP 10.244.2.4.80 > 200.100.100.1.43198: Flags [S.], seq 490740861, ack 3169002769,
+net0  Out ARP, Request who-has 200.100.100.1 tell 172.20.20.4, length 28
+net0  Out ARP, Request who-has 200.100.100.1 tell 172.20.20.4, length 28
+```
+
+[RFC 8950](https://datatracker.ietf.org/doc/html/rfc8950) explains why we do not need ARP in previous case.
+
+##### Egress traffic to LB IP and BGP route learning
+
+When pod in a node that has the routing entry due BGP peering
+```
+kind-worker:/# ip route
+    200.100.100.0/24 nhid 68 via inet6 fe80::2c5f:eff:fec4:cf7b dev net0 proto bgp metric 20
+kind-worker:/# ip route get 200.100.100.1
+    200.100.100.1 via inet6 fe80::2c5f:eff:fec4:cf7b dev net0 src 172.20.20.4 uid 0
+kind-worker:/# ip a s eth0
+324: eth0@if325: <BROADCAST,MULTICAST,UP,LOWER_UP> mtu 1500 qdisc noqueue state UP group default
+    link/ether 02:42:ac:14:14:04 brd ff:ff:ff:ff:ff:ff link-netnsid 0
+    inet 172.20.20.4/24 brd 172.20.20.255 scope global eth0
+       valid_lft forever preferred_lft forever
+```
+initiates a connection to the outside, the source IP of the packet is one of the other interfaces.
+What is the selection algorithm is unclear and kernel documentation could
+answer. The important information is that if IP address is added in the
+loopback interface, then that IP address is being used. There is not need for
+an IP in the peer interface.
+
+```
+kind-worker:/# ip a s lo
+1: lo: <LOOPBACK,UP,LOWER_UP> mtu 65536 qdisc noqueue state UNKNOWN group default qlen 1000
+    link/loopback 00:00:00:00:00:00 brd 00:00:00:00:00:00
+    inet 127.0.0.1/8 scope host lo
+       valid_lft forever preferred_lft forever
+    inet 12.12.12.12/32 scope global lo
+       valid_lft forever preferred_lft forever
+    inet6 ::1/128 scope host
+       valid_lft forever preferred_lft forever
+kind-worker:/# ip route get 200.100.100.1
+200.100.100.1 via inet6 fe80::cc61:c8ff:fefc:2dce dev net0 src 12.12.12.12 uid 0
+    cache
+```
+So the answer is IP address is needed, but configuring IP address in the loopback is enough
+to control it.
+
+##### Egress traffic to LB IP and static routes
+
+The interface needs IP address to setup source IP address and routing.
+
+## Test Plan
+
+### E2E test for {i,E}Unnumbered on primary interface
+
+```mermaid
+graph TD
+    subgraph Rack0
+        direction LR
+        subgraph ToR0
+            direction TB
+            A1[eth0] -.- A{Routing}
+            A -.- A3[eth00]
+            A -.- A4[eth01]
+
+            BGP(BGP)
+
+        end
+
+        %% Top Right Subgraph
+        subgraph Node1
+            direction TB
+            B1 --- B2(cni)
+            B2 --- B3
+            subgraph Pod1[pod]
+              B3[eth0]
+            end
+            BGP-N1(BGP process)
+        end
+
+        %% Bottom Right Subgraph
+        subgraph Node0
+            direction TB
+            C1 --- C2(cni)
+            C2 --- C3
+            subgraph Pod2[pod]
+              C3[eth0]
+            end
+            BGP-N0(BGP process)
+        end
+
+        %% Connections between subgraphs
+        A3 -- point-to-point ------ B1[eth0]
+        A4 -- point-to-point ------ C1[eth0]
+    end
+
+    %% Define styles for subgraphs and nodes
+    classDef node fill:#f0f0f0,stroke:#000000,stroke-width:1px;
+    class Node0 node;
+    class Node1 node;
+```
+
+
+
+There is no straightforward way to test unnumbered in the primary node
+interface. The testing setup (kind creation) creates bridge and connect
+containers. Ideally we need to setup a kind cluster, and replace the bridge
+with a container that runs FRR on each other side of the veth, and at the same
+time to provide switch/upstream. Until we find a way to automate that infra setup
+we will test that scenario manually.
+
+### E2E test for Unnumbered on secondary interface
+
+```mermaid
+graph TD
+    subgraph Rack0
+        direction LR
+        style Rack0 fill:#ffffff,stroke:#000000,stroke-width:2px
+
+        subgraph ToR0
+            style ToR0 fill:#ffffff,stroke:#000000,stroke-width:2px
+            direction TB
+            subgraph Default
+              direction LR
+              A1[net0] -.- A{Routing}
+              A -.- A3[eth00]
+              A -.- A4[eth01]
+            end
+            subgraph RED VRF*
+              direction LR
+              AA1[net00] -.- AA{Routing}
+              AA -.- AA3[eth00.vlanRED]
+              AA -.- AA4[eth01.vlanRED]
+              BGP-T0(BGP configured on RED)
+            end
+
+        end
+
+        subgraph Node0
+            direction TB
+            B1 --- B2(cni)
+            B2 --- B3
+            BB1
+            subgraph Pod1[pod]
+              B3[eth0]
+            end
+            BGP-N0(BGP process)
+        end
+
+        subgraph Node1
+            direction TB
+            C1 --- C2(cni)
+            C2 --- C3
+            CC1
+            subgraph Pod2[pod]
+              C3[eth0]
+            end
+            BGP-N1(BGP process)
+        end
+
+        %% Connections between subgraphs
+        A3 ------- B1[eth0]
+        AA3 -- BGP Peering ----- BB1[net0]
+        A4 ------- C1[eth0]
+        AA4 -- BGP Peering ----- CC1[net0]
+    end
+
+    %% Define styles for subgraphs and nodes
+    classDef node fill:#f0f0f0,stroke:#000000,stroke-width:1px;
+    class Node0 node;
+    class Node1 node;
+```
+
+We should modify the testing infra to include point-to-point links which means
+connect containers using VETH, and not through a bridge. We need NET-ADMIN cap
+(or sudo) when creating the VETH.
+
+#### Option 1 - have a separate test
+
+We could have an isolated test file that creates/deletes peer per that specific tests.
+In that option we do not to take resources from runners outside the runtime of
+the test. This new peering setup needs different configuration (not IPs but interface
+names) and therefore avoid extensive(?) refactoring of the help functions.
+
+#### Option 2 - extend the FRR Containers and have all test run on top
+
+Benefit of that would be that we will cover all the feature/scenarios (e.g. uncordon node).
+Reason not to do that would be add extra time, create many skip rules (e.g. if Multihop skip that Peer),
+extensive(?) refactor of the test functions.
+
+### An example ToR configuration when based on FRR
+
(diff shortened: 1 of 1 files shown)


design/unnumbered-bgp.md
```diff
@@ -0,0 +1,436 @@
...
+
+    %% Define styles for subgraphs and nodes
+    classDef node fill:#f0f0f0,stroke:#000000,stroke-width:1px;
+    class Node0 node;
+    class Node1 node;
+```
+
+## Design details
+
+The target `show running-config` in the FRR process running in the k8s node
```
oribon, line 161: can you make this more alike the config we have today for the speakers? also, can you add a few words about the need for peergroups (iirc it's because the usual directives do not work with the interface's name)

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,436 @@
...
+Peer groups 1, using 64 bytes of memory
+
+Neighbor                  V         AS   MsgRcvd   MsgSent   TblVer  InQ OutQ  Up/Down State/PfxRcd   PfxSnt Desc
+unnumbered-p2p-peer(net0) 4      65004       512       512        1    0    0 00:25:15            0        1 TOR
+
+Total number of neighbors 1
+```
+### Topics
+
+#### Remote-AS
```
oribon, line 310: not really sure how this relates here (it says here we can specify internal/external/asn which fits with any model), is there something special about this in regards to unnumbered?

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,436 @@
...
+
+#### frr-k8s side
+
+```go
+...
+
+type Neighbor struct {
+  Description string `json:"description"` // optional
+  Address string `json:"address"` // must be empty if interface != nil
+  Interfaces []string `json:"interface"` //
```
oribon, line 352: why is this an array?

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,436 @@
...
+### The API Extension
+
+#### frr-k8s side
+
+```go
+...
+
+type Neighbor struct {
+  Description string `json:"description"` // optional
+  Address string `json:"address"` // must be empty if interface != nil
```
oribon, line 351: does this really need to be empty? could both being specified solve anything for https://github.com/metallb/metallb/issues/2423?

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,436 @@
...
+  Interfaces []string `json:"interface"` //  
+  PeerGroup string `json:"peerGroup"` // optional, otherwise value <interface>-peergroup
```
oribon, line 353: is there a reason for exposing this to the user (in particular as part of "unnumbered")?

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,436 @@
...
+
+#### BGP RouterID
+
+//TODO is there a requirement for that?
+
+### Traffic forwarding
+
+FRR advertises IPv4 prefixes with IPv6 next hops using encoding specified in RFC 5549 (now RFC 8950).
+
+### The API Extension
```
oribon, line 342: nit: I think it would be clearer to say which fields you think should be added/modified, instead of the complete struct as it is today

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,436 @@
...
+
+```go
+...
+
+type Neighbor struct {
+  Description string `json:"description"` // optional
+  Address string `json:"address"` // must be empty if interface != nil
+  Interfaces []string `json:"interface"` //  
+  PeerGroup string `json:"peerGroup"` // optional, otherwise value <interface>-peergroup
+  RemoteAS string `json:"remoteAS"` // it can be "ASN number or internal or external [or auto]"
```
oribon, line 354: like the other comment, not sure this is an "unnumbered" thing

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,436 @@
...
+```
+
+#### MetalLB
+
+```go
+...
+type BGPPeerSpec struct {
+	MyASN uint32 `json:"myASN"`
+	ASN uint32 `json:"peerASN"` // or internal or external
+	Address string `json:"peerAddress"` // or interface name e.g. net0
```
oribon, line 375: what's the reason for address/interface being separate in frr-k8s but not here?

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,436 @@
...
+
+### User Stories
+
+There is direct link interface between the node (FRR process acting as BGP peer) and the
+ToR (acting as an BGP router).
+
+
+#### Story 1 - Unnumbered {i,E}BGP on primary node interface
+
+```mermaid
```
oribon, line 52: neat!

e2etest/bgptests/unnumber.go
```diff
@@ -0,0 +1,264 @@
...
+
+var _ = ginkgo.FDescribe("FRRK8S-MODE DUALSTACK - Unnumbered BGP", func() {
+	defer ginkgo.GinkgoRecover()
+	var (
+		testNamespace string
+		peers         []*frrcontainer.FRR
+	)
+
+	ginkgo.BeforeEach(func() {
+		c := frrcontainer.Config{
```
oribon, line 40: the goal is to have this as part of the infra setup to make sure everything works with unnumbered?

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,436 @@
...
+
+
+### Testing
+
+#### E2E tests
+
+- Modify the testing infra to include point-to-point links which means
+connect containers using VETH, and not through a bridge. We need NET-ADMIN cap (or sudo) when creating the VETH
+- Create/Delete peer per specific test not to take resources from runners outside the runtime of the test
+- This peer needs different configuration (not IPs, interface names), keep a raw config
```
oribon, line 426: iiuc this will all be covered in the initial implementation?

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,436 @@
...
+- There is no straightforward way to test unnumbered in the primary node interface. The testing setup (kind creation)
```
oribon, line 427: I wonder if there's some value in adding another worker+container that are peered using unnumbered (no p2p, but there will be no conflicts as they're the only routers in the network using it) to get some feedback on peering with the primary node interface? wdyt?

oribon (commented):
thanks! great work, did a first pass

design/unnumbered-bgp.md
karampok: I will go deeper on what we have today in the speaker, that part if not very clear.

About PeerGroups:
1. Peergroups is the only way I managed to make it working, and all reference/examples did make use of them. So I observe we need them but I did not find doc.
2. I think the case of having two ToRs switch will be a needed, so I am thinking that we can have PG with two interfaces assigned.

design/unnumbered-bgp.md
karampok: I thought that unnumber BGP referred to both interface without values, and ASN without values, but not be the case.

What only confuses me is that 
[neighbor PEER interface remote-as <internal|external|auto|ASN>](https://docs.frrouting.org/en/latest/bgp.html#clicmd-neighbor-PEER-interface-remote-as-internal-external-auto-ASN)

which  talks about unnumber, but when I added this directive  did not work  as expected. in my lab.

I would suggest that this doc documents the API changes for that also, but I will not work on the implementation.

design/unnumbered-bgp.md
karampok: To support the dual ToRs case.

design/unnumbered-bgp.md
karampok: Not really, it could be hidden and taken an internal value (if we decided using peergroups)

e2etest/bgptests/unnumber.go
karampok: My preference would be to create the new container the unnumber peering as part of the test, to avoid having a container running for the full rutime of the tests. Do you see a reason to move it to the infra setup?

design/unnumbered-bgp.md
karampok: Can you elaborate what you mean?

design/unnumbered-bgp.md
karampok: Do you mean to have a single worker as speaker, and have single external container frr connected on the same docker network bridge? 

My concern is that locally the linux bridge does not forwarding RA when LLA address is the source (at least in my fedora laptop).

For sure we need to verifiy how that works on primary interface, and especially when the interface has also a non-LLA address configured.

design/unnumbered-bgp.md
karampok: Not really, it could be hidden and taken an internal value (if we decided using peergroups)

design/unnumbered-bgp.md
oribon: > Peergroups is the only way I managed to make it working, and all reference/examples did make use of them. So I observe we need them but I did not find doc.

right, this is what I meant by "the usual directives do not work with the interface's name" :) let's add this to this section so it's clear why we suddenly need peer groups.

> I think the case of having two ToRs switch will be a needed, so I am thinking that we can have PG with two interfaces assigned.

I see, but I guess this could be achieved by just having 2 different peers?

design/unnumbered-bgp.md
oribon: >  would suggest that this doc documents the API changes for that also

I think it would be better to keep this focused on unnumbered, when we get to implementing that feature (either if it's before or after this is implemented) we can see how they work together.

design/unnumbered-bgp.md
oribon: so I think we should drop this for now so the doc stays focused on unnumbered, when/if we decide peer groups are valuable as part of the api we'll tackle them separately :)

e2etest/bgptests/unnumber.go
oribon: > Do you see a reason to move it to the infra setup?

to run all the existing tests against the "unnumbered" scenario (unless we decide we're good with having only a single test for unnumbered)

design/unnumbered-bgp.md
oribon: asking if these points will be covered or there's still a question about them, while the last point (about primary interfaces) still needs to be thought of

design/unnumbered-bgp.md
oribon: > My concern is that locally the linux bridge does not forwarding RA when LLA address is the source (at least in my fedora laptop).

oh, didn't you run into the "random" peering we talked about?

> Do you mean to have a single worker as speaker, and have single external container frr connected on the same docker network bridge?

right (adding another worker / designating one), then the "randomness" will be okay - since there are only 2 peers on the same network using unnumbered they'll always peer with each other

design/unnumbered-bgp.md
karampok: ok, I will go through to see how two neighbors would look like, and check again if Peer Group can be avoided

design/unnumbered-bgp.md
karampok: sounds good, I will remove that section

design/unnumbered-bgp.md
karampok: sounds good, I will not add any public API about PG

e2etest/bgptests/unnumber.go
karampok: I see your point, but "all existing" means double the time and mabye adding complexity in the setupinfra. But no strong opinion, we can come back to that  on a later step.

oribon: > means double the time

not sure about that, it would just be another container as part of the tests

> mabye adding complexity

probably, this pr is a good place to understand if it's reasonable or not :)

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,498 @@
...
+Linux sends and receives IPv4 packets over an interface that does not have an IPv4 address.
+It does not look we need to add address in the loopback.
+
+#### RA enabled on both ends of the link
+
+According to the design of unnumber BGP peering, we need to enable RAs.
+
+#### BGP RouterID
+
+//TODO is there a requirement for that?
```
oribon, line 317: wdym here?

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,498 @@
...
+Unnumbered BGP utilizes IPv6 link local address to automatically decide who to
+connect.
+
+## Limitation/Requirements
+
+- To use unnumbered BGP feature, there must be a link (point-to-point, directly
+connected) between the interfaces two BGP daemons are using
+- IPv6 must be enabled on that interfaces, and in specific LLA (link Local IPv6 address) enabled
+(but the use of IPv6 LLA does not require a full IPv6 deployment)
+- RAs must be enabled (it should be safe when p2p and no IP address on the interface TODO)
```
oribon, line 24: is there still a TODO here?

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,498 @@
...
+
+#### Option 1. Re-Use Address field
+
+We can re-use `Address string` to either be IP or interface, if not valid IP we
+indicate Unnumbered BGP peering. This looks the minimal API, but it will not
+have clean error path. E.g. if the user input is not valid IP, the
+implementation will shadow all the existing errors and will fail in the
+runtime.
+
+#### Option 2. Add a new Interface field
```
oribon, line 362: I think this is the most valid one, and one that matches FRR the most in a sense (since it seems there's a distinction between interfaces and addresses) without locking us. Also, the docs don't need to be as complicated (stating all of the cases) as it's simply following the what frr is doing, that is we can just shortly say, if address is not specified the peering happens over the interface (unnumbered), if address is specified it is assumed the it is lla.

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,498 @@
...
+// == extra FRR directive `neighbor fe80::a876:4dff:fe77:408 interface eth0`
+// 4. If address is LLA and empty interface, then fail
+// 5. If address is empty and interface has value then the BGP unnumber peering config will take place
+// ==  modify FRR directive `neighbor PEER interface remote-as <ASN>`
+// && replace any address directive to use interface e.g.  `neighbor net0 timers 180 540`
+// && enable RAs on this interface
+// 6. If address has value (non-LLA) and interface non empty, then fail
+```
+That option might be not ideal (subjective opinion) because documentation looks complicated,
+and user might have to do trial-and-error approach. Plus user needs to understand corner cases.
```
oribon, line 382: the user will need to understand the usecases as it does for frr, is there extra complexity beyond that?

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,498 @@
...
+}
+
+```
+
+##### MetalLB
+
+```go
+type BGPPeerSpec struct {
+  Address string `json:"peerAddress"`
+  Interface bool `json:"interface"`
```
oribon, line 401: I guess this should be string?

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,498 @@
...
+```yaml
+apiVersion: metallb.io/v1beta2
+kind: BGPPeer
+metadata:
+  name: peer-unnumber
+  namespace: metallb-system
+spec:
+  myASN: 64512
+  ASN: 645000
+  peerAddress: "" // empty, can be omitted
```
oribon, line 417: just omit here for the example

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,498 @@
...
+
+#### E2E tests
+
+- Modify the testing infra to include point-to-point links which means connect
+containers using VETH, and not through a bridge. We need NET-ADMIN cap (or sudo)
+when creating the VETH.
+
+- Create/Delete peer per that specific test not to take resources from runners
+outside the runtime of the test. This peer needs different configuration (not
+IPs and interface names), keep separate raw configuration.
```
oribon, line 483: not sure I understand, you're proposing a separate test for the unnumbered case or expanding the existing containers setup?

karampok: I propose a separate isolated test for that use case.
(but no strong argument against expanding container setup).
Do you have a preference here?

design/unnumbered-bgp.md
karampok: ok, will do

design/unnumbered-bgp.md
karampok: yes, will fix

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,498 @@
...
+  no bgp network import-check
+
+  neighbor net0 interface remote-as 65004
+  neighbor net0 description TOR
+  neighbor net0 port 179
+  neighbor net0 bfd
+  neighbor net0 bfd profile fornet0
+  neighbor net0 timers 180 540
+  neighbor net0 timers connect 30
+  neighbor net0 graceful-restart
```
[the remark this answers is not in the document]
karampok: I think I did, but I will double check, is important

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,498 @@
...
+
+## Design details
+
+The `show running-config` in the FRR process running in the k8s node
+looks like
+
+
+```
+interface net0
+  ipv6 nd ra-interval 10
```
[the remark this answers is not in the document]
karampok: As it seems https://github.com/FRRouting/frr/commit/4a04e5f7968d46e446bdb530724da18b90b7f5f8 we do not need at all the interface block, it works without as RA auto-enabled in the unnumber case. (and I do not see any reason to change RA interval in a p2p link)

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,498 @@
...
+
+        end
+
+        %% Top Right Subgraph
+        subgraph Node1
+            direction TB
+            B1 --- B2(cni)
+            B2 --- B3
+            subgraph Pod1[pod]
+              B3[eth0]
```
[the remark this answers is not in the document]
karampok: I wanted to point out that there a pod/cni part, and a BGP process that is in th host. I should add the BGP process in the first diagram. I think there is no harm to have it

design/unnumbered-bgp.md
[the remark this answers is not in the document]
oribon: 👍 (saw that you removed it completely)

design/unnumbered-bgp.md
oribon: I'd say if it's possible to have all the tests ran on unnumbered in a clean way it's preferable

design/unnumbered-bgp.md
karampok: no todo, removed

design/unnumbered-bgp.md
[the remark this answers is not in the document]
karampok: looks good, updated the doc and did my smoke test using route-maps and interface

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,498 @@
...
+  address-family ipv4 unicast
+    network 5.5.5.5/32
+    neighbor net0 activate
+ exit-address-family
+```
+
+* keep the format as in [neighborsession.tmpl](internal/bgp/frr/templates/neighborsession.tmpl)
+* instead of address (ip) to be interface name
+* enable RAs in that interface name
+* must exclude FRR directives that do not accept interface e.g.
```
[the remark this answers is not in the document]
karampok: I found that this directive `disable-connected-check` can be combined with inteface. So it must be excluded otherwise throughs errors. I have not find other directive

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,498 @@
...
+    exit-address-family
+    !
+    address-family ipv6 unicast
+        redistribute connected
+        neighbor MTLB activate
+    exit-address-family
+    exit
+```
+
+An example ToR configuration when based on Junos
```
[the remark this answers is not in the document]
karampok: imo yes, maybe not our task, but unless I see that real config in real setup, I am not sure our use cases are valid.

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,498 @@
...
+
+```
+//TODO
+```
+
+###  How a peering will look like
```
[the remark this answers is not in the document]
karampok: I find it nice when looking a such design doc, to get a feeling how a working setup looks like. Do you want me to remove it?

design/unnumbered-bgp.md
karampok: I have a lack of  BGP knowledge what is the general impact of this RouterID in our use cases?  Do you know?

design/unnumbered-bgp.md
karampok: imo yes :)

oribon: so I think it should be more detailed, because it is not really clear atm

design/unnumbered-bgp.md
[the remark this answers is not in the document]
oribon: not sure I follow, are there options we can't use for unnumbered or it just needs a different configuration option?

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,508 @@
...
+### Topics
+
+#### Loopback IPv4 address
+
+Linux sends and receives IPv4 packets over an interface that does not have an IPv4 address.
+It does not look we need to add address in the loopback.
+
+#### RA enabled on both ends of the link
+
+According to the design of unnumber BGP peering, we need to enable RAs. By
```
fedepaol, line 323: can you expand on what RA is?

design/unnumbered-bgp.md
fedepaol: frr sets the router id with the lowest ip of the interfaces. So far, we didn't have a need to do otherwise. If this is not relevant please remove the section as it's distracting.

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,508 @@
...
+> docker exec -it unnumbered-p2p-peer ip route
+5.5.5.5 nhid 21 proto bgp metric 20
+        nexthop via inet6 fe80::dcad:beff:feff:1160 dev eth00 weight 1
+        nexthop via inet6 fe80::dcad:beff:feff:1161 dev eth01 weight 1
+        nexthop via inet6 fe80::dcad:beff:feff:1162 dev eth02 weight 1
+```
+
+### The API Extension
+
+We have three option:
```
fedepaol, line 366: to do what? Can we have a premise saying what we want in frr? Just a couple of sentences

design/unnumbered-bgp.md
fedepaol: +1 on this

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,508 @@
...
+metadata:
+  name: peer-unnumber
+  namespace: metallb-system
+spec:
+  myASN: 64512
+  ASN: 645000
+  interface: net0
+```
+
+#### Option 3. Add new Unnumber bool field
```
fedepaol, line 428: Let's remove this, I think it's a bit confusing.

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,508 @@
...
+  name: peer-unnumber
+  namespace: metallb-system
+spec:
+  myASN: 64512
+  ASN: 645000
+  peerAddress: net0
+  unnumber: true
+```
+
+### Testing
```
fedepaol, line 469:
```suggestion
### Test plan
```

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,508 @@
...
+
+#### E2E test for [Unnumber on primary interface](design/unnumbered-bgp.md#story-1-unnumbered-iebgp-on-primary-node-interface)
+
+There is no straightforward way to test unnumbered in the primary node
+interface. The testing setup (kind creation) creates bridge and connect
+containers. Ideally we need to setup a kind cluster, and replace the bridge
+with a containers that runs FRR on each other side of the veth, and at the same
```
fedepaol, line 476:
```suggestion
with a container that runs FRR on each other side of the veth, and at the same
```

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,508 @@
...
+time to provide switch/upstream. Until we find a way to automate that infra setup
+we test that scenario manual.
+
+#### E2E test for [Unnumber on secondary interface](design/unnumbered-bgp.md#story-1-unnumbered-iebgp-on-secondary-node-interface)
+
+We should modify the testing infra to include point-to-point links which means
+connect containers using VETH, and not through a bridge. We need NET-ADMIN cap
+(or sudo) when creating the VETH.
+
+##### Option 1 - have a separate test
```
fedepaol, line 486: I'd start with this, the majority of the scenarios are functional and should (!!!) be orthogonal to how we establish the session

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,508 @@
...
+
+### The API Extension
+
+We have three option:
+
+#### Option 1. Re-Use Address field
+
+We can re-use `Address string` to either be IP or interface, if not valid IP we
+indicate Unnumbered BGP peering. This looks the minimal API, but it will not
+have clean error path. E.g. if the user input is not valid IP, the
```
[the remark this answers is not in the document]
karampok: done

design/unnumbered-bgp.md
karampok: ACK then, will put that as but we need to decide if to further https://github.com/metallb/metallb/issues/2423
and have this 
````
// 1. if address and no interface then as normal
// 2. if non valid address and no interface then fail
// 3. if address is LLA and interface not empty, then LLA support
// == extra FRR directive `neighbor fe80::a876:4dff:fe77:408 interface eth0`
// 4. If address is LLA and empty interface, then fail
// 5. If address is empty and interface has value then the BGP unnumber peering config will take place
// ==  modify FRR directive `neighbor PEER interface remote-as <ASN>`
// && replace any address directive to use interface e.g.  `neighbor net0 timers 180 540`
// && enable RAs on this interface
// 6. If address has value (non-LLA) and interface non empty, then fail
``` 
as the end API

design/unnumbered-bgp.md
karampok: This is my preference :), I will keep it a rejected option without code example

design/unnumbered-bgp.md
karampok: done

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,508 @@
...
+spec:
+  myASN: 64512
+  ASN: 645000
+  peerAddress: net0
+  unnumber: true
+```
+
+### Testing
+
+#### E2E test for [Unnumber on primary interface](design/unnumbered-bgp.md#story-1-unnumbered-iebgp-on-primary-node-interface)
```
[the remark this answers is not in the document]
karampok: done

design/unnumbered-bgp.md
karampok: Done

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,508 @@
...
+
+There is no straightforward way to test unnumbered in the primary node
+interface. The testing setup (kind creation) creates bridge and connect
+containers. Ideally we need to setup a kind cluster, and replace the bridge
+with a containers that runs FRR on each other side of the veth, and at the same
+time to provide switch/upstream. Until we find a way to automate that infra setup
+we test that scenario manual.
```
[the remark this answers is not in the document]
karampok: As discussed is not worth the effort to refactor how we set up the infra.

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,508 @@
...
+
+#### E2E test for [Unnumber on secondary interface](design/unnumbered-bgp.md#story-1-unnumbered-iebgp-on-secondary-node-interface)
```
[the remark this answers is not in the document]
karampok: done

design/unnumbered-bgp.md
karampok: Sounds good, i keep the text as is, one we have option 1 I would say we can merge, but I am in favor to then try to implement also option

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,508 @@
...
+
+Displayed  2 routes and 4 total paths
+> docker exec -it unnumbered-p2p-peer ip route
+5.5.5.5 nhid 21 proto bgp metric 20
+        nexthop via inet6 fe80::dcad:beff:feff:1160 dev eth00 weight 1
+        nexthop via inet6 fe80::dcad:beff:feff:1161 dev eth01 weight 1
+        nexthop via inet6 fe80::dcad:beff:feff:1162 dev eth02 weight 1
+```
+
+### The API Extension
```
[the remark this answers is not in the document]
karampok: done

design/unnumbered-bgp.md
karampok: Three options how the API could look like, but that was removed.

design/unnumbered-bgp.md
karampok: I was expected if you could explain me how the RouterID is important or not. Example when we have single IPv6 stack. But I have that for discussion, removed

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,508 @@
...
+Neighbor                  V         AS   MsgRcvd   MsgSent   TblVer  InQ OutQ  Up/Down State/PfxRcd   PfxSnt Desc
+unnumbered-p2p-peer(net0) 4      65004       512       512        1    0    0 00:25:15            0        1 TOR
+
+Total number of neighbors 1
+```
+### Topics
+
+#### Loopback IPv4 address
+
+Linux sends and receives IPv4 packets over an interface that does not have an IPv4 address.
```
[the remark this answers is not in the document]
karampok: TODO, explain better

karampok (commented):
thanks for the feedback

karampok:
@fedepaol  @oribon I have updated the doc and I think I have addressed all your feedback. Please take a look, thanks

design/unnumbered-bgp.md
oribon: not sure I understand - implementation wise the issue will be solved for free(?) and documentation wise it seems sufficient to document the `Interface` field something like "if specified without address then peering happens over the interface (unnumbered), if together with address then it is assumed to be LLA"?

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,541 @@
...
+  are the same across the cluster nodes)
+- Remove the task of network admins to setup up addressing on those interfaces
+  if it is needed only for BGP peering
+
+## Non Goals
+
+- Make it work over no point-to-point connection
+- Define/Discuss dual ToR architecture, on top of bond interfaces or other HA strategies
+- [Dynamic BGP Peering](https://ipwithease.com/dynamic-bgp-peering/) is a
+  feature different than unnumbered BGP and not supported by Metallb. Both
```
oribon, line 41: nit: remove the "not supported by MetalLB" (it might be in the future, and unrelated to this)

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,541 @@
...
+  feature reduce the configuration needed for the BGP peering. The main difference is
+  that Dynamic BGP peering still needs IP addresses and does not require point-to-point links.
+- Support `remote-as` to accept `<external|internal|auto>`
```
oribon, line 44: I'd also remove this (seems unrelated)

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,541 @@
...
+As network administrator, I want to reduce the size of the configuration in the ToR router.
+
+## The API Changes
+
+### Option 2. Add a new Interface field
+
+We can introduce a new field `Interface string` which if defined then Unnumber
+BGP takes place. In that case, the address must NOT have value. For
+completeness, we could use field might be used to support `LLA%zone` [user case](https://github.com/metallb/metallb/issues/2423).
+So at the end, the API documentation will look like
```
oribon, line 68: as the other comment, I think this could be a lot shorter (no need to specify everything and its consequence)

design/unnumbered-bgp.md
karampok: done

design/unnumbered-bgp.md
karampok: I agree, but I prefer to remove once @fedepaol take another feedback and he is ok with the consequence or not (we could decide no to support the LLA)

fedepaol: Let's focus on unnumbered and maybe implement LLA afterwards if we need / want to.
What's important here is that the API is compatible with that. If we want to mention it, let's add a subsection in the bottom

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,540 @@
...
+
+### Story 3
+
+As network administrator, I want to reduce the size of the configuration in the ToR router.
+
+## The API Changes
+
+### Option 2. Add a new Interface field
+
+We can introduce a new field `Interface string` which if defined then Unnumber
```
fedepaol, line 64: Since this is the proposed change, let's reword with `A new interface field will be introduced`

design/unnumbered-bgp.md
```diff
@@ -0,0 +1,540 @@
...
+As network administrator, I want to avoid configuring IP addresses for
+point-to-point connection just for the BPG peering reason.
+
+### Story 3
+
+As network administrator, I want to reduce the size of the configuration in the ToR router.
+
+## The API Changes
+
+### Option 2. Add a new Interface field
```
fedepaol, line 62: This is not an option anymore, but the proposed change

karampok: done

design/unnumbered-bgp.md
karampok: done

design/unnumbered-bgp.md
karampok: ok, removed the reference to LLA

fedepaol:
LGTM, thanks for this!

PR #2510, opened 2024-09-03
Spike on supporting LLAs

karampok:
/kind feature

Fix for #2423 

AFAIU there is not need for any API change to support that. We could allow the string `fe80::42:acff:fe12:6%eth0` to be properly parsed and if valid (=LLA and `%eth0` part which is called zone exists) to handle it (which is if Zone exist then add the extra FRR directive ` neighbor fe80::a876:4dff:fe77:408 interface eth0`

Under the hood:
1. Moving from net.IP to the netip.Addr helps greatly, and looks a nice improvement
2. E2E are also possible by finding the LLAs of the interface and use that instead of the global ipv address 

**Release note**:

-->

```release-note
Support for LLA (LinkLocalAddress) ipv6 addresses (e.g. `fe80::42:acff:fe12:6%eth0` )
```


karampok:
@fedepaol @oribon Please have a fast look here if you do like the direction and if yes then I can further work to undraft it and make it ready for the code review. Thank you

oribon (commented):
a general thought: imo it would be cleaner to just have addr + interface in the api (like proposed in the unnumbered pr) instead of supporting "IPv6 with a scoped addressing zone" in the peer's address - the main reasons being:
* it seems for frr that format is just an alias for specifying both address and interface
* when we do have an interface field it might be confusing (do I specify the interface as part of the address or in the interface field).
wdyt?

karampok:
After some conversation, we will proceed with the proposal from @oribon:
We will have the following API change

```
type Neighbor struct {
  Address string `json:"address"` //
  Interface string `json:"interface"` 
// 1. if address and no interface as normal
// 2. if address is LLA and interface then LLA support=extra directive `neighbor fe80::a876:4dff:fe77:408 interface eth0`  
// 3. If address is LLA and empty interface, validation is failing
// 4. If address is empty and interface has value then the BGP unnumber peering config will take place= `neighbor PEER interface remote-as <internal|external|auto|ASN> && other && enable RAs on this interface`                                         
}
```

 instead of allowing user to enter `fe80::1234%eth0` in the address field and no API change.

That will be reviewed again after the BGP unumber design/implementation

karampok:
Closing as that was just a spike

PR #2515, opened 2024-09-05
chore: update toolchain to go 1.22.7

mjnagel:
**Is this a BUG FIX or a FEATURE ?**: Fix/chore to update dependencies

/kind cleanup

**What this PR does / why we need it**: Updated to the latest 1.22.x version of Go for dockerfiles + toolchain, resolves several detected CVEs with trivy, etc.

**Release note**:

```release-note
NONE
```


karampok:
Hi, since go 1.21 when toolchain was introduced the value we have in the docker image should not be anymore important. The value in the go.mod should be the one define

```
tmp> pwd
/tmp
[I] kka@f-t14s /tmp> go version
go version go1.21.11 linux/amd64
[I] kka@f-t14s /tmp> z metallb
[I] kka@f-t14s ~/w/g/m/metallb (main ✖1)> head go.mod
module go.universe.tf/metallb

go 1.22.0

toolchain go1.22.4

require (
        github.com/davecgh/go-spew v1.1.1
        github.com/go-kit/log v0.2.1
        github.com/google/go-cmp v0.6.0
[I] kka@f-t14s ~/w/g/m/metallb (main ✖1)> go version
go version go1.22.4 linux/amd64
> cd -
[I] kka@f-t14s /tmp> go version
go version go1.21.11 linux/amd64
```

Having said that, it might be a better approach to simly add latest tag in build go docker to avoid future cores.


(But let's wait some feedback from the maintainers of metallb)

fedepaol:
> Having said that, it might be a better approach to simly add latest tag in build go docker to avoid future cores.
> 

We should never do this, as we wouldn't have reproducible builds and we wouldn't have control on what goes where.

fedepaol:
I'll try to explain, with the goal of being instructive.

> The reproducable go build is guaranteed by the go toolchain (go.mod, go.sum). The important note is that the go version of the docker build image does not define the actual go version, only go.mod does that (unless disabled on purpose).
> 
> I understand that "latest" tag might sound riskier than "go1.22.4" tag (both are tags that can change, is not sha) and any change (e.g size of the docker image) could make the ci flake.

https://go.dev/doc/modules/gomod-ref#go

"Indicates that the module was written assuming the semantics of the Go version specified by the directive." Does not say anything about the toolchain.

At go 1.21 or higher:

    The go line declares a required minimum version of Go to use with this module.
    The go line must be greater than or equal to the go line of all dependencies.
    The go command no longer attempts to maintain compatibility with the previous older version of Go.
    The go command is more careful about keeping checksums of go.mod files in the go.sum file.

The toolchain directive is also not stating which EXACT toolchain to use:

https://go.dev/doc/toolchain


> when using the go command bundled with Go 1.21.3 in a main module that says go 1.21.0, the go command uses Go
> 1.21.3. When the go or toolchain line is newer than the bundled toolchain, the go command runs the newer toolchain 
> instead.


This, as I read, is that a newer latest will always use the latest go and not the one asked by the toolchain.


We were bitten so many times by it, kind, libraries, cli tools, you name it.

All in all, I insist, **not using exact versions is a recipe for harming our future selves and we should pin everywhere we can.**. We should never try to be smart and try to make exceptions because it might work... because that can change too. This is for the code, and for the dependencies.

And even if the toolchain was the same, having a different image we build the binary on can have unexpected side effects, such as libc, to name the first one that comes to my mind (as we don't disable CGO iirc).

fedepaol:
I wonder if the govulncheck failure here is related to the bump in the go version or incidental. @mjnagel fancy fixing the errors? Otherwise we'll do (adn thanks for the PR!)

fedepaol:
> @fedepaol / @karampok updated the toolchain to match the dockerfiles. I also started on the gosec issues but I wasn't seeing all the same output/failures locally with gosec so it was difficult to iterate on the others. They do seem unrelated based on other PRs having the same failures (#2516). I'm pretty unfamiliar with the bgp code and a bit nervous to try and fix all those overflow issues 😅 - if you are willing to tackle the other ones that would be appreciated!

absolutely! Thanks a lot for the help!

fedepaol:
Filed https://github.com/metallb/metallb/pull/2525 for the gosec failures

fedepaol:
@mjnagel I fixed all the gosec warnings, mind removing your changes here to solve the conflict?

PR #2528, opened 2024-09-11
Safeconvert: add arm specific code for int to uint32

fedepaol:

>
> /kind bug
/kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

On arm, there is no need to check if an int is bigger than uint32 as it
will never be the case. Additionally, leaving the check would cause
building on arm to fail.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/safeconvert/convertuint32_arm.go
+++ b/internal/safeconvert/convertuint32_arm.go
@@ -1,0 +1,18 @@
+//go:build arm
+// +build arm
+
+// SPDX-License-Identifier:Apache-2.0
+
+package safeconvert
+
+import (
+	"fmt"
+)
+
+func IntToUInt32(toConvert int) (uint32, error) {
+	if toConvert < 0 {
+		return 0, fmt.Errorf("trying to convert negative value to uint32: %d", toConvert)
+	}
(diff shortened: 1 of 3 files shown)


internal/safeconvert/convertuint32_arm.go
```diff
@@ -0,0 +1,18 @@
+//go:build arm || arm64
+// +build arm arm64
```
zeeke, line 2: `arm64` should not be affected by this:

```
$ export GOARCH=arm; go build  -o speaker-$GOARC ./speaker 
# go.universe.tf/metallb/internal/safeconvert
internal/safeconvert/convert.go:34:17: math.MaxUint32 (untyped int constant 4294967295) overflows int

$ export GOARCH=arm64; go build  -o speaker-$GOARC ./speaker
$ echo $?
0
```
```suggestion
// +build arm
```

Can we be more selective?

fedepaol: good catch!

fedepaol: fixed

zeeke (commented):
LGTM

Issue #2526, opened 2024-09-11
CI flake: Unknown stream id 1, discarding message 


fedepaol:
### MetalLB Version

not relevant

### Deployment method

Not relevant

### Main CNI

Not relevant

### Describe the bug

We are seeing flakes on ci like https://github.com/metallb/metallb/actions/runs/10797181713/job/29948033121?pr=2525

The vtysh calls are failing with " Unknown stream id 1, discarding message". 

### To Reproduce

It's a flake!

### Expected Behavior

CI stable

### Additional Context

We should implement some retry mechanism to make the ci more stable

### I've read and agree with the following

- [X] I've checked all open and closed issues and my request is not there.
- [X] I've checked all open and closed pull requests and my request is not there.

### I've read and agree with the following

- [X] I've checked all open and closed issues and my issue is not there.
- [X] This bug is reproducible when deploying MetalLB from the main branch
- [X] I have read the [troubleshooting guide](https://metallb.universe.tf/troubleshooting/) and I am still not able to make it work
- [X] I checked the logs and MetalLB is not discarding the configuration as not valid
- [X] I enabled the debug logs, collected the information required from the cluster using the [collect script](https://raw.githubusercontent.com/metallb/metallb/main/troubleshooting/collect.sh) and will attach them to the issue
- [X] I will provide the definition of my service and the related endpoint slices and attach them to this issue

karampok:
Some notes:
https://kubernetes.io/blog/2024/08/20/websockets-transition/
1.31: Streaming Transitions from SPDY to WebSockets
and our kubectl version changed to v1.31 as part of this commit https://github.com/metallb/metallb/commit/88aa1b6969e265c7f673908cd1bec6a506f8711e

Reverting to older version is not ideal ofc.
I am thinking that we can exclude the stderr and take only the stdout here https://github.com/metallb/metallb/blob/main/e2etest/pkg/executor/executor.go#L70

karampok:
> I'd say let's retry it a few times in case of error. WDYT?

I am not sure what you suggest.

The actual error is nil, the output combines stderr/stdout, In that in that case
stderr = `E0910 18:08:54.853077  339136 websocket.go:296] Unknown stream id 1, discarding message`
stdout =
 ```
\n{\n\"default\":{\n \"vrfId\": 0,\n \"vrfName\": \"default\",\n \"tableVersion\": 0,\n \"routerId\": \"172.19.0.2\",\n  
[...]

PR #2536, opened 2024-09-13
E2E: ignore stderr if error is nil when kubectl exec

karampok:
/kind flake

#2526 

 ```release-note ```
```
NONE
```

### DIFF
--- a/e2etest/pkg/executor/executor.go
+++ b/e2etest/pkg/executor/executor.go
@@ -3,7 +3,9 @@
 package executor
 
 import (
+	"bytes"
 	"errors"
+	"fmt"
 	"os"
 	"os/exec"
 
@@ -67,8 +69,15 @@ func (p *podExecutor) Exec(cmd string, args ...string) (string, error) {
 		return "", errors.New("the kubectl parameter is not set")
 	}
 	fullargs := append([]string{"exec", p.name, "-n", p.namespace, "-c", p.container, "--", cmd}, args...)
-	out, err := exec.Command(Kubectl, fullargs...).CombinedOutput()
-	return string(out), err
+
+	c := exec.Command(Kubectl, fullargs...)
+	var stdoutBuf, stderrBuf bytes.Buffer
+	c.Stdout = &stdoutBuf
+	c.Stderr = &stderrBuf
+	if err := c.Run(); err != nil {
+		return "", fmt.Errorf("kubectl exec failed: %w, stderr: %s", err, stderrBuf.String())
+	}
+	return stdoutBuf.String(), nil
 }
 
 // add ephemeral container to deal with distroless image


e2etest/bgptests/validate.go
```diff
@@ -356,10 +356,14 @@ func validateServiceNotInRoutesForCommunity(c *frrcontainer.FRR, community strin
 func isRouteInjected(pods []*corev1.Pod, pairingFamily ipfamily.Family, routeToCheck, vrf string) (bool, string) {
 	for _, pod := range pods {
 		podExec, err := FRRProvider.FRRExecutorFor(pod.Namespace, pod.Name)
-		Expect(err).NotTo(HaveOccurred())
```
karampok, line 359: @fedepaol AFAIU the flake happens because even though that function is called as part of an Eventually in the main testing code, that expect just make the test fail. If we have not Expect, it retries.

I am not sure if I should do more refactor, e.g. the return "bool,string" is not ideal, it could be "string,error".

wdyt?

fedepaol: not sure I understood, but the failure can't be here as here we just fetch the executor (and it will fail only if it doesn't find the pods, for example).
The failure can be in any place we call .Exec , so that's where I'd try to hide error behind a few retries (if we are not able to do better).

karampok: >The failure can be in any place we call .Exec ,

what I see is that the executor.Exec is always wrapped in a  ginkgo Eventually logic
https://github.com/metallb/metallb/blob/main/e2etest/bgptests/validate.go#L338-L353 so even if fails a retry takes place.

As well as the function of interest https://github.com/metallb/metallb/blob/main/e2etest/bgptests/bgp.go#L1173-L1188 is part of an Eventually loop, just it has Expect instead of return error.

Therefore creating a retry logic into the call.Exec https://github.com/metallb/metallb/blob/main/e2etest/pkg/executor/executor.go#L91-L103 is a valid approach but ideally we need the code inside an Eventually/Consistently to not have other assertions, and keep the retry higher.

karampok: (I put the comment in the wrong Expect, but still that Expect could give a flake)

fedepaol: "Always" is a bold statement :-). We are calling it from many places, and I don't want to check all of them, nor worry about future ones needing to be wrapped in eventuallys. 

In particular, the job linked in the issue is failing here https://github.com/metallb/metallb/blob/main/e2etest/bgptests/validate.go#L362 where no retry is in place.

The caller must trust Exec will always work unless there is a syntax error and not to worry that it might fail temporarily. Because of this, if we won't find a better solution / the root cause of the failure, I am suggesting to do some sort of retry inside of it. And the retry should not leverage gomega, as we try to avoid gomega in util packages. 

But again, the retry is the last straw if we can't make kubectl exec work in any way (maybe bumping up kubectl? trying to file an issue against k8s?), or doing both so we have a reliable CI while we investigate?

e2etest/pkg/executor/executor.go
```diff
@@ -67,8 +69,18 @@ func (p *podExecutor) Exec(cmd string, args ...string) (string, error) {
...
+
+	c := exec.Command(Kubectl, fullargs...)
+	var stdoutBuf, stderrBuf bytes.Buffer
+	c.Stdout = &stdoutBuf
+	c.Stderr = &stderrBuf
+	if err := c.Run(); err != nil {
+		return "", fmt.Errorf("kubectl exec failed: %w, stderr: %s", err, stderrBuf.String())
+	}
+	if stderrBuf.String() != "" {
+		return "", fmt.Errorf("kubectl exec succeeded but stderr: %s", stderrBuf.String())
```
karampok, line 81: Do you think if the err is nil, but the stderr is not empty, should we return an error?

In our flake  the `kubectl exec` is not failing is just adding some stderr, so the final return string
is not a valid json when parsed from the next function (that parses json into route struct)

```
Failed to parse routes E0910 18:08:54.853077  339136 websocket.go:296] Unknown stream id 1, discarding message\n

{\n\"default\":{\n \"vrfId\": 0,\n \"vrfName\": \"default\",\n \"tableVersion\": 0,\n \"routerId\": \"172.19.0.2\",\n \"defaultLocPrf\": 100,\n \"localAS\": 64512,\n \"routes\": {  }  } \n,\n\"red\":{\n \"vrfId\": 5,\n \"vrfName\": \"red\",\n \"tableVersion\": 0,\n \"routerId\": \"172.31.0.4\",\n \"defaultLocPrf\": 100,\n \"localAS\": 64513,\n \"routes\": {  }  } \n}\n
````

For that flake I am not sure a retry logic makes sense given the error is nil

fedepaol: If that's the case, then I think the right thing to do is to return the sdterr only after a proper failure, and the std output otherwise.

karampok:
@fedepaol  this is ready, the small change we discussed, thanks

fedepaol:
Thanks!

PR #2537, opened 2024-09-16
Align hold/keepalive BGP timers with FRR

oribon:
/kind bug

The default holdtime of frr-k8s was different from the metallb one, specifically frr-k8s delegates the defaults to FRR instead of hard-coding them. This means that a bgppeer created without specifiying the hold time will conflict with an frrconfiguration created without specifying the hold time.
Here we align the speaker with this approach, having a hardcoded default only on native mode. In addition, we get the benefit of being able to differentiate between `0 0` timers or no timers, which have a different purpose in FRR.

**Release note**:

```release-note
Fix default BGPPeer hold/keepalive timers being incompatible with FRRConfigurations. We now delegate the defaults to FRR itself instead of hard-coding them, supporting 0 timers as well.
```


fedepaol:
we should definitely mention this in the release notes, as the default value is going to change.

Issue #2425, opened 2024-06-11
Disabling members list causes all nodes to be removed from speakers list


ncthompson:
### MetalLB Version

v0.14.5

### Deployment method

Charts

### Main CNI

Weave

### Kubernetes Version

v1.24.10

### Cluster Distribution

Rancher

### Describe the bug

Disabling speaker member list via the Helm chart by setting `speaker.memberlist.enabled=false` causes all speakers to not announce any l2 service. This seem to be a regression on a previous feature.

### To Reproduce

To replicate via installation set `speaker.memberlist.enabled=false` in the helm installation. You can then setup a basic l2 pool as described [here](https://github.com/metallb/metallb/blob/main/configsamples/deployment_l2.yaml)

You can add and run the following unit test in `speaker/l2_controller
[metallb_report.tgz](https://github.com/user-attachments/files/15789049/metallb_report.tgz)
_test.go`

```go
func TestUsableNodesEPSlicesWithSpeakerlistDisabled(t *testing.T) {
	// Create a speaker response with a null list as expected when speaker is disabled
	fakeSL := &fakeSpeakerList{
		speakers: nil,
	}
	c1, err := newController(controllerConfig{
		MyNode:  "iris1",
		Logger:  log.NewLogfmtLogger(log.NewSyncWriter(os.Stderr)),
		SList:   fakeSL,
		bgpType: bgpFrr,
	})
	if err != nil {
		t.Fatalf("creating controller: %s", err)
	}
	c1.client = &testK8S{t: t}
	advertisementsForNode := []*config.L2Advertisement{
		{
			Nodes: map[string]bool{
				"iris1": true,
				"iris2": true,
			},
		},
	}

	conf := &config.Config{
		Pools: &config.Pools{ByName: map[string]*config.Pool{
			"default": {
				CIDR:             []*net.IPNet{ipnet("10.20.30.0/24")},
				L2Advertisements: advertisementsForNode,
			},
		}},
	}

	svc := &v1.Service{
		Spec: v1.ServiceSpec{
			Type:                  "LoadBalancer",
			ExternalTrafficPolicy: v1.ServiceExternalTrafficPolicyTypeLocal,
		},
		Status: statusAssigned("10.20.30.1"),
	}
	eps := map[string][]discovery.EndpointSlice{
		"10.20.30.1": {
			{
				Endpoints: []discovery.Endpoint{
					{
						Addresses: []string{
							"2.3.4.5",
						},
[...]

PR #2549, opened 2024-09-27
Disable memberlist does not remove all nodes from speaker list

karampok:
/kind bug

#2425 

**Release note**:

-->

```release-note
Disable memberlist does not remove all nodes from speaker list
```


karampok:
Draft as i still need to undestand the memberlist.enable|disable feature.

Implementation is as suggested in the issue

karampok:
I could not see any downside of doing that. So given ci is green I would suggest that we merge.

speaker/layer2_controller.go
```diff
@@ -223,8 +223,12 @@ func poolMatchesNodeL2(pool *config.Pool, node string) bool {
 }
 
 func (c *layer2Controller) speakersForPool(pool *config.Pool, nodes map[string]*v1.Node) map[string]bool {
+	usableSpeakers := c.sList.UsableSpeakers()
+	if usableSpeakers == nil {
+		return nil
```
fedepaol, line 228: here I would be more mindful and IF we want to match the previous logic, we should assume that if member list is not available (which is technically different from usableSpeakers returning nil), we iterate over all the nodes.

This has a very important flaw though: metallb will behave as if all the nodes have a running speaker (which is something I am not sure we really want).

So, I would say that if we want to fix this we need to take into account looking where the speakers are running in order to avoid to elect one node without a speaker.

fedepaol:
Sorry for the delay :sweat: 

The offending commit is almost 2 years old: https://github.com/metallb/metallb/commit/cac2caa55f95f65d51ce095bd1f79c819d88aba5#diff-f31007be42c7bbf0ff6dee5806c2a2458bd48ab3923571590a3473038d598c18L154

We need to make a call here:

- we can restore the old behaviour (using all the nodes without checking if a speaker is really running there
- we can use the k8s api to see which / what speakers are running where (with the extra complexity of a reconciliation of those pods)
- we can leave things as they are and say memberlist is the only way to consume l2

(or we can approach incrementally).

@karampok @oribon thoughts?

fedepaol:
> Sorry for the delay 😓
> 
> The offending commit is almost 2 years old: [cac2caa#diff-f31007be42c7bbf0ff6dee5806c2a2458bd48ab3923571590a3473038d598c18L154](https://github.com/metallb/metallb/commit/cac2caa55f95f65d51ce095bd1f79c819d88aba5#diff-f31007be42c7bbf0ff6dee5806c2a2458bd48ab3923571590a3473038d598c18L154)
> 
> We need to make a call here:
> 
>     * we can restore the old behaviour (using all the nodes without checking if a speaker is really running there
> 
>     * we can use the k8s api to see which / what speakers are running where (with the extra complexity of a reconciliation of those pods)
> 
>     * we can leave things as they are and say memberlist is the only way to consume l2
> 
> 
> (or we can approach incrementally).
> 
> @karampok @oribon thoughts?

ping! Also, @karampok mind fixing the conflicts?

speaker/layer2_controller.go
```diff
@@ -223,8 +223,12 @@ func poolMatchesNodeL2(pool *config.Pool, node string) bool {
 }
 
 func (c *layer2Controller) speakersForPool(l log.Logger, name string, pool *config.Pool, nodes map[string]*v1.Node) map[string]bool {
+	usableSpeakers := c.sList.UsableSpeakers()
+	if usableSpeakers == nil {
+		return nil
```
fedepaol, line 228: getting back to this (sorry for the delay-2):
Here we are giving the `nil` value a semantic: it means memberlist is disabled. Can we have UsableSpeakers return a struct containing the map plus a "considerAllNodes" flag (maybe with a better name)?
At the same time, we should return that struct and use that instead of nil in the following calls.
This will give more clarity and less doubts on what the code wants to do.

fedepaol: ping :-)

karampok: I am not sure we should extend the logic. IMO the scope of the the other logic (like usableNode which I rename) seems different. Please take another look, thanks!

speaker/layer2_controller.go
```diff
@@ -61,10 +60,9 @@ func usableNodes(eps []discovery.EndpointSlice, speakers map[string]bool) []stri
 				continue
 			}
 			nodeName := *ep.NodeName
-			if speakers != nil {
-				if hasSpeaker := speakers[nodeName]; !hasSpeaker {
-					continue
-				}
+			//DISCUSS: we can keep if speaker !=nil to be safe but that is not a valid code path
```
fedepaol, line 63: what I was suggesting is to have this function accept SpeakerListInfo instead of the map, and to rely on the Disabled field instead of having a nil map.

speaker/layer2_controller.go
```diff
@@ -122,11 +117,13 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce
 	})
 
 	// Are we first in the list? If so, we win and should announce.
-	if len(availableNodes) > 0 && availableNodes[0] == c.myNode {
+	owner := availableNodes[0] // we have returned if len()==0
```
fedepaol, line 120: can you split the refactoring bits in a separate commit so the actual change is easier to follow?

speaker/layer2_controller.go
```diff
@@ -93,13 +91,10 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce
...
 
-	// we select the nodes with at least one matching l2 advertisement
-	forPool := c.speakersForPool(l, name, pool, nodes)
-	var availableNodes []string
+	speakerMap := c.speakersForPool(l, name, pool, nodes)
+	availableNodes := convertMapToSlice(speakerMap)
 	if svc.Spec.ExternalTrafficPolicy == v1.ServiceExternalTrafficPolicyTypeLocal {
-		availableNodes = usableNodes(eps, forPool)
-	} else {
-		availableNodes = nodesWithActiveSpeakers(forPool)
```
fedepaol, line 102: why you removed this?

speaker/layer2_controller.go
```diff
@@ -46,11 +47,9 @@ func (c *layer2Controller) SetConfig(log.Logger, *config.Config) error {
...
 }
 
-// usableNodes returns all nodes that have at least one fully ready
-// endpoint on them.
-// The speakers parameter is a map containing all the nodes with active speakers.
-// If the speakers map is nil, it is ignored.
-func usableNodes(eps []discovery.EndpointSlice, speakers map[string]bool) []string {
+// onlyNodesWithEndpoint returns all nodes that have at least one fully ready
+// endpoint on them and that have a speaker according to the speaker map param.
+func onlyNodesWithEndpoint(eps []discovery.EndpointSlice, speakers map[string]bool) []string {
```
fedepaol, line 52: let's call this "nodesWithEndpoints"

speaker/layer2_controller.go
fedepaol: found it

speaker/layer2_controller.go
```diff
@@ -93,13 +91,10 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce
 		return "notOwner"
 	}
 
-	// we select the nodes with at least one matching l2 advertisement
-	forPool := c.speakersForPool(l, name, pool, nodes)
-	var availableNodes []string
+	speakerMap := c.speakersForPool(l, name, pool, nodes)
+	availableNodes := convertMapToSlice(speakerMap)
```
fedepaol, line 95: the previous name of the function made the logic clearer, I'd stick to that

speaker/layer2_controller.go
karampok: I might be confused

But this function receives a map which is never nil
1. if memberList disabled, the  arg into that function is the full list of nodes 
2. if memberList enabled, the arg into that function is a map which can be empty by design but not nil

speaker/layer2_controller.go
karampok: done

speaker/layer2_controller.go
karampok: I do not see why `usableNodes` which just transforms the map to array looks better, but done

speaker/layer2_controller.go
karampok: that small refactor took place to allow me to log a message which will be usefull  if someone runs memberlist disabled, with speaker with node selector.  
I would find better as part of this commit but I can split it also (or drop it if you think is not important)

speaker/layer2_controller.go
fedepaol: the old version was nodesWithActiveSpeakers, so it's clear what we are picking

karampok: correctly my bad

speaker/layer2_controller.go
fedepaol: now I got the logic of your implementation, it's fine as it is. Thanks.

internal/speakerlist/speakerlist.go
```diff
@@ -53,6 +61,8 @@ func New(logger log.Logger, nodeName, bindAddr, bindPort, secret, namespace, lab
 
 	if labels == "" || bindAddr == "" {
 		level.Info(logger).Log("op", "startup", "msg", "not starting fast dead node detection (memberlist), need ml-bindaddr / ml-labels config")
+		level.Info(logger).Log("op", "startup", "msg", "when memberlist disabled requires speaker pods to run on all nodes")
```
fedepaol, line 64: "member list address not provided, disabling speaker discovery. Please note that with this configuration MetalLB assumes all the nodes have running speakers".

I wonder if we should document this somewhere.

speaker/layer2_controller.go
fedepaol: Fair enough.
Let's remove the comment though

speaker/layer2_controller.go
```diff
@@ -122,11 +116,13 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce
...
 	// Are we first in the list? If so, we win and should announce.
-	if len(availableNodes) > 0 && availableNodes[0] == c.myNode {
+	owner := availableNodes[0] // we have returned if len()==0
+	if owner == c.myNode {
+		level.Debug(l).Log("event", "shouldannounce", "protocol", "l2", "node", c.myNode, "service", name)
 		return ""
 	}
 
-	// Either not eligible, or lost the election entirely.
+	level.Debug(l).Log("event", "skipping should announce l2", "node", c.myNode, "service", name, "reason", "not owner", "owner", owner)
```
fedepaol, line 125: I have doubts about these. I think they are already covered by the "serviceWithdrawn" / Service announcing logs outside. The only extra value provided is the owner name, which _maybe_ we can append to the return value. I'd say let's remove for now.

speaker/layer2_controller.go
```diff
@@ -223,8 +218,15 @@ func poolMatchesNodeL2(pool *config.Pool, node string) bool {
 }
 
 func (c *layer2Controller) speakersForPool(l log.Logger, name string, pool *config.Pool, nodes map[string]*v1.Node) map[string]bool {
+	sl := c.sList.UsableSpeakers()
+	eligibleNodes := maps.Keys(sl.Nodes)
+	if sl.Disabled {
+		// when memberlist disabled we consider all nodes eligible under the assumption
```
fedepaol, line 224:
```suggestion
		// when memberlist is disabled we consider all nodes eligible under the assumption that
```

charts/metallb/values.yaml
```diff
@@ -262,6 +262,7 @@ speaker:
   logLevel: info
   tolerateMaster: true
   memberlist:
+    # When enabled: false, speaker pods must run on all nodes, no nodeSelector
```
fedepaol, line 265:
```suggestion
    # When enabled: false, the speaker pods must run on all nodes
```

fedepaol: we can enforce the mutual exclusion with helm rules in the templates

karampok: we have already a policy, is that what you mean by rule?
 https://github.com/metallb/metallb/blob/651eb25245a054ab76698ecc399a040b97267851/charts/metallb/policy/speaker.rego#L11-L16 which does not even allow to disable the memberlist. Not sure the rules/policies are working though

speaker/layer2_controller.go
karampok: reverted

speaker/layer2_controller.go
karampok: removed

charts/metallb/values.yaml
fedepaol: nope, it's in the chart. Try to look for `fail` in the charts, there are a few examples.

internal/speakerlist/speakerlist.go
karampok: Added something that should appear in https://github.com/metallb/metallb/tree/main/charts/metallb in the Description

fedepaol:
LGTM, thanks!

fedepaol:
sent to merge queue

PR #2552, opened 2024-09-30
Add support for "Dynamic ASN"

oribon:

>
/kind feature

Adds support for FRR's "Dynamic ASN", allowing users to specify whether the peer is ibgp or ebgp without explicitly setting its asn. Leverages https://docs.frrouting.org/en/latest/bgp.html#clicmd-neighbor-PEER-remote-as-internal

**Release note**:

```release-note
Add DynamicASN field for a BGPPeer, which allows the speaker to detect the AS number to use without explicitly setting it. The new field is mutually exclusive with the existing PeerASN field, and one of them must be specified for any given BGPPeer. 
```


fedepaol:
lgtm

PR #2568, opened 2024-10-22
Tasks: warn if not able to export kind logs

fedepaol:

>
> /kind bug
/kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

Warning instead of failing if we are not able to export the kind logs. Useful when
setting the report path but running against a running cluster instead of kind.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -1340,7 +1340,7 @@ def e2etest(
     )
 
     if export != None:
-        run("{} export logs {}".format(kind_path, export))
+        run("{} export logs {}".format(kind_path, export), warn=True)
 
     if testrun.failed:
         raise Exit(message="E2E tests failed", code=testrun.return_code)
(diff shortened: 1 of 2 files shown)


tasks.py
```diff
@@ -1340,7 +1340,7 @@ def e2etest(
     )
 
     if export != None:
-        run("{} export logs {}".format(kind_path, export))
+        run("{} export logs {}".format(kind_path, export), warn=True)
```
karampok, line 1343: It possible also to print the command? I think there is echo true something?

karampok (approved):

tasks.py
fedepaol: yep but should be part of the warning I guess

PR #2583, opened 2024-11-06
internal/config: Replace k8s.io/utils/strings/slices by Go stdlib slices

tklauser:

/kind cleanup

Use the functionality provided by the Go standard library `slices` package since Go 1.21 instead of the deprecated functions in the `k8s.io/utils/strings/slices` package.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -19,6 +19,7 @@ import (
 	"fmt"
 	"net"
 	"reflect"
+	"slices"
 	"sort"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


tklauser:
> Thanks! Sending to merge queue. How did you find this, out of curiosity?

I did the original change in `k8s.io/utils/strings/slices` deprecating most of the functions (see https://github.com/kubernetes/utils/pull/315) and now checked for prominent importers of the package here: https://pkg.go.dev/k8s.io/utils/strings/slices?tab=importedby

PR #2585, opened 2024-11-07
#364: Reject service IPs overlap with node IPs

### LINKED ISSUE #364: Detect and reject configurations where the service IP range overlaps with node IPs
Probably due to inadequate documentation, people think that they can define service IPs that overlap with node IPs. This breaks things in multiple ways (ARP poisoning in L2 mode, wrong kube-proxy rules executed in all modes...), and is not at all a supported mode of operation.

The controller should detect and reject such configurations, as well as refuse to allocate IPs if they overlap with node IPs.

tyiying:
**Is this a BUG FIX or a FEATURE ?**: BUG FIX #364

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

Probably due to inadequate documentation, people think that they can define service IPs that overlap with node IPs. This breaks things in multiple ways (ARP poisoning in L2 mode, wrong kube-proxy rules executed in all modes...), and is not at all a supported mode of operation.
This PR add a webhook validation to check IPAddressPool IP range, if it is overlap with node IPs, reject the configuration of IPAddressPool

- Webhook validation rejection for IPAddressPool has explicit error message so it is easy for user to adjust the configuration. Also webhook validation on IPAddressPool does the rejection before automatic IP allocation take place at service configuration stage. 

**Release note**:
```release-note
Don't allow IPAddressPools providing IPs overlapping with node IPs.
```


controller/service.go
```diff
@@ -220,6 +221,16 @@ func (c *controller) allocateIPs(key string, svc *v1.Service) ([]net.IP, error)
...
 			return nil, fmt.Errorf("requested loadBalancer IP(s) %q does not match the ipFamily of the service", desiredLbIPs)
 		}
+
+		nodes := []v1.Node{}
+		for _, node := range c.nodes {
+			nodes = append(nodes, *node)
+		}
+
+		nodeIPs := k8snodes.NodeIPsForFamily(nodes, serviceIPFamily)
+		if err = checkSvcIPsOverlapWithNodeIPs(desiredLbIPs, nodeIPs); err != nil {
```
fedepaol, line 231: I would argue that this will never happen, because even if the service asks for a node ip, that ip will never be available in a pool. Let's remove the controller bits and focus only on the configuration part.

controller/main.go
```diff
@@ -147,6 +148,11 @@ func (c *controller) SetPools(l log.Logger, pools *config.Pools) controllers.Syn
 	return controllers.SyncStateReprocessAll
 }
 
+func (c *controller) SetNode(l log.Logger, node *v1.Node) controllers.SyncState {
+	c.nodes[node.Name] = node
```
fedepaol, line 152: nit: although I am asking to remove this, writing it to let you know that we should be careful to consider that this map is not aligned with the content of the informer cache, and only expands. If a node gets removed, its ips will still be not valid.

fedepaol (commented):
Looks good! Just a nit on redundant check over the service ip that can be simplified.

fedepaol:
Also (but in this case is not that bad because the code is quite compact), in general we try to separate the pr in various commits, the most straightforward being splitting the logic from the e2es

controller/service.go
tyiying: ok. will remove.

controller/main.go
tyiying: I see. As long as the removed nodes' IP aren't made available in the pool for the services, there won't be any risk of conflict.

controller/service.go
tyiying: This comment is resolved. Checking for service LB ip against node IPs is removed.

controller/main.go
tyiying: Comment is resolved. Code is removed.

e2etest/webhookstests/webhooks.go
```diff
@@ -106,6 +113,51 @@ var _ = ginkgo.Describe("Webhooks", func() {
...
+		ginkgo.DescribeTable("IPAddressPool with overlapping addresses of the nodes",
+			func(ipFamily ipfamily.Family) {
+				nodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+				Expect(err).NotTo(HaveOccurred())
+				nodeIps, err := k8s.NodeIPsForFamily(nodes.Items, ipFamily, "")
+				Expect(err).NotTo(HaveOccurred())
+				for _, nodeIPString := range nodeIps {
+					nodeIP := net.ParseIP(nodeIPString)
+					var nodeCIDRs []string
+					if nodeIP.To4() != nil {
```
fedepaol, line 126: we now the family already, no need to do this if. Also, it should be enough just to pick the first node, no need to iterate over all

fedepaol:
one small nit and I think it's good to go! Thanks!

e2etest/webhookstests/webhooks.go
tyiying: not sure I understand you. do you mean this if check at line 126? This check is needed to differentiate ipv4 and ipv6 since they have different CIDRMask. 

I agree, we don't need to iterate all nodes ip, first one is enough. will change.

e2etest/webhookstests/webhooks.go
```diff
@@ -106,6 +113,51 @@ var _ = ginkgo.Describe("Webhooks", func() {
...
+								Spec: metallbv1beta1.IPAddressPoolSpec{
+									Addresses: nodeCIDRs,
+								},
+							},
+						},
+					}
+					err = ConfigUpdater.Update(resources)
+					Expect(err).To(HaveOccurred())
+					Expect(err.Error()).To(ContainSubstring("contains nodeIp"))
+					break
```
tyiying, line 155: here I break out the loop after the first node.

e2etest/webhookstests/webhooks.go
fedepaol: what I mean is, we have ipFamily as parameter, so instead of parsing the node ip and trying to understand if it's v4 or v6, we should rely on that.

e2etest/webhookstests/webhooks.go
fedepaol: but then why building the nodCIDRs? It's always going to be one right?

fedepaol: or, even doing the loo in first instance, if we always break

e2etest/webhookstests/webhooks.go
tyiying: ok. what happen if ipfamily is ipfamily.DualStack? it could return both ipv4 and ipv6 nodeIPs.

fedepaol: ipfamily is a parameter we are driving here, and we have only v4 and v6 iiuc

fedepaol: I mean ipFamily is what you are setting here, https://github.com/metallb/metallb/pull/2585/files#diff-d32fb1c70f00b7179f34c7f6d38e974556efc26ffa22cc0d808991e07f13140eR158 in line 158

e2etest/webhookstests/webhooks.go
tyiying: agree. I will change.

e2etest/webhookstests/webhooks.go
```diff
@@ -106,6 +113,50 @@ var _ = ginkgo.Describe("Webhooks", func() {
...
+						},
+					}
+					err = ConfigUpdater.Update(resources)
+					Expect(err).To(HaveOccurred())
+					Expect(err.Error()).To(ContainSubstring("contains nodeIp"))
+				}
+			},
+			ginkgo.Entry("IPV4", ipfamily.IPv4),
+			ginkgo.Entry("IPV6", ipfamily.IPv6),
+			ginkgo.Entry("DUALSTACK", ipfamily.DualStack),
```
fedepaol, line 158: I don't think there's value in testing dualstack here (plus nodesip[0] is always ipv4)

fedepaol:
@tyiying I took the liberty to apply the last couple of changes myself, as all the work was done and I wanted to push this forward.
Sending to the merge queue, thanks for the contribution!

tyiying:
> @tyiying I took the liberty to apply the last couple of changes myself, as all the work was done and I wanted to push this forward. Sending to the merge queue, thanks for the contribution!

@fedepaol Thanks! please let me know if anything else need to be done.

PR #2587, opened 2024-11-08
:seedling: Mquhuy/support prefer dual stack policy

### LINKED ISSUE #2482: metallb controller fails to allocate IP address to dual stack services (PreferDualStack) if only IPv4 is available
### MetalLB Version

0.14.5

### Deployment method

Charts

### Main CNI

calico

### Kubernetes Version

_No response_

### Cluster Distribution

_No response_

### Describe the bug

Related issue:  https://github.com/metallb/metallb/issues/1339 

here is the service:

apiVersion: v1
kind: Service
metadata:
  name: nginx-ingress
  namespace: nginx-ingress
spec:
  allocateLoadBalancerNodePorts: true
  clusterIP: 10.0.0.79
  clusterIPs:
  - 10.0.0.79
  - fd00::d899
  externalTrafficPolicy: Local
  healthCheckNodePort: 30272
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  - IPv6
  ipFamilyPolicy: PreferDualStack
  ports:
  - name: http
    nodePort: 30122
    port: 80
    protocol: TCP
    targetPort: http
  - name: https
    nodePort: 32678
    port: 443
    protocol: TCP
    targetPort: https
  selector:
    app.kubernetes.io/instance: nginx-ingress
    app.kubernetes.io/name: nginx-ingress
  sessionAffinity: None
  type: LoadBalancer
status:
  loadBalancer: {}
Because the service is dual stack, but metallb has no IPv6 configured in any pool, the controller fails to allocate an IPv6 address and stops there. 

### To Reproduce

Steps:
1. Create IPaddresspool pool1 only for IPv4 family.
2. In case service created with addresspool annotation as pool1. and ipFamilyPolicy is 'PreferDualStack'.
3. Expected is the service get IP address assigned  from ipv4 family only. But actually service stuck in pending state for LBIP.
4. The same behavior can be seen without addresspool annotation if all pools are only IPv4 family.

### Expected Behavior

As mentioned above

### Additional Context

If there is only single IP family is available in metallb ipaddresspool (no IPv6 address defined) :
1. In case of preferDualstack metallb controller should allocate that IP 
2. In case of RequireDualStack it should not allocate any IP and status will be in pending state.

Currently with respect to metallb controller there is no difference between RequireDualStack and preferDualstack and both are treated as mentioned above as  'RequireDualStack'.

mquhuy:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

Fixes #2482 

**Release note**:

```release-note

```


internal/allocator/allocator_test.go
oribon, line 1: same, can you please add tests for prefer?

e2etest/l2tests/assignment.go
```diff
@@ -678,4 +678,157 @@ var _ = ginkgo.Describe("IP Assignment", func() {
 			Expect(err).NotTo(HaveOccurred())
 		})
 	})
+	ginkgo.Context("PREFER-DUALSTACK - only 1 stack available", func() {
```
oribon, line 681: nit: I think we can remove the "only 1 stack available"?

e2etest/l2tests/assignment.go
```diff
@@ -678,4 +678,157 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+			}, 5*time.Minute, 1*time.Second).Should(Equal(2))
+			ginkgo.By("validate LoadBalancer IPs are allocated from poolDual")
+			err = config.ValidateIPInRange([]metallbv1beta1.IPAddressPool{poolDual}, jigservice.GetIngressPoint(
+				&svc1.Status.LoadBalancer.Ingress[0]))
+			Expect(err).NotTo(HaveOccurred())
+			err = config.ValidateIPInRange([]metallbv1beta1.IPAddressPool{poolDual}, jigservice.GetIngressPoint(
+				&svc1.Status.LoadBalancer.Ingress[1]))
+			Expect(err).NotTo(HaveOccurred())
+		})
+		ginkgo.It("Should prefer primary ip family among single-stack pools", func() {
```
oribon, line 741: can you add another case for when we use default (ipv4) as primary?

e2etest/l2tests/assignment.go
```diff
@@ -678,4 +678,157 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+			}, 5*time.Minute, 1*time.Second).Should(Equal(1))
+			ginkgo.By("validate LoadBalancer IP is allocated from pool1")
+			err = config.ValidateIPInRange([]metallbv1beta1.IPAddressPool{pool1}, jigservice.GetIngressPoint(
+				&svc1.Status.LoadBalancer.Ingress[0]))
+			Expect(err).NotTo(HaveOccurred())
+			ginkgo.By("Updating pool1 to include both v4 and v6 addresses")
+			pool1.Spec.Addresses = []string{v4PoolAddresses, v6PoolAddresses}
+			err = ConfigUpdater.Update(config.Resources{Pools: []metallbv1beta1.IPAddressPool{pool1}})
+			Expect(err).NotTo(HaveOccurred())
+			time.Sleep(2 * time.Minute)
```
oribon, line 820: we shouldn't use sleeps, eventually should cover these cases

e2etest/l2tests/assignment.go
oribon, line 1: can you also add tests for `prefer -> require` removes the existing ip(s), and require -> prefer allocates one ip instead of none?

oribon (commented):
thanks a lot for all of your efforts!! imo looks better than the other approach 😄

controller/service.go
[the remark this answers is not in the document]
mquhuy: Thank you, that's a good find!

controller/service.go
```diff
@@ -144,6 +145,30 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
...
+	// If svc currently has 1 ip and policy PreferDualStack, try assigning ip from the missing family and same pool
+	if len(lbIPs) == 1 && familyPolicy == v1.IPFamilyPolicyPreferDualStack {
+		level.Info(l).Log("event", "tryAssignAdditionalIP", "msg", "familyPolicy is PreferDualStack, trying to assign additional ip")
+		existingFamily := ipfamily.ForAddress(lbIPs[0])
+		additionalFamily := ipfamily.IPv4
+		if existingFamily == ipfamily.IPv4 {
+			additionalFamily = ipfamily.IPv6
+		}
+		currentPool := c.ips.Pool(key)
+		newIP, err := c.ips.AllocateFromPoolForFamilyAndPolicy(key, svc, additionalFamily, currentPool, k8salloc.Ports(svc), SharingKey(svc), k8salloc.BackendKey(svc))
```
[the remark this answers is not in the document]
mquhuy: Sure, I'll add the policy to make it clearer. Thanks for pointing it out!

I'm not sure if I understand your second point correctly, do you mean that when we use Assign, the ip that was assigned from the existingFamily will be lost? But we are only assigning the additional Family here?

internal/allocator/allocator.go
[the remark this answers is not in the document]
mquhuy: It would be a very long line. Can we keep it as one-per-line?

internal/allocator/allocator.go
```diff
@@ -245,89 +247,240 @@ func (a *Allocator) Unassign(svc string) {
...
-		return nil, fmt.Errorf("unknown pool %q", poolName)
 	}
+	return poolIps
+}
 
-	ips := []net.IP{}
-	ipfamilySel := make(map[ipfamily.Family]bool)
+// SelectIPsForFamilyPolicy returns a slice of IPs from the given poolIPs, which are suitable for the
+// given service ipfamily and IPFamilyPolicy.
+func (a *Allocator) SelectIPsForFamilyAndPolicy(
```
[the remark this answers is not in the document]
mquhuy: Yes, I think that's a good idea.

internal/allocator/allocator.go
```diff
@@ -245,89 +247,240 @@ func (a *Allocator) Unassign(svc string) {
...
+		}
+	case v1.IPFamilyPolicyPreferDualStack:
+		if hasIPv4 && hasIPv6 {
+			return []net.IP{ipv4, ipv6}, nil
+		}
+		if hasIPv4 {
+			return []net.IP{ipv4}, nil
+		}
+		if hasIPv6 {
+			return []net.IP{ipv6}, nil
```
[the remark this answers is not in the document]
mquhuy: Because `if hasIPv4 && hasIPv6 {` is wrong, then there can be either ipv4 or ipv6 (or none) in the map. And we will pick the ip in both cases.

internal/allocator/allocator.go
```diff
@@ -245,89 +247,240 @@ func (a *Allocator) Unassign(svc string) {
...
+		if serviceIPFamilyPolicy != v1.IPFamilyPolicyPreferDualStack {
 			continue
 		}
-		ip := a.getIPFromCIDR(cidr, pool.AvoidBuggyIPs, svcKey, ports, sharingKey, backendKey)
-		if ip != nil {
-			ips = append(ips, ip)
-			delete(ipfamilySel, cidrIPFamily)
+		if hasPrimaryIP {
+			primaryIPPoolCandidates = append(primaryIPPoolCandidates, poolIps)
+		} else if hasSecondaryIP {
```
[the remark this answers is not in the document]
mquhuy: The idea with the slice is to loop through all pools before deciding. It could be that one of the later pools has dual stack, and we shouldn't return the first one that has any of these condition true.

internal/allocator/allocator.go
```diff
@@ -245,89 +247,240 @@ func (a *Allocator) Unassign(svc string) {
...
+	svc *v1.Service,
+	serviceIPFamily ipfamily.Family,
+	ports []Port,
+	sharingKey, backendKey string,
+) ([]net.IP, error) {
 	if alloc := a.allocated[svcKey]; alloc != nil {
 		if err := a.Assign(svcKey, svc, alloc.ips, ports, sharingKey, backendKey); err != nil {
 			return nil, err
 		}
 		return alloc.ips, nil
```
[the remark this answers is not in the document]
mquhuy: The difference here is the ipFamily we use to Allocate. If what you meant is regarding the function name, I think `Assign` is a general name, and doesn't imply "reassign everything", especially as we have the ipfamily as a param.

I'm open for suggestion though.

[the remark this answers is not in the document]
mquhuy: Not sure I follow, did we use this function for that purpose?

internal/allocator/allocator.go
```diff
@@ -245,89 +247,240 @@ func (a *Allocator) Unassign(svc string) {
...
+	}
+	ips, err = a.AllocateBestPool(allPools, svcKey, svc, serviceIPFamily, ports, sharingKey, backendKey)
+	if err == nil {
+		return ips, nil
 	}
 
 	return nil, errors.New("no available IPs")
 }
 
+// AllocateFromPool assigns an available IP from pool to service.
```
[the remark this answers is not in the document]
mquhuy: Not sure I follow this either? This function behavior doesn't change. Also, imo the usage of this function is general enough that we don't need to comment that when we define it. Maybe when we called it? (although even in that case I don't think the comment is correct).

controller/controller_test.go
```diff
@@ -673,14 +674,16 @@ func TestControllerMutation(t *testing.T) {
 			desc: "simple dual-stack LoadBalancer",
 			in: &v1.Service{
 				Spec: v1.ServiceSpec{
-					Type:       "LoadBalancer",
-					ClusterIPs: []string{"1.2.3.4", "3000::1"},
+					Type:           "LoadBalancer",
+					ClusterIPs:     []string{"1.2.3.4", "3000::1"},
+					IPFamilyPolicy: &IPFamilyPolicyRequireDualStack,
```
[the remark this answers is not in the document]
mquhuy: Why would it be the default? The default is singlestack?

e2etest/l2tests/assignment.go
mquhuy: Oops. Sorry.

internal/allocator/allocator_test.go
mquhuy: Will do.

controller/service.go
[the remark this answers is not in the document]
oribon: this in `Assign`:
```go
	alloc := &alloc{
		pool:  pool.Name,
		ips:   ips,
		ports: make([]Port, len(ports)),
		key:   *sk,
	}
	copy(alloc.ports, ports)
	a.assign(svcKey, alloc)
```
makes me think this would override the existing cache, without preserving the existing ip

controller/service.go
```diff
@@ -144,6 +145,30 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
...
+		if err != nil {
+			c.client.Errorf(svc, "AdditionalAssignFailed", "invalid requested loadbalancer IP: %s", err)
```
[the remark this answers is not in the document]
oribon: > even though it doesn't happen here (yet).

so I'd say we should go with what's cleaner now if the case of "non-nil IP with an error" can not happen

internal/allocator/allocator.go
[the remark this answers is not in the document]
oribon: that's just a personal preference and seems more aligned with the rest of the code, if you're against it you can keep it this way.

internal/allocator/allocator.go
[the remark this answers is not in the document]
oribon: > This function behavior doesn't change

right, just thought that it might be worth to add a comment now that we have more "Allocate*" functions.

internal/allocator/allocator.go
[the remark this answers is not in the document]
oribon: no because we didn't have that extra logic, so I think it would make the code a bit more readable

internal/allocator/allocator.go
```diff
@@ -245,89 +247,240 @@ func (a *Allocator) Unassign(svc string) {
...
+func (a *Allocator) AllocateBestPool(
+	pools []*config.Pool,
+	svcKey string,
+	svc *v1.Service,
+	serviceIPFamily ipfamily.Family,
+	ports []Port,
+	sharingKey, backendKey string,
+) ([]net.IP, error) {
+	poolIps, err := a.SelectPoolIPsForService(pools, svcKey, svc, serviceIPFamily, ports, sharingKey, backendKey)
+	if err == nil {
```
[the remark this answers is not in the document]
oribon: not sure I understand, if we get an error we will return a generic error anyways at the end, so why not returning it?

controller/controller_test.go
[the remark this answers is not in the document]
oribon: I thought when the cluster is dualstack and the user does not specify a policy then it is filled with require, is that not the case?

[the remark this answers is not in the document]
mquhuy: I don't think that should happen. We should give the svc dualstack only if it explicitly asks.

internal/allocator/allocator.go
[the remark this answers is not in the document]
mquhuy: So your suggestion is?

internal/allocator/allocator.go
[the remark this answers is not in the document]
mquhuy: I was meant to return the generic error if we cannot find a pool too, but I guess that's not necessary. I'll change it.

controller/service.go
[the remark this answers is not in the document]
mquhuy: It doesn't happen here, but that doesn't mean it cannot happen.
IMO the better way is to check the error if there is one.

controller/service.go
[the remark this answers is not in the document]
mquhuy: You mean in `preferDualStack` case? We do not use this function there, though?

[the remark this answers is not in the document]
mquhuy: I added the comment in line 166 of service.go

internal/allocator/allocator.go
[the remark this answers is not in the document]
mquhuy: Thank you. Tbh it doesn't matter now to me, but it did when I had to edit this function. :sweat_smile:

internal/allocator/allocator.go
```diff
@@ -245,89 +289,222 @@ func (a *Allocator) Unassign(svc string) {
...
-
-	switch serviceIPFamily {
-	case ipfamily.DualStack:
-		ipfamilySel[ipfamily.IPv4], ipfamilySel[ipfamily.IPv6] = true, true
-	default:
-		ipfamilySel[serviceIPFamily] = true
-	}
-
+// getFreeIPsFromPool determines, with best effort, an ipv4 and an ipv6 available from the provided pool.
+func (a *Allocator) getFreeIPsFromPool(
```
oribon, line 293: not sure I understand how this function returns free ips, and not just some ips in the pool (that can already be taken)?

controller/service.go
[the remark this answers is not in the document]
mquhuy: (1) Correct. `poolIps` will be filled with 1 ipv4 and 1 ipv6. Both are free, so the ipv4 is definitely not the one we assigned.

(2) We don't use the original family, but the family of the missing family, not the original ipfamily. In this case, the missing family would be ipv6, as we already had an ipv4. `selectIPsForFamilyAndPolicy` will return just the ipv6 part of the `poolIps`.

(3) The cache is only updated with ipv6, hence lbIPs will only have 2 ips.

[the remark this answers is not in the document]
oribon: > We don't use the original family, but the family of the missing family. In this case, the missing family would be ipv6, as we already had an ipv4. selectIPsForFamilyAndPolicy will return just the ipv6 part of the poolIps.

but the policy passed is still "prefer", so `selectIPsForFamilyAndPolicy` will return both ips?

> The cache is only updated with ipv6, hence lbIPs will only have 2 ips.

iiuc `Assign` overwrites the cache with the `ips` it is given, so not sure how it is "only updates with ipv6", and lbIPs won't be with dupes

it seems there is some flaw with what's happening here, I'll write it up in a different comment

oribon:
it seems there is a flaw with how we attempt to add another ip, did a short try with this pr locally:

1) start with a clean dev-env, apply the service:
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: my-pod
  namespace: metallb-system
  labels:
    app.kubernetes.io/name: MyApp
spec:
  containers:
  - name: nginx
    image: nginx:1.14.2
    ports:
    - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: my-service
  namespace: metallb-system
spec:
  selector:
    app.kubernetes.io/name: MyApp
  ports:
    - protocol: TCP
      port: 80
      targetPort: 80
  type: LoadBalancer
  ipFamilyPolicy: "PreferDualStack"
```
2) apply a pool with only ipv4:
```yaml
kind: IPAddressPool
metadata:
  name: addresspool
  namespace: metallb-system
spec:
  addresses:
  - 192.168.1.100/32
```

at this point, the service flaps between having the ip, to having it "twice":
```yaml
status:
  loadBalancer:
    ingress:
    - ip: 192.168.1.100
      ipMode: VIP
```
and:
```yaml
status:
  loadBalancer:
    ingress:
    - ip: 192.168.1.100
      ipMode: VIP
    - ip: 192.168.1.100
      ipMode: VIP
```

and the controller reconciles constantly:
```
k logs -n metallb-system controller-545fb745d-qd56k | grep "nolbIPsIPFamily" | wc -l
1373
```

3. update the address pool with an ipv6 range AND extend the existing v4 to include the assigned ip:
```yaml
kind: IPAddressPool
metadata:
  name: addresspool
  namespace: metallb-system
spec:
  addresses:
  - 192.168.1.0/24
  - fc00:f853:0ccd:e799::/124
```
and the service loses its existing ipv4:
```yaml
  loadBalancer:
    ingress:
    - ip: 192.168.1.0
      ipMode: VIP
    - ip: 'fc00:f853:ccd:e799::'
      ipMode: VIP
```

mquhuy:
Hi @oribon, 

IMO the "Additional ip should be assigned when the assigned 1-stack pool becomes dual-stack" e2e test should have covered the issue, but I agree that I was not aware the `Assign()` does the unassign first.

I added function `AllocateIPFromPoolForAdditionalFamily`, which hopefully makes it clearer what the intention is. Could you please have a look?

I think reconciling constantly is expected though. Otherwise we would never be able to detect if the pool has become dual stack.

mquhuy:
@oribon I tested your scenario as well and it seems to be normal to me. There is no "nolbIPsIPFamily" logs either.

internal/allocator/allocator_test.go
oribon, line 1: can you please add tests for the `AllocateFromPoolForAdditionalFamily` scenarios?

e2etest/l2tests/assignment.go
```diff
@@ -678,4 +697,267 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
 	})
+	ginkgo.Context("PREFER DUALSTACK", func() {
+		const v4PoolAddresses = "192.168.10.100/32"
+		const v6PoolAddresses = "fc00:f853:0ccd:e799::/124"
+		const v4PoolAddresses2 = "192.168.11.100/32"
+		const v6PoolAddresses2 = "fc00:f853:0ccd:e800::/124"
+		IPFamilyPolicyPreferDualStack := v1.IPFamilyPolicyPreferDualStack
+		IPFamilyPolicyRequireDualStack := v1.IPFamilyPolicyRequireDualStack
+		ginkgo.It("Should select dual-stack pool if available", func() {
+			poolv4 := metallbv1beta1.IPAddressPool{
```
oribon, line 708: can you declare them (poolv4/v6/dual) once as part of the context?

e2etest/l2tests/assignment.go
```diff
@@ -678,4 +697,267 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+			}
+			resources := config.Resources{
+				Pools: []metallbv1beta1.IPAddressPool{poolv6, poolv4},
+			}
+			err := ConfigUpdater.Update(resources)
+			Expect(err).NotTo(HaveOccurred())
+			ginkgo.By("Creating the service with PreferDualStack policy")
+			svc1, _ := service.CreateWithBackend(cs, testNamespace, "svc-test-ns-pool-2", func(svc *v1.Service) {
+				svc.Spec.IPFamilyPolicy = &IPFamilyPolicyPreferDualStack
+				svc.Spec.IPFamilies = []v1.IPFamily{v1.IPv4Protocol, v1.IPv6Protocol}
```
oribon, line 788: should we remove this? I think we should be as close as possible to what a user will do which is keeping this empty?

e2etest/l2tests/assignment.go
```diff
@@ -678,4 +697,267 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+			}, 5*time.Minute, 1*time.Second).Should(Equal(2))
+			ginkgo.By("validate LoadBalancer IPs are allocated from poolDual")
+			err = config.ValidateIPInRange([]metallbv1beta1.IPAddressPool{poolDual}, jigservice.GetIngressPoint(
+				&svc1.Status.LoadBalancer.Ingress[0]))
+			Expect(err).NotTo(HaveOccurred())
+			err = config.ValidateIPInRange([]metallbv1beta1.IPAddressPool{poolDual}, jigservice.GetIngressPoint(
+				&svc1.Status.LoadBalancer.Ingress[1]))
+			Expect(err).NotTo(HaveOccurred())
+		})
+		ginkgo.It("Should prefer primary ip ipv4 family among single-stack pools", func() {
```
oribon, line 761: add "by default" to the end of the test name

e2etest/l2tests/assignment.go
```diff
@@ -678,4 +697,267 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+				return len(svc1.Status.LoadBalancer.Ingress)
+			}, 5*time.Minute, 1*time.Second).Should(Equal(1))
+			ginkgo.By("validate LoadBalancer IP is allocated from pool1")
+			err = config.ValidateIPInRange([]metallbv1beta1.IPAddressPool{pool1}, jigservice.GetIngressPoint(
+				&svc1.Status.LoadBalancer.Ingress[0]))
+			Expect(err).NotTo(HaveOccurred())
+			firstIP := svc1.Status.LoadBalancer.Ingress[0].String()
+			ginkgo.By("Updating pool1 to include both v4 and v6 addresses")
+			pool1.Spec.Addresses = []string{v4PoolAddresses, v6PoolAddresses}
```
oribon, line 880: can we modify this to resemble what I commented on? that is, we also extend the ipv4 range (but still include the original ip), to be sure we aren't preserving the original ipv4 just because there is only a single one in the pool

e2etest/l2tests/assignment.go
```diff
@@ -678,4 +697,267 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+				svc1, err = cs.CoreV1().Services(svc1.Namespace).Get(context.Background(), svc1.Name, metav1.GetOptions{})
+				Expect(err).NotTo(HaveOccurred())
+				return len(svc1.Status.LoadBalancer.Ingress)
+			}, 5*time.Minute, 1*time.Second).Should(Equal(1))
+			ginkgo.By("validate LoadBalancer IP is allocated from pool1")
+			err = config.ValidateIPInRange([]metallbv1beta1.IPAddressPool{pool1}, jigservice.GetIngressPoint(
+				&svc1.Status.LoadBalancer.Ingress[0]))
+			Expect(err).NotTo(HaveOccurred())
+		})
+	})
```
oribon, line 962: can you please add another case for when there is an ipv4 pool only, the service with prefer gets a single stack ipv4, then we add a dualstack pool and verify that the service ip didn't jump to the other pool?

oribon (commented):
good work! looks a lot cleaner now. apart from the comments (which are pretty minor) looks good to me! thanks a lot
@fedepaol can you please also take a look?

controller/controller_test.go
```diff
@@ -705,12 +710,112 @@ func TestControllerMutation(t *testing.T) {
...
+					},
+				},
+				Spec: v1.ServiceSpec{
+					Type:           "LoadBalancer",
+					ClusterIPs:     []string{"1.2.3.4"},
+					IPFamilyPolicy: &IPFamilyPolicyPreferDualStack,
+				},
+				Status: statusAssigned([]string{"1.2.3.0"}),
+			},
+		},
```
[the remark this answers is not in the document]
mquhuy: Shouldn't the "Should prefer primary ipv6 ip family among single-stack pools" case have covered that?

internal/allocator/allocator_test.go
mquhuy: There's already "Additional ip should be assigned when the assigned 1-stack pool becomes dual-stack"

internal/allocator/allocator.go
```diff
@@ -245,16 +289,175 @@ func (a *Allocator) Unassign(svc string) {
...
+		ip := a.getIPFromCIDR(cidr, pool.AvoidBuggyIPs, svcKey, ports, sharingKey, backendKey)
+		if ip != nil {
+			poolIps[cidrIPFamily] = ip
+		}
+	}
+	return poolIps
+}
+
+// findBestPoolForService returns the ipPool corresponding with the most suitable pool for a serviceIPFamily.
+func (a *Allocator) findBestPoolForService(
```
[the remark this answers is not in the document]
mquhuy: I tried that out but after that the unit tests started failing: https://github.com/metallb/metallb/actions/runs/11949071913/job/33307798029?pr=2587

I think there might be some issue with this method, although I cannot find out which. If this isn't very important details, could I investigate it later and probably make another PR for it?

e2etest/l2tests/assignment.go
mquhuy: No, we cannot. These pools are configured to the `testNamespace` namespace, but the value of `testNamespace` is randomized in every test case.

mquhuy:
Hi @oribon . It took a few tries, but I finally made all the changes that you requested (except for the ones that I explained in in-line comments).

If it looks all good, could we get it merged? Downstream is really in-need of it, and I'm currently on vacation. If there's some minor detail that you still would like to change, could we maybe take it later?

e2etest/l2tests/assignment.go
```diff
@@ -678,4 +698,271 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+			err = ConfigUpdater.Update(resources)
+			Expect(err).NotTo(HaveOccurred())
+			// Sleep to let the svc jump if it wants
+			time.Sleep(2 * time.Minute)
+			ginkgo.By("Validating that the originally assigned IP hasn't changed")
+			Eventually(func() bool {
+				svc1, err = cs.CoreV1().Services(svc1.Namespace).Get(context.Background(), svc1.Name, metav1.GetOptions{})
+				Expect(err).NotTo(HaveOccurred())
+				return svc1.Status.LoadBalancer.Ingress[0].String() == firstIP
+			}, 1*time.Minute, 1*time.Second).Should(BeTrue())
```
oribon, line 965: don't use sleeps, this should be a `Consistently`

oribon (approved):
the core logic seems good, other than that `time.Sleep` lgtm
thanks a lot for your efforts and patience!
@fedepaol can you also take a look please?

e2etest/l2tests/assignment.go
mquhuy: Thank you! I've updated it.

internal/allocator/allocator.go
```diff
@@ -245,16 +299,178 @@ func (a *Allocator) Unassign(svc string) {
...
+	if primaryIPPoolCandidate != nil {
+		return primaryIPPoolCandidate, nil
+	}
+	if secondaryIPPoolCandidate != nil {
+		return secondaryIPPoolCandidate, nil
+	}
+	return nil, fmt.Errorf("no suitable pool for %s IPFamily", serviceIPFamily)
+}
+
+// allocateBestPool picks the most suitable pool and tries to allocate its ips.
```
fedepaol, line 382: please invert the order of callee / caller (https://kentcdodds.com/blog/newspaper-code-structure)

internal/allocator/allocator.go
```diff
@@ -54,6 +54,60 @@ type alloc struct {
 	key
 }
 
+// ipPool represents a mapping of IP families (IPv4, IPv6, or both) to their respective IP addresses.
```
fedepaol, line 57: this deserves a separate file I think. Also, we should think about the naming because it's not a "pool" and it's easy to mismatch it to the other "pool of ips" (and I know naming is hard :-) ).
Let's call this "allocation"

internal/allocator/allocator.go
```diff
@@ -54,6 +54,60 @@ type alloc struct {
...
+// - Just an IPv4 address
+// - Just an IPv6 address
+// - Both IPv4 and IPv6 addresses for dual-stack support.
+//
+// Key:
+// - ipfamily.IPv4: Corresponds to the IPv4 address allocated for the service.
+// - ipfamily.IPv6: Corresponds to the IPv6 address allocated for the service.
+type ipPool struct {
+	PoolName string
+	IPMap    map[ipfamily.Family]net.IP
```
fedepaol, line 69: if this is (at most) a 2 keys map, I'd suggest to have to explicit fields and a get / set method with ip and  family as parameters. It will make it easier to understand what this object is about.

e2etest/l2tests/assignment.go
```diff
@@ -36,6 +38,16 @@ var (
 	}
 )
 
+func updateServiceIPFamilyPolicy(cs clientset.Interface, namespace, serviceName string, policy *v1.IPFamilyPolicy) error {
```
fedepaol, line 41: nit: util functions on the bottom (or in their own package)

internal/allocator/allocator.go
```diff
@@ -263,69 +479,79 @@ func (a *Allocator) AllocateFromPool(svcKey string, svc *v1.Service, serviceIPFa
...
+}
 
-	switch serviceIPFamily {
-	case ipfamily.DualStack:
-		ipfamilySel[ipfamily.IPv4], ipfamilySel[ipfamily.IPv6] = true, true
-	default:
-		ipfamilySel[serviceIPFamily] = true
+// AllocateFromPoolForFamilyAndPolicy assigns an available IP from pool to service, with a specified serviceIPFamily
+// and suitable for the service IPFamilyPolicy.
+func (a *Allocator) AllocateFromPoolForFamilyAndPolicy(
```
fedepaol, line 492: there are a few too many Allocate* functions. Let's squash this with the one above as it always ends up in here and this is short enough.

internal/allocator/allocator.go
```diff
@@ -245,16 +299,178 @@ func (a *Allocator) Unassign(svc string) {
...
+		return primaryIPPoolCandidate, nil
+	}
+	if secondaryIPPoolCandidate != nil {
+		return secondaryIPPoolCandidate, nil
+	}
+	return nil, fmt.Errorf("no suitable pool for %s IPFamily", serviceIPFamily)
+}
+
+// allocateBestPool picks the most suitable pool and tries to allocate its ips.
+func (a *Allocator) allocateBestPool(
```
fedepaol, line 383: let's call this allocateFromPools

fedepaol (commented):
Thanks for this! I left a few comments, but they are mostly about readability and naming, nothing major (still need to go over e2es)

e2etest/l2tests/assignment.go
```diff
@@ -678,4 +698,269 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+			ObjectMeta: metav1.ObjectMeta{Name: "test-ns-dualstack-pool"},
+			Spec: metallbv1beta1.IPAddressPoolSpec{
+				Addresses: []string{
+					v4PoolAddresses2,
+					v6PoolAddresses2,
+				},
+			},
+		}
+		ginkgo.BeforeEach(func() {
+			poolv4.Spec.AllocateTo = &metallbv1beta1.ServiceAllocation{Namespaces: []string{testNamespace}}
```
fedepaol, line 732: do we need the allocation?

e2etest/l2tests/assignment.go
```diff
@@ -678,4 +698,269 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+				svc1, err = cs.CoreV1().Services(svc1.Namespace).Get(context.Background(), svc1.Name, metav1.GetOptions{})
+				Expect(err).NotTo(HaveOccurred())
+				return len(svc1.Status.LoadBalancer.Ingress)
+			}, 5*time.Minute, 1*time.Second).Should(Equal(1))
+			ginkgo.By("validate LoadBalancer IP is allocated from pool1")
+			err = config.ValidateIPInRange([]metallbv1beta1.IPAddressPool{pool1}, jigservice.GetIngressPoint(
+				&svc1.Status.LoadBalancer.Ingress[0]))
+			Expect(err).NotTo(HaveOccurred())
+		})
+		ginkgo.It("Svc assigned from single-stack pool should not jump to dual-stack pool when it's available", func() {
```
fedepaol, line 927: Can we see what happens if we remove an IP from a dual stack service (primary / secondary)? What is the expected behavior that we want?

e2etest/l2tests/assignment.go
mquhuy: I think we do, `testnamespace` is changed for every It, so it's impossible to do the Allocation beforehands. Or maybe there's something I'm missing here?

e2etest/l2tests/assignment.go
```diff
@@ -678,4 +698,269 @@ var _ = ginkgo.Describe("IP Assignment", func() {
 			Expect(err).NotTo(HaveOccurred())
 		})
 	})
+	ginkgo.Context("PREFER DUALSTACK", func() {
```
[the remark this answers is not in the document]
mquhuy: This is already in the e2e commit?

e2etest/l2tests/assignment.go
mquhuy: idk. What do you think should be the expected behavior?

mquhuy: Imo in this case the ips should be reassigned? In case we have a dualstack pool available?

mquhuy:
Hi @fedepaol. Thank you so much for the comments. I've now fixed most/all of the issues you pointed out.
The only thing I'm not sure is the dualstack->singlestack behavior that you mentioned. Could you please check my response there and see if it makes senses to you?

e2etest/l2tests/assignment.go
fedepaol: if you don't specify the selector, the pool will apply to all the services. Not sure if I am missing something myself :D

e2etest/l2tests/assignment.go
fedepaol: yeah I don't have a strong opinion. On one side, we are promising the ip of the service not to change, on the other we change the ip with a good one in case the old one goes away.

I'd say we reassign if a dual stack pool is available, but we keep it as it is (cleaning one ip) if no other dual stack pool is available. Would it make sense? @oribon thoughts?

controller/service.go
```diff
@@ -189,6 +211,30 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 	return nil
 }
 
+// serviceFamilyChanged determines if lbIP has different ipfamily
+// than what it's supposed to have.
+func (c *controller) serviceFamilyChanged(
```
fedepaol, line 216: nit: this can be a function, no need to be a method

internal/allocator/allocation.go
```diff
@@ -0,0 +1,122 @@
...
+// Key:
+// - ipfamily.IPv4: Corresponds to the IPv4 address allocated for the service.
+// - ipfamily.IPv6: Corresponds to the IPv6 address allocated for the service.
+type Allocation struct {
+	PoolName string
+	IPV4     net.IP
+	IPV6     net.IP
+}
+
+func newAllocation(poolName string) *Allocation {
```
fedepaol, line 29: nit (sorry I didn't notice this before): no need to have a constructor, we can just use literal init with the poolname

internal/allocator/allocation.go
```diff
@@ -0,0 +1,122 @@
...
+	return &Allocation{
+		PoolName: poolName,
+		IPV4:     nil,
+		IPV6:     nil,
+	}
+}
+
+func (a *Allocation) getIPV4() net.IP {
```
fedepaol, line 37: my suggestion (which would have made this more compatible with the previous) was to have getIPForFamily(family) / setIPForFamily(family) methods. But this is fine too

internal/allocator/allocation.go
```diff
@@ -0,0 +1,122 @@
...
+	ip := a.IPV4
+	if ip == nil {
+		return nil
+	}
+	if ip.To4() == nil {
```
fedepaol, line 42: I have mixed feelings about this check. I'd say we should trust who filled this (still us) so in case of error let it explode.

internal/allocator/allocator.go
```diff
@@ -245,16 +245,178 @@ func (a *Allocator) Unassign(svc string) {
...
+) *Allocation {
+	allocation := newAllocation(pool.Name)
+	for _, cidr := range pool.CIDR {
+		cidrIPFamily := ipfamily.ForCIDR(cidr)
+		if ip := allocation.selectIPForSingleStack(cidrIPFamily); ip != nil {
+			continue
+		}
+		ip := a.getIPFromCIDR(cidr, pool.AvoidBuggyIPs, svcKey, ports, sharingKey, backendKey)
+		if ip != nil {
+			if cidrIPFamily == ipfamily.IPv4 {
```
fedepaol, line 264: yes with the ipfamily variant this would be simpler

internal/allocator/allocator.go
```diff
@@ -552,3 +705,12 @@ func (a *Allocator) checkSharing(svc string, ip string, ports []Port, sk *key) e
 	}
 	return nil
 }
+
+// determineServiceIPPolicy determines the IPFamilyPolicy of a given svc.
+func determineServiceIPPolicy(svc *v1.Service) v1.IPFamilyPolicy {
```
fedepaol, line 710: nit; ipPolicyForService

fedepaol (commented):
This is shaping up really nicely! Left a couple of few other nits

e2etest/l2tests/assignment.go
oribon: sounds good to me (re-allocating or keeping if possible)

mquhuy:
Thank you @oribon. These were all great suggestion. I've implemented them all.

fedepaol:
looks good! Last ask is to add a couple of test for that "removed" scenario and eventually hammer it.
Thanks for the patience!

mquhuy:
> looks good! Last ask is to add a couple of test for that "removed" scenario and eventually hammer it.
> 
> Thanks for the patience!

Sorry but could you please elaborate it a little? I have added a test which verified that after the ipv4 range is removed from the stack, the service will still get ipv6 (changes from 2 ips to 1 ip). Is that (one of) the scenario(s) that you are thinking of? What else should I test there?

fedepaol:
> > looks good! Last ask is to add a couple of test for that "removed" scenario and eventually hammer it.
> > Thanks for the patience!
> 
> Sorry but could you please elaborate it a little? I have added a test which verified that after the ipv4 range is removed from the stack, the service will still get ipv6 (changes from 2 ips to 1 ip). Is that (one of) the scenario(s) that you are thinking of? What else should I test there?

I was talking about what we discussed here https://github.com/metallb/metallb/pull/2587#discussion_r1860021064
If it looses one ip and a dual stack is available, we should move the ips.

e2etest/l2tests/assignment.go
mquhuy: Sure, it's just that in our e2e tests, there's always a selector (to the testnamespace) in all services so I didn't want to change that. If that's acceptable I can remove the allocation.

mquhuy:
> I was talking about what we discussed here [#2587 (comment)](https://github.com/metallb/metallb/pull/2587#discussion_r1860021064) If it looses one ip and a dual stack is available, we should move the ips.

Thank you. Actually, that's kinda causing an issue.
The current way of checking the ips is as a whole. We try an idempotent Assign() on the current lbIPs, and if there's an error, we clear the state and try to reassign. If there's another dualstack pool available at this stage, that pool will be picked, but if there's not, there's no way we can ensure to pick the exact ip that the service is using (of the remaining stack).

I have been thinking of how to implement this behavior, but the method I found requires a change in the way convergeBalancer() works, which imo deserves its own PR. Maybe I'm missing something here? Could you please give me a hint?

fedepaol:
> > I was talking about what we discussed here [#2587 (comment)](https://github.com/metallb/metallb/pull/2587#discussion_r1860021064) If it looses one ip and a dual stack is available, we should move the ips.
> 
> Thank you. Actually, that's kinda causing an issue. The current way of checking the ips is as a whole. We try an idempotent Assign() on the current lbIPs, and if there's an error, we clear the state and try to reassign. If there's another dualstack pool available at this stage, that pool will be picked, but if there's not, there's no way we can ensure to pick the exact ip that the service is using (of the remaining stack).
> 
> I have been thinking of how to implement this behavior, but the method I found requires a change in the way convergeBalancer() works, which imo deserves its own PR. Maybe I'm missing something here? Could you please give me a hint?
>

 
This is also true for a service with one single IP where the pool suddenly gets ips for the other family, right? Which can be extended to "the service ips remain the same as long as the ips (or the ip) are available. If the condition doesn't apply anymore, a reallocation happens". Is this correct? I need to think about it.

On one side, I'd like to say we shouldn't add more complexity and just document the behavior, on the other I am afraid this kinds of defeats the purpouse of this PR a bit. One service that becomes single stack, stays single stack forever. Need to think a bit about this.

fedepaol:
Deleted a reply, was looking at the wrong checkout :sweat_smile: 
Will have a look at the code again tomorrow, hope I'll be a bit sharper.

Thanks again :-)

mquhuy:
> This is also true for a service with one single IP where the pool suddenly gets ips for the other family, right? Which can be extended to "the service ips remain the same as long as the ips (or the ip) are available. If the condition doesn't apply anymore, a reallocation happens". Is this correct? I need to think about it.
> 
> On one side, I'd like to say we shouldn't add more complexity and just document the behavior, on the other I am afraid this kinds of defeats the purpouse of this PR a bit. One service that becomes single stack, stays single stack forever. Need to think a bit about this.

In case the svc initially gets single stack, but the pool gets extended, then the svc will get an additional ip from the newly added stack (while the ip from the original stack stays still). This was one of the agreed behaviors we had from the start, and we have a new e2e test to cover that case. We won't go into the problem that once the svc becomes single stack, it stays single stack.

The only issue now is that if a svc is initially assigned with dualstack, then the pool loses a stack, then the ip initially assigned from that stack will be considered malfunctioned, and by default, if one of the assigned ips is malfunctioned, then we reassign the whole thing, so there's no guarantee that (if there's no other dual stack pool) the remaining stack will get the same ip it got before. I would say that this isn't an "issue" per se, so I agree with you that we should just document it.

fedepaol:
Also, @mquhuy , mind adding a very small section under "advanced address pool configuration" (in another pr)?

Thanks!

mquhuy:
> Also, @mquhuy , mind adding a very small section under "advanced address pool configuration" (in another pr)?
> 
> 
> 
> Thanks!

Sure! Thank you both @fedepaol and @oribon for being patient with me 😄 It's been a long journey and I have learned a lot about metallb. Looking forward to contributing more to the project!

Issue #2431, opened 2024-06-13
config metallb bgp ipv4 ippool with native mode failed


changluyi:
### MetalLB Version

0.14.5

### Deployment method

Operator

### Main CNI

kube-ovn

### Describe the bug

it is a dual cluster,  and I run metallb with native mode,  if you first create a dual ipaddressPool  A , and associate the ipaddressPool A to l2advertisement, and secondly create a ipv4 ipaddressPool B, and associate the ipaddressPool  to a bgpadvertisement, it will get a webhook error 

- ` pool "A" has ipv6 CIDR fd01::/120, native bgp mode does not support ipv6`

### To Reproduce

1. create ipaddressPool  A and l2advertisement with below:
```
kind: IPAddressPool
metadata:
  name: test
  namespace: metallb-system
spec:
  addresses:
  - 1.1.1.1/24
  - fd01::03/120
  autoAssign: true
  avoidBuggyIPs: true
---
apiVersion: metallb.io/v1beta1
kind: L2Advertisement
metadata:
  generation: 1
  name: test
  namespace: metallb-system
spec:
  ipAddressPools:
  - test
```

2. create ipaddressspool test2
```
apiVersion: metallb.io/v1beta1
kind: IPAddressPool
metadata:
  name: test2
  namespace: metallb-system
spec:
  addresses:
  - 4.4.4.4/24
  autoAssign: true
  avoidBuggyIPs: true
```

3. associate test2 to bgpadvertisement
```
apiVersion: metallb.io/v1beta1
kind: BGPAdvertisement
metadata:
  name: test2
  namespace: metallb-system
spec:
  ipAddressPools:
  - test2
```

webhook will raise error:
```
Error from server (Forbidden): error when creating "bgpadv.yaml": admission webhook "bgpadvertisementvalidationwebhook.metallb.io" denied the request: pool "test" has ipv6 CIDR fd01::/120, native bgp mode does not support ipv6
```

### Expected Behavior

when test2 ippool associate with bgpadvertisement , should not check test1's ipv6 cidr.

### Additional Context

![image](https://github.com/metallb/metallb/assets/47097611/64b1a4b3-9352-4bce-9dbf-a3bb2a47e767)
the related code is above, 
I think the code inspection at this location needs to be more accurate

### I've read and agree with the following

- [X] I've checked all open and closed issues and my request is not there.
- [X] I've checked all open and closed pull requests and my request is not there.

### I've read and agree with the following

[...]

PR #2593, opened 2024-11-08
IPV6 bgp validation: validate bgpadvertisement selected pools

fedepaol:

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

We can have l2 ipv6 addresses when using metallb in native mode. The important thing to consider is, not having any v6 based pool selected by a bgp advertisement. Here we narrow down the scope of the checking function.

Fixes https://github.com/metallb/metallb/issues/2431

**Release note**:

```release-note
BGP native mode: narrow down the validation so that ipv6 pools are not valid only if there is a bgp advertisement referencing them.
```

### DIFF
--- a/internal/config/validation.go
+++ b/internal/config/validation.go
@@ -7,9 +7,12 @@ import (
 
 	"errors"
 
+	metallbv1beta1 "go.universe.tf/metallb/api/v1beta1"
 	metallbv1beta2 "go.universe.tf/metallb/api/v1beta2"
 	"go.universe.tf/metallb/internal/bgp/community"
 	"go.universe.tf/metallb/internal/ipfamily"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 )
 
 type Validate func(ClusterResources) error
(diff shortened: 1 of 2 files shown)


internal/config/validation.go
```diff
@@ -69,8 +72,18 @@ func findIPv6BGPAdvertisement(c ClusterResources) error {
...
+	bgpSelectors, err := poolSelectorsForBGP(c)
+	if err != nil {
+		return err
+	}
+
 	for _, p := range c.Pools {
 		for _, cidr := range p.Spec.Addresses {
+			if !bgpSelectors.matchesPool(p) {
+				continue
+			}
```
oribon, line 85: can we have this above the cidr loop?

internal/config/validation.go
```diff
@@ -203,3 +216,45 @@ func hasBFDEcho(peer *Peer, bfdProfiles map[string]*BFDProfile) bool {
...
+	}
+	for _, l := range s.byLabels {
+		if l.Matches(labels.Set(p.Labels)) {
+			return true
+		}
+	}
+	return false
+}
+
+func poolSelectorsForBGP(c ClusterResources) (poolSelector, error) {
```
oribon, line 241: I think we should also handle the scenario where the bgpadv does not have pools/selectors?

fedepaol: we do, in the `matchesPool` above we always match if there are no selectors

oribon (approved):
lgtm

PR #2595, opened 2024-11-11
E2E: add changes to support unnumbered BGP and better test BFD

karampok:
Important change is that the parsing neighbors does not fail if address IP is nil, as now this is happy path for unnumbered BGP.

Also, adds
- opendelay timer in peers so local frr starts the peering
- BFD passive mode in peering
- increased debug level in peers
- function to create point-to-point connection
- function to join IPv4, IPv6 routes for family agnostic check that is useful to unnumbered

kind bug

To support testing unnumber, to increase testing for BFD

-->

```release-note
NONE
```


fedepaol:
why kind bug?

karampok:
is not really bug, neither feature. Any specific tag you prefer for that PR?

karampok:
@oribon this PR also need for the unnumber feature, thanks.

e2etest/pkg/frr/container/container.go
```diff
@@ -198,7 +205,10 @@ func configureContainer(cfg Config, configDir string) (*FRR, error) {
 		Network:        cfg.Network,
 	}
 
-	if cfg.Network == hostNetwork {
+	switch cfg.Network {
+	case noNetwork:
+		break
```
oribon, line 210: can we be explicit here instead?
```go
frr.Ipv4 = ""
frr.Ipv6 = ""
```

e2etest/pkg/frr/container/container.go
```diff
@@ -365,3 +374,30 @@ func containerIsRunning(containerName string) error {
 
 	return nil
 }
+
+func SetupP2PPeer(frrImage string, node corev1.Node) (*FRR, error) {
```
oribon, line 378: seems the functions care only about the container's name, not the entire corev1.Node. wdyt about
```go
CreateP2PPeerFor(containerName, frrImage string)?
```

e2etest/pkg/frr/container/container.go
```diff
@@ -365,3 +374,30 @@ func containerIsRunning(containerName string) error {
...
+	c := Config{
+		Name:    fmt.Sprintf("unnumbered-p2p-%d", time.Now().Unix()),
```
oribon, line 380: wdyt about naming the peer after the container it is peered with? 
```go
fmt.Sprintf("p2p-%s", containerName)
```

e2etest/pkg/frr/config/unnumbered.go
```diff
@@ -0,0 +1,97 @@
...
+				neighbor MTLB activate
+		exit-address-family
+		!
+		address-family ipv6 unicast
+				redistribute connected
+				neighbor MTLB activate
+		exit-address-family
+	exit`
+
+func WirePeer(peerName string, n corev1.Node) error {
```
oribon, line 61: I'd say this belongs to the container package (has nothing to do with frr), and be renamed something like `wireContainers`. Also, can you add a comment explaining what it does (+ is there some sort of relation between the containers, or what matters is that there are 2 containers?)

e2etest/pkg/frr/container/container.go
```diff
@@ -365,3 +374,30 @@ func containerIsRunning(containerName string) error {
...
+	}
+
+	peer := peers[0]
+
+	if err := frrconfig.WirePeer(peer.Name, node); err != nil {
+		return nil, fmt.Errorf("wire the peer failed - %w", err)
+	}
+	if err := peer.UpdateBGPConfigFile(frrconfig.UnnumberedPeerFRRConfig); err != nil {
+		return nil, err
+	}
```
oribon, line 401: imo this part does not belong as part of this function, we should keep this package around "containers setup", and call the frr bgp config parts where needed in the tests / where it makes more sense

e2etest/pkg/frr/parse.go
```diff
@@ -197,9 +207,9 @@ func ParseNeighbours(vtyshRes string) ([]*Neighbor, error) {
 	res := make([]*Neighbor, 0)
 	for k, n := range toParse {
 		ip := net.ParseIP(k)
-		if ip == nil {
-			return nil, fmt.Errorf("failed to parse %s as ip", ip)
-		}
+		// if ip == nil {
+		// 	return nil, fmt.Errorf("failed to parse %s as ip", ip)
+		// }
```
oribon, line 212: no need to keep this around

e2etest/pkg/frr/parse.go
```diff
@@ -197,9 +207,9 @@ func ParseNeighbours(vtyshRes string) ([]*Neighbor, error) {
 	res := make([]*Neighbor, 0)
 	for k, n := range toParse {
 		ip := net.ParseIP(k)
```
oribon, line 209: we have been treating the ip in some places as the id of the neighbor in some sense, and now it would be lost (or cause mess) in unnumbered (e.g where we use `res[n.IP.String()]`) because the field would be nil, wdyt would be the right way to support it?

e2etest/pkg/frr/config/unnumbered.go
```diff
@@ -0,0 +1,97 @@
...
+	"bytes"
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"go.universe.tf/e2etest/pkg/executor"
+	corev1 "k8s.io/api/core/v1"
+)
+
+const UnnumberedPeerFRRConfig = `
```
oribon, line 15: I'd expect this to be a template

oribon (commented):
can you split the commits to what's related to unnumbered and the general improvements?

karampok:
> can you split the commits to what's related to unnumbered and the general improvements?

Now that I have include BFD with unnumbered peering I am using everything (re-wording commit and PR title)

e2etest/pkg/frr/parse.go
karampok: Technicallly is not nil, is string `<nil>`
```
func (ip IP) String() string {
	if len(ip) == 0 {
		return "<nil>"
	}
``` which makes sure there is no panicking or anything. So if you see `<nil>` is the unnumber test.

Alternatively I could add another field `.ID` which is string and can be either `ip` or `interface` and refactor all references, but not sure is worth it.

What do you prefer? Let me know if you have a better suggestion. Thanks

File: e2etest/pkg/frr/parse.go (hunk 6 above)
Comment (karampok) [reply e2etest/pkg/frr/parse.go]: done

File: e2etest/pkg/frr/container/container.go (hunk 5 above)
Comment (karampok) [reply e2etest/pkg/frr/container/container.go]: done, will be added in the ffr-k8s repo in the test

File: e2etest/pkg/frr/config/unnumbered.go (hunk 4 above)
Comment (karampok) [reply e2etest/pkg/frr/config/unnumbered.go]: done
- it does matter because we hardcode the mac in the node part, which is required to calculate the LLA ipv6 address
- having defined mac/LLA is very helpful for debuging

File: e2etest/pkg/frr/container/container.go (hunk 3 above)
Comment (karampok) [reply e2etest/pkg/frr/container/container.go]: done

File: e2etest/pkg/frr/container/container.go (hunk 2 above)
Comment (karampok) [reply e2etest/pkg/frr/container/container.go]: done

File: e2etest/pkg/frr/container/container.go (hunk 1 above)
Comment (karampok) [reply e2etest/pkg/frr/container/container.go:211]: done

File: e2etest/pkg/frr/config/unnumbered.go (hunk 4 above)
Comment (oribon) [reply e2etest/pkg/frr/config/unnumbered.go]: I meant in the same file as the `SetupP2PPeer` (which imo should still be called `CreateP2PPeerFor`). Also should we have a const for that mac/lla if that's the case?

File: e2etest/pkg/frr/parse.go (hunk 7 above)
Comment (karampok) [reply e2etest/pkg/frr/parse.go:213]: I have not been able to find a right way, we could refactor (e.g IP to be string and holds the eth0 value, but then IP is not making sense, so we would need IP to peer naming change). We could say if "<nil>" use the routeID?

File: e2etest/pkg/container/routes.go [hunk 9]
```diff
@@ -93,3 +98,140 @@ func SetupVRFForNetwork(containerName, vrfNetwork, vrfName, vrfRoutingTable stri
...
+			Family string `json:"family"`
+			Host   string `json:"host"`
+		} `json:"via,omitempty"`
+		Dev string `json:"dev"`
+	}
+
+	// `ip -4 route show dev eth0` does not work if multiple nexthops
+	// 200.200.200.0/24 nhid 49 proto bgp metric 20
+	//       nexthop via 172.18.0.5 dev eth0 weight 1
+	//       nexthop via 172.18.0.6 dev eth0 weight 1
```
oribon, line 176: nit: unnecessary

e2etest/pkg/container/routes.go
```diff
@@ -93,3 +98,140 @@ func SetupVRFForNetwork(containerName, vrfNetwork, vrfName, vrfRoutingTable stri
...
+	cmd := "ip %s --json route show proto bgp" // dev %s
```
oribon, line 177: declare the cmd as args []string inside the loop,

e2etest/pkg/container/routes.go
```diff
@@ -93,3 +98,140 @@ func SetupVRFForNetwork(containerName, vrfNetwork, vrfName, vrfRoutingTable stri
...
+	for _, proto := range []string{"-4", "-6"} {
+		out, err := exc.Exec("bash", "-c", fmt.Sprintf(cmd, proto))
```
oribon, line 179: is the "bash -c" needed?

e2etest/pkg/container/routes.go
```diff
@@ -93,3 +98,140 @@ func SetupVRFForNetwork(containerName, vrfNetwork, vrfName, vrfRoutingTable stri
...
+		}
+
+		var routes []IPRoute
+		err = json.Unmarshal([]byte(out), &routes)
+		if err != nil {
+			return nil, fmt.Errorf("failed to parse JSON output: %w", err)
+		}
+
+		for _, r := range routes {
+			prefix, err := netip.ParsePrefix(r.Dst)
```
oribon, line 191: nit: prefix -> dst

e2etest/pkg/container/routes.go
```diff
@@ -93,3 +98,140 @@ func SetupVRFForNetwork(containerName, vrfNetwork, vrfName, vrfRoutingTable stri
...
+
+// GetIPv6Link return the LinkLocalAddress (LLA) of a specific interface.
+func GetIPv6Link(containerName, dev string) (string, error) {
+	exc := executor.ForContainer(containerName)
+	return netdev.LinkLocalAddressForDevice(exc, dev)
+}
+
+// BGPRoutes executes `ip route` command to fetch BGP routes
+// and return a map of map leveraging netip.{Prefix,Addr} which
+// allow direct comparison.
```
oribon, line 151: can you make it clearer that what is returned is a map of routes -> their set of nexthops, and the role of "dev" here? also, using bool instead of struct{} is a bit cleaner

e2etest/pkg/container/routes.go
```diff
@@ -93,3 +98,140 @@ func SetupVRFForNetwork(containerName, vrfNetwork, vrfName, vrfRoutingTable stri
...
+		}
+
+		for _, r := range routes {
+			prefix, err := netip.ParsePrefix(r.Dst)
+			if err != nil {
+				return nil, fmt.Errorf("invalid prefix %s: %w", r.Dst, err)
+			}
+
+			if _, exists := routeMap[prefix]; !exists {
+				routeMap[prefix] = make(map[netip.Addr]struct{})
```
oribon, line 197: do we store destinations regardless if they have a nexthop via the dev?

e2etest/pkg/frr/config/unnumbered.go
```diff
@@ -0,0 +1,67 @@
...
+exit`
+
+type RouterConfigUnnumbered struct {
+	ASNLocal      uint32
+	ASNRemote     uint32
+	Hostname      string
+	Interface     string
+	RouterID      string
+	ToAdvertiseV4 string
+	ToAdvertiseV6 string
```
oribon, line 52: can each of these be a []string?

e2etest/pkg/container/routes.go
```diff
@@ -93,3 +98,140 @@ func SetupVRFForNetwork(containerName, vrfNetwork, vrfName, vrfRoutingTable stri
...
+	netNSCntA = bytes.TrimSpace(netNSCntA)
+
+	netNSCntB, err := exec.Command(executor.ContainerRuntime,
+		"inspect", "-f", "{{ .NetworkSettings.SandboxKey }}", containerB).CombinedOutput()
+	if err != nil {
+		return fmt.Errorf("%s - %w", netNSCntB, err)
+	}
+	netNSCntB = bytes.TrimSpace(netNSCntB)
+
+	c := fmt.Sprintf("ip link add %s netns %s type veth peer name %s", dev, netNSCntA, dev)
```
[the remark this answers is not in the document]
oribon: it is more readable

e2etest/pkg/container/routes.go
oribon: right, it's not something critical, just what we did in other places in the code iirc (also, did you try using `k8s.io/apimachinery/pkg/util/sets` for this?).
also, the:
```
// The return json from the kernel varies depend if ipv4 with single entry, or
// ipv4 with multiple entries or ipv6. The use of maps, and using keys the types
// netip.{Prefix,Addr} allows direct comparison.
```
part is unnecessary

karampok: Removed the part. I have not really tried to use the set from the k8s.io. I think the test function is okay as is using the standard lib but now strong opinion. Would you like that I refactor to use the `k8s.io/apimachinery/pkg/util/set`?

oribon: not a big deal, we can leave it as is

PR #2605, opened 2024-11-22
Add unnumbered BGP peering for frrk8s backend 

karampok:
/kind feature

**Release note**:

```release-note
Add unnumbered BGP peering for frrk8s backend 
```


internal/bgp/frr/frr.go
```diff
@@ -254,17 +253,10 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
 
 		neighborName := NeighborName(s.PeerAddress, s.PeerASN, s.DynamicASN, s.VRFName)
 		if neighbor, exist = rout.neighbors[neighborName]; !exist {
-			host, port, err := net.SplitHostPort(s.PeerAddress)
-			if err != nil {
-				return nil, err
+			if s.PeerPort == 0 {
```
oribon, line 256: can we block this sooner when the config is parsed?

internal/bgp/frrk8s/frrk8s.go
```diff
@@ -48,10 +47,22 @@ type session struct {
 	logger         log.Logger
 }
 
-// sessionName() defines the format of the key of the 'sessions' map in
-// the 'frrState' struct.
+// sessionName() defines the format of a map key.
 func sessionName(s session) string {
-	baseName := fmt.Sprintf("%d@%s-%d@%s", s.PeerASN, s.PeerAddress, s.MyASN, s.SourceAddress)
+	f := func(ASN uint32, dynamicASN string) string {
```
oribon, line 52: can we just use the contents of this function to set asn? this is a bit confusing, especially since it's used once

internal/bgp/frrk8s/frrk8s.go
```diff
@@ -227,23 +238,15 @@ func (sm *sessionManager) updateConfig() error {
...
-			if err != nil {
-				return err
-			}
-			portUint16 := uint16(portUint)
-
 			if !reflect.DeepEqual(s.PasswordRef, corev1.SecretReference{}) && s.Password != "" {
 				return fmt.Errorf("invalid session with password and secret set: %s", sessionName(*s))
 			}
 
+			if s.PeerPort == 0 {
```
oribon, line 247: same

oribon (commented):
can you update the `DiscardFRROnly` func to block interface?
also, if we decide this is only for frr-k8s I guess we should block it in frr mode only as well (currently they share the validation)

e2etest/bgptests/unnummbered.go
```diff
@@ -0,0 +1,198 @@
...
+	"github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+var FRRImage string
+var _ = ginkgo.Describe("IPV4 FRRK8S-MODE Unnumbered BGP", func() {
```
oribon, line 37: why is this v4 only?

e2etest/bgptests/unnummbered.go
```diff
@@ -0,0 +1,198 @@
...
+			}
+
+			err = ConfigUpdater.Update(resources)
+			Expect(err).NotTo(HaveOccurred(), "apply the CR in k8s api failed")
+
+			nodeP2PContainer := executor.ForContainer(nodeWithP2PConnection.Name)
+			nodeLLA, err := netdev.LinkLocalAddressForDevice(nodeP2PContainer, iface)
+			Expect(err).NotTo(HaveOccurred())
+			// peerLLA, err := netdev.LinkLocalAddressForDevice(externalP2PContainer, iface)
+			// Expect(err).NotTo(HaveOccurred())
```
oribon, line 133: leftover?

e2etest/bgptests/unnummbered.go
```diff
@@ -0,0 +1,198 @@
...
+			ginkgo.By("Checking the service gets an ip assigned")
+			for _, i := range svc.Status.LoadBalancer.Ingress {
+				ingressIP := jigservice.GetIngressPoint(&i)
+				err = config.ValidateIPInRange(resources.Pools, ingressIP)
+				Expect(err).NotTo(HaveOccurred())
+			}
+
+			ginkgo.By(fmt.Sprintf("validating the p2p peer %s received the routes from node", externalP2PContainer.Name))
+			validatePrefixViaFor(externalP2PContainer, iface, nodeLLA, prefixSend)
+
```
oribon, line 148: is it possible to check that the external client can reach the service?

e2etest/bgptests/unnummbered.go
```diff
@@ -0,0 +1,198 @@
...
+	ginkgo.Context("FRR when applying unnumbered config", func() {
+		ginkgo.It("Sessions should be established", func() {
+			rc := frrconfig.RouterConfigUnnumbered{
+				ASNLocal:      metalLBASN,
+				ASNRemote:     metalLBASN,
+				Hostname:      "tor1",
+				Interface:     "net1",
+				RouterID:      "1.1.1.1",
+				ToAdvertiseV4: []string{"1.1.1.0/24"},
+				ToAdvertiseV6: []string{"1111::/64"},
```
oribon, line 79: does this do something?

e2etest/bgptests/unnummbered.go
```diff
@@ -0,0 +1,198 @@
...
+
+func validateUnnumberedBGPPeering(peer *frrcontainer.FRR, nodeLLA string) {
+	ginkgo.By(fmt.Sprintf("validating BGP peering to %s", peer.Name))
+	Eventually(func() error {
+		neighbors, err := frr.NeighborsInfo(peer)
+		if err != nil {
+			return err
+		}
+		for _, n := range neighbors {
+			if n.BGPNeighborAddr == nodeLLA && n.Connected { //&& n.BFDInfo.Status == "Up" {
```
oribon, line 165: can we add bfd? 😄

e2etest/bgptests/unnummbered.go
```diff
@@ -0,0 +1,198 @@
...
+			ginkgo.By("validating the node and p2p container peered")
+			validateUnnumberedBGPPeering(externalP2PContainer, nodeLLA)
+
+			svc, _ := testservice.CreateWithBackend(cs, testNamespace, "external-local-lb")
+			ginkgo.By("Checking the service gets an ip assigned")
+			for _, i := range svc.Status.LoadBalancer.Ingress {
+				ingressIP := jigservice.GetIngressPoint(&i)
+				err = config.ValidateIPInRange(resources.Pools, ingressIP)
+				Expect(err).NotTo(HaveOccurred())
+			}
```
oribon, line 144: let's verify that the service got the prefixSend (it makes the test clearer)

internal/config/config_test.go
```diff
@@ -3499,6 +3499,104 @@ func TestParse(t *testing.T) {
...
+						NodeSelectors: []labels.Selector{labels.Everything()},
+						EBGPMultiHop:  false,
+					},
+				},
+				Pools:       &Pools{ByName: map[string]*Pool{}},
+				BFDProfiles: map[string]*BFDProfile{},
+			},
+		},
+		{
+			desc: "unnumbered peer with address without interface ok",
```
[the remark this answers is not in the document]
oribon: yep

internal/bgp/frrk8s/frrk8s.go
oribon: yeah I mean block it from ever getting to this situation (blocking the bad config earlier, not calling the session code with port 0).

e2etest/bgptests/unnummbered.go
oribon: fair enough, we can keep it this way

karampok:
> can you update the `DiscardFRROnly` func to block interface? also, if we decide this is only for frr-k8s I guess we should block it in frr mode only as well (currently they share the validation)

The plan is to proceed next to support that in the frr backend AFAIK, not sure worth spliting the validation.

e2etest/bgptests/unnummbered.go
karampok: Given I configure using the BGPPeer CR with Interface, adding the route learning is not possible (I think)
I clean up all related code.

e2etest/bgptests/unnummbered.go
karampok: Was supposed to be part of the route learning but removed.

e2etest/bgptests/unnummbered.go
karampok: Given that prefixSend includes mask `/32, /128`, doing that check the result does not look cleaner.* IMO we can let that because is only even do not have check as the real check is following. I take another look once I have the per stack test

karampok: Added the check we can keep that if you like it more (or revert)

speaker/bgp_controller.go
```diff
@@ -60,6 +59,15 @@ type peer struct {
 	session bgp.Session
 }
 
+func (p peer) id() string {
+	ret := ""
```
oribon, line 63: can this be `ret := p.cfg.Iface`, if the addr is not nil then `ret = p.cfg.Addr.String()` and `return ret`?

speaker/bgp_controller.go
```diff
@@ -218,22 +226,28 @@ func (c *bgpController) syncPeers(l log.Logger) error {
...
 			var routerID net.IP
 			if p.cfg.RouterID != nil {
 				routerID = p.cfg.RouterID
 			}
 
+			peerAddr := ""
+			if p.cfg.Addr != nil {
+				peerAddr = p.cfg.Addr.String()
+			}
+			level.Info(l).Log("event", "peerAdded", "peer", peerAddr+p.cfg.Iface, "msg", "peer configured, starting BGP session")
```
oribon, line 246: move this back up (using the id func)

internal/bgp/native/native.go
```diff
@@ -98,9 +98,14 @@ func (sm *sessionManager) SyncExtraInfo(extras string) error {
 
 func (sm *sessionManager) SetEventCallback(func(interface{})) {}
 
+func (s *session) PeerLabelFor() string {
```
oribon, line 101: yeah the "for" can be removed (in response to your comment 😅)

internal/bgp/frrk8s/frrk8s.go
```diff
@@ -48,13 +47,23 @@ type session struct {
...
 
-// sessionName() defines the format of the key of the 'sessions' map in
-// the 'frrState' struct.
+// sessionName() defines the format of a map key.
 func sessionName(s session) string {
-	baseName := fmt.Sprintf("%d@%s-%d@%s", s.PeerASN, s.PeerAddress, s.MyASN, s.SourceAddress)
+	asn := strconv.FormatUint(uint64(s.PeerASN), 10)
+	if v := s.DynamicASN; v != "" {
+		asn = v
+	}
```
oribon, line 55: nit: `if s.DynamicASN != ""` and `asn = s.DynamicASN` looks cleaner

internal/bgp/frrk8s/frrk8s.go
```diff
@@ -48,13 +47,23 @@ type session struct {
...
+
+	peer := s.PeerAddress
+	if v := s.PeerInterface; v != "" {
```
oribon, line 58: same

karampok: done

internal/bgp/frrk8s/frrk8s.go
karampok: (not sure) but done

internal/bgp/native/native.go
karampok: done

speaker/bgp_controller.go
karampok: done

internal/config/validation.go
```diff
@@ -54,6 +54,9 @@ func DiscardFRROnly(c ClusterResources) error {
 		if p.Spec.DynamicASN != "" {
 			return fmt.Errorf("peer %s has dynamicASN set on native bgp mode", p.Spec.Address)
 		}
+		if p.Spec.Interface != "" {
```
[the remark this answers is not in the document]
karampok: done

speaker/bgp_controller.go
karampok: done

internal/bgp/frrk8s/frrk8s.go
```diff
@@ -48,13 +47,23 @@ type session struct {
...
-// the 'frrState' struct.
+// sessionName() defines the format of a map key.
 func sessionName(s session) string {
-	baseName := fmt.Sprintf("%d@%s-%d@%s", s.PeerASN, s.PeerAddress, s.MyASN, s.SourceAddress)
+	asn := strconv.FormatUint(uint64(s.PeerASN), 10)
+	if s.DynamicASN != "" {
+		asn = s.DynamicASN
+	}
+
+	peer := s.PeerAddress
```
fedepaol, line 57: nit: peerName

internal/bgp/native/native.go
```diff
@@ -98,9 +98,14 @@ func (sm *sessionManager) SyncExtraInfo(extras string) error {
 
 func (sm *sessionManager) SetEventCallback(func(interface{})) {}
 
+func (s *session) PeerLabel() string {
```
fedepaol, line 101: no need for this to be a method. Let's have a `peer` (or let's find a better name) private field we set in `NewSession` above

speaker/bgp_controller.go
```diff
@@ -60,6 +59,15 @@ type peer struct {
 	session bgp.Session
 }
 
+func (p peer) id() string {
```
fedepaol, line 62: same as above, can we set this when we create an instance of `peer` ?

speaker/bgp_controller.go
```diff
@@ -218,22 +226,28 @@ func (c *bgpController) syncPeers(l log.Logger) error {
...
 			// Session doesn't exist, but should be running. Create
 			// it.
-			level.Info(l).Log("event", "peerAdded", "peer", p.cfg.Addr, "msg", "peer configured, starting BGP session")
+			level.Info(l).Log("event", "peerAdded", "peer", p.id(), "msg", "peer configured, starting BGP session")
 			var routerID net.IP
 			if p.cfg.RouterID != nil {
 				routerID = p.cfg.RouterID
 			}
 
+			peerAddr := "" // we need because otherwise the value will "<nil>"
```
fedepaol, line 243: you can use `ptr.Deref` here

e2etest/bgptests/dump.go
```diff
@@ -12,19 +12,20 @@ import (
 	"github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"go.universe.tf/e2etest/pkg/frr"
+	frrcontainer "go.universe.tf/e2etest/pkg/frr/container"
 	"go.universe.tf/e2etest/pkg/metallb"
 	clientset "k8s.io/client-go/kubernetes"
 )
 
-func dumpBGPInfo(basePath, testName string, cs clientset.Interface, namespace string) {
+func dumpBGPInfo(basePath, testName string, cs clientset.Interface, namespace string, containers []*frrcontainer.FRR) {
```
fedepaol, line 20: this change to the testing framework should go in a separate commit.
Also, let's have this accept an "extraContainers" variadic argument

e2etest/bgptests/unnummbered.go
```diff
@@ -0,0 +1,200 @@
...
+			ASNLocal:  metalLBASN,
+			ASNRemote: metalLBASN,
+			Hostname:  "tor1",
+			Interface: iface,
+			RouterID:  "1.1.1.1",
+		}
+
+		var err error
+		ginkgo.By(fmt.Sprintf("creating p2p %s:%s -- %s:external-p2p-container", nodeWithP2PConnection.Name, iface, iface))
+		externalP2PContainer, err = frrcontainer.CreateP2PPeerFor(nodeWithP2PConnection.Name, iface, FRRImage)
```
fedepaol, line 84: externalP2P should be local and not local. Also, removing the container in defer is fragile, compared to having it done in an `AfterSomething`

e2etest/pkg/container/routes.go
```diff
@@ -173,7 +173,14 @@ func BGPRoutes(exc executor.Executor, dev string) (map[netip.Prefix]map[netip.Ad
 		for _, r := range routes {
 			dst, err := netip.ParsePrefix(r.Dst)
 			if err != nil {
-				return nil, fmt.Errorf("invalid prefix %s: %w", r.Dst, err)
+				// `ip route` command does not print the prefix when /32 or /128
```
fedepaol, line 176: this is definetely too nested :P
Let's move to a parseIPRouteIP function where we can escape early.
Also, the fail and retry approach with the two length is not great, let's check if an IP is v6 or v4 and behave accordingly (or error)

internal/config/config.go
```diff
@@ -389,9 +392,19 @@ func peerFromCR(p metallbv1beta2.BGPPeer, passwordSecrets map[string]corev1.Secr
 	if p.Spec.ASN == p.Spec.MyASN && p.Spec.EBGPMultiHop {
 		return nil, errors.New("invalid ebgp-multihop parameter set for an ibgp peer")
 	}
-	ip := net.ParseIP(p.Spec.Address)
-	if ip == nil {
-		return nil, fmt.Errorf("invalid BGPPeer address %q", p.Spec.Address)
+	var ip net.IP
+	if p.Spec.Address != "" && p.Spec.Interface == "" {
```
fedepaol, line 396: let's move this below the following two checks, which are more validation.
There you can remove the `p.Spec.Interface == ""` too as you already enforced that above.

karampok: done

e2etest/bgptests/unnummbered.go
```diff
@@ -0,0 +1,200 @@
...
+	})
+
+	ginkgo.AfterEach(func() {
+		if ginkgo.CurrentSpecReport().Failed() {
+			k8s.DumpInfo(Reporter, ginkgo.CurrentSpecReport().LeafNodeText)
+			dumpBGPInfo(ReportPath, ginkgo.CurrentSpecReport().LeafNodeText, cs, testNamespace, []*frrcontainer.FRR{externalP2PContainer})
+		}
+	})
+
+	ginkgo.DescribeTable("Session is established and route is advertised", func(prefixSend []string, iface string, tweak func(svc *corev1.Service)) {
```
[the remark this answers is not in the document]
karampok: done
prefixSend, iface is how we named in frr-k8s unnumbered test.
Not sure to be honest what are proper names

e2etest/bgptests/dump.go
karampok: I see the the `namespace string` is not being used, should I let it /remove it?

e2etest/bgptests/unnummbered.go
karampok: Done 

Can you please expand why the defer is fragile?

speaker/bgp_controller.go
karampok: TIL thanks,
I suppose you speak about
https://pkg.go.dev/k8s.io/utils/ptr#Deref it does not work when we have net.IP 
```
(speaker/bgp_controller.go:245:32: in call to ptr.Deref, type "net".IP of p.cfg.Addr does not match *T (cannot infer T) )
```

speaker/bgp_controller.go
karampok: done

internal/bgp/frrk8s/frrk8s.go
karampok: done

e2etest/bgptests/dump.go
karampok: maybe in another commit

karampok:
@fedepaol please take a look, thank you

e2etest/bgptests/unnummbered.go
fedepaol: because if you ctrl-c the test execution, you'll lose the defer while the after is always executed (unless you ctrl-c / ctrl-c)

speaker/bgp_controller.go
fedepaol: right, I thoguht this was a *string

fedepaol:
LGTM, thanks!

PR #2624, opened 2024-12-04
Fix(tests): resolve flakiness in service pool reassignment test

mquhuy:
The test verifies that a service (svc) assigned to a dual-stack pool correctly switches to a new dual-stack pool when its original pool is replaced by a single-stack pool. Previously, the test used `Eventually` to assert that the svc maintained 2 IPs, which caused flakiness since the svc could still hold old IPs temporarily.

To fix this, the test now also verifies that the svc's IPs are different from the original ones, ensuring a proper pool switch before final IP count validation. This prevents premature test completion and ensures accurate reassignment detection.

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design
/kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note

```

### DIFF
--- a/e2etest/l2tests/assignment.go
+++ b/e2etest/l2tests/assignment.go
@@ -977,6 +977,7 @@ var _ = ginkgo.Describe("IP Assignment", func() {
 				Expect(err).NotTo(HaveOccurred())
 				return len(svc1.Status.LoadBalancer.Ingress)
 			}, 5*time.Minute, 1*time.Second).Should(Equal(2))
+			originallyAssignedIps := getServiceIps(cs, svc1.Namespace, svc1.Name)
 			ginkgo.By("Validate LoadBalancer IPs are allocated from pool1")
 			err = config.ValidateIPInRange([]metallbv1beta1.IPAddressPool{pool1}, jigservice.GetIngressPoint(
 				&svc1.Status.LoadBalancer.Ingress[0]))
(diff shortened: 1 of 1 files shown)


oribon:
lgtm, but I think the commit message references something that is not there
```
...
We fix this by adding the new pool before replacing
the original pool with a single-stack one
```
should we remove that part?

mquhuy:
> lgtm, but I think the commit message references something that is not there
> 
> ```
> ...
> We fix this by adding the new pool before replacing
> the original pool with a single-stack one
> ```
> 
> should we remove that part?

Sorry for that. I meant to add that change as well but for some reason it was not pushed. :sweat: Fixed now.

oribon:
> I meant to add that change as well 

not sure I understand why it is needed though, did you see it not working without it? to me it seems the eventually you added should be enough since the config should ultimately converge

mquhuy:
> > I meant to add that change as well
> 
> not sure I understand why it is needed though, did you see it not working without it? to me it seems the eventually you added should be enough since the config should ultimately converge

I have tried running the test a few times, and the flake mostly comes from the fact that the new dualstack pool is added at the same time when the current dualstack pool becomes single stack.

It could happen that when the config converges, the new dualstack pool is not available yet, so metallb only sees that the current pool has become single-stack. By default, it reassigns the svc to the same pool. After the new dualstack pool is available, the designed behavior is that the svc won't be reassigned.

The change now is that we add the new dualstack pool first, so when the current pool gets shrunk to single-stack and triggers reassignment, the new pool is already there, hence the svc is guaranteed to be assigned to it.

oribon (approved):
I see, thanks for clarifying. it might still be a bit racey this way, but seems good enough for now. thanks a lot for fixing this!

PR #2627, opened 2024-12-05
E2E: add function to fetch BGP routes using iproute2

karampok:
The function returns a map of destination prefixes to their next-hop addresses, supporting both IPv4 and IPv6 routes with single or multiple next-hops, and uses netip.{Prefix,Addr} to allow direct comparison.

 /kind feature

We cannot use the normal vtysh approach because FRR is restarting

Same function exists in https://github.com/metallb/metallb/pull/2595 as needed by the unnumbered BGP testing

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etest/pkg/container/routes.go
+++ b/e2etest/pkg/container/routes.go
@@ -3,7 +3,9 @@
 package container
 
 import (
+	"encoding/json"
 	"fmt"
+	"net/netip"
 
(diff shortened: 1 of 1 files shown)


karampok:
@fedepaol this one as part of https://github.com/metallb/frr-k8s/pull/231 thanks

PR #2674, opened 2025-02-03
Exclude node given pattern in a configmap

karampok:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

Exclude means that pod is running but shouldAnnounce blocks the advertisements when the node
matces the patter in the config map e.g. annotation or labels or conditions

This will replace the behavior of stoping the advertisements when node becomes unschedulable.

**Release note**:

```release-note

```


speaker/main.go
```diff
@@ -531,6 +544,19 @@ func compareIPs(ips1, ips2 []net.IP) bool {
 	return true
 }
 
+func parseExcludeNodePattern() (*v1.Node, error) {
```
fedepaol, line 547: is there a reason why the pattern is represented by a full fledged node? I think this deserves its own struct (possibly with a "matches" method)

speaker/bgp_controller.go
```diff
@@ -174,16 +174,22 @@ func (c *bgpController) ShouldAnnounce(l log.Logger, name string, _ []net.IP, po
 		return "notOwner"
 	}
 
-	if err := k8snodes.IsNodeAvailable(nodes[c.myNode]); err != nil {
```
fedepaol, line 177: nit: I think it'd be cleaner (and more cherry-pick / backport friendly) to have a revert commit of the previous logic

e2etest/pkg/frr/provider/provider.go
fedepaol, line 1: can you expand the commit message here?

e2etest/bgptests/cordon_node.go
```diff
@@ -19,7 +19,7 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 )
 
-var _ = ginkgo.Describe("BGP Cordon Node", func() {
+var _ = ginkgo.Describe("BGP CordonNode", func() {
```
fedepaol, line 22: does this test have value anymore?

e2etest/bgptests/exclude_node.go
```diff
@@ -0,0 +1,128 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+var _ = ginkgo.Describe("FRR BGP ExcludeNode IPV4", func() {
+	var (
+		ann           = [2]string{"ocp-mc", "rebooting"}
+		allNodes      *corev1.NodeList
+		excludeCM     = "excludenodepattern" // configmap name is hardcoded to manifests
```
fedepaol, line 28: let's make a const then.

e2etest/pkg/k8s/nodes.go
```diff
@@ -82,6 +82,15 @@ func SelectorsForNodes(nodes []v1.Node) []metav1.LabelSelector {
 	return selectors
 }
 
+func AddAnnotationToNode(nodeName, key, value string, cs clientset.Interface) {
```
fedepaol, line 85: nit: annotateNode

e2etest/bgptests/exclude_node.go
```diff
@@ -0,0 +1,128 @@
...
+		cs            clientset.Interface
+		svc           *corev1.Service
+		testNamespace string
+	)
+
+	restartSpeakerPods := func() {
+		err := metallb.RestartSpeakerPods(cs)
+		Expect(err).NotTo(HaveOccurred())
+		Eventually(func() error {
```
fedepaol, line 37: how about embedding this waiting logic in `RestartSpeakerPods` ?

e2etest/bgptests/exclude_node.go
```diff
@@ -0,0 +1,128 @@
...
+		testNamespace, err = k8s.CreateTestNamespace(cs, "bgp")
+		Expect(err).NotTo(HaveOccurred())
+
+		allNodes, err = cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+		Expect(err).NotTo(HaveOccurred())
+
+		_, svc = setupBGPService(cs, testNamespace, ipfamily.IPv4, []string{v4PoolAddresses}, FRRContainers, func(svc *corev1.Service) {
+			testservice.TrafficPolicyCluster(svc)
+		})
+		testservice.ValidateDesiredLB(svc)
```
fedepaol, line 91: what's the purpouse of validating this in BeforeEach?

internal/k8s/controllers/node_controller.go
fedepaol, line 1: not sure I understood the last part of the commit message. Mind clarifying?

charts/metallb/templates/speaker.yaml
fedepaol, line 1: nit on commit: extend the speaker manifest

charts/metallb/templates/speaker.yaml
```diff
@@ -191,6 +191,10 @@ spec:
             defaultMode: 256
             name: metallb-excludel2
       {{- end }}
+        - name: excludenodes
+          configMap:
+            name: excludenodepattern
```
fedepaol, line 196: I'd call this nodesexclusionpatterns

speaker/main.go
karampok: because I would like not create extra structs/methods and leverage the k8s library. Plus I wanted to keep full fledged node  in case we need to extend in the future. E.g disable when noschedulable (which I do think it makes sense under some condition), or node special condition or taint or whatever is in the v1.Node stuct.

Nevertheless, if we agree that we strictly care only about annotation and labels, then I could drop that implementation and have custom struct with parsing and matches method. Would you like me to do that?

fedepaol: Yes I think it's confusing, as a user would expect an exact match. Let's keep just a map / two maps.

e2etest/bgptests/exclude_node.go
karampok: I would prefer to keep it like that, because I think is cleaner to keep ginkgo{Expect,Eventually} asserts on the test file . If I embed I should use go (not-ginkgo) functions.

Plus that will break the other test using that (Graceful Restart) that expects to return so to check other stuff in between

charts/metallb/templates/speaker.yaml
karampok: Done

charts/metallb/templates/speaker.yaml
karampok: done

e2etest/bgptests/exclude_node.go
karampok: removed

e2etest/bgptests/cordon_node.go
karampok: removed

karampok:
@fedepaol  this PR has been forgotten, needs some code review

PR #2681, opened 2025-02-06
build: Fetch kubectl only if it is not present

dlipovetsky:
/kind bug

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

The --short flag is not supported by recent kubectl clients, so the version is not reported, and kubectl is fetched even if it is already on the filesystem.

Although kubectl exits with a non-zero status, this is not caught by the get_command_version function.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -1530,7 +1530,7 @@ def fetch_kubectl():
     curl_command = "curl -o {} -LO https://dl.k8s.io/release/{}/bin/$(go env GOOS)/$(go env GOARCH)/kubectl".format(
         kubectl_path, kubectl_version
     )
-    get_version_command = f"{kubectl_path} version --short"
+    get_version_command = f"{kubectl_path} version"
     fetch_dependency(
         kubectl_path,
         kubectl_version,


tasks.py
```diff
@@ -1530,7 +1530,7 @@ def fetch_kubectl():
     curl_command = "curl -o {} -LO https://dl.k8s.io/release/{}/bin/$(go env GOOS)/$(go env GOARCH)/kubectl".format(
         kubectl_path, kubectl_version
     )
-    get_version_command = f"{kubectl_path} version --short"
+    get_version_command = f"{kubectl_path} version"
```
karampok, line 1533:
```
 kubectl version
Client Version: v1.31.0
Kustomize Version: v5.4.2
Server Version: v1.31.0
```
we probably need to isolate only the Client Version

dlipovetsky: I think what you're asking is already done in https://github.com/metallb/metallb/blob/17aafdea571b097012fafe758d921a388fff9f46/tasks.py#L1538-L1539

fedepaol:
LGTM, sending to merge

Issue #2403, opened 2024-05-17
Services having the same "allow-shared-ip" value are not updated when one of them is updated


AlexeyKasatkin:
### MetalLB Version

0.14.5

### Deployment method

Charts

### Main CNI

Calico v3.27.0

### Kubernetes Version

v1.27.10

### Describe the bug

There are two services that share the same IP address and have the same `externalTrafficPolicy` value:

```
---
apiVersion: v1
kind: Service
metadata:
  annotations:
    metallb.universe.tf/allow-shared-ip: foo
  name: test-udp-svc
  namespace: test
spec:
  allocateLoadBalancerNodePorts: true
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  loadBalancerIP: 1.2.3.4
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: pdns-udp
    port: 53
    protocol: UDP
    targetPort: 53
  selector:
    application: someapp
    component: somecomponent
  sessionAffinity: None
  type: LoadBalancer
---
apiVersion: v1
kind: Service
metadata:
  annotations:
    metallb.universe.tf/allow-shared-ip: foo
  name: test-tcp-svc
  namespace: test
spec:
  allocateLoadBalancerNodePorts: true
  externalTrafficPolicy: Cluster
  internalTrafficPolicy: Cluster
  loadBalancerIP: 1.2.3.4
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: pdns-tcp
    port: 53
    protocol: TCP
    targetPort: 53
  selector:
    application: someapp
    component: somecomponent
  sessionAffinity: None
  type: LoadBalancer
```

The services have the external IP address assigned and are accessible.
Change `externalTrafficPolicy` value for both services from Cluster to Local.
In result, the 1st service that has been changed, remains having no external IP assigned.
Though, the service that was changed later, have the external IP assigned as expected.

```
test-tcp-svc                 LoadBalancer   10.233.62.135   1.2.3.4        53:32123/TCP                                                        13m
test-udp-svc                 LoadBalancer   10.233.54.13    <pending>      53:31244/UDP                                                        14m
```

From `test-udp-svc` events:
```
  Normal   IPAllocated       75s                metallb-controller  Assigned IP ["1.2.3.4"]
[...]

PR #2692, opened 2025-02-14
Reprocess services if shareability of a given service changes

fedepaol:

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

If the sharing criteria (ETP, endpoint selector) of a given service change, it might mean that the IP can be shared with other services.
Because of this,  we reprocess all the services.

**Release note**:

```release-note
Reprocess the services and reassign IPs if the sharing criteria of a given service changes (ie, the external traffic policy changes and suddenly another service in pending becomes compatible).
```

Fixes https://github.com/metallb/metallb/issues/2403

### DIFF
--- a/e2etest/l2tests/assignment.go
+++ b/e2etest/l2tests/assignment.go
@@ -19,6 +19,7 @@ import (
 	metallbv1beta1 "go.universe.tf/metallb/api/v1beta1"
 
 	jigservice "go.universe.tf/e2etest/pkg/jigservice"
+	corev1 "k8s.io/api/core/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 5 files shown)


e2etest/l2tests/assignment.go
```diff
@@ -221,6 +225,99 @@ var _ = ginkgo.Describe("IP Assignment", func() {
...
+
+			checkHasLBIP(svc, ip)
+			checkHasLBIP(svc1, ip)
+
+			ginkgo.By("changing etp policy of the second service to local")
+			jig1.UpdateService(context.Background(), func(svc *corev1.Service) {
+				svc.Spec.ExternalTrafficPolicy = corev1.ServiceExternalTrafficPolicyLocal
+			})
+
+			ginkgo.By("checking the second service loses its ip")
```
oribon, line 295: does the first service also lose its ip (should it?)?

fedepaol: nope, the first keeps it. By changing the second, we are making it's ip not compabile anymore with the status of the system as a whole. I think the current behavior is correct, in preserving what is not changing and removing the IP from the "offending" one

oribon: fair enough, should we test this behavior as well?

fedepaol: you mean ensuring the first service stays the same, I guess. It makes sense, I'll do that

Issue #2679, opened 2025-02-06
Invalid caBundle should be removed from conversion webhook kustomize patch for BGPPeers CRD


dlipovetsky:
### MetalLB Version

0.14.9

### Deployment method

Charts

### Main CNI

n/a

### Kubernetes Version

1.31.0

### Cluster Distribution

n/a

### Describe the bug

Kubernetes v1.31.0 and newer validates the caBundle field.
> CustomResourceDefinition objects created with non-empty caBundle fields which are invalid or do not contain any certificates will not appear in discovery or serve endpoints until a valid caBundle is provided. Updates to CustomResourceDefinition are no longer allowed to transition a valid caBundle field to an invalid caBundle field, because this breaks serving of the existing CustomResourceDefinition. (https://github.com/kubernetes/kubernetes/pull/124061, [@Jefftree](https://github.com/Jefftree)) [SIG API Machinery]
> -- https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.31.md#api-change-2

We should remove the invalid caBundle field here:
https://github.com/metallb/metallb/commit/776cefecc29788fabca23cb487185ac5c4333c70#L13

### To Reproduce

N/A

### Expected Behavior

N/A

### Additional Context

To be honest, I'm not even sure the value has any effect. When remove it and run `inv generatemanifests`, I see no changes. But it seems wise to remove it all the same.

I have a branch ready to create the PR: https://github.com/dlipovetsky/metallb/commits/remove-invalid-cabundle/

### I've read and agree with the following

- [x] I've checked all open and closed issues and my request is not there.
- [x] I've checked all open and closed pull requests and my request is not there.

### I've read and agree with the following

- [x] I've checked all open and closed issues and my issue is not there.
- [x] This bug is reproducible when deploying MetalLB from the main branch
- [x] I have read the [troubleshooting guide](https://metallb.universe.tf/troubleshooting/) and I am still not able to make it work
- [x] I checked the logs and MetalLB is not discarding the configuration as not valid
- [x] I enabled the debug logs, collected the information required from the cluster using the [collect script](https://raw.githubusercontent.com/metallb/metallb/main/troubleshooting/collect.sh) and will attach them to the issue
- [x] I will provide the definition of my service and the related endpoint slices and attach them to this issue

fedepaol:
Odd. I think we had a reason for bothering adding the field...
Need to dig / remember

dlipovetsky:
[...]

PR #2695, opened 2025-02-14
fix: Remove invalid caBundle in conversion webhook configuration

dlipovetsky:
/kind fix

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

Removes a kustomize patch that applies an invalid caBundle; that's not allowed in v1.31.x (https://github.com/kubernetes/kubernetes/pull/124061)

Note that the patch has no effect; removing it has no effect on the chart. Please see https://github.com/metallb/metallb/issues/2679#issuecomment-2659922510. So this could realy be categorized as cleanup.

**Release note**:

```release-note
NONE
```


fedepaol:
So, I had something I wasn't able to understand, which is, we are already running 1.31 in CI (an it works!).
I dug a bit deeper, and I believe the file you are fixing is not used at all. 

The conversion is embedded in the crds, [here](https://github.com/metallb/metallb/blob/main/config/crd/patches/crd-conversion-patch-bgppeers.yaml#L11) (also, note the comment about the need for a valid caBundle).
It also resonates with your comment of the all-in-one manifests not being regenerated.

Now, my question is: why did you file the issue in first instance? Is there some configuration / deployment that I am missing that leverages that infamous leftover?

If there's no real issue, the right thing to do is to zap the file. If there's a scenario where this is affected, I am probably missing it.

Thanks!

dlipovetsky:
> I dug a bit deeper, and I believe the file you are fixing is not used at all.

Yes, I suspected that, but I wasn't certain, so I was more conservative when I said this in the description above:

> Note that the patch has no effect; removing it has no effect on the chart. 

> If there's no real issue, the right thing to do is to zap the file.

Sounds good to me. Should I add a commit here, or open a separate PR? If I use this PR, then I'll also update the PR title.

dlipovetsky:
I replaced the commit, and renamed the PR.

PR #2716, opened 2025-03-12
fix: webhook to handle interface field of BGPPeer resource

karampok:
The commit extends the webhook validation to ensure BGPPeer resources with the interface field set are being validated as previously the .address was. When  Unnumbered BGP, the address field can be empty.

>
 /kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
NONE
```

### DIFF
--- a/internal/config/validation.go
+++ b/internal/config/validation.go
@@ -3,9 +3,8 @@
 package config
 
 import (
-	"fmt"
-
 	"errors"
+	"fmt"
 
 	metallbv1beta1 "go.universe.tf/metallb/api/v1beta1"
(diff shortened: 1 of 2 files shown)


fedepaol:
If we did not release unnumbered yet, let's remove the release note.

internal/config/validation.go
fedepaol, line 1: Nit on commit message: When using Unnumbered BGP, the address field can be empty and validation must be performed against the interface.

karampok: done

PR #2717, opened 2025-03-12
Bump golangci linter

karampok:
/kind cleanup

**Release note**:

```release-note
NONE
```


karampok:
What do you think about enabling also the https://github.com/mvdan/gofumpt as linter?

PR #2734, opened 2025-03-26
Add Deutsche Telekom's Das SCHIFF / T-CaaS as an adopter

chdxD1:
Add Deutsche Telekom AG as a adopter of MetalLB.

We are a heavy user of the project in our horizontal telco cloud stack, using the metallb-controller in all instances and metallb-speaker in some of the deployments.


fedepaol:
Thanks :heart:

PR #2741, opened 2025-04-06
Add webhookSecretName support to the metallb helm chart

splaspood:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

Recently webhookSecretName was added as webhook-secret to the controller cli arguments.   This PR modifies the helm chart to allow overriding this value via helm.

**Release note**:

```release-note
Added controller.webhookSecretMode to helm chart values to allow setting --webhook-secret for controller runtime.
```

### DIFF
--- a/charts/metallb/templates/controller.yaml
+++ b/charts/metallb/templates/controller.yaml
@@ -188,7 +188,7 @@ spec:
       - name: cert
         secret:
           defaultMode: 420
-          secretName: metallb-webhook-cert
+          secretName: {{ .Values.controller.webhookSecretName | quote }}
       {{- if .Values.prometheus.controllerMetricsTLSSecret }}
       - name: metrics-certs
         secret:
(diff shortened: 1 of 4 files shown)


splaspood:
I was working on this, but after resolving the webhook/cert related conflicts, I encountered others.   I figure this means what I'm doing is not the 'right thing(tm)'.   

Ultimately what I'm trying to do is have two distinct metallb instances running in my cluster, one for 'external' stuff (on an external node) and the other for internal.

fedepaol:
> I was working on this, but after resolving the webhook/cert related conflicts, I encountered others. I figure this means what I'm doing is not the 'right thing(tm)'.
> 
> Ultimately what I'm trying to do is have two distinct metallb instances running in my cluster, one for 'external' stuff (on an external node) and the other for internal.

please note that metallb is not meant to run multi-instances... but with all the provided selectors (node, advertisement, pools) you should be able to express two different configs for two different set of workloads / nodes.

PR #2745, opened 2025-04-10
Add Kiratech Training as an adopter

rascasoft:
This extends the list of adopters with the Kiratech Training project, where MetalLB is used as the reference technology in every lab that requires a Load Balancer. MetalLB is essential to several of our exercises, and every student learns how it works and the benefits of using this technology in (but not limited to) bare metal environments.

/kind documentation

**Release note**:

```release-note
NONE
```


fedepaol:
Thanks!

PR #2746, opened 2025-04-10
frr-k8s controller: implement cleanup on pod shutdown

oribon:
Currently when the speaker shuts down it can leave a stale FRRConfiguration. Here we register a cleanup Runnable with the manager, so that we delete the node's FRRConfiguration when the manager signals a shutdown (when the context is cancelled).

 /kind bug

**Release note**:

```release-note
Updated the speaker so that it removes the FRRConfiguration it is managing when its pod is shutdown.
```


oribon:
this is more of a poc
TBD: this will cause a disruption to the connection (e.g gracefulrestart scenarios) if the pod is destined to restart and not deleted permanently, so it might be favorable to keep the current behavior. if we decide to drop this I guess it's worth to document the behavior (user needs to manually delete configuration if they remove the speaker from a node)

Issue #2704, opened 2025-02-27
Reconsider the behavior of the DisableMP flag


fedepaol:
### Is your feature request related to a problem?

When moving to FRR, we inadvertently enabled the possibility to advertise the v6 addresses over a v4 session (and viceversa). This was totally unintentional, we tried to amend and then we found out that users (at least one!) were already relying on it (see https://github.com/metallb/metallb/issues/1402) so we reverted to the original behavior in https://github.com/metallb/metallb/pull/1444.

Then https://github.com/metallb/metallb/issues/1736 and https://github.com/metallb/metallb/issues/2106 were raised, so we made the behavior optional (but the original, with v6 over v4 enabled).

This is causing confusion and works in a limited set of scenarios, because FRR doesn't have a guarantee of what the v6 nexthop should be for a session established over v6.
 
In retrospective,  this should've been the opt-in behavior and not the disable one.
Additionally, disableMP is misleading as the multiprotocol bgp extension is used for much more than this.

### Describe the solution you'd like

- Deprecation of the disableMP flag
- A breaking change in the behavior, where the default is v4 over v4 and v6 over v6
- A new flag (advertise v4 over v6 and viceversa), clearly stating the non deterministic behavior (so single interface nodes can work, and they are the exception)
- The changes reflected in frr-k8s too

### I've read and agree with the following

- [x] I've checked all open and closed issues and my request is not there.
- [x] I've checked all open and closed pull requests and my request is not there.

fedepaol:
cc @oribon

oribon:
> A new flag (advertise v4 over v6 and viceversa), clearly stating the non deterministic behavior (so single interface nodes can work, and they are the exception)

would there be any functionality change, or this is simply about renaming it to something more proper with better documentation?

fedepaol:
> > A new flag (advertise v4 over v6 and viceversa), clearly stating the non deterministic behavior (so single interface nodes can work, and they are the exception)
> 
> would there be any functionality change, or this is simply about renaming it to something more proper with better documentation?

The default behavior would change (and would be by default what disablemp = true provides).
It's a breaking change, yes, but for good.

fedepaol:
Closed as completed

PR #2766, opened 2025-05-13
WIP FRR: deprecate disableMP

fedepaol:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

As described in metallb/metallb#2704, the genesis of the disableMP flag is a bit unfortunate.

"When moving to FRR, we inadvertently enabled the possibility to advertise the v6 addresses over a v4 session (and viceversa). This was totally unintentional, we tried to amend and then we found out that users (at least one!) were already relying on it.

This is causing confusion and works in a limited set of scenarios, because FRR doesn't have a guarantee of what the v6 nexthop should be for a session established over v6."

Because of this, we rollback the behavior, enabling an ipfamily to a neighbor connected with the corresponding ip family.

We also add a flag "dualstackfamily" to enable the behavior for those scenarios where a user might want to enable a dual stack family for a neighbor.

**Release note**:

```release-note
Deprecate the disableMP flag, and make the v4 over v4 and v6 over v6 behavior the default. Additionally, add a dualStackFamily flag per neighbor, where we enable the v4 family over a v6 link and viceversa.
```


internal/bgp/frr/frr.go
```diff
@@ -385,22 +372,41 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
...
-			n.LocalPrefsV6 = sets.List(properties.LocalPrefsV6)
-		}
-
 		config.Routers = append(config.Routers, toAdd)
 	}
 	return config, nil
 }
 
+func frrIPFamily(ipFamily ipfamily.Family) string {
+	if ipFamily == "ipv6" {
```
oribon, line 381: nit: you can use the Family const here

internal/bgp/frr/frr.go
```diff
@@ -287,96 +273,97 @@ func (sm *sessionManager) createConfig() (*frrConfig, error) {
...
 	}
 
 	for _, r := range sortMap(routers) {
+		for _, n := range r.neighbors {
+			n.PrefixesV4 = sets.List(n.prefixesV4Set)
+			n.PrefixesV6 = sets.List(n.prefixesV6Set)
+			for k, m := range n.PrefixesModifiers {
+				m.Prefixes = sets.List(n.PrefixesModifiers[k].prefixesSet)
+				n.PrefixesModifiers[k] = m
+			}
```
oribon, line 365: is there a reason not to have this above where we handle the neighbor? (also, feels like the prefixesV*Set can be dropped as a field if we do that)

karampok (commented):
> > because FRR doesn't have a guarantee of what the v6 nexthop should be for a session established over v6.

I am not sure I understand that, can you expand thanks

karampok:
that is interesting, when you peer over v6, is the FRR properly picking the ipv4 address of the interface?

fedepaol:
> that is interesting, when you peer over v6, is the FRR properly picking the ipv4 address of the interface?

I would argue that there's no "proper interface address" here as the session is just a mean to advertise prefixes.

internal/config/validation.go
```diff
@@ -50,6 +50,9 @@ func DiscardFRROnly(c ClusterResources) error {
 		if p.Spec.DisableMP {
 			return fmt.Errorf("peer %s has disable MP flag set on native bgp mode", p.Spec.Address)
 		}
+		if p.Spec.DualStackAddressFamily {
```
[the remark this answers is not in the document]
fedepaol: this sounds reasonable though. In native mode we don't even have v6 support, so this does not make sense at all

internal/bgp/frr/frr.go
fedepaol: yep, it's because the current implementation runs under the assumption that we could have multiple sessions associated to the same neighbor. So it's better to sort once the negihbors and advertisements are squashed

oribon (commented):
/lgtm

PR #2773, opened 2025-05-26
Add networkpolicies 

karampok:
/kind feature

Adding networkpolicies support, and that is an opt-in feature.
E2E test are executed with network policies enabled.

**Release note**:

```release-note
 Add network policy support to Helm charts and manifests

    - Network policies are disabled by default. User can enable them in Helm
      by setting the value `networkpolicies.enabled` true. When deploying
      using manifests the user should modify the config/native/kustomization.yaml
      to include the network policies.

    - User can install a default deny network policy rule if needed. In Helm
      by setting the chart value `networkpolicies.defaultDeny` to true. In
      manifest by modifying accordingly the config/networkpolicies/kustomization.yaml
      to include or exclude the deny_all policy.
```

### DIFF
--- a/e2etest/netpoltests/networkpolicies.go
+++ b/e2etest/netpoltests/networkpolicies.go
@@ -1,0 +1,76 @@
+// SPDX-License-Identifier:Apache-2.0
+package netpoltests
+
+import (
+	"net"
+
+	"github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"go.universe.tf/e2etest/pkg/executor"
+	"go.universe.tf/e2etest/pkg/k8s"
+	"go.universe.tf/e2etest/pkg/k8sclient"
+	"go.universe.tf/e2etest/pkg/metallb"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+)
+
+const agnhostImage = "registry.k8s.io/e2e-test-images/agnhost:2.47"
+
+var _ = ginkgo.Describe("Networkpolicies", func() {
+	var (
+		cs         clientset.Interface
+		probe      *corev1.Pod
+		controller *corev1.Pod
+	)
+
+	ginkgo.BeforeEach(func() {
+		cs = k8sclient.New()
+
+		probeNamespace, err := k8s.CreateTestNamespace(cs, "test-netpol")
+		Expect(err).NotTo(HaveOccurred())
+		ginkgo.DeferCleanup(func() {
+			err = k8s.DeleteNamespace(cs, probeNamespace)
+			Expect(err).NotTo(HaveOccurred())
+		})
+		tpod := &corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "netpol-test",
+				Namespace: probeNamespace,
+			},
+			Spec: corev1.PodSpec{
+				Containers: []corev1.Container{{
+					Name:  "netpol",
+					Image: agnhostImage,
+				}},
+			},
+		}
+		probe, err = k8s.CreatePod(cs, tpod)
+		Expect(err).NotTo(HaveOccurred())
+
+		controller, err = metallb.ControllerPod(cs)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	ginkgo.It("only allowed traffic", func() {
+		ginkgo.By("checking ingress - probe pod to any port on the controller other than webhook should be timeout")
+		controllerIP := controller.Status.PodIP
+		probeExec := executor.ForPod(probe.Namespace, probe.Name, "netpol")
+		out, err := probeExec.Exec("./agnhost", "connect", net.JoinHostPort(controllerIP, "7472"), "--timeout", "5s")
+		Expect(err).To(HaveOccurred())
+		Expect(out).To(ContainSubstring("TIMEOUT"), out)
+		out, err = probeExec.Exec("./agnhost", "connect", net.JoinHostPort(controllerIP, "9443"), "--timeout", "5s")
+		Expect(err).NotTo(HaveOccurred(), out) // Until we manage to allow traffic only from k8s API
+
+		ginkgo.By("checking egress - controller pod to any port on the probe other port than API should be timeout")
+		probeIP := probe.Status.PodIP
+		ctrlExec, err := executor.ForPodDebug(cs, controller.Namespace, controller.Name, "controller", agnhostImage)
+		Expect(err).NotTo(HaveOccurred())
+		out, err = ctrlExec.Exec("./agnhost", "connect", net.JoinHostPort(probeIP, "8080"), "--timeout", "5s")
+		Expect(err).To(HaveOccurred())
+		Expect(out).To(ContainSubstring("TIMEOUT")) // If no netpol then REFUSED
+		out, err = ctrlExec.Exec("./agnhost", "connect", net.JoinHostPort(probeIP, "6443"), "--timeout", "5s")
+		Expect(err).To(HaveOccurred())
+		Expect(out).To(ContainSubstring("REFUSED")) // until we restrict only to k8s svc
+	})
+})
--- a/e2etest/pkg/metallb/metallb.go
+++ b/e2etest/pkg/metallb/metallb.go
@@ -4,12 +4,11 @@ package metallb
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"slices"
 	"time"
-
-	"errors"
 
 	. "github.com/onsi/gomega"
 	"go.universe.tf/e2etest/pkg/k8s"
--- a/tasks.py
+++ b/tasks.py
@@ -580,7 +580,6 @@ apiServer:
         if with_prometheus:
             prometheus_values = (
                 "--set prometheus.serviceMonitor.enabled=true "
-                "--set prometheus.secureMetricsPort=9120 "
                 "--set speaker.frr.secureMetricsPort=9121 "
                 "--set prometheus.serviceAccount=prometheus-k8s "
                 "--set prometheus.namespace=monitoring "
@@ -612,7 +611,7 @@ apiServer:
         run(
             "helm install metallb charts/metallb/ --set controller.image.tag=dev-{} "
             "--set speaker.image.tag=dev-{} --set speaker.logLevel=debug "
-            "--set networkpolicies.enabled=true --set networkpolicies.defaultDeny=true "
+            "--set networkpolicies.enabled=true --set networkpolicies.defaultDeny=true --set prometheus.secureMetricsPort=9120 "
             "--set controller.logLevel=debug {} {} --namespace metallb-system".format(
                 architecture, architecture, prometheus_values, frr_values
             ),
--- a/e2etest/e2etest_suite_test.go
+++ b/e2etest/e2etest_suite_test.go
@@ -30,6 +30,7 @@ import (
 
 	"go.universe.tf/e2etest/bgptests"
 	"go.universe.tf/e2etest/l2tests"
+	_ "go.universe.tf/e2etest/netpoltests"
 	testsconfig "go.universe.tf/e2etest/pkg/config"
 	"go.universe.tf/e2etest/pkg/executor"
 	frrprovider "go.universe.tf/e2etest/pkg/frr/provider"
--- a/e2etest/go.work.sum
+++ b/e2etest/go.work.sum
@@ -594,6 +594,7 @@ github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a/go.mod h1:l
 github.com/benbjohnson/clock v1.1.0/go.mod h1:J11/hYXuz8f4ySSvYwY0FKfm+ezbsZBKZxNJlLklBHA=
 github.com/bgentry/speakeasy v0.1.0/go.mod h1:+zsyZBPWlz7T6j88CTgSN5bM796AkVf0kBD4zp0CCIs=
 github.com/bketelsen/crypt v0.0.3-0.20200106085610-5cbc8cc4026c/go.mod h1:MKsuJmJgSg28kpZDP6UIiPt0e0Oz0kqKNGyRaWEPv84=
+github.com/blang/semver/v4 v4.0.0/go.mod h1:IbckMUScFkM3pff0VJDNKRiT6TG/YpiHIM2yvyW5YoQ=
 github.com/cenkalti/backoff/v4 v4.1.1/go.mod h1:scbssz8iZGpm3xbr14ovlUdkxfGXNInqkPWOWmG2CLw=
 github.com/cenkalti/backoff/v4 v4.2.1/go.mod h1:Y3VNntkOUPxTVeUxJ/G5vcM//AlwfmyYozVcomhLiZE=
 github.com/census-instrumentation/opencensus-proto v0.2.1/go.mod h1:f6KPmirojxKA12rnyqOA5BBL4O983OfeGPqjHWSTneU=
--- a/e2etest/pkg/k8s/pods.go
+++ b/e2etest/pkg/k8s/pods.go
@@ -6,9 +6,11 @@ import (
 	"bytes"
 	"context"
 	"io"
+	"time"
 
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
 )
 
@@ -40,7 +42,8 @@ func PodLogs(cs clientset.Interface, pod *corev1.Pod, podLogOpts corev1.PodLogOp
 
 // PodIsReady returns the given pod's PodReady and ContainersReady condition.
 func PodIsReady(p *corev1.Pod) bool {
-	return podConditionStatus(p, corev1.PodReady) == corev1.ConditionTrue && podConditionStatus(p, corev1.ContainersReady) == corev1.ConditionTrue
+	return podConditionStatus(p, corev1.PodReady) == corev1.ConditionTrue &&
+		podConditionStatus(p, corev1.ContainersReady) == corev1.ConditionTrue
 }
 
 // podConditionStatus returns the status of the condition for a given pod.
@@ -56,4 +59,29 @@ func podConditionStatus(p *corev1.Pod, condition corev1.PodConditionType) corev1
 	}
 
 	return corev1.ConditionUnknown
+}
+
+// CreatePod creates the given Pod and waits until it is running and ready.
+// This ensures that status fields such as the Pod IP are populated and can be used by the caller.
+func CreatePod(cs clientset.Interface, p *corev1.Pod) (*corev1.Pod, error) {
+	pod, err := cs.CoreV1().Pods(p.GetObjectMeta().GetNamespace()).Create(context.TODO(), p, metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return waitForPodReady(cs, pod.Namespace, pod.Name)
+}
+
+func waitForPodReady(cs clientset.Interface, ns, name string) (*corev1.Pod, error) {
+	var ret *corev1.Pod
+	err := wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 20*time.Second, false, func(context.Context) (bool, error) {
+		var err error
+		ret, err = cs.CoreV1().Pods(ns).Get(context.TODO(), name, metav1.GetOptions{})
+		if err != nil {
+			return false, nil
+		}
+
+		isReady := (ret.Status.Phase == corev1.PodRunning) && (PodIsReady(ret))
+		return isReady, nil
+	})
+	return ret, err
 }


charts/metallb/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,47 @@
...
+metadata:
+  name: metallb-controller
+  namespace: {{ .Release.Namespace | quote }}
+spec:
+  podSelector:
+    matchLabels:
+      {{- include "metallb.selectorLabels" . | nindent 6 }}
+      app.kubernetes.io/component: controller
+  egress:
+    - ports:
```
oribon, line 27: can you add to the commit message why this exact ports (and same for ingress)?

config/networkpolicies/controller.yaml
```diff
@@ -0,0 +1,24 @@
...
+metadata:
+ name: controller
+ namespace: metallb-system
+spec:
+ podSelector:
+   matchLabels:
+    app: metallb
+    component: controller
+ egress:
+  - ports:
```
oribon: why we don't need 443 here?

controller/main.go
oribon, line 1: I don't think this is necessary (nor visible/informative enough), I think we should just drop it. (if our networkpolicies caused an error then it is a bug, and if custom policies were used we can't shield users from a misconfig)

charts/metallb/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,49 @@
...
+        - protocol: TCP
+          port: 6443
+        - protocol: TCP
+          port: 443
+  ingress:
+    - ports:
+        {{- if .Values.controller.webhookMode }}
+        - protocol: TCP
```
fedepaol: I think it's gonna be more future-proof if we use named ports. IIUC the spec allows it https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.33/#networkpolicyport-v1-networking-k8s-io

If we don't have these ports under a name, we can add them.

charts/metallb/README.md
```diff
@@ -68,6 +68,7 @@ Kubernetes: `>= 1.19.0-0`
 | imagePullSecrets | list | `[]` |  |
 | loadBalancerClass | string | `""` |  |
 | nameOverride | string | `""` |  |
+| networkpolicies.defaultDeny | bool | `false` |  |
```
fedepaol, line 72: what's the use case for this vs disabling / enabling netpols at all?

controller/main.go
fedepaol: agree, it's not worth the extra complexity

charts/metallb/templates/networkpolicy.yaml
```diff
@@ -0,0 +1,47 @@
...
+  namespace: {{ .Release.Namespace | quote }}
+spec:
+  podSelector:
+    matchLabels:
+      {{- include "metallb.selectorLabels" . | nindent 6 }}
+      app.kubernetes.io/component: controller
+  egress:
+    - ports:
+        - protocol: TCP
+          port: 6443
```
fedepaol: these should be available to override, as we don't have control over them. Also, the speaker reaches out the apiserver too, but I guess we are not adding it as it's host networked. Can you add an explicit commetn about it?

fedepaol:
Also, please enrich the release notes a bit saying what we are going to block. Also also, the network policies are enabled if you deploy with manifests / kustomize

controller/main.go
karampok: okay, commit dropped

charts/metallb/README.md
karampok: For example if you install metallb in a different namespace, where other things might exist you  want your policies but not a default-deny that will break the other things. In OCP, the plan is eventually external component to install default rules and not added by the components. If you like we can exclude it from the helm API (but in that case the metallb-operator will  add it somehow)

config/networkpolicies/controller.yaml
karampok: bz 
```
 oc get svc kubernetes -o yaml
apiVersion: v1
kind: Service
metadata:
  creationTimestamp: "2025-06-10T08:42:56Z"
  labels:
    component: apiserver
    provider: kubernetes
  name: kubernetes
  namespace: default
  resourceVersion: "195"
  uid: 8809a655-eee9-4e1d-863a-3147dbe44424
spec:
  clusterIP: 10.96.0.1
  clusterIPs:
  - 10.96.0.1
  internalTrafficPolicy: Cluster
  ipFamilies:
  - IPv4
  ipFamilyPolicy: SingleStack
  ports:
  - name: https
    port: 443
    protocol: TCP
    targetPort: 6443
  sessionAffinity: None
  type: ClusterIP
```
how the controller speaks is through the svc and the networkpolicies seems to operate on the target port (at least in kind cluster and in OCP). I cleaned up the port egress port 443 from the helm chart .

Having said that I will search deeper if there is standardized documenation how the  CNI network should work on the targetPort or not.

charts/metallb/templates/networkpolicy.yaml
karampok: done
(nice tip)

charts/metallb/templates/networkpolicy.yaml
karampok: Updated the commit message

karampok:
@fedepaol @oribon  I think I have addressed all your comments, ptal thank you

charts/metallb/README.md
fedepaol: makes sense, thanks!

fedepaol:
Looks good! Would it make sense to add a test ensuring that the network policies are effectively working?
I am thinking about curling to / from a different pod.

karampok:
> Looks good! Would it make sense to add a test ensuring that the network policies are effectively working? I am thinking about curling to / from a different pod.

That is a bit tricky.
There is a default-deny, if our network-policy are not correct then all test are failling because operator never healthy.  If the kind cluster does not support netpol (which it does), we could have not-related metallb-test to verify it (but I think it not very meaningful).  
1. We could have have one other pod (`kubectl run pod `) in the namespace to try to reach the metrics port to verify that ingress traffic is blocked (but that mean we should have add a check `if netpol enable and default deny exist`, and we should logic to grub the pod IP and such)
2. I cannot thing of good way to test the egress, it should have like `kubectl debug pod` to have a shell in the controller network namespace 

Is that something you have  in mind?

(update maybe we could have that test in the operator,)

IMO it's not worth the extra complexity but we can do it

fedepaol:
> > Looks good! Would it make sense to add a test ensuring that the network policies are effectively working? I am thinking about curling to / from a different pod.
> 
> That is a bit tricky. There is a default-deny, if our network-policy are not correct then all test are failling because operator never healthy. If the kind cluster does not support netpol (which it does), we could have not-related metallb-test to verify it (but I think it not very meaningful).

Not sure I understood this part.

> 
>     1. We could have have one other pod (`kubectl run pod `) in the namespace to try to reach the metrics port to verify that ingress traffic is blocked (but that mean we should have add a check `if netpol enable and default deny exist`, and we should logic to grub the pod IP and such)
> 
>     2. I cannot thing of good way to test the egress, it should have like `kubectl debug pod` to have a shell in the controller network namespace
> 
> 
> Is that something you have in mind?
> 
> (update maybe we could have that test in the operator,)
> 
> IMO it's not worth the extra complexity but we can do it

I am not comfortable in saying "it works" if we don't have a way to verify it.
What I was suggesting was to spin up a pod that we try to reach from the controller and that tries to curl the controller, to ensure that the netpol we are configuring are really effective.

karampok:
>  we try to reach from the controller

how do you suggest to do that given we have distroless container without a shell?

fedepaol:
> > we try to reach from the controller
> 
> how do you suggest to do that given we have distroless container without a shell?

https://metallb.io/troubleshooting/#how-to-debug-the-speaker-and-the-controller-containers

e2etest/pkg/executor/executor.go
oribon, line 1: not sure I understand the rationale of this change (is it to be able to run the same debug twice?)?
if that's the reason, it seems a bit overcomplex (vs what we had before), is it possible (if reasonable) to just randomize the name of the ephemeral container instead when Execing?

e2etest/pkg/metallb/metallb.go
```diff
@@ -150,3 +150,39 @@ func FRRK8SPods(cs clientset.Interface, namespace string) ([]*corev1.Pod, error)
 	}
 	return res, nil
 }
+func CreatePod(cs clientset.Interface, p *corev1.Pod) (*corev1.Pod, error) {
+	pod, err := cs.CoreV1().Pods("default").Create(context.TODO(), p, metav1.CreateOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return WaitForPodReady(cs, pod.Namespace, pod.Name)
+}
```
oribon: this seems a bit opinionated (the "default" part at least, I guess it should be p.Namespace), I think it makes sense to omit this function and the next, creating/deleting directly where needed

e2etest/netpoltests/networkpolicies.go
```diff
@@ -0,0 +1,154 @@
...
+					break
+				}
+			}
+			Expect(found).To(BeTrue(), "Policy %s/%s not found", wantPolicy.Namespace, wantPolicy.Name)
+		}
+
+		tpod := &corev1.Pod{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "netpol-test",
+				Namespace: "default", // probe pod must not be in the metallb-system namespace
```
oribon: can we create a new ns for this test (and remove it in the defer)?

tasks.py
```diff
@@ -580,7 +580,6 @@ def dev_env(
         if with_prometheus:
             prometheus_values = (
                 "--set prometheus.serviceMonitor.enabled=true "
-                "--set prometheus.secureMetricsPort=9120 "
```
oribon, line 583: why this change?

karampok: because when we execute the frr-k8s external this is the single tet that is NOT  using this value so only in that single test I cannot have the same template to compare the netpolicies I got versus what I expect.  I see no harm to always have that in our test. 

```
  ingress:
    - ports:
        - protocol: TCP
          {{- if .Values.prometheus.secureMetricsPort }}
          port: metricshttps
          {{ else }}
          port: {{ .Values.prometheus.metricsPort }}
          {{- end }}
        {{- if .Values.controller.webhookMode }}
        - protocol: TCP
          port: webhook-server
        {{- end }}

```
alternative I could also have a new (kind of duplicate) helm value called networkpolicies.metricsPort

e2etest/netpoltests/networkpolicies.go
karampok: Sure I can move into a random new namespace, I see not argument for/against though

e2etest/pkg/executor/executor.go
karampok: Debug container is a container attached to the pod, and we can do from
there normal `kubectl exec` commands. In that way we avoid running any
`-it` argument neither adding more than one debug container per target
container.

Having the go client instead of kubectl is more flexible and is required
to catch errors as well as to wait for the debug pod to be ready

(from the commit message).
  I think that is more recommended implementation and not so complex imo.

Having an -interactive shell when doing not interactive call wrong (if I remove it it fails)
Return values of the command is more straightforward.
Speed also is  reason ( exec is faster) if we want to extend the number of debug command
Debug container cannot be removed so kind the dynamic name will leak as many as calls as many containers
Having a go-client implementation is better (we still use kubectl but we can start not using it)

e2etest/netpoltests/networkpolicies.go
karampok: I updated to use test-namespace

e2etest/pkg/metallb/metallb.go
karampok: yes, I fixed it, it was mistake. I could skip the function but I should skip also the delete function and then the main code will look more packed ( I think I also got feedback to hide those under function in the past)

e2etest/pkg/executor/executor.go
oribon: what I mean is why specifically we need this change, what are we doing now that we couldn't do before? is it the `if we want to extend the number of debug command` part?

e2etest/netpoltests/networkpolicies.go
oribon: > I see not argument for/against though

we should avoid polluting namespaces (even if not very probable)

tasks.py
oribon: but what is actually changing? it seems like you're just moving that part down (or I'm missing something)

e2etest/pkg/metallb/metallb.go
oribon: fair enough (though I guess we can now omit the Delete part since deleting the namespace takes care of that)

tasks.py
karampok: https://github.com/metallb/metallb/blob/795410790c7ab8f5ee8e1b33999df1b916469462/.github/workflows/ci.yaml#L223-L223 moving that part down it will be run also for that test,

oribon: so why not put it there instead? 😅

e2etest/pkg/executor/executor.go
karampok: So if I revert and have in test code
```
		ctrlExec := executor.ForPodDebug(controller.Namespace, controller.Name, "controller", agnhostImage)
		out, err = ctrlExec.Exec("./agnhost", "connect", net.JoinHostPort(probeIP, "8080"), "--timeout", "5s")
		Expect(err).NotTo(HaveOccurred()) # this error is nil, which in the refactor case is not nil as expected
		Expect(out).To(ContainSubstring("TIMEOUT")) // If no netpol then REFUSED
```
then the test is failing
```

SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS
------------------------------
Networkpolicies only allowed traffic
/home/kka/workspace/github.com/metallb/wts/up-np-metallb/e2etest/netpoltests/networkpolicies.go:89
  STEP: checking ingress - probe pod to any port on the controller other than webhook should be timeout @ 06/19/25 12:56:10.11
  STEP: checking egress - controller pod to any port on the probe other port than API should be timeout @ 06/19/25 12:56:15.324
  [FAILED] in [It] - /home/kka/workspace/github.com/metallb/wts/up-np-metallb/e2etest/netpoltests/networkpolicies.go:104 @ 06/19/25 12:56:15.88
• [FAILED] [19.888 seconds]
Networkpolicies [It] only allowed traffic
/home/kka/workspace/github.com/metallb/wts/up-np-metallb/e2etest/netpoltests/networkpolicies.go:89

  [FAILED] Expected
      <string>: Targeting container "controller". If you don't see processes from this container it may be because the container runtime doesn't support this feature.
      Unable to use a TTY - input is not a terminal or the right kind of file
      If you don't see a command prompt, try pressing enter.
      warning: couldn't attach to pod/controller-765bbcb6fc-dzcv4, falling back to streaming logs: error stream protocol error: unknown error

  to contain substring
      <string>: TIMEOUT
  In [It] at: /home/kka/workspace/github.com/metallb/wts/up-np-metallb/e2etest/netpoltests/networkpolicies.go:104 @ 06/19/25 12:56:15.88
------------------------------
SSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSSS
--
````

Comment (karampok) [reply e2etest/pkg/executor/executor.go:1]: which is hiding the real stdout from the process, and me trying to fix it I refactor, did not dig deeper (as refactor seemed the right approach)

Comment (karampok) [reply e2etest/pkg/executor/executor.go:1]:
```
/home/kka/workspace/github.com/metallb/wts/up-np-metallb/build/kubectl debug -n metallb-system -c controller-debugger-x --target=controller --image=registry.k8s.io/e2e-test-images/agnhost:2.47 controller-765bbcb6fc-dzcv4 -- ./agnhost connect 10.244.1.7:8080 --timeout 5s
Targeting container "controller". If you don't see processes from this container it may be because the container runtime doesn't support this feature.
```
just running the command I do not see the stdout

Comment (karampok) [reply e2etest/pkg/executor/executor.go:1]: with -it (which we should not have in theory) I can see but there is some time waiting (kind the go code does not wait but it returns)
```
/home/kka/workspace/github.com/metallb/wts/up-np-metallb/build/kubectl debug -it -n metallb-system -c controller-debugger-xxx --target=controller --image=registry.k8s.io/e2e-test-images/agnhost:2.47 controller-765bbcb6fc-dzcv4 -- ./agnhost connect 10.244.1.7:8080 --timeout 5s
Targeting container "controller". If you don't see processes from this container it may be because the container runtime doesn't support this feature.
If you don't see a command prompt, try pressing enter.
 --- here is hanging  for a bit ---

TIMEOUT
Session ended, the ephemeral container will not be restarted but may be reattached using 'kubectl attach controller-765bbcb6fc-dzcv4 -c controller-debugger-xxx -i -t' if it is still running
```

File: e2etest/netpoltests/networkpolicies.go (hunk 7 above)
Comment (karampok) [reply e2etest/netpoltests/networkpolicies.go]: done

File: e2etest/pkg/executor/executor.go
Comment (karampok) [reply e2etest/pkg/executor/executor.go:1]: Re-word the commit description with our chat findings

File: e2etest/pkg/executor/executor.go [hunk 9]
```diff
@@ -86,27 +92,69 @@ type podDebugExecutor struct {
 	name      string
 	container string
 	image     string
+	ephemeral string
+	cs        clientset.Interface
 }
 
-func ForPodDebug(namespace, name, container, image string) Executor {
-	return &podDebugExecutor{
+func ForPodDebug(cs clientset.Interface, namespace, podName, targetContainer string) (Executor, error) {
```
oribon: can we pass the image as input like before instead of hardcoding agnhost?

e2etest/pkg/executor/executor.go
```diff
@@ -86,27 +92,69 @@ type podDebugExecutor struct {
...
-		name:      name,
-		container: container,
-		image:     image,
+		name:      podName,
+		container: targetContainer,
+		ephemeral: "debugger-" + targetContainer,
+		cs:        cs,
+	}
+
+	ephemeralName := "debugger-" + targetContainer
```
oribon: nit: drop this and use ret.ephemeral (or drop references to ret.ephemeral, put this in ret and always use it)

e2etest/pkg/executor/executor.go
```diff
@@ -86,27 +92,69 @@ type podDebugExecutor struct {
 	name      string
 	container string
 	image     string
+	ephemeral string
+	cs        clientset.Interface
```
oribon, line 94: this is unused

e2etest/pkg/executor/executor.go
```diff
@@ -86,27 +92,69 @@ type podDebugExecutor struct {
...
+
+	if !containerExists {
+		pod.Spec.EphemeralContainers = append(pod.Spec.EphemeralContainers, ephemeralContainer)
+		_, err = cs.CoreV1().Pods(namespace).UpdateEphemeralContainers(ctx, podName, pod, metav1.UpdateOptions{})
+		if err != nil {
+			return nil, fmt.Errorf("failed to create ephemeral container: %w", err)
+		}
 	}
+
+	return ret, wait.PollUntilContextTimeout(context.Background(), 1*time.Second, 20*time.Second, false, func(context.Context) (bool, error) {
```
oribon: nit: can the wait be before the return, and the usual if `err != nil` after?

e2etest/netpoltests/networkpolicies.go
```diff
@@ -0,0 +1,161 @@
...
+		out, err = ctrlExec.Exec("./agnhost", "connect", net.JoinHostPort(probeIP, "8080"), "--timeout", "5s")
+		Expect(err).To(HaveOccurred())
+		Expect(out).To(ContainSubstring("TIMEOUT")) // If no netpol then REFUSED
+		out, err = ctrlExec.Exec("./agnhost", "connect", net.JoinHostPort(probeIP, "6443"), "--timeout", "5s")
+		Expect(err).To(HaveOccurred())
+		Expect(out).To(ContainSubstring("REFUSED")) // until we restrict only to k8s svc
+	})
+})
+
+func expectedPolicies() ([]networkingv1.NetworkPolicy, error) {
```
oribon: I think this is unnecessary (checking that the objs are there), e.g for the same reasons we don't check the webhook objects are there before running the webhook e2es. I guess we can keep it if you see there's value in this

e2etest/netpoltests/networkpolicies.go
```diff
@@ -0,0 +1,161 @@
...
+				Containers: []corev1.Container{{
+					Name:  "netpol",
+					Image: agnhostImage,
+				}},
+			},
+		}
+		probe, err = metallb.CreatePod(cs, tpod)
+		Expect(err).NotTo(HaveOccurred())
+		ginkgo.DeferCleanup(func() {
+			err = metallb.DeletePod(cs, probe)
```
oribon: nit: this is redundant (taken care by deleting the namespace

config/networkpolicies/controller.yaml
```diff
@@ -0,0 +1,24 @@
...
+  podSelector:
+    matchLabels:
+     app: metallb
+     component: controller
+  egress:
+    - ports:
+      - protocol: TCP
+        port: 6443
+  ingress:
+    - ports:
```
oribon, line 17: can we also have here monitoring?

karampok: why? I added as a commented, in all case this is not enabled by default, t user must actively change the manifest (and we should not recommend use the non https port)

e2etest/netpoltests/networkpolicies.go
karampok: done

e2etest/netpoltests/networkpolicies.go
karampok: I have mixed feeling, I will wait @fedepaol feedback

e2etest/pkg/executor/executor.go
karampok: don

e2etest/pkg/executor/executor.go
karampok: fixed the ephemeral, the cs is being used

e2etest/pkg/executor/executor.go
karampok: done

e2etest/pkg/executor/executor.go
karampok: done

e2etest/l2tests/interface_selector.go
```diff
@@ -25,7 +25,7 @@ import (
 	"go.universe.tf/e2etest/pkg/status"
 )
 
-const agnostImage = "registry.k8s.io/e2e-test-images/agnhost:2.45"
+const agnhostImage = "registry.k8s.io/e2e-test-images/agnhost:2.47"
```
fedepaol: nit: do we need this change? In any case, not belongs to this commit

e2etest/pkg/executor/executor.go
```diff
@@ -86,27 +90,73 @@ type podDebugExecutor struct {
...
+	containerExists := false
+	for _, ec := range pod.Spec.EphemeralContainers {
+		if ec.Name == ret.ephemeral {
+			containerExists = true
+			break
+		}
 	}
+
+	if !containerExists {
+		pod.Spec.EphemeralContainers = append(pod.Spec.EphemeralContainers, ephemeralContainer)
```
fedepaol: I'd move this to a "addEphemeralContainerToPod" function

e2etest/pkg/executor/executor.go
fedepaol: only thing I still don't understand from the discussion is why the current test using it is failing. If the command was never returning the right value, I am not able to justify the mac assertion here https://github.com/metallb/metallb/blob/c759397ca9215faa49c549e384f9fb3a7c7c448e/e2etest/l2tests/interface_selector.go#L160

You probably chatted about this, but the conclusion in the commit message seem either not accurate or the current test is volkswagening

```
therefore it
returns neither the correct output (= only the output from the
command we want to run), neither the error (if the command fails the
kubectl debug actually returns ok).
```

e2etest/pkg/metallb/metallb.go
```diff
@@ -150,3 +149,26 @@ func FRRK8SPods(cs clientset.Interface, namespace string) ([]*corev1.Pod, error)
 	}
 	return res, nil
 }
+
+func CreatePod(cs clientset.Interface, p *corev1.Pod) (*corev1.Pod, error) {
```
fedepaol: Let's move these under pkg/k8s

e2etest/netpoltests/networkpolicies.go
fedepaol: Agree on this being redundant, we want to ensure the behavior, not the configuration that leads to the behavior.

e2etest/pkg/executor/executor.go
karampok: Done (if I understood your ask correctly)

e2etest/pkg/metallb/metallb.go
karampok: Done

e2etest/pkg/executor/executor.go
karampok: the observation is that for the command that returns instantly (whoami/ifconfig), it works just fine. It
does not work fine for the commands that will timeout/exit. I re-wrote the commit so it is clear that the above sentence is only valid for the specific type of commans (which is agnhost connect  --timeout).

e2etest/netpoltests/networkpolicies.go
karampok: removed

e2etest/l2tests/interface_selector.go
karampok: removed it from the commit

karampok (commented):
@fedepaol  I updated according to your comments, let me know I missed something

fedepaol:
LGTM

PR #2786, opened 2025-06-11
Add dual-stack IP family support to webhook service

### LINKED ISSUE #2785: [helm chart] metallb-webhook-service is missing 'spec.ipFamilyPolicy' and 'spec.ipFamilies'
### MetalLB Version

0.14.9

### Deployment method

Charts

### Main CNI

calico

### Kubernetes Version

_No response_

### Cluster Distribution

_No response_

### Describe the bug

**Helm chart version** - v0.14.9
**K8S Cluster** - Dual stack cluster (IPv6 preference) with calico CNI
**Service yaml** - https://github.com/metallb/metallb/blob/main/charts/metallb/templates/webhooks.yaml#L129

metallb-webhook-service is missing the 'spec.ipFamilyPolicy' and 'spec.ipFamilies' (https://github.com/metallb/metallb/blob/main/charts/metallb/templates/webhooks.yaml#L129)

As a result when metallb is deployed in dual stack cluster, We are seeing only IPv6 assigned to the metallb services also ipFamilyPolicy is coming as SingleStack. As per kubernetes doc (https://kubernetes.io/docs/concepts/services-networking/dual-stack/#dual-stack-options-on-new-services)

> Service which does not explicitly define .spec.ipFamilyPolicy,  When its created, Kubernetes assigns a cluster IP for the Service from the first configured service-cluster-ip-range and sets the .spec.ipFamilyPolicy to SingleStack.

kubectl describe command on the metallb-webhook-service
```
service/metallb-webhook-service
  IP Family Policy: SingleStack
  IP Families: IPv6
  IP: fd10:100:96::8d71
  IPs: fd10:100:96::8d71
  Port: <unset>  443/TCP
  TargetPort: 9443/TCP
  Endpoints: [fd10:100:64:3::d]:9443
```

### To Reproduce

1. Install the metallb v0.14.9 via Helm chart on dual stack cluster
2. Describe the service "metallb-webhook-service"

### Expected Behavior

"metallb-webhook-service" should provide user option to pass the ipFamilyPolicy and ipFamilies (list) from values.yaml

### Additional Context

https://kubernetes.io/docs/concepts/services-networking/dual-stack/#dual-stack-options-on-new-services

karampok:
Configure webhook service with PreferDualStack IP family policy.

According to
https://kubernetes.io/docs/concepts/services-networking/dual-stack/#dual-stack-options-on-new-services

```
PreferDualStack: Allocates both IPv4 and IPv6 cluster IPs for the
Service when dual-stack is enabled. If dual-stack is not enabled or
supported, it falls back to single-stack behavior.`
```
/kind bug

Fixes #2785

**Release note**:

```release-note
Configure webhook service with PreferDualStack IP family policy.

```

### DIFF
--- a/charts/metallb/templates/webhooks.yaml
+++ b/charts/metallb/templates/webhooks.yaml
@@ -134,6 +134,7 @@ metadata:
   labels:
     {{- include "metallb.labels" . | nindent 4 }}
 spec:
+  ipFamilyPolicy: PreferDualStack
   ports:
   - port: 443
     targetPort: 9443
--- a/config/manifests/metallb-frr-k8s-prometheus.yaml
+++ b/config/manifests/metallb-frr-k8s-prometheus.yaml
@@ -2895,6 +2895,7 @@ metadata:
   name: metallb-webhook-service
   namespace: metallb-system
 spec:
+  ipFamilyPolicy: PreferDualStack
   ports:
   - port: 443
     targetPort: 9443
--- a/config/manifests/metallb-frr-k8s.yaml
+++ b/config/manifests/metallb-frr-k8s.yaml
@@ -2820,6 +2820,7 @@ metadata:
   name: metallb-webhook-service
   namespace: metallb-system
 spec:
+  ipFamilyPolicy: PreferDualStack
   ports:
   - port: 443
     targetPort: 9443
--- a/config/manifests/metallb-frr-prometheus.yaml
+++ b/config/manifests/metallb-frr-prometheus.yaml
@@ -1975,6 +1975,7 @@ metadata:
   name: metallb-webhook-service
   namespace: metallb-system
 spec:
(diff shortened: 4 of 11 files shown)


karampok:
@fedepaol or @oribon  can you please take a look, thank you

fedepaol:
The change makes sense / won't hurt, but can you explain what's the effect of having the webhook service being single ipfamily (especially in the context described in the issue, which is primary = v6?)
Having the issue being raised as bug and @tylerauerbeck 's comment makes me think that something would not work in this scenario. but I am not sure I understood what it is

karampok:
No I cannot  explain the effect tbh, in theory the k8s API server is the only component that  speaks to that svc, and that can take place in either stacks. It does sound legit though for a dual cluster the service to be dualstack. Let's wait to author of the issue to come back.

PR #2795, opened 2025-06-30
Handle better the webhook mode

karampok:
/kind cleanup

There is an obvious bug for networkpolicies that should exist when not disabled.
Refactor the networkpolicies test to be able to run when deployed by operator

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etest/netpoltests/networkpolicies.go
+++ b/e2etest/netpoltests/networkpolicies.go
@@ -2,7 +2,9 @@
 package netpoltests
 
 import (
+	"fmt"
 	"net"
+	"strconv"
 
 	"github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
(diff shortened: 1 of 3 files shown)


e2etest/netpoltests/networkpolicies.go
```diff
@@ -53,24 +54,58 @@ var _ = ginkgo.Describe("Networkpolicies", func() {
...
+		"8888": "TIMEOUT", // Add random port to check for being blocked
+	}
+	for _, c := range pod.Spec.Containers {
+		for _, p := range c.Ports {
+			switch p.Name {
+			case "webhook-server", "metricshttps":
+				ret[fmt.Sprintf("%d", p.ContainerPort)] = "" // Agnhost connect return empty on success
+			default: // "monitoring" port is not whitelisted in the network policy
+				ret[fmt.Sprintf("%d", p.ContainerPort)] = "TIMEOUT"
+			}
```
oribon: why do we need this? I thought the netpols will allow all ports to be reachable now?

e2etest/netpoltests/networkpolicies.go
```diff
@@ -53,24 +54,58 @@ var _ = ginkgo.Describe("Networkpolicies", func() {
...
 })
+
+func ingressPorts(pod *corev1.Pod) map[string]string {
+	ret := map[string]string{
+		"8888": "TIMEOUT", // Add random port to check for being blocked
+	}
+	for _, c := range pod.Spec.Containers {
+		for _, p := range c.Ports {
+			switch p.Name {
+			case "webhook-server", "metricshttps":
```
oribon: if the other comment makes sense, we don't need the switch case

e2etest/netpoltests/networkpolicies.go
```diff
@@ -53,24 +54,58 @@ var _ = ginkgo.Describe("Networkpolicies", func() {
...
+				Expect(err).To(HaveOccurred())
+				Expect(out).To(ContainSubstring(action), fmt.Sprintf("connect to port %s returned %s not %s", port, out, action))
+				continue
+			}
+			Expect(err).NotTo(HaveOccurred())
+		}
 	})
 })
+
+func ingressPorts(pod *corev1.Pod) map[string]string {
```
oribon: I think it would be cleaner to just create the map inside the test, this function hides some non-trivial assumptions when looking from above (the purpose of map[string]string for example)

e2etest/netpoltests/networkpolicies.go
```diff
@@ -53,24 +54,58 @@ var _ = ginkgo.Describe("Networkpolicies", func() {
...
+			default: // "monitoring" port is not whitelisted in the network policy
+				ret[fmt.Sprintf("%d", p.ContainerPort)] = "TIMEOUT"
+			}
+		}
+	}
+
+	return ret
+}
+
+func egressPorts() map[string]string {
```
oribon: same, let's just declare the map as part of the test

e2etest/netpoltests/networkpolicies.go
karampok: I am not sure what you mean? Can you expand?
In the test we do not allow the non https metrics  to be accessible outside the pod network namespace, thre is a commented config/networkpolicies/kustomization entry if a user wants to do that.

e2etest/netpoltests/networkpolicies.go
karampok: done

e2etest/netpoltests/networkpolicies.go
karampok: done

e2etest/netpoltests/networkpolicies.go
oribon: ah sorry, so let's modify this `for` to allow all ports, with only one `if` to check if the current port is the monitoring one, and in that case have a TIMEOUT. (also would be useful to have a comment beside it `// non https metrics is not exposed`)

e2etest/netpoltests/networkpolicies.go
```diff
@@ -53,24 +54,56 @@ var _ = ginkgo.Describe("Networkpolicies", func() {
 	})
 
 	ginkgo.It("only allowed traffic", func() {
-		ginkgo.By("checking ingress - probe pod to any port on the controller other than webhook should be timeout")
+		ginkgo.By("checking ingress - probe pod to any port on the controller other than webhook,metricshttps should be timeout")
+
+		ingressPorts := func(pod *corev1.Pod) map[string]string {
```
oribon: and no need for this function, just create and modify the map

karampok: done

oribon:
lgtm

PR #2802, opened 2025-07-11
bgp: allow duplicated bgppeer per neighbor ip

### LINKED ISSUE #2367: Relax the single BGPPeer per neighbor IP constraint

The frr implementation of metallb rejects any duplicate BGPPeer (i.e., peers that have the same IP for the same VRF).
Ideally, this uniqueness should be enforced on per node basis. Because it was impossible to detect (and maintain) upfront (and because it was the early stage of the frr implementation), we decided to block any overlapping IPs regardless of the fact that they might influence the behaviour of a distinct set of nodes.

This prevents some useful features, such as using different ASN / SRC address per nodes / set of nodes, or having custom timers per nodes (because a set of nodes could be on a different region with different latencies).

What we propose here is to relax the constraint, trying to intercept misconfigurations in a best effort manner (with the webhooks) and blacklisting the neighbor ip (acting as if it was not configured at all) whenever there are multiple bgppeers eligible for the same node (on the same vrf), basically relying on the user to do the right configuration and resulting in a configuration error if this is not the case.

The user feedback should also improve whenever we implement https://github.com/metallb/metallb/issues/2156

This should also be extended to frr-k8s.

### Additional context

_No response_

clwluvw:

/kind feature

Ensure that duplicate BGP peers in FRR mode have nodeSelectors to distinguish them.
Also, add a runtime check in the BGP controller to detect duplicate peers by address or interface, and return a clear error if found.

Fixes: https://github.com/metallb/metallb/issues/2367

**Release note**:
```release-note
Allow duplicated BGPPeer per neighbor IP
```

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -257,9 +257,6 @@ func For(resources ClusterResources, validate Validate) (*Config, error) {
 	}
 
 	cfg.BGPExtras = bgpExtrasFor(resources)
-	if err != nil {
-		return nil, err
-	}
 
 	err = validateConfig(cfg)
 	if err != nil {
--- a/internal/config/validation.go
+++ b/internal/config/validation.go
@@ -146,19 +146,22 @@ func DontValidate(c ClusterResources) error {
 // any options that are available only in the native implementation.
 func DiscardNativeOnly(c ClusterResources) error {
 	if len(c.Peers) > 1 {
-		peerAddr := make(map[string]bool)
+		peerAddrHasNodeSelectors := make(map[string]bool) // if a peer has nodeSelectors, we can have multiple peers with the same address
 		routerID := c.Peers[0].Spec.RouterID
 		peer0 := peerIdentifier(c.Peers[0].Spec)
-		peerAddr[peer0] = true
+		peerAddrHasNodeSelectors[peer0] = len(c.Peers[0].Spec.NodeSelectors) > 0
 		for _, p := range c.Peers[1:] {
 			if p.Spec.RouterID != routerID {
 				return fmt.Errorf("peer %s has RouterID different from %s, in FRR mode all RouterID must be equal", p.Spec.RouterID, c.Peers[0].Spec.RouterID)
 			}
 			peerID := peerIdentifier(p.Spec)
-			if _, ok := peerAddr[peerID]; ok {
-				return fmt.Errorf("peer %s already exists, FRR mode doesn't support duplicate BGPPeers", p.Spec.Address)
+			if hadNodeSelectors, ok := peerAddrHasNodeSelectors[peerID]; ok {
+				// make sure there is at least one nodeSelector for this peer
+				if len(p.Spec.NodeSelectors) == 0 || !hadNodeSelectors {
+					return fmt.Errorf("duplicate peer %s has no nodeSelectors, in FRR mode each duplicate peer must have nodeSelectors to differentiate", p.Spec.Address)
+				}
 			}
-			peerAddr[peerID] = true
+			peerAddrHasNodeSelectors[peerID] = len(p.Spec.NodeSelectors) > 0
 		}
 	}
 	for _, p := range c.Peers {
--- a/internal/config/validation.go
+++ b/internal/config/validation.go
@@ -5,6 +5,7 @@ package config
 import (
 	"errors"
 	"fmt"
+	"reflect"
 
 	metallbv1beta1 "go.universe.tf/metallb/api/v1beta1"
 	metallbv1beta2 "go.universe.tf/metallb/api/v1beta2"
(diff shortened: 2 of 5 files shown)


clwluvw:
ping @fedepaol @oribon :)

clwluvw:
SYN @fedepaol @oribon :)

clwluvw:
ping @fedepaol @oribon :)

internal/config/config.go
```diff
@@ -252,9 +252,6 @@ func For(resources ClusterResources, validate Validate) (*Config, error) {
 	}
 
 	cfg.BGPExtras = bgpExtrasFor(resources)
-	if err != nil {
```
fedepaol, line 260: makes sense, but not related to this change, please add it in its own commit

internal/config/validation.go
```diff
@@ -146,19 +146,22 @@ func DontValidate(c ClusterResources) error {
...
-		peerAddr[peer0] = true
+		peerAddrHasNodeSelectors[peer0] = len(c.Peers[0].Spec.NodeSelectors) > 0
 		for _, p := range c.Peers[1:] {
 			if p.Spec.RouterID != routerID {
 				return fmt.Errorf("peer %s has RouterID different from %s, in FRR mode all RouterID must be equal", p.Spec.RouterID, c.Peers[0].Spec.RouterID)
 			}
 			peerID := peerIdentifier(p.Spec)
-			if _, ok := peerAddr[peerID]; ok {
-				return fmt.Errorf("peer %s already exists, FRR mode doesn't support duplicate BGPPeers", p.Spec.Address)
+			if hadNodeSelectors, ok := peerAddrHasNodeSelectors[peerID]; ok {
```
fedepaol: checking if the other peer has a selector is not enough, we should ensure that there are no overlapping nodes, and not sure we can do that at this stage

fedepaol: also, to make this right, we should allow compatible peers hitting the same node.
So if peerA and peerB are the same, it's fine to allow them if they select the same node, we shouldn't if they are incompatible

speaker/bgp_controller.go
```diff
@@ -208,8 +208,9 @@ func (c *bgpController) ShouldAnnounce(l log.Logger, name string, _ []net.IP, po
 // implying that the set of running BGP sessions may need tweaking.
 func (c *bgpController) syncPeers(l log.Logger) error {
 	var (
-		errs          int
-		needUpdateAds bool
+		errs           int
+		needUpdateAds  bool
+		duplicatePeers = make(map[string]bool) // address/interface -> true if already seen
```
fedepaol: there's something I don't like about the current design, and this is is going to overcomplicate it.
In SetConfig here https://github.com/metallb/metallb/blob/main/speaker/bgp_controller.go#L108 we check for obsolete peers and we remove them before calling syncPeers.
In syncPeers, we kind of do the same activity https://github.com/metallb/metallb/blob/main/speaker/bgp_controller.go#L229 because syncPeers might be called also from the node listener.

What I'd suggest to do here is, wrap the newPeers / oldPeers logic in a function that additinally:
- checks for the node eligibilty (if it matches or not the current node)
- deduplicate the list, so newPeers already contains each peer exactly once

Then, we call the function in both SetConfig and SetNode, making syncPeers dumber.
Hope this make sense

fedepaol (changes requested):

fedepaol:
sorry for the delay, left some comments.

Also, the hardest part here is the validation. We need to check for the validity of each node, but we can't prevent a valid configuration become invalid because somebody changes a label to a node (and there's not much we can do about that, because we certainly don't want to add a webhook over the node resource)

fedepaol:
also, ignore copilot's comments

internal/config/validation.go
clwluvw: > we should allow compatible peers hitting the same node.

@fedepaol I'm having a challenge on how we can define compatible peers. Could you please comment on `arePeersCompatible()`?

clwluvw:
Unstale me please

clwluvw:
unstale me

clwluvw:
unstale me

fedepaol:
Will try to revamp the review on this, sorry I didn't notice it was upgraded

clwluvw:
unstale me

internal/config/validation.go
fedepaol: sorry if I reply so late. I think we are rejecting multiple peers with same asn number / ip address, so it should be fine to ensure they are not configured on overlapping nodes

clwluvw: Thanks @fedepaol. I've simplified validateDuplicatePeers to just reject any duplicate peers whose node selectors might overlap, and removed the arePeersCompatible check entirely.

fedepaol:
Superseeded by https://github.com/metallb/metallb/pull/2980

PR #2810, opened 2025-07-30
MetalLB: translate disableMP to frr-k8s

fedepaol:

>
> /kind bug
/kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

Even if disableMP is not handled anymore, the frr-k8s version still has the field and if the field is set, we should honor the request of the user in translating it.

**Release note**:

```release-note
Translate the disablemp flag to frrconfiguration.
```


oribon:
lgtm

Issue #2203, opened 2023-12-06
Automated test focus / skip


fedepaol:
### Is your feature request related to a problem?

A new contributor might find daunting that when we run e2e we run ALL The tests, whereas what we run depends on the bgp mode and on the ip family (and on having or not prometheus there).

### Describe the solution you'd like

I envision a mode we run dev-env with that checks the bgp type, the ip family and what else and provides the right focuses.
This has to be an opt-in feature. We want the CI to run with specific focuses that are hardcoded and driven from outside to avoid starting to have volkswagen tests (https://github.com/auchenberg/volkswagen)

We should document this new mode and promote it as the default when developing locally.

### I've read and agree with the following

- [X] I've checked all open and closed issues and my request is not there.
- [X] I've checked all open and closed pull requests and my request is not there.

fedepaol:
An easier way would be to panic in inv e2etest if no focus is provided, and print out the table of focus-es we need depending on the environment

sbshah97:
How do we go about this task?

fedepaol:
Hey @sbshah97 , I changed my mind when writing this, and instead of automation, we could go simpler and do the following

- running `inv e2etest` with no focus - skip parameter fails and prints the set of skips that must be applied depending on the environment, as a nicely formatted table
- the content of the table must mirror what we do in CI https://github.com/metallb/metallb/blob/main/.github/workflows/ci.yaml#L256

Something along the line of:
- ipv4 - frr - no prometheus | SKIP="FRRK8S-MODE|IPV6|DUALSTACK"

and so on.

This should help first time contributors in understanding what should they do when they want to test MetalLB locally.

fedepaol:
@sbshah97 are you interested in working on this?

Additionally, it would make sense to maintain / build the list of skips in a single configuration file that is used both by CI and by the aforemenitioned logic.

roopeshsn:
Hi, @fedepaol! I am interested in working on this issue. I am a first-time contributor to MetalLB. As of my understanding after reading your above comments, running `inv e2etest` with the `skip` flag (`inv e2etest --skip BGP`) should fail, right? and a table of tests that should be skipped based on the environment should be outputted. In the CI for E2E, it seems that the tests are skipped based on `bgp-type` and `ip-family` flags. Could you clarify on this?

fedepaol:
[...]

PR #2836, opened 2025-08-29
Tasks: Automatic Test Selection (Auto-Focus)

rauhersu:
A feature, this is a new behaviour to run e2etests

/kind feature

This is an implementation of the [issue#2203](https://github.com/metallb/metallb/issues/2203):

This PR improves the e2e testing workflow by making intelligent test selection the default behavior and streamlining the user experience. 

E2E tests now automatically detect the development environment configuration and skip irrelevant tests                                                                                                                                                         
Eliminates the need to manually specify test filters for most common scenarios                                                                                                                                                                                 
Users can opt-out with --auto-focus-disable flag when manual control is needed      

**Release note**:

```release-note
    e2etests are run with auto test detection by default
```


oribon (commented):
thanks for submitting this!
having this recommend the set of tests to skip (the relevant flags) instead of running automatically will make it easier for the user to understand what's happening, see @fedepaol's 2 comments https://github.com/metallb/metallb/issues/2203#issuecomment-1886739372
which the way I see it, when no flags are set at all `inv e2etests` fails, with a message like: "the current environment is ..., to run all relevant tests use ..."

in addition, it is mentioned in the issue to have a maintained list of infra <-> tests, so that both CI and tasks.py can use it, did you attempt implementing it?

rauhersu:
Hey @oribon ! Thanks for the quick feedback. Let me answer again.

> in addition, it is mentioned in the issue to have a maintained list of infra <-> tests, so that both CI and tasks.py can use it, did you attempt implementing it?

I have added a new commit introducing `test-filters.yaml` as a configuration file for skip patterns: once the dev env is detected by `tasks.py`, those skip patterns will be selected according to that config, instead of programmatically (`tasks.py`) as the previous changes were doing. This file works as a single source of truth for skip patterns, so `ci.yaml` is using this mechanism as well.  See the modified `README.md `explaining these details.

> which the way I see it, when no flags are set at all inv e2etests fails, with a message like: "the current environment is ..., to run all relevant tests use ..."

So far I have maintained the `--autofocus-dry-run` I introduced. 
The user is expected to follow this wow:
```
# Check the skip patterns detected via auto-focus
$ inv e2etest --auto-focus-dry-run                                                                                                                                                                                                                             
...                                                                                                                                                                                                        
Auto-detected environment: BGP Type: frr-k8s, IP Family: dual, Protocol: bgp, Prometheus: True                                                                                                                                                              
Auto-skip patterns: FRR-MODE|IPV6|L2      

🏁 Dry-run complete. Exiting without running tests.
# Exec the tests with those  skip patterns if we are happy
$ inv e2etest
```

I believe that you are proposing getting the tests patterns and next manually introduce them, something like:
```
# Check the skip patterns detected via auto-focus
$ inv e2etest                                                                                                                                                                                                                              
...                                                                                                                                                                                                        
Auto-detected environment: BGP Type: frr-k8s, IP Family: dual, Protocol: bgp, Prometheus: True                                                                                                                                                              
Auto-skip patterns: FRR-MODE|IPV6|L2      
Exiting without running tests.

# Exec the tests manually including those skip patterns if we are happy
$ inv e2etest --skip "FRR-MODE|IPV6|L2"
```

Imho, that's a slightly more verbose wow, but it's true the 'dry run' mode was not specified in the description of the task. Let me know if we prefer this wow.

oribon:
right, I like the latter more (or, providing no flags should fail). also, it should fill the rest of the flags as well on the suggestion (bgp-mode, etc.)

rauhersu:
New changes available for reviewing.

1. I have re-written the history to squash commits with meaningful changes for reviewers.
2. I have documented this feature on  [e2etests/README.md](https://github.com/metallb/metallb/pull/2836/files#diff-fcdf9a4969bc608b247c4c0d0e76b4fbbc2733ede15f35f47c1219879ddcd7c9).

e2etest/test-filters.yaml
```diff
@@ -0,0 +1,40 @@
+# This file defines which test patterns to skip based on environment configuration
+test_filters:
```
oribon: nit: I think we can drop this (`filters = config["test_filters"]` seems odd when the file is called test-filters)

tasks.py
```diff
@@ -38,6 +39,95 @@
...
+        protocol: Protocol type (bgp, layer2) - optional
+        with_prometheus: Whether Prometheus is present (bool) - optional
+
+    Returns:
+        List of skip pattern strings
+    """
+    config = load_test_filters()
+    filters = config["test_filters"]
+    skip_patterns = []
+
```
oribon: let's add a validation for the args, failing if they don't match what we support. then we can drop the conditionals
```python
if x in filters[x]:
   ...
```

tasks.py
```diff
@@ -38,6 +39,95 @@
...
+                matches = False
+                break
+            elif key == "ip_family" and ip_family != value:
+                matches = False
+                break
+
+        # If condition matches, add the skip patterns
+        if matches:
+            if "skip" in special_case:
+                skip_patterns.extend(special_case["skip"])
```
oribon: let's start with hardcoding the single "special case" that we have (which is native + ipv6) instead of this logic (and the part in the file). if/when it becomes necessary we can enhance

tasks.py
```diff
@@ -38,6 +39,95 @@
...
+
+    # Remove duplicates and return
+    return list(set(skip_patterns))
+
+
+def generate_skip_string(bgp_type, ip_family, protocol=None, with_prometheus=None):
```
oribon: let's have the logic here as part of the previous function, no need to separate

tasks.py
```diff
@@ -1643,3 +1771,182 @@ def get_command_version(get_version_command: str, version_prefix: str) -> Option
...
+            "protocol": detect_protocol(),
+        }
+
+    except Exception as e:
+        raise Exit(message=f"Error: Could not auto-detect environment config: {e}")
+
+    return config
+
+
+def generate_test_filters(config):
```
oribon: this function is redundant (it just unpacks `config`), let's just use `generate_skip` directly where needed

tasks.py
```diff
@@ -1643,3 +1771,182 @@ def get_command_version(get_version_command: str, version_prefix: str) -> Option
...
+
+
+def detect_prometheus():
+    """Detect if Prometheus is available by checking monitoring namespace."""
+    prom_result = run(f"{kubectl_path} get namespace monitoring", hide=True, warn=True)
+    return prom_result.ok
+
+
+def detect_ip_family(namespace="metallb-system"):
+    """Detect IP family from ipaddresspool resource (dual, ipv4, ipv6)."""
```
oribon: there is no guarantee (or reason) that an ipaddresspool exists in a clean cluster, I think we should opt to checking the node/pods internal ips (something like https://kubernetes.io/docs/tasks/network/validate-dual-stack/#validate-node-addressing)

tasks.py
```diff
@@ -1643,3 +1771,182 @@ def get_command_version(get_version_command: str, version_prefix: str) -> Option
...
+        return "ipv6"
+    elif has_ipv4:
+        return "ipv4"
+    else:
+        raise Exception(
+            "Cannot detect IP family from ipaddresspool resource: unmatched ip family addresses"
+        )
+
+
+def detect_protocol(namespace="metallb-system"):
```
oribon: for the same reasons, but this time it is not really detectable 😅 I think we should drop this and let the user decide

e2etest/README.md
oribon, line 1: imo the changes here a bit excessive, can we narrow this down to documenting the new flag concisely?

oribon (commented):
thanks! looks a lot better

e2etest/test-filters.yaml
rauhersu: Done.

Removed the top-level `test_filters`.

tasks.py
rauhersu: Done.

Removed the generic `special_cases` configuration from YAML and implemented the native BGP + IPv6 special case directly in Python.

tasks.py
rauhersu: Done.

Refactored IP family detection to use node `podCIDRs` instead of `ipaddresspool` resources, as suggested in the comment above.

tasks.py
rauhersu: Done.

Used a 'best effort' approach to implement the protocol detection: if it will be auto-detected via `bgppeer`  or `l2advertisement` resources if configured. Otherwise, we are not able to auto-detect and need to make the user decide and provide a `--protocol` parameter for helping to auto-detect.

tasks.py
rauhersu: Done.

Removed that function.

tasks.py
rauhersu: I believe we need that function and it might be not redundant, instead of using  `generate_skip_patterns` directly:  the `ci.yaml` workflow needs a external call to fetch the 'skip patterns' automatically according to this new feature and that is done via 'run', fetching the corresponding parameters from the stdout so this function serves as an 'adapter' for the CI side of this feature.

Notice I have renamed this function to `generate_ci_skip_patterns`, intending to provide a better description of its purpose.

Cc: @karampok

e2etest/README.md
rauhersu: Done.

tasks.py
rauhersu: Done.

Created `validate_test_filter_parameters()`

.github/workflows/ci.yaml
```diff
@@ -229,18 +229,19 @@ jobs:
           HELM_FLAGS=""
           echo '/etc/frr/core-%e.%p.%h.%t' | sudo tee /proc/sys/kernel/core_pattern
           if [ ${{ matrix.deployment }} = "helm" ]; then export SPEAKER_SELECTOR="app.kubernetes.io/component=speaker" && export CONTROLLER_SELECTOR="app.kubernetes.io/component=controller"; fi
+
+          # Generate skip patterns using invoke task
+          SKIP_PATTERNS=$(inv generate-ci-skip-patterns "${{ matrix.bgp-type }}" "${{ matrix.ip-family }}")
```
karampok: IMO I will not  change the workflow ci. I would prefer to let it hardcoded the values to make sure in the future we do not silently change something.

e2etest/README.md
```diff
@@ -42,6 +42,56 @@ The test suite will run the appropriate tests against the cluster.
...
+## Step 2: Run Tests with Recommended Parameters
+
+Copy and run the recommended command from Step 1:
+
+```bash
+# Run the recommended command (this actually runs the tests)
+inv e2etest --skip "IPV6|DUALSTACK|L2" --bgp-mode frr
+```
+
+# E2E Tests auto detection in CI
```
karampok: no need to document, the workflow ci file should be the source of truth

e2etest/README.md
```diff
@@ -42,6 +42,56 @@ The test suite will run the appropriate tests against the cluster.
...
+
+Example output:
+```
+📊 Auto-detected environment: BGP Type: frr, IP Family: ipv4, Protocol: bgp, Prometheus: True
+⏭️  Auto-skip patterns: IPV6|DUALSTACK|L2
+
+The recommended test execution command is:
+$ inv e2etest --skip "IPV6|DUALSTACK|L2" --bgp-mode frr
+
+Exiting without running tests.
```
karampok: you could run the test using `--dry-run` also to show how many test will run
```
Ran 46 of 174 Specs in 0.003 seconds
SUCCESS! -- 46 Passed | 0 Failed | 0 Pending | 128 Skipped
PASS

Ginkgo ran 1 suite in 1.109130194s
Test Suite Passed
[I] kka@x1g12 ~/g/m/metallb (rauhersu/main ✔)> inv e2etest --skip "IPV4|DUALSTACK|FRRMODE|L2" --bgp-mode frr-k8s --ginkgo-params " -v --focus 'BGP' --dry-run"
````

File: tasks.py [hunk 11]
```diff
@@ -1250,6 +1351,46 @@ def e2etest(
...
+            f"\n📊 Auto-detected environment: BGP Type: {env_config['bgp_type']}, "
+            f"IP Family: {env_config['ip_family']}, "
+            f"Protocol: {env_config['protocol']}, "
+            f"Prometheus: {env_config['with_prometheus']}"
+        )
+
+        if skip_patterns:
+            print(f"⏭️  Auto-skip patterns: {skip_patterns}")
+        else:
+            print("✅ No tests will be skipped")
```
karampok: when else can happen?

e2etest/README.md
karampok: I would say that the test should run,  no reason just to  print and  to have to copy

tasks.py
```diff
@@ -1250,6 +1351,46 @@ def e2etest(
...
+
+        print(
+            f"\nThe recommended test execution command is:\n"
+            f"$ {recommended_cmd}\n\n"
+            f"Exiting without running tests."
+        )
+
+        return
+
+    if bgp_mode == "":
```
karampok, line 1294: When that can happen?

tasks.py
```diff
@@ -1643,3 +1785,182 @@ def get_command_version(get_version_command: str, version_prefix: str) -> Option
...
+    if protocol_override:
+        if protocol_override not in ["bgp", "layer2"]:
+            raise ValueError(
+                f"Invalid protocol override '{protocol_override}'. Supported: bgp, layer2"
+            )
+        return protocol_override
+
+    # Attempt automatic detection
+    bgppeer_result = run(
+        f"{kubectl_path} get bgppeer -n {namespace} "
```
karampok: the chances are that there will be always empty, since we run that on new kind clusters.
 I will not try to guess the protocol. It should always given.

karampok (commented):
Some comments from my side.

Please also squash in single commit your PR. Thanks

.github/workflows/ci.yaml
fedepaol: Agree. CI must be willingly pick what to choose and what not. The reason for this is to make the life easier to newcomers so they won't get surprised if a ton of tests fail locally.

e2etest/README.md
fedepaol: Agree, let's remove  this section

tasks.py
```diff
@@ -1231,6 +1330,8 @@ def e2etest(
     junit_report="junit-report.xml",
     host_bgp_mode="ibgp",
     frr_k8s_namespace="metallb-system",
+    auto_detect_env=False,
```
fedepaol: the ultimate goal of this work is to make the newcomers' life easier. So, I'd default to true here, and I would set this as false in ci

tasks.py
```diff
@@ -38,6 +39,102 @@
 yq_version = "v4.45.1"
 
 
+def load_test_filters():
```
fedepaol: please move these to the bottom of the file, and let's have all the autoselection utility functions in the same section of the file.

tasks.py
```diff
@@ -1250,6 +1351,46 @@ def e2etest(
...
+
+        print(
+            f"\n📊 Auto-detected environment: BGP Type: {env_config['bgp_type']}, "
+            f"IP Family: {env_config['ip_family']}, "
+            f"Protocol: {env_config['protocol']}, "
+            f"Prometheus: {env_config['with_prometheus']}"
+        )
+
+        if skip_patterns:
+            print(f"⏭️  Auto-skip patterns: {skip_patterns}")
```
fedepaol: call me old style, but please no emojis

rauhersu: Done.

tasks.py
rauhersu: Done

tasks.py
rauhersu: Done

tasks.py
rauhersu: Neither my comment above nor this discussion does no longer apply: the ci uses its own logic to run the tests (no auto detection) as the latest feedback indicated.

tasks.py
rauhersu: `--protocol` is a required parameter for auto detection, as the latest feedback indicated.

.github/workflows/ci.yaml
rauhersu: Done.

The CI has its own workflow, no auto detection there.

e2etest/README.md
rauhersu: Done.

No CI documentation.

e2etest/README.md
rauhersu: Done.

I have modified the previous behaviour: the tests will run by default whenever auto detection is enabled, which is enabled by default as Federico suggested too.

https://github.com/metallb/metallb/pull/2836#discussion_r2419261462

tasks.py
rauhersu: It was a dead code path. Removed.

tasks.py
rauhersu: If `--auto-detect-env=false` and the user does not provide that parameter.

tasks.py
rauhersu: Done.

`--protocol` is a mandatory parameter for auto detection.

e2etest/README.md
```diff
@@ -13,29 +13,77 @@ with other types of bgp backend, you specify that with `--bgp-type` parameter. F
...
+Running tests with auto-detected configuration...
+[Test execution begins...]
 ```
 
+## **Manual Mode**
+To use manual parameters exclusively, disable auto-detection:
+
+```bash
+# Disable auto-detection and run with explicit parameters
+inv e2etest --auto-detect-env=false --skip "IPV6|DUALSTACK" --bgp-mode frr
```
karampok, line 50: I prefer the auto-detect-env=false to be the default.

Because person  A that wants autodetect will run   `inv e2etest --bgp-mode frr`
(short command) and person  B that does not want autodetect will run  `inv e2etest --auto-detect-env=false --skip "IPV6|DUALSTACK" --bgp-mode frr` (long command)

Note:  I am opinionated because I am always running the test with focus
```
inv e2etest --skip "IPV6|DUALSTACK|FRR-MODE|L2" --bgp-mode frr-k8s --ginkgo-params " -v --focus 'ConfigurationState'  [--dry-run]"
```
and having always to carry extra `-- auto-detect-env=false` will be not ideal to me.

WDYT @oribon @fedepaol

karampok: I just observed this https://github.com/metallb/metallb/pull/2836#discussion_r2419261462 so I suppose is okay to have that default behavior

oribon:
@rauhersu sorry this takes so long, can you please resolve the conflicts?

Issue #2663, opened 2025-01-24
Ensure somehow that when adding a feature, the relevant doc is added in hte same PR


fedepaol:
### Is your feature request related to a problem?

We often oversee the need of documenting a new feature, so we end up with MetalLB bringing new features that lack their documentation counterpart.

### Describe the solution you'd like

We should have some kind of check / reminder both for the user AND the reviewer.
A non blocking CI check that verifies the website was touched when adding a feature PR is the best, a blunt checkbox to fill is probably better than nothing.

### I've read and agree with the following

- [x] I've checked all open and closed issues and my request is not there.
- [x] I've checked all open and closed pull requests and my request is not there.

PR #2841, opened 2025-09-03
introduce a new documention check workflow...

rauhersu:
Addresses #2663: documentation reminder for feature PRs. See scripts/README_DOC_CHECK.md for a detailed explanation and examples.

- Add doc-check.yaml workflow (non-blocking) that runs on PRs with feature changes to help ensure features are documented. It leverages 'check_docs_local.py' to implement the documentation checks.
- Detect API changes, new config samples, Helm changes, etc.
- Add local testing script and documentation.
- Modify the PR template with documentation checklist.

/kind documentation

**Release note**:

```release-note
   NONE
```


.github/pull_request_template.md
```diff
@@ -28,6 +28,20 @@
 
 **Special notes for your reviewer**:
 
+**Documentation Checklist** *(for feature PRs)*:
+<!--  
+If this PR adds new user-facing features, please ensure documentation is updated:
+- API changes should be reflected in generated docs (run `inv generateapidocs`)
```
oribon: I guess we can remove this, it is already enforced by ci :)

rauhersu: Done.

Removed.

fedepaol:
I think this is a bit too much for what we need to do.

A diff from where we are merging and a check that api, internal are changed are enough imo.
The script does a lot more, but I am not sure it's worth the complexity and the maintenance cost it's bringing. I'd add a simple bash script that does the check (and please no emojis!)

PR #2856, opened 2025-10-07
Configuration status

karampok:
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
Add configuratonStatus CRD
```

### DIFF
--- a/design/crd-configstate.md
+++ b/design/crd-configstate.md
@@ -1,0 +1,282 @@
+# ConfigurationState CRD
+
+As a cluster user, I want to verify that MetalLB CRs successfully applied to the API
+are also successfully processed by MetalLB components. This CR surfaces configuration
+errors currently hidden in logs, making them visible via standard kubectl commands.
+
+
+## Configuration Errors
+
+We define configuration errors as errors caused solely by user input that can only be
+resolved by correcting that input. This is in contrast to infrastructure errors (such as
+missing RBAC permissions to watch resources) which can be fixed by administrators through
+deployment changes or code bug fixes.
+
+**Note:** While webhooks cover most configuration errors, they cannot be solely relied upon.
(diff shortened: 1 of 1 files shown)


fedepaol:
I'll make a generic comment since the PR is still wip.

I am not sure I like this approach, and this is not exactly what was drafted in the related [design proposal](https://github.com/metallb/metallb/blob/main/design/crd-status.md#configuration-status).
What the proposal said was either to cut corners making one single global configuration status managed by the controller (because the controller is responsible of the majority of the configuration) OR a per speaker configuration (which might make sense now with all the node selectors we have). 

With one singleton status, we may have a gigantic status, fed by all the controllers of all the speakers, becoming a lot of data to carry around and to redistribute to clients. A per node / per speaker instance is more scalable imo.

Regarding having one condition per controller, what we are giving away here is the knowledge of the real error. A configuration is the result of many things: 
- the nodes
- the configuration bits
- possibly, the result of applying the frrk8s configuration

By hooking more or less blindly in each controller (even some that are unnecessary, he service controller and the bgp status ones are the first two I noticed), we are giving away the knowledge that the lower levels have, so the same error might float differently because a node event was processed before of a config event or viceversa.

All in all, we should consider if we prefer to stick to the structure of the other "status" controllers, which are waken by a signal and they go fetch the current status. This would allow us to retry whenever writing the new configurations status fails, and would have a single place where we write the configuration, instead of having multiple scattered writers.

karampok:
Thanks for the feedback.

I do not understand your points tbh :).  But I suggest as next step  that I add a design document where I  can expand my approach along side what is written https://github.com/metallb/metallb/blob/main/design/crd-status.md#configuration-status.

karampok:
https://github.com/metallb/metallb/pull/2881

PR #2863, opened 2025-10-15
Add readOnlyRootFilesystem=true to containers missing it

cgoncalves:
/kind bug

`readOnlyRootFilesystem` prevents containers from writing to the root filesystem, reducing attack surface and improving security posture by limiting potential malicious file modifications and ensuring immutable container runtime.

`allowPrivilegeEscalation=false` prevents containers from gaining additional privileges beyond those initially granted, further hardening the security posture by blocking privilege escalation attacks.

**Release note**:

```release-note
Added `readOnlyRootFilesystem: true` and `allowPrivilegeEscalation: false` to containers missing these security flags, enhancing container security posture without affecting functionality.
```

### DIFF
--- a/config/frr/speaker-patch.yaml
+++ b/config/frr/speaker-patch.yaml
@@ -23,6 +23,12 @@ spec:
           emptyDir: {}
         - name: metrics
           emptyDir: {}
+        - name: frr-tmp
+          emptyDir: {}
+        - name: frr-lib
+          emptyDir: {}
+        - name: frr-log
+          emptyDir: {}
       initContainers:
         # Copies the initial config files with the right permissions to the shared volume.
         - name: cp-frr-files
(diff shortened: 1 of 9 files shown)


fedepaol:
Thanks! Mind adding this also to the kustomize based installation under config?

config/frr/speaker-patch.yaml
```diff
@@ -100,8 +111,13 @@ spec:
               mountPath: /etc/frr
             - name: reloader
               mountPath: /etc/frr_reloader
+            - name: frr-log
```
fedepaol, line 114: is this related to this change?

cgoncalves: Yes. Logs are written to `/var/log/frr` therefore it needs write permissions which requires a volume because the root filesystem is read-only.

fedepaol:
thanks!

PR #2870, opened 2025-10-21
Design document for configuration state CR

karampok:
/kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

This is more for discussion imo but of course we reach a doc file we can merge

**Release note**:

```release-note
NONE
```

### DIFF
--- a/design/crd-configstate.md
+++ b/design/crd-configstate.md
@@ -1,0 +1,282 @@
+# ConfigurationState CRD
+
+As a cluster user, I want to verify that MetalLB CRs successfully applied to the API
+are also successfully processed by MetalLB components. This CR surfaces configuration
+errors currently hidden in logs, making them visible via standard kubectl commands.
+
+
+## Configuration Errors
+
+We define configuration errors as errors caused solely by user input that can only be
+resolved by correcting that input. This is in contrast to infrastructure errors (such as
+missing RBAC permissions to watch resources) which can be fixed by administrators through
+deployment changes or code bug fixes.
+
+**Note:** While webhooks cover most configuration errors, they cannot be solely relied upon.
+In corner cases (e.g., many concurrent API calls), two conflicting CRs could be stored in the
+API. Another common case is that the webhook explicitly ignores transient errors related to
+resource ordering by resetting fields that could cause such errors (via the
+`resetTransientErrorsFields` function).
+
+**Note:** This CRD specifically does not cover errors that arise when the configuration
+applied by the speaker to the `frr-k8s` daemon conflicts with other external configurations
+within that daemon.
+
+## API
+
+One CRD definition, Multiple CRs, Namespaced.
+
+```bash
+$ kubectl get configurationstates -n metallb-system
+NAME                    RESULT    LASTERROR   AGE
+controller              Valid                 5m
+speaker-kind-worker     Valid                 5m
+speaker-kind-worker2    Valid                 5m
+...
+
+# Query by component type using labels
+$ kubectl get configurationstates -n metallb-system -l metallb.io/component-type=speaker
+
+# Query by node using labels
+$ kubectl get configurationstates -n metallb-system -l metallb.io/node-name=kind-worker
+```
+
+```go
+// ConfigurationState is a status-only CRD that reports configuration validation results
+// from MetalLB components. The type and node information are conveyed through labels
+// rather than spec fields, following Kubernetes best practices for status resources.
+// Labels:
+//   - metallb.io/component-type: "controller" or "speaker"
+//   - metallb.io/node-name: node name (only for speaker)
+
+// ConfigurationStateStatus defines the observed state of ConfigurationState.
+type ConfigurationStateStatus struct {
+	// Result indicates the configuration validation result.
+	// Possible values:
+	//   - "Valid": Configuration is successfully validated.
+	//   - "Invalid": Configuration has errors.
+	//   - "Unknown": Component has not reported state (e.g., during initialization or after a crash).
+	// +optional
+	// +kubebuilder:validation:Enum=Valid;Invalid;Unknown
+	Result string `json:"result,omitempty"`
+
+	// LastError contains the error message from the last reconciliation failure.
+	// This field is empty when Result is "Valid".
+	// +optional
+	LastError string `json:"lastError,omitempty"`
+}
+
+```
+
+* Initially, users will query individual ConfigurationState CRs per component
+* Aggregation can be addressed in a future proposal/issue if needed
+* Users can leverage label selectors to query specific subsets (e.g., all speakers, specific node)
+* API does not provide the summary of the "input configuration", only the error
+  which should indicate to the user which CR must be changed
+
+## Controller side
+
+For Controller (Deployment - single instance) there are three reconcilers
+
+- PoolReconciler - yes, but only config errors, that error should appear in configState
+- PoolStatusReconciler - no config errors
+- ServiceReconciler - no config errors
+
+```yaml
+apiVersion: metallb.io/v1beta1
+kind: ConfigurationState
+metadata:
+    name: controller
+    namespace: metallb-system
+    labels:
+        metallb.io/component-type: controller
+status:
+    result: "Valid"
+    lastError: ""
+
+# Failed example
+apiVersion: metallb.io/v1beta1
+kind: ConfigurationState
+metadata:
+    name: controller
+    namespace: metallb-system
+    labels:
+        metallb.io/component-type: controller
+status:
+    result: "Invalid"
+    lastError: "failed to parse configuration: CIDR \"192.168.10.100/32\" in pool \"client2-pool\" overlaps with already defined CIDR \"192.168.10.0/24\""
+```
+
+## Speaker side
+
+For Speaker (DaemonSet - one pod per node) there are
+
+- ConfigReconciler - yes, but only config errors, that error should appear in configState
+- NodeReconciler - no config errors
+- FRRK8sReconciler - no config errors
+- ServiceReconciler - no config errors
+- ServiceBGPStatusReconciler - no config errors
+- Layer2StatusReconciler - no config errors
+
+```yaml
+apiVersion: metallb.io/v1beta1
+kind: ConfigurationState
+metadata:
+    name: speaker-kind-worker
+    namespace: metallb-system
+    labels:
+        metallb.io/component-type: speaker
+        metallb.io/node-name: kind-worker
+status:
+    result: "Valid"
+    lastError: ""
+
+# Failed example:
+apiVersion: metallb.io/v1beta1
+kind: ConfigurationState
+metadata:
+    name: speaker-kind-worker2
+    namespace: metallb-system
+    labels:
+        metallb.io/component-type: speaker
+        metallb.io/node-name: kind-worker2
+status:
+    result: "Invalid"
+    lastError: "peer peer1 referencing non existing bfd profile my-bfd-profile"
+```
+
+## Implementation
+
+Ideally like FRRNodeState pattern in metallb/frrk8s repo, but might not be
+ideal because the ConfigurationStateReconciler needs references to ConfigReconciler
+and NodeReconciler in-memory result. Alternative implementation based on
+condition instead of channel can be evaluated. TBD and discussed during
+the implementation PR.
+
+The `res := r.Handler(r.Logger, cfg)` will not be refactored to return
+an error. When handler returns SyncStateNoRetry result, that will be reported
+in the ConfigurationState status.
+
+E2E tests for transient errors can run without disabling webhooks. Tests for
+other validation errors might require disabling the webhook.
+
+## List of Configuration Errors that pass Webhooks
+
+### Transient Errors
+
+Transient errors occur due to interdependencies between CRDs where resources
+reference other resources that don't exist yet. These errors are temporary and
+resolve automatically when the missing resource is created. The webhook strips
+fields that can cause transient errors to avoid making assumptions based on
+object creation ordering.
+
+- BFD Profile Reference
+  - When: BGPPeer references a BFD profile that doesn't exist
+  - Error: `peer %s referencing non existing bfd profile %s`
+  - Code: [internal/config/config.go:290](https://github.com/metallb/metallb/blob/main/internal/config/config.go#L290)
+
+- Password Secret Reference
+  - When: BGPPeer references a password secret that doesn't exist
+  - Error: `secret ref not found for peer config %q/%q`
+  - Code: [internal/config/config.go:496](https://github.com/metallb/metallb/blob/main/internal/config/config.go#L496)
+
+- Community Alias Reference
+  - When: BGPAdvertisement references a community alias that doesn't exist
+  - Error: Invalid community format
+  - Code: [internal/config/config.go](https://github.com/metallb/metallb/blob/main/internal/config/config.go)
+
+**Given** a MetalLB installation in namespace `metallb-system`
+
+**When** user applies a BGPPeer that references a non-existent BFD profile:
+```yaml
+apiVersion: metallb.io/v1beta2
+kind: BGPPeer
+metadata:
+  name: peer1
+  namespace: metallb-system
+spec:
+  myASN: 64512
+  peerASN: 64513
+  peerAddress: 192.168.1.1
+  bfdProfile: my-bfd-profile  # References profile that doesn't exist
+```
+
+**Then** speaker/configReconciler fails to load configuration
+
+   ```bash
+   kubectl apply -f bgppeer.yaml
+   # Output: bgppeer.metallb.io/peer1 created
+
+   kubectl logs -n metallb-system -l component=speaker -c speaker --tail=50 | grep "error"
+   # Output: {"caller":"config_controller.go:140","controller":"ConfigReconciler","error":"peer peer1 referencing non existing bfd profile my-bfd-profile","level":"error"}
+   ```
+
+### Other Validation Errors
+
+Other validation errors can occur for example when the secret exists
+but has incorrect content. The user must fix the secret to resolve these
+errors. The webhook strips the `passwordSecret` field to avoid accessing secret
+content during validation. The reconciler validates the actual secret content.
+
+- Secret Type Mismatch
+  - When: BGPPeer references a secret that exists but has wrong type (not `kubernetes.io/basic-auth`)
+  - Code: [internal/config/config.go:500](https://github.com/metallb/metallb/blob/main/internal/config/config.go#L500)
+
+- Password Field Missing
+  - When: BGPPeer references a secret that exists but doesn't have a `password` field
+  - Code: [internal/config/config.go:505](https://github.com/metallb/metallb/blob/main/internal/config/config.go#L505)
+
+- IPv6 Pool with BFD Echo
+  - When: Pool has IPv6 CIDR, BGPAdvertisement references BGPPeer with BFDProfile that has echo mode enabled
+  - Code: [internal/config/validation.go](https://github.com/metallb/metallb/blob/main/internal/config/validation.go)
+
+**Given** a MetalLB installation in namespace `metallb-system`
+
+**When** user applies a Secret with wrong type and BGPPeer that references it:
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: bgp-password
+  namespace: metallb-system
+type: Opaque  # Wrong type - should be kubernetes.io/basic-auth
+stringData:
+  password: "mypassword123"
+---
+apiVersion: metallb.io/v1beta2
+kind: BGPPeer
+metadata:
+  name: peer-with-secret
+  namespace: metallb-system
+spec:
+  myASN: 64512
+  peerASN: 64513
+  peerAddress: 192.168.1.2
+  passwordSecret:
+    name: bgp-password
+```
+
+**Then** speaker/configReconciler fails to load configuration
+
+   ```bash
+   kubectl apply -f secret-and-peer.yaml
+   # Output:
+   # secret/bgp-password created
+   # bgppeer.metallb.io/peer-with-secret created
+
+   kubectl logs -n metallb-system -l component=speaker -c speaker --tail=50 | grep "error"
+   # Output: {"caller":"config_controller.go:140","controller":"ConfigReconciler","error":"parsing peer peer-with-secret secret type mismatch on \"metallb-system\"/\"bgp-password\", type \"kubernetes.io/basic-auth\" is expected \nfailed to parse peer peer-with-secret password secret","level":"error"}
+   ```
+
+## References
+
+- https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/918/files
+- Openshift Operators are doing a lot conditions to report overall state (Degraded) https://github.com/openshift/library-go/blob/91376e1b394e6eddd36c4421c44e2bb00503e3bf/pkg/operator/apiserver/controller/workload/workload.go#L181
+- **[Kubernetes API Conventions - Status Fields](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#spec-and-status)**:
+  Official guidelines for status subresources and conditions
+- **[Knative Conditions](https://github.com/knative/pkg/blob/main/apis/condition_types.go)**:
+  Standard condition types and structures used across Knative projects
+- **[Knative Condition Set Manager](https://github.com/knative/pkg/blob/main/apis/condition_set.go)**:
+  High-level condition management with automatic Ready condition calculation
+- **[Controller Pitfalls: Report Status and Conditions](https://ahmet.im/blog/controller-pitfalls/#report-status-and-conditions)**:
+  Real-world guidance on implementing conditions correctly, emphasizing observedGeneration importance


design/crd-status.md
```diff
@@ -108,6 +110,26 @@ status:
 Given that the configuration is composed by multiple CRs, there are few cases
 where a given configuration is invalid because of a single CR (i.e. invalid IP formatting).
 
+//TODO
```
fedepaol: given the eventual consistency nature of the cluster, and the fact that a configuration is composed by different pieces, it's not granted that the webhook will reliably intercept all the invalid configuration combinations. 
One example is two clients applying two overlapping ipaddresspools.

design/crd-configstatus.md
```diff
@@ -0,0 +1,348 @@
...
+  - lastTransitionTime: "2025-10-10T09:13:47Z"
+    message: ""
+    reason: SyncStateSuccess
+    status: "True"
+    type: speaker-kind-worker/frrk8sReconcilerValid
+  - lastTransitionTime: "2025-10-10T09:13:50Z"
+    message: ""
+    reason: SyncStateSuccess
+    status: "True"
+    type: speaker-kind-worker/configReconcilerValid
```
fedepaol: the user won't care about the configReconciler result (or the pool reconciler, or the node reconciler).
What we need to expose is the information of whether the configuration applied is digested or not.

So, I think in general the original proposal makes more sense here. One single configuration result (per server, at this point)

design/crd-configstatus.md
```diff
@@ -0,0 +1,348 @@
...
+but Ready condition is not false).
+
+## User Integration
+
+The aggregate Ready condition provides a simple way to check MetalLB's to check if there is no error:
+
+```bash
+kubectl get configurationstatus config-status -n metallb-system
+# NAME            READY   REASON
+# config-status   True    AllComponentsReady
```
fedepaol: we can still have an aggregate even if we split the configuration in multiple bits

design/crd-configstatus.md
```diff
@@ -0,0 +1,348 @@
...
+
+Even we can add all reconcilers (which should be ok), we must have
+
+* Controller (Deployment - single instance)
+    * PoolReconciler:  Condition: `controller/poolReconcilerValid`
+    * ServiceReconciler:  Condition: `controller/serviceReconcilerValid`
+
+* Speaker (DaemonSet - one pod per node)
+    * ConfigReconciler: Condition: `speaker-<nodename>/configReconcilerValid`
+    * ServiceReconciler: Condition: `speaker-<nodename>/serviceReconcilerValid`
```
fedepaol: I don't think this makes sense wrt the configuration

design/crd-configstatus.md
```diff
@@ -0,0 +1,348 @@
...
+    * NodeReconciler: Condition: `speaker-<nodename>/nodeReconcilerValid`
+    * FRRK8sReconciler Condition: `speaker-<nodename>/frrk8sReconcilerValid`
```
fedepaol: this is orthogonal to the configuration and probably requires to be addressed subsequently

design/crd-configstatus.md
```diff
@@ -0,0 +1,348 @@
...
+- Must review stale conditions, are they removed?
+
+# FEEDBACK from Fede
+
+https://github.com/metallb/metallb/pull/2856#issuecomment-3388777537
+
+> I am not sure I like this approach, and this is not exactly what was drafted
+> in the related design proposal.
+
+We need to clarify the drafted proposal, questions on the other doc
```
fedepaol: the questions have little to do with the proposal though, as if the webhook was able to catch any error, the only thing left would've been the frrk8s errors (which are not configuration errors)

design/crd-configstatus.md
```diff
@@ -0,0 +1,348 @@
...
+
+> What the proposal said was either to cut corners making one single global
+> configuration status managed by the controller (because the controller is
+> responsible of the majority of the configuration) OR a per speaker
+> configuration (which might make sense now with all the node selectors we have).
+
+This proposal does have "single global config status" AND each speaker adds each status.
```
fedepaol: it's a single global object, which becomes messy if there are 1000 nodes in the cluster

design/crd-configstatus.md
```diff
@@ -0,0 +1,348 @@
...
+
+Q: how the "node selectors" are important, AFAIC if there speaker it posts its status,
```
fedepaol: because if we didn't have node selectors, we could have had one single central point (the controller) in charge to validate that all the configuration was valid.

design/crd-configstatus.md
```diff
@@ -0,0 +1,348 @@
...
+> controllers of all the speakers, becoming a lot of data to carry around and to
+> redistribute to clients. A per node / per speaker instance is more scalable
+> imo.
+
+Status size is fixed (3 condition for the controller +
+N times 3 conditions per speaker) and the user should only check the summary.
+A solution which would list per node status and the user must "grep" the not valid
+is less ideal
+
+Q: What is "redistribute to clients"?
```
fedepaol: whoever puts a watch on that (every speaker, in case of 1000 nodes).

design/crd-configstatus.md
```diff
@@ -0,0 +1,348 @@
...
+
+> Regarding having one condition per controller, what we are giving away here is
+> the knowledge of the real error. A configuration is the result of many things:
+> * the nodes
+> * the configuration bits
+> * possibly, the result of applying the frrk8s configuration
+
+Q: what is "real" error example?
```
fedepaol: see my answer on the original design proposal

design/crd-configstatus.md
```diff
@@ -0,0 +1,348 @@
...
+
+Q: do you mean by the configuration bits the configuration of  all `kubectl api-resources --api-group=metallb.io -o name`
```
fedepaol: as I wrote above, one configuration might be valid for one speaker but not for another (because of node selectors)

design/crd-configstatus.md
```diff
@@ -0,0 +1,348 @@
...
+AFAIC The validation if overall config is ok or not, it means if there is any
+error somewhere or do you mean to actively check (e.g. vtysh show bgp config)?
+
+> By hooking more or less blindly in each controller (even some that are
+> unnecessary, he service controller and the bgp status ones are the first two I
+> noticed), we are giving away the knowledge that the lower levels have, so the
+> same error might float differently because a node event was processed before of
+> a config event or viceversa.
+
+'hooking blindly' is not "unnecessary", e.g if the poolreconciler loses rbac to
```
fedepaol: but this is not an error in the configuration, which is what this is about.

design/crd-configstatus.md
```diff
@@ -0,0 +1,348 @@
...
+list something, that error will be shown, in all other case the condition will be ok.
+It is for free to post condition.
+
+Q: Can we define the "giving away the lower levels have? and "float differently"?
```
fedepaol: By reporting the errors from all the controllers, we are giving up control on what we expose (because we are exposing everything) instead of being deliberate on what's really a configuration error (vs for example an rbac error). 

I'd like this to observe the same pattern we have in the other status controllers.
One status object changed whenever it's needed, a signal to one single controller in charge of changing it.
This pattern is more solid than changing the status from multiple different places.

design/crd-configstatus.md
```diff
@@ -0,0 +1,348 @@
...
+
+At the end what would be the description of an e2e test for that case you have in mind?
+
+> All in all, we should consider if we prefer to stick to the structure of the
+> other "status" controllers, which are waken by a signal and they go fetch the
+> current status. This would allow us to retry whenever writing the new
+> configurations status fails, and would have a single place where we write the
+> configuration, instead of having multiple scattered writers.
+
+Q: Can we define "fetch the current status"? How can controller that wakes up from signal find out the status? ask frrk8s?
```
fedepaol: check how frrk8s' configuration status works https://github.com/metallb/frr-k8s/blob/main/internal/controller/frrconfiguration_controller.go

design/crd-configstatus.md
```diff
@@ -0,0 +1,348 @@
...
+
+Q: "When the configuration status fails", what is that in code? Reconciler loop runs?
+
+"instead of having multiple scattered writers"  this sounds very bad, but is not
```
fedepaol: it's not only about the impact on the api server, it's about having one single place we write the configuration from

design/crd-configstatus.md
```diff
@@ -0,0 +1,348 @@
...
+due to condition ServerSide condition patching.
+
+
+Q: Channel approach is the following?
+
+```
+Having speaker code have a controller ConfigStatus which create
+`config-status-speaker-on-node1`, and it wakes up on a channel event.
+Other controller (only frrk8s?) when runs it always send an event.
```
fedepaol: see https://github.com/metallb/frr-k8s/blob/main/internal/controller/frrconfiguration_controller.go

design/crd-configstatus.md
karampok: There is no watch on the speaker, there is single watch on controller which aggregates. The rest simply patch their condition.

design/crd-configstatus.md
karampok: The user will only see the information in the READY condition, the other is internal iimplementation

Do you suggest the following

```
kubectl get configurationstatus -n metallb-system
# NAME            READY   REASON
config-status-controller   True    ""
config-status-worker-1   True    ""
config-status-worker-3   True    ""
.
.
config-status-worker-3   false    ""
.
.
config-status-worker-N   True    ""
```

Like what we have

```
k get frrnodestates
NAME                 AGE
kind-control-plane   86m
kind-worker          86m
kind-worker2         86m
```

design/crd-configstatus.md
karampok: If we have configstatus per speaker. Who will aggregate the results?

design/crd-configstatus.md
karampok: ok

design/crd-configstatus.md
karampok: So if frrk8s has an external config, and the metallb config is failing because frrk8swebhook and frrk8sreconciler logs an error. Is  this orthogonal?

design/crd-configstatus.md
karampok: Why is messier if we have  one object with 1000 condition instead of 1000 objects?

design/crd-configstatus.md
karampok: hypothetically, let's say we do have  nodeselectors, how would the controller in central point validate the configuration?

design/crd-configstatus.md
karampok: ok

design/crd-configstatus.md
karampok: I understand you want to see an exact implementation of `frrk8sv1beta1.FRRNodeStateStatus`

design/crd-configstatus.md
karampok: > really a configuration error

One example is two clients applying two overlapping ipaddresspools. -->  
is this real error? I cannot have it because it is blocked by webhook 

```   
 admission webhook "ipaddresspoolvalidationwebhook.metallb.io" denied the request: CIDR "192.168.10.100/30" in pool "pool-2" overlaps with already defined CIDR "192.168.10.0/24"```

Webhook from frrk8s --> is this "real error?

Do we have other examples I can use for the e2e tests?

File: design/crd-configstatus.md (hunk 15 above)
Comment (karampok) [reply design/crd-configstatus.md]: > we write configuration from

Can we have verbatim example of the configuration, is it the final FRR configuration?
```
  lastReloadResult: success
  runningConfig: |
    Building configuration...

    Current configuration:
    !
    frr version 9.1_git
    frr defaults traditional
    hostname kind-worker
    log file /etc/frr/frr.log
    log timestamp precision 3
    no ipv6 forwarding
````

File: design/crd-status.md (hunk 1 above)
Comment (karampok) [reply design/crd-status.md]: This example is not passing web hook. Is there another example?

What I try is
```
		ginkgo.By("Applying two IPAddressPools with overlapping IP ranges")
		resources := config.Resources{
			Pools: []metallbv1beta1.IPAddressPool{
				{
					ObjectMeta: metav1.ObjectMeta{
						Name: "pool-1",
					},
					Spec: metallbv1beta1.IPAddressPoolSpec{
						Addresses: []string{
							"192.168.10.0/24",
						},
					},
				},
				{
					ObjectMeta: metav1.ObjectMeta{
						Name: "pool-2",
					},
					Spec: metallbv1beta1.IPAddressPoolSpec{
						Addresses: []string{
							"192.168.10.100-192.168.10.200",
						},
					},
				},
			},
		}

```

Comment (fedepaol) [reply design/crd-status.md]: I wrote "two clients"

File: design/crd-configstatus.md (hunk 2 above)
Comment (fedepaol) [reply design/crd-configstatus.md]: yes, this is what I am suggesting

File: design/crd-configstatus.md (hunk 3 above)
Comment (fedepaol) [reply design/crd-configstatus.md]: the controller for example, collecting one global config with one single status.

File: design/crd-configstatus.md (hunk 5 above)
Comment (fedepaol) [reply design/crd-configstatus.md]: imo yes. Because this is meant to intercept errors in metallb configuration. It's similar to when we generate a configuration and frr says it's not valid

File: design/crd-configstatus.md (hunk 8 above)
Comment (fedepaol) [reply design/crd-configstatus.md]: you asked how the node selectors are important. My comment is the answer to that.

File: design/crd-configstatus.md (hunk 15 above)
Comment (fedepaol) [reply design/crd-configstatus.md]: I don't understand the comment

File: design/crd-configstatus.md (hunk 14 above)
Comment (fedepaol) [reply design/crd-configstatus.md]: I am saying that model proved to work well, with one single writer in the whole process, and the other status controlelrs in metallb work in the same way.

File: design/crd-configstatus.md (hunk 8 above)
Comment (karampok) [reply design/crd-configstatus.md]: not arguing that are not important. Just trying to udnerstand what you mean by saying "validate the configuration"

File: design/crd-configstatus.md (hunk 15 above)
Comment (karampok) [reply design/crd-configstatus.md]: When I have failure and I execute
```
kubectl get configurationstatus -n metallb-system
# NAME            READY   REASON
config-status-worker-3   false    "HERE"
````
what is the string value that you want to see in HERE,
do you mean the frrconfiguration?

File: design/crd-status.md (hunk 1 above)
Comment (karampok) [reply design/crd-status.md]: Do you mean this?
```
		ginkgo.By("Client 1: Applying first IPAddressPool")
		resources1 := config.Resources{
			Pools: []metallbv1beta1.IPAddressPool{
				{
					ObjectMeta: metav1.ObjectMeta{
						Name: "client1-pool",
					},
					Spec: metallbv1beta1.IPAddressPoolSpec{
						Addresses: []string{
							"192.168.10.0/24",
						},
					},
				},
			},
		}

		err := ConfigUpdater.Update(resources1)

		ginkgo.By("Client 2: Applying second IPAddressPool with overlapping range")
		resources2 := config.Resources{
			Pools: []metallbv1beta1.IPAddressPool{
				{
					ObjectMeta: metav1.ObjectMeta{
						Name: "client2-pool",
					},
					Spec: metallbv1beta1.IPAddressPoolSpec{
						Addresses: []string{
							"192.168.10.100-192.168.10.200",
						},
					},
				},
			},
		}

		err = ConfigUpdater.Update(resources2)
```

File: design/crd-configstatus.md (hunk 13 above)
Comment (fedepaol) [reply design/crd-configstatus.md]: I replied above. Again, the webhooks cannot be trusted to cover everything as they don't cover eventual consistency, and the state they observe might be incomplete

File: design/crd-configstatus.md (hunk 7 above)
Comment (fedepaol) [reply design/crd-configstatus.md]: because the majority of those conditions are just noise for the user.

File: design/crd-configstatus.md (hunk 9 above)
Comment (fedepaol) [reply design/crd-configstatus.md]: fair enough, but having one gigantic object is still not ideal

File: design/crd-configstatus.md (hunk 3 above)
Comment (karampok) [reply design/crd-configstatus.md]: > collecting one global config with one single status.

Do you mean watching the configstatus per server/speaker,  and updating its own configstatus

```
kubectl get configurationstatus -n metallb-system
# NAME            READY   REASON
config-status-controller   False    "Worker 50 has invalid config" <----  THIS
config-status-worker-1   True    ""
config-status-worker-3   True    ""
.
.
config-status-worker-50   false    ""
.
.
config-status-worker-N   True    ""
```

File: design/crd-status.md (hunk 1 above)
Comment (fedepaol) [reply design/crd-status.md]: I mean two distinct clients calling the webhook in parallel.

File: design/crd-configstatus.md (hunk 3 above)
Comment (fedepaol) [reply design/crd-configstatus.md]: nope, because the contrller might have its own configuration status (different from the speakers, because the logic and the types of resources it watches are different).

I am saying that, as an extra, we might have a global config telling us if everything is fine or not. But this can be an extra, and wasn't intended in the original design.

File: design/crd-configstatus.md (hunk 15 above)
Comment (fedepaol) [reply design/crd-configstatus.md]: I want a message saying why the configuratoin failed

File: design/crd-configstatus.md [hunk 17]
```diff
@@ -0,0 +1,348 @@
...
+**only patches its own condition**. 
+
+Even we can add all reconcilers (which should be ok), we must have
+
+* Controller (Deployment - single instance)
+    * PoolReconciler:  Condition: `controller/poolReconcilerValid`
+    * ServiceReconciler:  Condition: `controller/serviceReconcilerValid`
+
+* Speaker (DaemonSet - one pod per node)
+    * ConfigReconciler: Condition: `speaker-<nodename>/configReconcilerValid`
```
karampok: So this is the reconciler that we only care?

```
	cfg, err := toConfig(resources, r.ValidateConfig)
	if err != nil {
		configStale.Set(1)
		syncResult = SyncStateError
		level.Error(r.Logger).Log("controller", "ConfigReconciler", "error", "failed to parse the configuration", "error", err)
		syncError = fmt.Errorf("failed to parse configuration: %w", err)
		return ctrl.Result{}, nil
	}

````

https://github.com/metallb/metallb/blob/621e1bceeff96f07b6520cdc780fd28d90533a6f/internal/k8s/controllers/config_controller.go#L137

File: design/crd-configstatus.md (hunk 14 above)
Comment (fedepaol) [reply design/crd-configstatus.md]: I checked the sriov operator proposal. They have both conditions and explicit fields in the more complicated object: https://github.com/k8snetworkplumbingwg/sriov-network-operator/pull/918/files#diff-f5bf2694480f9bffddea2134e0f099b121c0b5aef2d2e3a6c8e281349ed6f21cR165


So this is probably something we can do to, particularly for the frr status / frr k8s status bits.
We can have a condition for the whole config, but at the same time I think we should have a global "config ok / failed" field plus a message.

Thoughts?

**Comment by karampok** [comment]
@fedepaol I updated the document, let me know what you think, thanks

File: design/crd-configstatus.md (hunk 14 above)
Comment (karampok) [reply design/crd-configstatus.md]: I miss which frr k8s bits you refer, let's have a new comment in the updated design doc with what you suggest

File: design/crd-configstate.md [hunk 18]
```diff
@@ -0,0 +1,275 @@
...
+```go
+// ConfigurationStateSpec defines the desired state of ConfigurationState.
+type ConfigurationStateSpec struct {
+	// Type identifies whether this is a controller or speaker instance.
+	// +kubebuilder:validation:Enum=Controller;Speaker
+	Type string `json:"type"`
+
+	// NodeName is set when Type is "Speaker" to identify which node this speaker is running on.
+	// +optional
+	NodeName string `json:"nodeName,omitempty"`
```
fedepaol: I am probably biased by "user driven" specs, but I would've expected the CRD to be a status only CRD, having a representative name plus having labels to allow the user to query on.

The spec part is normally assigned to the "desired status" while here we are using it as informative. Does it make sense?

design/crd-configstate.md
```diff
@@ -0,0 +1,275 @@
...
+	ValidConfig bool `json:"validConfig,omitempty"`
+
+	// LastError contains the error message from the last reconciliation failure.
+	// Empty when ValidConfig is true.
+	// +optional
+	LastError string `json:"lastError,omitempty"`
+
+```
+
+* Aggregator of the status will not be implemented, user will look into each one CR
```
fedepaol: let's leave the door open and say we can work on it in a new proposal / issue

design/crd-configstate.md
```diff
@@ -0,0 +1,275 @@
...
+status:
+    validConfig: false
+    lastError: "failed to parse configuration: invalid BGP peer address"
+```
+
+## Implementation
+
+Ideally like FRRNodeState pattern it metallb/frrk8s repo, but might not be
+ideal because for ConfigurationStateReconciler needs references to ConfigReconciler
+and NodeReconciler in-memory result. Alternative implementation based on
```
fedepaol, line 152: I think it's fine to have it condition based, but if we go this route we'll need a controller who watches its own status to do the aggregation

design/crd-configstate.md
karampok: yes it does, I will update the doc saying add add the type/node  labels instead.

design/crd-configstate.md
karampok: i will rephrase to say maybe in the future

design/crd-configstate.md
```diff
@@ -0,0 +1,275 @@
...
+   # Output:
+   # secret/bgp-password created
+   # bgppeer.metallb.io/peer-with-secret created
+
+   kubectl logs -n metallb-system -l component=speaker -c speaker --tail=50 | grep "error"
+   # Output: {"caller":"config_controller.go:140","controller":"ConfigReconciler","error":"parsing peer peer-with-secret secret type mismatch on \"metallb-system\"/\"bgp-password\", type \"kubernetes.io/basic-auth\" is expected \nfailed to parse peer peer-with-secret password secret","level":"error"}
+   ```
+
+
+//TODO Configuration errors in speaker/ConfigReconciler/Handler
```
karampok: @fedepaol  me and Claude have not been able to find  input that gives  configuration error,. We will do some deeper next week but if you know a scenario, that  would be nice (looking at least one to decide if handler needs refactor or not).

design/crd-configstate.md
```diff
@@ -0,0 +1,275 @@
...
+## Configuration Errors
+
+We define configuration errors as errors caused solely by user input that can only be
+resolved by correcting that input. This is in contrast to infrastructure errors (such as
+missing RBAC permissions to watch resources) which can be fixed by administrators through
+deployment changes or code bug fixes.
+
+**Note:** Webhooks are covering almost all configuration errors but we can not
+rely because in corner cases (many calls in parallel) could two conflicted CRs
+stored in the API.
```
oribon: I'd say another (more common) point is that the webhook explicitly ignores errors related to resource orderings (the `resetTransientErrorsFields` func)

design/crd-configstate.md
```diff
@@ -0,0 +1,275 @@
...
+	// +optional
+	NodeName string `json:"nodeName,omitempty"`
+}
+
+// ConfigurationStateStatus defines the observed state of ConfigurationState.
+type ConfigurationStateStatus struct {
+	// ValidConfig indicates whether the configuration is valid.
+	// True when all reconcilers report success, False otherwise.
+	// +optional
+	ValidConfig bool `json:"validConfig,omitempty"`
```
oribon: I'd have this as a string, that for now will only be Valid/Invalid (to be more flexible)

design/crd-configstate.md
```diff
@@ -0,0 +1,275 @@
...
+resolved by correcting that input. This is in contrast to infrastructure errors (such as
+missing RBAC permissions to watch resources) which can be fixed by administrators through
+deployment changes or code bug fixes.
+
+**Note:** Webhooks are covering almost all configuration errors but we can not
+rely because in corner cases (many calls in parallel) could two conflicted CRs
+stored in the API.
+
+**Note:** Specifically do not cover the case where the applied frrk8sconfig from
+speaker gives error, because of other external config in the frrk8s daemon.
```
oribon: why? (sorry if this has already been answered)

fedepaol: because the focus of this is user provided configuration. When we have this in place, it should be relatively easy to extend it with a condition or a field representing the status of frrconfiguration

design/crd-configstate.md
karampok: Do you have a usecase that we would like to extend to other values? I

Given I cannot find a nice name (validConfig is valid/invalid does not sounds as good as validConfig true), I prefer to say, but no strong opinion

karampok: I could see a value for "unknown" on second thought,
What do you think of   the name Result?
```
  type ConfigurationStateStatus struct {
      // Result indicates the configuration validation result.
      // Possible values: "Valid", "Invalid", "Unknown"
      // +kubebuilder:validation:Enum=Valid;Invalid;Unknown
      Result string `json:"result,omitempty"`
```

karampok: added a commit https://github.com/metallb/metallb/pull/2870/commits/ff30a0e123c1846d981e70481ffdec5850c10f0b if we like the result I squash,

design/crd-configstate.md
```diff
@@ -0,0 +1,283 @@
...
+kind: ConfigurationState
+metadata:
+    name: speaker-kind-worker2
+    namespace: metallb-system
+    labels:
+        metallb.io/component-type: speaker
+        metallb.io/node-name: kind-worker2
+status:
+    result: "Invalid"
+    lastError: "failed to parse configuration: invalid BGP peer address"
```
fedepaol: is this a real example? I would expect the name of the bgppeer and the address (not to be addressed here). Providing more insightful errors is out of scope, but if this was made up please alignn to the real error

design/crd-configstate.md
```diff
@@ -0,0 +1,283 @@
...
+
+Ideally like FRRNodeState pattern it metallb/frrk8s repo, but might not be
+ideal because for ConfigurationStateReconciler needs references to ConfigReconciler
+and NodeReconciler in-memory result. Alternative implementation based on
+condition instead of channel can be evaluated. TBD and discussed during
+the implementation PR.
+
+`res := r.Handler(r.Logger, cfg)` might need refactor to return error.
+
+**Note:** Subset of E2E test will work only with disabling the webhook
```
fedepaol: given the transient error considerations, I think we can test without disabling the webhooks, right?

design/crd-configstate.md
fedepaol: Yeah, looking at the code it seems like we are not returning (apart from corner cases). I'd say let's hook the error logic into the SyncStateNoRetry result and bubble up a generic configuration error there so it's future proof (in case the handler is changed).

design/crd-configstate.md
oribon: what would be an example for unknown? (`Result` seems fine to me)
about 
```
Do you have a usecase that we would like to extend to other values?
```
it's just that more than often bool fields end up needing to be refactored, and I like the explicitness (and flexibility) of strings 😅

design/crd-configstate.md
karampok: No just made up error

design/crd-configstate.md
karampok: for the transient errors, correct no webhook disable is needed (and probably we are ok)

karampok:
I have updated all the comments (plus some typos/ wording suggested by AI).  Thanks

There are three TODOs, which needs to be investigated but not sure will change the gist of the proposal.

design/crd-configstate.md
```diff
@@ -0,0 +1,288 @@
...
+   # Output:
+   # secret/bgp-password created
+   # bgppeer.metallb.io/peer-with-secret created
+
+   kubectl logs -n metallb-system -l component=speaker -c speaker --tail=50 | grep "error"
+   # Output: {"caller":"config_controller.go:140","controller":"ConfigReconciler","error":"parsing peer peer-with-secret secret type mismatch on \"metallb-system\"/\"bgp-password\", type \"kubernetes.io/basic-auth\" is expected \nfailed to parse peer peer-with-secret password secret","level":"error"}
+   ```
+
+
+//TODO configuration errors in speaker/nodecontroller `internal/k8s/controllers/node_controller.go`
```
fedepaol: Let's remove / fix the TODOs and I think this can be merged.
There's no need to outline all the possible errors here as long as we cover them in the implementation

fedepaol:
Thanks!

karampok:
that was fast, thank you

Issue #2806, opened 2025-07-22
false-positive MetalLBBGPSessionDown firing


miminar:
### MetalLB Version

v0.15.2

### Deployment method

Charts

### Main CNI

kube-ovn

and the same happens on another cluster with cilium

### Kubernetes Version

v1.32.6

### Cluster Distribution

k0s

### Describe the bug

Since the update of metallb to v0.15.2 (from v0.14.9), we get a false-positive alert MetalLBBGPSessionDown for all the BGPPeers.

### To Reproduce

Deploy the chart with values:

```yaml
  controller:
    image:
      # renovate: datasource=docker depName=quay.io/metallb/controller versioning=docker
      tag: main   # same with v0.15.2 
    logLevel: debug
  speaker:
    frr:
      enabled: false
    image:
      # renovate: datasource=docker depName=quay.io/metallb/speaker versioning=docker
      tag: main   # same with v0.15.2 
    logLevel: debug
  frrk8s:
    enabled: false
  prometheus:
    serviceMonitor:
      enabled: true
    prometheusRule:
      enabled: true
    rbacPrometheus: false
```

Additional resources:
```
---
apiVersion: metallb.io/v1beta2
kind: BGPPeer
metadata:
  name: rou-lca-1
  namespace: metallb-system
spec:
  ebgpMultiHop: true
  myASN: 4250001028
  peerASN: 4250000126
  peerAddress: 10.206.128.9
---
apiVersion: metallb.io/v1beta2
kind: BGPPeer
metadata:
  name: rou-lca-2
  namespace: metallb-system
spec:
  ebgpMultiHop: true
  myASN: 4250001028
  peerASN: 4250000126
  peerAddress: 10.206.128.10
```

### Expected Behavior

No MetalLBBGPSessionDown alert firing when the BGP sessions are established. No duplication of peers in the timeseries (IP and IP:Port).

### Additional Context

Time series for `metallb_bgp_session_up{job=~"metallb.*"} == 0`:

```

metallb_bgp_session_up{container="speaker", endpoint="metrics", instance="10.205.202.10:7472", job="metallb", namespace="metallb-system", peer="10.206.128.10", pod="metallb-speaker-nkvkw", service="metallb-speaker-monitor-service"}
[...]

PR #2879, opened 2025-11-03
Restore bgp metrics for native mode

fedepaol:

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

Restoring the metrics with the original format. Reverts bce07617862b630a321b04f716efe7089adc4f47 for the native bgp implementation where we were referring the peer with its address.

The effect here is that we had the original setup with the peeraddress as 0, and all the other events with the peer name. Here we restore to what we had before the given commit.

Fixes https://github.com/metallb/metallb/issues/2806

**Release note**:

```release-note
Fix never changing session down metric with peer address in native mode. All the other events are generated with address:port, here we restore to the original format with the address only.
```


oribon (approved):
lgtm

PR #2881, opened 2025-11-05
Add ConfigurationState CRD to expose configuration errors

karampok:
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
  Add ConfigurationState CRD to expose configuration errors (when user-defined CRs cannot be reconciled) for each
  MetalLB component
```

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -240,7 +240,7 @@ jobs:
           if [ "${{ matrix.ip-family }}" == "ipv6" ] && [ ${{ matrix.bgp-type }} == "native" ]; then SKIP="$SKIP|BGP"; fi
           if [ "${{ matrix.bgp-type }}" == "frr" ]; then SKIP="$SKIP|FRRK8S-MODE"; fi
           if [ "${{ matrix.bgp-type }}" == "frr-k8s" ]; then SKIP="$SKIP|FRR-MODE"; fi
-          if [ "${{ matrix.bgp-type }}" == "frr-k8s-external" ]; then SKIP="$SKIP|FRR-MODE|metrics"; FRRK8NAMESPACE="--frr-k8s-namespace=frr-k8s-system"; fi
+          if [ "${{ matrix.bgp-type }}" == "frr-k8s-external" ]; then SKIP="$SKIP|FRR-MODE|metrics|Networkpolicies"; FRRK8NAMESPACE="--frr-k8s-namespace=frr-k8s-system"; fi
 
           echo "Skipping $SKIP"
           echo "With VRF $WITH_VRF"
@@ -300,10 +300,13 @@ jobs:
       - name: Create multi-node K8s Kind Cluster
         run: |
           cd ${GITHUB_WORKSPACE}/metallboperator
+          which kind
+          kind version
           ./hack/kind-multi-node-cluster-without-registry.sh
           kind load docker-image quay.io/metallb/speaker:dev-amd64
           kind load docker-image quay.io/metallb/controller:dev-amd64
           export KUBECONFIG=${HOME}/.kube/config
+          kubectl version
 
       - name: Deploy Metal LB Operator
         run: |
--- a/.github/workflows/composite/setup/action.yaml
+++ b/.github/workflows/composite/setup/action.yaml
@@ -15,7 +15,7 @@ runs:
         sudo apt-get update
         sudo apt-get install python3-pip arping ndisc6
         sudo pip3 install -r dev-env/requirements.txt
-        GO111MODULE="on" go get sigs.k8s.io/kind@v0.19.0
+        GO111MODULE="on" go install sigs.k8s.io/kind@v0.30.0
         go install sigs.k8s.io/controller-tools/cmd/controller-gen@v0.7.0
         #NOTE: This pulls the ginkgo version that is pinned in go.mod
         go install github.com/onsi/ginkgo/v2/ginkgo
--- a/e2etest/netpoltests/networkpolicies.go
+++ b/e2etest/netpoltests/networkpolicies.go
@@ -57,13 +57,25 @@ var _ = ginkgo.Describe("Networkpolicies", func() {
 	ginkgo.It("only allowed traffic", func() {
 		ginkgo.By("checking ingress - probe pod to any port on the controller other than webhook,metricshttps should be timeout")
 
+		// Check if controller has secure metrics port
+		hasSecureMetrics := false
+		for _, c := range controller.Spec.Containers {
+			for _, p := range c.Ports {
+				if p.Name == "metricshttps" {
+					hasSecureMetrics = true
+					break
+				}
+			}
+		}
+
 		ingressPorts := map[string]string{
 			"8888": "TIMEOUT", // Add random port to check for being blocked
 		}
 		for _, c := range controller.Spec.Containers {
 			for _, p := range c.Ports {
 				port := strconv.Itoa(int(p.ContainerPort))
-				if p.Name == "monitoring" { // the non-https port is not whitelisted in the network policies
+				if p.Name == "monitoring" && hasSecureMetrics {
+					// monitoring port should be blocked when secure metrics are available
 					ingressPorts[port] = "TIMEOUT"
 					continue
 				}


internal/k8s/controllers/config_controller.go
```diff
@@ -57,6 +60,15 @@ func (r *ConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctr
 var requestHandler = func(r *ConfigReconciler, ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	level.Info(r.Logger).Log("controller", "ConfigReconciler", "start reconcile", req.String())
 	defer level.Info(r.Logger).Log("controller", "ConfigReconciler", "end reconcile", req.String())
+
+	var resultErr error
+	errType := ErrorTypeNone
+	defer func() {
+		if err := r.reportCondition(ctx, resultErr, errType); err != nil {
```
karampok: I was conservative but I am thinking here to just return error if fails so back-off controller will rerun
(just to cover the unique case the CR is not yet created)

e2etest/configurationstatetests/configurationstate.go
```diff
@@ -0,0 +1,352 @@
...
+		Expect(err).NotTo(HaveOccurred())
+
+		ginkgo.By("Verifying status has valid result")
+		Eventually(func() error {
+			return stateMatches(stateName, validStatus)
+		}, 60*time.Second, 5*time.Second).Should(Succeed())
+	})
+
+	ginkgo.It("controller should have invalid result when pool CIDR overlaps", func() {
+		// Disable webhooks because they share the same validation implementation (config.For())
```
karampok: I think we should remove the e2e that requires disabling the web hooks.

api/v1beta1/configurationstate_types.go
```diff
@@ -0,0 +1,86 @@
...
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// ConfigurationStateResult represents the validation result of a MetalLB configuration.
+// +kubebuilder:validation:Enum=Valid;Invalid;Unknown
+type ConfigurationStateResult string
+
+const (
+	// ConfigurationStateResultValid indicates configuration is successfully validated.
```
fedepaol: nit: I'd be less verbose and call it ConfigurationResult

internal/k8s/controllers/config_controller.go
fedepaol: yep, I think we should return error if we fail to write

internal/k8s/controllers/config_controller.go
```diff
@@ -254,3 +271,44 @@ func (r *ConfigReconciler) getSecrets(ctx context.Context) (map[string]corev1.Se
...
+		// Always set to now (tracks last reconciliation, not last status change).
+		LastTransitionTime: metav1.Now(),
+	}
+
+	if resultErr != nil {
+		condition.Status = metav1.ConditionFalse
+		condition.Message = resultErr.Error()
+	}
+
+	configStatus := &metallbv1beta1.ConfigurationState{
```
fedepaol, line 302: wouldn't it be better to have name / namespace as configreconciler fields here, and get before patching?
The reference feels a bit convoluted

internal/k8s/controllers/config_controller.go
```diff
@@ -254,3 +271,44 @@ func (r *ConfigReconciler) getSecrets(ctx context.Context) (map[string]corev1.Se
 	}
 	return secretsMap, nil
 }
+
+func (r *ConfigReconciler) reportCondition(ctx context.Context, resultErr error, errType string) error {
+	if r.ConfigStateRef == nil {
+		return nil
+	}
+
+	owner := "configReconciler"
```
fedepaol: either a constant or use the string directly below

internal/k8s/controllers/configuration_state_controller.go
```diff
@@ -0,0 +1,125 @@
...
+	return "ConfigurationStateReconciler"
+}
+
+func (r *ConfigurationStateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	level.Info(r.Logger).Log("controller", r, "start reconcile", req.String())
+	defer level.Info(r.Logger).Log("controller", r, "end reconcile", req.String())
+
+	var configStatus metallbv1beta1.ConfigurationState
+	if err := r.Get(ctx, req.NamespacedName, &configStatus); err != nil {
+		if !apierrors.IsNotFound(err) {
```
fedepaol: can you try to denest this by handling isNotFound then create, otherwise handle the error?

internal/k8s/controllers/configuration_state_controller.go
```diff
@@ -0,0 +1,125 @@
...
+			level.Error(r.Logger).Log("controller", r, "error", "failed to get ConfigurationState", "error", err)
+			return ctrl.Result{}, err
+		}
+
+		if err := r.Create(ctx, r.ConfigStateRef); err != nil {
```
fedepaol: same here, let's pass name and namespace to the reconciler

internal/k8s/controllers/configuration_state_controller.go
```diff
@@ -0,0 +1,125 @@
...
+	result := metallbv1beta1.ConfigurationStateResultUnknown
+	lastError := ""
+
+	// Only aggregate if there are conditions reported
+	if len(configStatus.Status.Conditions) > 0 {
+		result = metallbv1beta1.ConfigurationStateResultValid
+
+		var msgs []string
+		for _, cond := range configStatus.Status.Conditions {
+			if cond.Status == metav1.ConditionFalse {
```
fedepaol: this assumes that status = false is an error condition, which might change depending on the conditions we are adding. Would it make sense to have a "isConditionErrorType" filter function over the type where we check the current conditions we have? If the filter fails we continue

internal/k8s/controllers/configuration_state_controller.go
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	result := metallbv1beta1.ConfigurationStateResultUnknown
+	lastError := ""
+
+	// Only aggregate if there are conditions reported
+	if len(configStatus.Status.Conditions) > 0 {
+		result = metallbv1beta1.ConfigurationStateResultValid
+
+		var msgs []string
```
fedepaol: let's call this errorMessages

internal/k8s/controllers/configuration_state_controller.go
```diff
@@ -0,0 +1,125 @@
...
+		}
+
+		if len(msgs) > 0 {
+			result = metallbv1beta1.ConfigurationStateResultInvalid
+			lastError = strings.Join(msgs, "\n")
+		}
+	}
+
+	configStatus.Status.Result = result
+	configStatus.Status.LastError = lastError
```
fedepaol: this is not the last error though, it's a collection of errors.

internal/k8s/controllers/pool_controller.go
```diff
@@ -41,12 +43,22 @@ type PoolReconciler struct {
...
 }
 
 func (r *PoolReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
 	level.Info(r.Logger).Log("controller", "PoolReconciler", "start reconcile", req.String())
 	defer level.Info(r.Logger).Log("controller", "PoolReconciler", "end reconcile", req.String())
+
+	var resultErr error
+	errType := ErrorTypeNone
+	defer func() {
+		if err := r.reportCondition(ctx, resultErr, errType); err != nil {
```
fedepaol: same consideration about returning the error we made above

e2etest/configurationstatetests/configurationstate.go
```diff
@@ -0,0 +1,352 @@
+// SPDX-License-Identifier:Apache-2.0
+
+// Package configurationstatetests ...
```
fedepaol: ??

e2etest/configurationstatetests/configurationstate.go
fedepaol: Agree

internal/k8s/controllers/config_controller.go
karampok: done (for that I have to have named returns)

e2etest/configurationstatetests/configurationstate.go
karampok: done

api/v1beta1/configurationstate_types.go
karampok: done

internal/k8s/controllers/config_controller.go
karampok: I have name/namespace.
The get is not needed

internal/k8s/controllers/config_controller.go
karampok: done

internal/k8s/controllers/configuration_state_controller.go
karampok: probably done

internal/k8s/controllers/configuration_state_controller.go
karampok: not done (I will do)

internal/k8s/controllers/configuration_state_controller.go
karampok: done, changed the API

internal/k8s/controllers/pool_controller.go
karampok: done

e2etest/configurationstatetests/configurationstate.go
karampok: done (was there to avoid some gopls warnings, added proper sentence)

internal/k8s/controllers/config_controller.go
```diff
@@ -40,23 +43,33 @@ const bgpExtrasConfigName = "bgpextras"
...
-var requestHandler = func(r *ConfigReconciler, ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+var requestHandler = func(r *ConfigReconciler, ctx context.Context, req ctrl.Request) (result ctrl.Result, retError error) {
 	level.Info(r.Logger).Log("controller", "ConfigReconciler", "start reconcile", req.String())
 	defer level.Info(r.Logger).Log("controller", "ConfigReconciler", "end reconcile", req.String())
+
+	var conditionErr error
+	defer func() {
+		if err := r.reportCondition(ctx, conditionErr); err != nil {
+			level.Error(r.Logger).Log("controller", "ConfigReconciler", "message", "failed to report condition", "error", err)
+			retError = errors.Join(retError, err)
```
fedepaol: please let's add a comment here saying this is returned. Or let's call it handlerError. I am afraid this is hiding a bit the behavior.
Also, shouldn't we have naked returns? I think this is shadowed by the explicit returns we have

internal/k8s/controllers/configuration_state_controller.go
```diff
@@ -0,0 +1,133 @@
...
+	return "ConfigurationStateReconciler"
+}
+
+func (r *ConfigurationStateReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.Result, error) {
+	level.Info(r.Logger).Log("controller", r, "start reconcile", req.String())
+	defer level.Info(r.Logger).Log("controller", r, "end reconcile", req.String())
+
+	var configStatus metallbv1beta1.ConfigurationState
+	if err := r.Get(ctx, req.NamespacedName, &configStatus); err != nil {
+		if apierrors.IsNotFound(err) {
```
fedepaol: I think here we can go a level above like
err := r.Get{}
if apierrors.IsNotFound(err) {
}
if err != nil {
}

internal/k8s/controllers/configuration_state_controller.go
```diff
@@ -0,0 +1,133 @@
...
+		level.Error(r.Logger).Log("controller", r, "error", "failed to get ConfigurationState", "error", err)
+		return ctrl.Result{}, err
+	}
+
+	result := metallbv1beta1.ConfigurationResultUnknown
+	if len(configStatus.Status.Conditions) > 0 {
+		result = metallbv1beta1.ConfigurationResultValid
+	}
+
+	var msgs []string
```
fedepaol: nit (non blocking): I'd call this errorMessages

internal/k8s/controllers/values.go
```diff
@@ -35,4 +35,15 @@ const (
 	LabelServiceNamespace = "metallb.io/service-namespace"
 )
 
-var errRetry = errors.New("event handling failed, retrying")
+// Error types for condition reporting.
+const (
+	ErrorTypeConfiguration = "configurationError"
+	ErrorTypeInfra         = "infraError"
+	ErrorTypeNone          = "Reconciled"
```
fedepaol, line 42: let's keep all of them starting with capital or not

e2etest/configurationstatetests/configurationstate.go
```diff
@@ -0,0 +1,249 @@
...
+			return stateMatches(stateName, wantStatus)
+		}, 30*time.Second, 5*time.Second).Should(Succeed())
+
+		ginkgo.By("Recreating secret with correct type")
+		err = ConfigUpdater.Client().Delete(context.Background(), secret) // field is immutable, need to delete first
+		Expect(err).NotTo(HaveOccurred())
+		secret.ResourceVersion = ""
+		secret.UID = ""
+		secret.Type = corev1.SecretTypeBasicAuth
+		err = ConfigUpdater.Client().Create(context.Background(), secret)
```
fedepaol, line 133: I think if we don't ensure the secret is deleted, this might flake failing because teh secret is still being deleted

internal/k8s/controllers/config_controller.go
karampok: yes, I was also unsure about. I was investigating that, according to https://go.dev/ref/spec#Defer_statements
```
/ f returns 42
func f() (result int) {
	defer func() {
		// result is accessed after it was set to 6 by the return statement
		result *= 7
	}()
	return 6
}
```

Also in another part
```
A "return" statement that specifies results sets the result parameters before any deferred functions are executed.
```
AFAICT no shadowing  should take place, but I can try to rename retError to handlerError (or reconcilerError or resultError?)

internal/k8s/controllers/configuration_state_controller.go
karampok: Do you like the API name ErrorSummary?
```
	configStatus.Status.ErrorSummary = strings.Join(errorMessages, "\n")
```

internal/k8s/controllers/config_controller.go
karampok: done

internal/k8s/controllers/configuration_state_controller.go
karampok: that was what you meant by denest :).  done

internal/k8s/controllers/configuration_state_controller.go
karampok: done, but let me know you want to change the API also

internal/k8s/controllers/values.go
karampok: done

e2etest/configurationstatetests/configurationstate.go
karampok: done

internal/k8s/controllers/configuration_state_controller.go
fedepaol: yep it's fine

fedepaol:
LGTM, thanks

fedepaol:
@karampok one last ask, can you give the PR a better title?

karampok:
ci failure seems flake

fedepaol:
one last pass rebased with main and I'll merge

PR #2889, opened 2025-11-24
allocator: enforce no mutations of config

oribon:
The pool_controller was constantly reloading services even though
the actual config did not change (e.g events triggered by ns labels changing),
caused by reflect.DeepEqual returning false.
The flow goes:
  (1) pool_controller parses the config
  (2) compares with the current cached config, if they differ:
  (3) passes it to the allocator's SetPools
  (4) pool_controller caches the parsed config
Ideally, the config should not change between (1) and (4), but
in practice the allocator mutates the config when it runs ipaddr.NewPrefix(...),
making the parsed config deviate from the original config.
Here we fix this by creating a copy of the ip before running NewPrefix,
and add a unit test that ensures the allocator does not mutate the config.

/kind bug

**Release note**:

```release-note
bugfix: enforce allocator does not mutate the config, enabling the pool controller to not re-trigger redundant configuration.
```

### DIFF
--- a/internal/allocator/allocator.go
+++ b/internal/allocator/allocator.go
@@ -627,7 +627,8 @@ func poolCount(p *config.Pool) (int64, int64, int64) {
 		}
 		sz := int64(math.Pow(2, float64(b-o)))
 
-		cur := ipaddr.NewCursor([]ipaddr.Prefix{*ipaddr.NewPrefix(cidr)})
+		cidrCopy := copyCIDR(cidr)
+		cur := ipaddr.NewCursor([]ipaddr.Prefix{*ipaddr.NewPrefix(cidrCopy)})
 		firstIP := cur.First().IP
 		lastIP := cur.Last().IP
 
@@ -697,7 +698,8 @@ func (a *Allocator) getIPFromCIDR(cidr *net.IPNet, avoidBuggyIPs bool, svc strin
 		sharing: sharingKey,
 		backend: backendKey,
 	}
-	c := ipaddr.NewCursor([]ipaddr.Prefix{*ipaddr.NewPrefix(cidr)})
+	cidrCopy := copyCIDR(cidr)
+	c := ipaddr.NewCursor([]ipaddr.Prefix{*ipaddr.NewPrefix(cidrCopy)})
 	for pos := c.First(); pos != nil; pos = c.Next() {
 		if avoidBuggyIPs && ipConfusesBuggyFirmwares(pos.IP) {
 			continue
@@ -760,5 +762,18 @@ func (a *Allocator) updatePoolStats(p *config.Pool) {
 		AvailableIPv6: ipv6 - int64(len(a.poolIPV6InUse[p.Name])),
 		AssignedIPv4:  int64(len(a.poolIPV4InUse[p.Name])),
 		AssignedIPv6:  int64(len(a.poolIPV6InUse[p.Name])),
+	}
+}
+
+// copyCIDR creates a new copy of a CIDR, useful for passing to functions that
+// mutate it (e.g. ipaddr.NewPrefix).
+func copyCIDR(cidr *net.IPNet) *net.IPNet {
+	ip := make(net.IP, len(cidr.IP))
+	copy(ip, cidr.IP)
+	mask := make(net.IPMask, len(cidr.Mask))
+	copy(mask, cidr.Mask)
+	return &net.IPNet{
+		IP:   ip,
+		Mask: mask,
 	}
 }
--- a/internal/allocator/allocator_test.go
+++ b/internal/allocator/allocator_test.go
@@ -82,71 +82,77 @@ func noopCallback(_ string) {}
 
 func TestAssignment(t *testing.T) {
 	alloc := New(noopCallback)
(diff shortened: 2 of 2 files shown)


fedepaol:
Mind expanding the flow / sequence of actions here?
What's the resource being changed here? Where is the deepequal failing?
I had a hard time following it from your description.

oribon:
@fedepaol updated the commit message, lmk if it needs more clarifications

fedepaol:
so the labels were on the pool?

oribon:
which labels?

fedepaol:
> the pool_controller was constantly reloading services even though
> the config did not change (events triggered by ns labels changing),

so the trigger is namespace label changes?

oribon:
ah yes, but this is just an example (I guess there are other events that do not affect the actual config)

fedepaol:
> ah yes, but this is just an example (I guess there are other events that do not affect the actual config)

yes but the pool controller is triggered by pools. What I don't get is the e2e workflow of this

oribon:
```go
	return ctrl.NewControllerManagedBy(mgr).
		For(&metallbv1beta1.IPAddressPool{}).
		Watches(&metallbv1beta1.Community{}, &handler.EnqueueRequestForObject{}).
		Watches(&corev1.Namespace{}, &handler.EnqueueRequestForObject{}).
		WithEventFilter(p).
		Complete(r)
```
it gets triggered by any of these - a namespace can be targeted by a pool, then change its labels. (not sure though what community is doing there)

fedepaol:
> ```go
> 	return ctrl.NewControllerManagedBy(mgr).
> 		For(&metallbv1beta1.IPAddressPool{}).
> 		Watches(&metallbv1beta1.Community{}, &handler.EnqueueRequestForObject{}).
> 		Watches(&corev1.Namespace{}, &handler.EnqueueRequestForObject{}).
> 		WithEventFilter(p).
> 		Complete(r)
> ```
> 
> it gets triggered by any of these - a namespace can be targeted by a pool, then change its labels. (not sure though what community is doing there)

what I don't understand is you are saying we are in the pool controller, but the function you changed (getFreeIPsFromPool) is used when we assign, not when we convert a config

fedepaol:
and imo we should never go there if the config is not moving, because of 
```
	cfg, err := toConfig(resources, r.ValidateConfig)
	if err != nil {
		configStale.Set(1)
		level.Error(r.Logger).Log("controller", "PoolReconciler", "error", "failed to parse the configuration", "error", err)
		return ctrl.Result{}, nil
	}

	level.Debug(r.Logger).Log("controller", "PoolReconciler", "rendered config", dumpConfig(cfg))
	if reflect.DeepEqual(r.currentConfig, cfg) {
		level.Debug(r.Logger).Log("controller", "PoolReconciler", "event", "configuration did not change, ignoring")
		return ctrl.Result{}, nil
	}
```

oribon:
> what I don't understand is you are saying we are in the pool controller, but the function you changed (getFreeIPsFromPool) is used when we assign, not when we convert a config

right, but what I tried to explain in the commit message is that when we are in step 3 (allocator doing its stuff, running assign / unassign / poolCount) the config (which is a pointer with a bunch of pointer fields) is mutated by that prefix func, hence by the time we got to step 4 it is not the same config as it was in step 1 -> we are stuck in a loop where the parsed config is different than what is stored in the cache.

> and imo we should never go there if the config is not moving, because of

this is exactly the "puzzling" part of this bug, the config should not move UNLESS something between parsing and caching changes it.

internal/allocator/allocator.go
```diff
@@ -697,7 +697,7 @@ func (a *Allocator) getIPFromCIDR(cidr *net.IPNet, avoidBuggyIPs bool, svc strin
 		sharing: sharingKey,
 		backend: backendKey,
 	}
-	c := ipaddr.NewCursor([]ipaddr.Prefix{*ipaddr.NewPrefix(cidr)})
+	c := ipaddr.NewCursor([]ipaddr.Prefix{*prefixFor(cidr)})
```
fedepaol: Let's use a copyCIDR function instead of prefixFor, with a comment that newprefix mutates it

oribon: I'm not against changing, but the reason I went with this approach is CIDRs are abundant across the code but prefixes are only used here, so it felt cleaner to wrap them instead of having to think about "does the cidr need to be copied" everytime

fedepaol: yep, but your wrapped function basically copies plus returns a prefix, doing a bit more than advertised for no apparent reason. If we copy (with a copyCIDR function) it's more immediate to understand what's happening imo

oribon: > doing a bit more than advertised for no apparent reason

well, to me it seems it "creates a new ipaddr.Prefix from a CIDR without mutating it" as advertised 😅  changed to copyCIDR

fedepaol: can you make it in two separate lines? Copy creation and usage? :bow:

oribon: done

fedepaol:
LGTM

PR #2891, opened 2025-12-03
allocator: return error when pool has no available IPs

oribon:
We explicitly return an error when there is no IPs in the pool instead of relying on the caller to handle nil ips correctly. As a side-effect, the AllocationFailed event changes from: `Failed to allocate IP for "service-a": ["<nil>"] is not allowed in config for service` to:
`Failed to allocate IP for "service-a": no available IPs in pool x`

When service-a requests an IP from a pool that doesn't have any left.

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
allocation: fix the AllocationFailed event to indicate when the pool doesn't have any IPs left.
```

### DIFF
--- a/controller/controller_test.go
+++ b/controller/controller_test.go
@@ -1297,6 +1297,9 @@ func TestDeleteRecyclesIP(t *testing.T) {
 	if k.gotService(svc2) != nil {
 		t.Fatal("SetBalancer svc2 mutated svc2 even though it should not have allocated")
 	}
+	if !k.loggedWarning {
+		t.Fatal("SetBalancer svc2 should have produced a warning event when pool has no available IPs")
+	}
 	k.reset()
 
 	// Deleting the first LB should tell us to reprocess all services.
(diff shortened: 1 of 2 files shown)


karampok:
lgtm to the code, would it possible to add a unit test?

oribon:
@karampok the case for "can't get an ip from a pool that has no available ips" is already covered (e.g `s5 can't get an IPv4` in `allocator_test.go`), is there anything else you think is missing?

karampok:
> @karampok the case for "can't get an ip from a pool that has no available ips" is already covered (e.g `s5 can't get an IPv4` in `allocator_test.go`), is there anything else you think is missing?

I was more looking to see something that covers the event itself probably. Like if I remove the implementation a test will fail.

oribon:
@karampok added an assertion that an event is produced (although without checking its content, because we don't do that anywhere else afaik)

PR #2892, opened 2025-12-03
Add namespaceOverwrite to helm chart

1fabi0:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

This PR adds a value to the Helm chart that allows users to override the MetalLB namespace. This enables support for umbrella charts, allowing MetalLB to be installed in a different namespace than the parent chart.

If using a custom namespace, the `crds` and `frr-k8s` components must also be configured with the matching namespace overwrite.  
I'd like to document the requirement that `crds` and `frr-k8s` must match this custom namespace, but I'm not sure if this belongs in the _README.md_ of the helm chart, the _values.yaml_ comments, or elsewhere. Could you point me to the preferred location?  

Support for overriding the `frr-k8s` namespace is not included in this PR. I will open a seperate PR there.

**Release note**:

```release-note
Add namespaceOverwrite to helm chart
```

### DIFF
--- a/charts/metallb/charts/crds/templates/crds.yaml
+++ b/charts/metallb/charts/crds/templates/crds.yaml
@@ -338,7 +338,7 @@ spec:
       clientConfig:
         service:
           name: metallb-webhook-service
-          namespace: {{ include "crds.namespace" }}
+          namespace: {{ include "crds.namespace" . }}
           path: /convert
       conversionReviewVersions:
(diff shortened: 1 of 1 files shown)


1fabi0:
I will take care of the DCO after documentation 😅

Issue #2895, opened 2025-12-12
e2e tests for frr-k8s fail


andreaskaris:
### MetalLB Version

upstream

### Deployment method

Charts

### Main CNI

upstream kind

### Describe the bug

In https://github.com/metallb/frr-k8s/actions/runs/20138674053/job/57804252558?pr=373, the following E2E fails:
```
  [FAIL] BGP validate FRR running configuration [It] Full BFD profile
```
Also see https://github.com/metallb/frr-k8s/pull/374

the reason being that since:
https://github.com/metallb/frr-k8s/commit/80c504ca87dff7c0f7ff30048b74a6aa13e9867c#diff-6fd3f52e34bdc8290b0ff8ad87370e978a9b0347542d79f9fc197d162b0bd740L102-L300

logging is to stdout, but metallb e2e wants the log to be present:
https://github.com/metallb/metallb/blob/dc46c0f505a19971a3fd846c97c44fc45cccf69a/e2etest/bgptests/bgp.go#L1650

To run this manually:

in frr-k8s repo:
```
 make docker-build IMG=quay.io/metallb/frr-k8s:main
```

in frr-k8s repo:
```
          helm_args="\
          --set prometheus.serviceMonitor.metricRelabelings[0].sourceLabels=\{__name__\} \
          --set prometheus.serviceMonitor.metricRelabelings[0].regex=\"frrk8s_bgp_(.*)\" \
          --set prometheus.serviceMonitor.metricRelabelings[0].targetLabel=\"__name__\" \
          --set prometheus.serviceMonitor.metricRelabelings[0].replacement=\"metallb_bgp_\\\$\$1\" \
          --set prometheus.serviceMonitor.metricRelabelings[1].sourceLabels=\{__name__\} \
          --set prometheus.serviceMonitor.metricRelabelings[1].regex=\"frrk8s_bfd_(.*)\" \
          --set prometheus.serviceMonitor.metricRelabelings[1].targetLabel=\"__name__\" \
          --set prometheus.serviceMonitor.metricRelabelings[1].replacement=\"metallb_bfd_\\\$\$1\" \
          "
          KIND_CLUSTER_NAME="kind" NAMESPACE=metallb-system LOGLEVEL=debug IP_FAMILY="ipv4" HELM_ARGS=${helm_args} make "deploy-helm"
```

in metallb repo:
```
         cd charts/metallb
[...]

PR #2648, opened 2025-01-09
WIP: fix: change log output for frr container

mkleku81:
Proposal for a solution to the issue https://github.com/metallb/metallb/issues/2644
In the main branch I see exactly the same approach, frr logs can consume all node disk space and lead to a disk preassure state. So I recommend changing the place where the frr container logs are sent in speaker pod. Changing destination place frr logs from /etc/frr/frr.log to stdout will allow to natively manage the space of frr container logs, which will be available using 
`kubectl logs -c frr < speaker_pod_name >`
without any  workaround  like [this](https://github.com/metallb/metallb/blob/main/charts/metallb/templates/speaker.yaml#L390-L403).   

/kind bug

```release-note
Allow rotation for the logs of the frr container
```

PR #2896, opened 2025-12-12
E2E: Handle different frr log configurations between frr-k8s and metallb

andreaskaris:
The test for BFD profile configuration was failing with frr-k8s because it expects a different log statement format. frr-k8s uses "log stdout" while metallb's frr uses "log file /etc/frr/frr.log" as of commit 80c504ca87dff7c0f7ff30048b74a6aa13e9867c in the frr-k8s repository.

Initially, I wanted to fix this with an IsFRRK8s flag in the E2Es to distinguish between the two modes and checks for the appropriate log statement accordingly. See https://github.com/metallb/metallb/pull/2896/changes/cf723da7c62da1d81ba68e6473479494d3ea7ca6

However, given that upstream metallb runs the frr-k8s tests as well, but pinned to v0.0.21, we cannot do this, either.

Instead, use a simple `Or` to unblock the frr-k8s CIs until metallb and frr-k8s E2Es use the same version of frr-k8s with log stdout; and ideally until metallb's frr also uses log stdout

Fixes https://github.com/metallb/metallb/issues/2895

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
/kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note

```

### DIFF
--- a/e2etest/bgptests/bgp.go
+++ b/e2etest/bgptests/bgp.go
@@ -1647,7 +1647,10 @@ var _ = ginkgo.Describe("BGP", func() {
 					return cfgStr
 				}, 1*time.Minute).Should(
 					And(
-						ContainSubstring("log file /etc/frr/frr.log"),
+						Or(
+							ContainSubstring("log file /etc/frr/frr.log"),
+							ContainSubstring("log stdout"),
+						),
 						WithTransform(substringCount("\n profile fullbfdprofile1"), Equal(1)),
 						ContainSubstring("receive-interval 93"),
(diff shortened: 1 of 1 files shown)


oribon:
thanks! I think the right thing would be to have https://github.com/metallb/metallb/pull/2648 (as you suggested) instead of implementing a temporary thing, can you review it please?
We can release frr-k8s sometimes this week and bump in MetalLB

andreaskaris:
> can you review it please?

yep!

PR #2908, opened 2025-12-23
Simplify hasHealthyEndpoint logic and fix edge cases with reused IPs

tarabrind:

/kind bug

In environments like **KubeVirt**, virtual machines migrating between nodes can lead to multiple endpoints sharing the same IP address. For example, during migration, an old pod might stay in a `Completed` state on the source node while a new one is `Running` on the destination node, both sharing the same IP.

The previous implementation of `hasHealthyEndpoint` in `bgp_controller.go` was problematic because it tracked readiness via an intermediate map keyed by the endpoint's IP address (`ready := map[string]bool{}`). This led to redundant logic (readiness belongs to the Endpoint, not the IP) and potential fragility when IPs are reused across different pod lifecycles.

**Real-world Scenario:**

Here is an example of such a state from a live cluster:

**Pods:**

```shell
# kubectl get pods -o wide
NAME                                READY   STATUS      RESTARTS   AGE    IP             NODE           NOMINATED NODE   READINESS GATES
virt-launcher-ubuntu-vm-one-pp6sq   0/1     Completed   0          4d2h   10.88.10.7     virtlab-ap-1   <none>           1/1
virt-launcher-ubuntu-vm-one-sjxcw   1/1     Running     0          34s    10.88.10.7     virtlab-ap-2   <none>           1/1
```

**EndpointSlice entries:**

```yaml
# kubectl get endpointslices.discovery.k8s.io vm-ssh-xk6g6 -o yaml
addressType: IPv4
apiVersion: discovery.k8s.io/v1
endpoints:
- addresses:
  - 10.88.10.7
  conditions:
    ready: false
    serving: false
    terminating: false
  nodeName: virtlab-ap-1
  targetRef:
    kind: Pod
    name: virt-launcher-ubuntu-vm-one-pp6sq
    namespace: default
    uid: 938eea92-6b6c-4df2-9edb-452aa20d727b
- addresses:
  - 10.88.10.7
  conditions:
    ready: true
    serving: true
    terminating: false
  nodeName: virtlab-ap-2
  targetRef:
    kind: Pod
    name: virt-launcher-ubuntu-vm-one-sjxcw
    namespace: default
    uid: 10df3525-1eab-460b-b881-3187899de323
kind: EndpointSlice
...
```

This PR refactors `hasHealthyEndpoint` to use a simplified map-based approach that groups readiness by IP address. This ensures that:

- Health is evaluated correctly even when multiple **Endpoints** share the same IP (KubeVirt scenario).
- If at least one endpoint for a given IP is healthy, the IP is considered ready.
- Found and updated one test case in `TestBGPSpeakerEPSlices` ("Endpoint on our node has some unready ports") that was asserting an incorrect behavior for the KubeVirt scenario (it expected no announcement if ANY endpoint for an IP was unready).
- The logic is cleaner than the original version while being more robust against IP reuse than a simple "any-ready" check.

While this refactoring aligns the BGP controller's intent with the Layer2 controller's [`activeEndpointExists`](https://github.com/metallb/metallb/blob/main/speaker/layer2_controller.go), I intentionally kept a simplified map-based approach instead of a pure early return. This was done to:

1. Maintain compatibility with the existing BGP test suite (specifically `TestBGPSpeakerEPSlices`), which expects per-IP readiness evaluation.
2. Ensure that "Ready" endpoints correctly shadow "Unready" ones for the same IP (fixing the KubeVirt migration issue) while keeping the code structure familiar to the original implementation.

**Release note**:

```release-note
Simplify hasHealthyEndpoint logic in BGP controller to correctly handle endpoint health when IP addresses are reused (e.g., during KubeVirt migrations).
```

### DIFF
--- a/charts/metallb/README.md
+++ b/charts/metallb/README.md
@@ -61,6 +61,7 @@ Kubernetes: `>= 1.19.0-0`
 | controller.webhookMode | string | `"enabled"` |  |
 | crds.enabled | bool | `true` |  |
 | crds.validationFailurePolicy | string | `"Fail"` |  |
+| frr-k8s.prometheus.rbacProxy.repository | string | `"registry.k8s.io/kubebuilder/kube-rbac-proxy"` |  |
 | frrk8s.enabled | bool | `false` |  |
 | frrk8s.external | bool | `false` |  |
 | frrk8s.namespace | string | `""` |  |
@@ -104,7 +105,7 @@ Kubernetes: `>= 1.19.0-0`
 | prometheus.prometheusRule.staleConfig.labels.severity | string | `"warning"` |  |
 | prometheus.rbacPrometheus | bool | `true` |  |
 | prometheus.rbacProxy.pullPolicy | string | `nil` |  |
-| prometheus.rbacProxy.repository | string | `"gcr.io/kubebuilder/kube-rbac-proxy"` |  |
+| prometheus.rbacProxy.repository | string | `"registry.k8s.io/kubebuilder/kube-rbac-proxy"` |  |
 | prometheus.rbacProxy.tag | string | `"v0.12.0"` |  |
 | prometheus.scrapeAnnotations | bool | `false` |  |
 | prometheus.serviceAccount | string | `""` |  |
@@ -129,7 +130,7 @@ Kubernetes: `>= 1.19.0-0`
 | speaker.frr.enabled | bool | `true` |  |
 | speaker.frr.image.pullPolicy | string | `nil` |  |
 | speaker.frr.image.repository | string | `"quay.io/frrouting/frr"` |  |
-| speaker.frr.image.tag | string | `"10.4.1"` |  |
+| speaker.frr.image.tag | string | `"10.5.1"` |  |
 | speaker.frr.metricsPort | int | `7473` |  |
 | speaker.frr.resources | object | `{}` |  |
 | speaker.frrMetrics.resources | object | `{}` |  |
--- a/charts/metallb/values.yaml
+++ b/charts/metallb/values.yaml
@@ -55,7 +55,7 @@ prometheus:
 
   # the image to be used for the kuberbacproxy container
   rbacProxy:
-    repository: gcr.io/kubebuilder/kube-rbac-proxy
+    repository: registry.k8s.io/kubebuilder/kube-rbac-proxy
     tag: v0.12.0
     pullPolicy:
 
@@ -338,7 +338,7 @@ speaker:
     enabled: true
     image:
       repository: quay.io/frrouting/frr
-      tag: 10.4.1
+      tag: 10.5.1
       pullPolicy:
     metricsPort: 7473
     resources: {}
@@ -378,6 +378,11 @@ frrk8s:
   enabled: false
   external: false
   namespace: ""
+
+frr-k8s:
+  prometheus:
+    rbacProxy:
+      repository: registry.k8s.io/kubebuilder/kube-rbac-proxy
 
 # networkpolicies
 networkpolicies:
--- a/config/frr-k8s/kustomization.yaml
+++ b/config/frr-k8s/kustomization.yaml
@@ -16,3 +16,6 @@ patchesJson6902:
       kind: ClusterRole
       name: metallb-system:speaker
     path: clusterrole-patch.yaml
+images:
+  - name: gcr.io/kubebuilder/kube-rbac-proxy
+    newName: registry.k8s.io/kubebuilder/kube-rbac-proxy
--- a/config/frr/speaker-patch.yaml
+++ b/config/frr/speaker-patch.yaml
@@ -35,7 +35,7 @@ spec:
           securityContext:
             runAsUser: 100
             runAsGroup: 101
-          image: quay.io/frrouting/frr:10.4.1
+          image: quay.io/frrouting/frr:10.5.1
           command: ["/bin/sh", "-c", "cp -rLf /tmp/frr/* /etc/frr/"]
           volumeMounts:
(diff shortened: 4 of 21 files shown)


tarabrind:
I've pushed a second commit to fix the CI failures. The `TestBGPSpeakerEPSlices` suite had a case that expected no announcement if any endpoint for an IP was unready. I've updated the logic to ensure a healthy endpoint takes precedence (crucial for KubeVirt migrations) and adjusted the test case accordingly. PR description is updated with details.

tarabrind:
Hi @fedepaol, @oribon! Gentle ping on this one. Let me know if you need anything else from my side.

oribon:
sorry for the late reply!
I validated (together with ai) that what you propose aligns with k8s:
https://kubernetes.io/docs/concepts/services-networking/endpoint-slices/#duplicate-endpoints 
points to the [`EndpointSliceCache code within kube-proxy`](https://github.com/kubernetes/kubernetes/blob/master/pkg/proxy/endpointslicecache.go#L241-246) which says:
```
		// On the other hand, there maybe also be two *different* Endpoints (i.e.,
		// with different targetRefs) that point to the same IP, if the pod
		// network reuses the IP from a terminating pod before the Pod object is
		// fully deleted. In this case we want to prefer the running pod over the
		// terminating one. (If there are multiple non-terminating pods with the
		// same podIP, then the result is undefined.)
```
I haven't gone over the exact changes, but ccing @fedepaol to also give this a look as he has done most (or all) of the work regarding how we treat epslices

speaker/bgp_controller.go
```diff
@@ -146,13 +145,9 @@ func hasHealthyEndpoint(eps []discovery.EndpointSlice, filterNode func(*string)
 				continue
 			}
 			for _, addr := range ep.Addresses {
-				if _, ok := ready[addr]; !ok && epslices.EndpointCanServe(ep.Conditions) {
-					// Only set true if nothing else has expressed an
-					// opinion. This means that false will take precedence
-					// if there's any unready ports for a given endpoint.
+				if epslices.EndpointCanServe(ep.Conditions) {
```
fedepaol: this is hard to follow compared to returning early. The semantic here is, if at least one endpoint with this address is ready, then let's consider it ready.

So, I'd flip the logic to 

```go
if if _, ok := ready[addr]; ok { // continue, in case of multiple endpoint slices with the same address, one is enough
   continue
}

if epslices.EndpointCanServe() {
   ready[addr] = true
}
ready[addr] = false
```

This is way more explicit and easy to understand imo

fedepaol:
I'd add a unit test for this scenario.
Also, please describe why you are doing this change in the commit message.
Thanks!

speaker/bgp_controller.go
tarabrind: Thanks for the feedback! I've refactored `hasHealthyEndpoint` to correctly handle reused IPs as suggested.

tarabrind:
> I'd add a unit test for this scenario. Also, please describe why you are doing this change in the commit message. Thanks!

I also added a new unit test `TestHasHealthyEndpoint` covering these edge cases (duplicate IPs with mixed readiness states).

speaker/bgp_controller.go
```diff
@@ -146,22 +145,16 @@ func hasHealthyEndpoint(eps []discovery.EndpointSlice, filterNode func(*string)
...
-					// opinion. This means that false will take precedence
-					// if there's any unready ports for a given endpoint.
-					ready[addr] = true
-				}
-				if !epslices.EndpointCanServe(ep.Conditions) {
-					ready[addr] = false
+				if ready[addr] {
+					continue
 				}
+				ready[addr] = epslices.EndpointCanServe(ep.Conditions)
```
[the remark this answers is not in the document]
tarabrind: This is existing code and outside the scope of this PR, but I can fix it if maintainers prefer.

tarabrind:
"Hello @oribon! I've updated PR (fixed merge conflicts). It’s been open for a while, just wanted to make sure it hasn't fallen off the radar. Would appreciate a review when possible!"

oribon:
@tarabrind can you please squash the commits?

oribon:
@tarabrind can you please squash the commits?

tarabrind:
> @tarabrind can you please squash the commits?

Done! I've squashed the commits and force-pushed the changes.

speaker/bgp_controller.go
oribon, line 139: nit: sorry if I'm missing something, but can this function be simplified further to:

```go
func hasHealthyEndpoint(eps []discovery.EndpointSlice, filterNode func(*string) bool) bool {
    for _, slice := range eps {
        for _, ep := range slice.Endpoints {
            node := ep.NodeName
            if filterNode(node) {
                continue
            }
            if epslices.EndpointCanServe(ep.Conditions) {
                return true
            }
        }
    }
    return false
}
```
not sure I understand why the ready map is needed with this new behavior

tarabrind: > not sure I understand why the ready map is needed with this new behavior

Thanks for the suggestion! That makes perfect sense.
I used a version very similar to your code snippet. The only minor change I made was adding the check `len(ep.Addresses) > 0`:

```go
if len(ep.Addresses) > 0 && epslices.EndpointCanServe(ep.Conditions) {
    return true
}
```

In the previous implementation, the `for _, addr := range ep.Addresses` loop implicitly skipped all endpoints that had not yet been assigned IP addresses. By adding `len(ep.Addresses) > 0`, we ensure that we maintain exactly the same behavior and do not accidentally treat an endpoint without IP addresses as a valid target.

I squashed the commits.

oribon:
lgtm, thanks!

PR #2912, opened 2025-12-29
fix/documentation: bumping readme badges from 0.0.0 to latest of helm chart

cloudscape-germany:
/kind documentation

Chart and App version showed 0.0.0

Please evaluate if we can automate this (via Actions?)

**Release note**:

```release-note
Cosmetic Fix of readme badges from 0.0.0 to latest of helm chart
```


tylerauerbeck:
👍🏻 This LGTM. Although probably in the long run want to loop this into the release process so that this gets bumped whenever a new version is cut.

oribon:
thanks! I'd say at the very least we should document this in `release-process.md`, maybe even extending `gen_relnotes.sh` with the logic to bump the README

cloudscape-germany:
I'll take a look. Feel free to be faster.

oribon:
thanks! no rush :)

fedepaol:
I'd say teh right thing to do is to sed our way in inv release (here https://github.com/metallb/metallb/blob/main/tasks.py#L896) as we do for the other versions

cloudscape-germany:
> I'd say teh right thing to do is to sed our way in inv release (here https://github.com/metallb/metallb/blob/main/tasks.py#L896) as we do for the other versions

Thanks for this. Do that in a seperate PR?

cloudscape-germany:
@fedepaol Can you merge this and we open new issue for inv release feature or provide alternative way forward.

cloudscape-germany:
Badges still showing 0.0.0 on Public GitHub Page

PR #2917, opened 2026-01-06
L2/BGP Advertisements Service Selectors

oribon:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
Extend the advertisement resources with a ServiceSelector, allowing an advertisement to target only a set of services within its selected pools.
```

### DIFF
--- a/speaker/layer2_controller.go
+++ b/speaker/layer2_controller.go
@@ -137,7 +137,7 @@ func (c *layer2Controller) SetBalancer(l log.Logger, name string, lbIPs []net.IP
 	updateStatus := false
 	adsForService := l2AdsForService(pool.L2Advertisements, c.myNode, svc)
 	for _, lbIP := range lbIPs {
-		ipAdv := ipAdvertisementFor(lbIP, c.myNode, adsForService)
+		ipAdv := ipAdvertisementFor(lbIP, adsForService)
 		if !ipAdv.MatchInterfaces(ifs...) {
 			level.Warn(l).Log("op", "SetBalancer", "protocol", "layer2", "service", name, "IPAdvertisement", ipAdv,
 				"localIfs", ifs, "msg", "the specified interfaces used to announce LB IP don't exist")
@@ -180,12 +180,9 @@ func (c *layer2Controller) SetEventCallback(callback func(interface{})) {
 	// Do nothing
 }
 
-func ipAdvertisementFor(ip net.IP, localNode string, l2Advertisements []*config.L2Advertisement) layer2.IPAdvertisement {
+func ipAdvertisementFor(ip net.IP, l2Advertisements []*config.L2Advertisement) layer2.IPAdvertisement {
 	ifs := sets.Set[string]{}
 	for _, l2 := range l2Advertisements {
-		if matchNode := l2.Nodes[localNode]; !matchNode {
-			continue
-		}
 		if l2.AllInterfaces {
 			return layer2.NewIPAdvertisement(ip, true, sets.Set[string]{})
 		}
--- a/speaker/layer2_controller_test.go
+++ b/speaker/layer2_controller_test.go
@@ -1929,122 +1929,61 @@ func TestIPAdvertisementFor(t *testing.T) {
 	tests := []struct {
 		desc             string
 		ip               net.IP
-		localNode        string
 		l2Advertisements []*config.L2Advertisement
 		expect           layer2.IPAdvertisement
 	}{
 		{
-			desc:      "Not specify the IPPool related interfaces",
-			ip:        net.IP{192, 168, 10, 3},
-			localNode: "nodeA",
+			desc: "AllInterfaces set",
+			ip:   net.IP{192, 168, 10, 3},
 			l2Advertisements: []*config.L2Advertisement{
 				{
-					Nodes: map[string]bool{
-						"nodeB": true,
-						"nodeA": true,
-					},
 					Interfaces:    []string{},
 					AllInterfaces: true,
 				},
 			},
 			expect: layer2.NewIPAdvertisement(net.IP{192, 168, 10, 3}, true, sets.Set[string]{}),
-		}, {
-			desc:      "LocalNode doesn't match L2Advertisement",
-			ip:        net.IP{192, 168, 10, 3},
-			localNode: "nodeA",
+		},
+		{
+			desc: "Single L2Advertisement with interfaces",
+			ip:   net.ParseIP("2000:12"),
 			l2Advertisements: []*config.L2Advertisement{
 				{
-					Nodes: map[string]bool{
-						"nodeB": true,
-					},
-					Interfaces: []string{"eth0", "eth1"},
-				}, {
-					Nodes: map[string]bool{
-						"nodeA": false,
-						"nodeC": true,
-					},
-					Interfaces: []string{"eth3", "eth4"},
-				},
-			},
-			expect: layer2.NewIPAdvertisement(net.IP{192, 168, 10, 3}, false, sets.Set[string]{}),
-		}, {
-			desc:      "LocalNode only match one L2Advertisement",
-			ip:        net.ParseIP("2000:12"),
-			localNode: "nodeA",
-			l2Advertisements: []*config.L2Advertisement{
-				{
-					Nodes: map[string]bool{
-						"nodeB": true,
-						"nodeA": false,
-					},
-					Interfaces: []string{"eth0"},
-				}, {
-					Nodes: map[string]bool{
-						"nodeC": true,
-						"nodeA": true,
-					},
 					Interfaces: []string{"eth1", "eth2"},
 				},
 			},
 			expect: layer2.NewIPAdvertisement(net.ParseIP("2000:12"), false, sets.New("eth1", "eth2")),
-		}, {
-			desc:      "LocalNode match multi-L2Advertisement",
-			ip:        net.IP{192, 168, 10, 3},
-			localNode: "nodeA",
+		},
+		{
+			desc: "Multiple L2Advertisements - interfaces merged",
+			ip:   net.IP{192, 168, 10, 3},
 			l2Advertisements: []*config.L2Advertisement{
 				{
-					Nodes: map[string]bool{
-						"nodeB": true,
-						"nodeA": true,
-					},
 					Interfaces: []string{"eth0", "eth1"},
 				}, {
-					Nodes: map[string]bool{
-						"nodeC": true,
-						"nodeA": true,
-					},
 					Interfaces: []string{"eth0", "eth2"},
-				}, {
-					Nodes: map[string]bool{
-						"nodeC": true,
-						"nodeD": true,
-					},
-					Interfaces: []string{"eth4", "eth3"},
 				},
 			},
 			expect: layer2.NewIPAdvertisement(net.IP{192, 168, 10, 3}, false, sets.New("eth0", "eth1", "eth2")),
-		}, {
-			desc:      "LocalNode match multi-L2Advertisement, and one of them not specify interfaces",
-			ip:        net.IP{192, 168, 10, 3},
-			localNode: "nodeA",
+		},
+		{
+			desc: "Multiple L2Advertisements - one has AllInterfaces",
+			ip:   net.IP{192, 168, 10, 3},
 			l2Advertisements: []*config.L2Advertisement{
 				{
-					Nodes: map[string]bool{
-						"nodeB": true,
-						"nodeA": true,
-					},
 					Interfaces: []string{"eth0", "eth1"},
 				}, {
-					Nodes: map[string]bool{
-						"nodeC": true,
-						"nodeA": true,
-					},
 					AllInterfaces: true,
-				}, {
-					Nodes: map[string]bool{
-						"nodeC": true,
-						"nodeD": true,
-					},
-					Interfaces: []string{"eth4", "eth3"},
 				},
 			},
 			expect: layer2.NewIPAdvertisement(net.IP{192, 168, 10, 3}, true, sets.Set[string]{}),
 		},
 	}
 	for _, test := range tests {
-		r := ipAdvertisementFor(test.ip, test.localNode, test.l2Advertisements)
-		if !r.Equal(&test.expect) {
-			t.Errorf("%s: expect interfaces is %+v, but the result is %+v", test.desc, test.expect, r)
-		}
+		t.Run(test.desc, func(t *testing.T) {
+			r := ipAdvertisementFor(test.ip, test.l2Advertisements)
+			if !r.Equal(&test.expect) {
+				t.Errorf("expect %+v, but got %+v", test.expect, r)
+			}
+		})
 	}
 }


e2etest/bgptests/service_selector.go
```diff
@@ -0,0 +1,227 @@
...
+						Expect(found).To(BeFalse(), "Non-matching service IP %s should not be advertised to %s", ip.IP, c.Name)
+					}
+				}
+			}
+		},
+		ginkgo.Entry("IPV4", ipfamily.IPv4, "192.168.10.0/24"),
+		ginkgo.Entry("IPV6", ipfamily.IPv6, "fc00:f853:0ccd:e799::/116"),
+	)
+
+	ginkgo.DescribeTable("Multiple service selectors with OR logic",
```
karampok: AFAICT that can have only this and remove the first one, i think they test the same (the second one cannot be green, while the first is failing)

e2etest/bgptests/service_selector.go
```diff
@@ -0,0 +1,227 @@
...
+			ginkgo.By("Checking that other service is not advertised")
+			for _, c := range FRRContainers {
+				frrRoutesV4, frrRoutesV6, err := frr.Routes(c)
+				Expect(err).NotTo(HaveOccurred())
+
+				for _, ip := range otherSvc.Status.LoadBalancer.Ingress {
+					if pairingIPFamily == ipfamily.IPv4 {
+						_, found := frrRoutesV4[ip.IP]
+						Expect(found).To(BeFalse(), "Other service IP %s should not be advertised to %s", ip.IP, c.Name)
+					} else {
```
karampok: nit: not having else

e2etest/bgptests/service_selector.go
```diff
@@ -0,0 +1,227 @@
...
+				})
+			defer testservice.Delete(cs, backendSvc)
+
+			otherSvc, _ := testservice.CreateWithBackend(cs, testNamespace, "other-svc",
+				func(s *corev1.Service) {
+					s.Labels = map[string]string{"tier": "other"}
+				})
+			defer testservice.Delete(cs, otherSvc)
+
+			allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
```
karampok: nit: allNodes can go to before each

api/v1beta1/bgpadvertisement_types.go
```diff
@@ -64,6 +64,13 @@ type BGPAdvertisementSpec struct {
 	// When empty, the loadbalancer IP is announced to all the BGPPeers configured.
 	// +optional
 	Peers []string `json:"peers,omitempty"`
+
+	// ServiceSelectors limits the set of services that will be advertised via this advertisement.
```
karampok, line 69: 1. Can you expand why mutually exclusive?
2.  Did you consider having a CEL check here so user will not be able to add a non valid CR? ( Is CEL better/worse than doing webhook validation?)
3. Should we  also have documented the OR logic?

speaker/bgp_controller.go
```diff
@@ -517,6 +525,49 @@ func poolMatchesNodeBGP(pool *config.Pool, node string) bool {
...
+	for _, ad := range ads {
+		if !ad.Nodes[node] {
+			continue
+		}
+		// Empty selectors = match all services
+		if len(ad.ServiceSelectors) == 0 {
+			result = append(result, ad)
+			continue
+		}
+		// Check if any selector matches (OR logic)
```
karampok, line 541: Do we need OR logic?
Can we not have in the label logic
```
serviceSelectors:
  - matchExpressions:
    - key: env
      operator: In
      values: [prod, staging]
```

internal/config/config.go
```diff
@@ -1062,11 +1105,45 @@ func containsAdvertisement(advs []*L2Advertisement, toCheck *L2Advertisement) bo
...
 		}
+		if !selectorsEqual(adv.ServiceSelectors, toCheck.ServiceSelectors) {
+			continue
+		}
 		return true
 	}
 	return false
 }
 
+// selectorsEqual compares two slices of selectors, assuming both are sorted.
```
karampok: Can we use  a lib?
```
(asking CLAUDE)
  import "k8s.io/apimachinery/pkg/api/equality"

  func selectorsEqual(a, b []metav1.LabelSelector) bool {
      return equality.Semantic.DeepEqual(a, b)
  }

  This works on the original []metav1.LabelSelector slices directly, avoiding:
  - Conversion to []labels.Selector
  - Manual sorting
  - String comparison iteration


```

internal/config/config.go
```diff
@@ -889,21 +922,31 @@ func validateBGPAdvPerPool(adv *BGPAdvertisement, pool *Pool) error {
...
 	}
 
 	return nil
 }
 
-func advertisementsAreCompatible(newAdv, adv *BGPAdvertisement, pool *Pool) bool {
+// BGPAdvertisementsHaveCompatibleLocalPref checks if two BGP advertisements have compatible local preferences,
+// assuming both target at least one shared service.
+func BGPAdvertisementsHaveCompatibleLocalPref(newAdv, adv *BGPAdvertisement, pool *Pool) bool {
+	if newAdv.LocalPref == adv.LocalPref {
```
andreaskaris, line 948: nit: Is this related to the service selectors? Or should this be in a different commit that explains why the LocalPref equality was missing before and what this fixes?

internal/config/config_test.go
```diff
@@ -3825,3 +3825,396 @@ func FuzzParseCIDR(f *testing.F) {
...
+						Spec: v1beta1.IPAddressPoolSpec{
+							Addresses: []string{"1.2.3.0/24"},
+						},
+					},
+				},
+				BGPAdvs: []v1beta1.BGPAdvertisement{
+					{
+						ObjectMeta: metav1.ObjectMeta{Name: "adv1"},
+						Spec: v1beta1.BGPAdvertisementSpec{
+							AggregationLength: ptr.To[int32](32),
```
andreaskaris, line 4126: add another test  for 						
```
AggregationLengthV6: ptr.To[int32](128),
```

Or just test for both values explicitly set in this test:
```
$ git diff
diff --git a/internal/config/config_test.go b/internal/config/config_test.go
index 606689d39..c0ca986e3 100644
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -4040,7 +4040,8 @@ func TestServiceSelectors(t *testing.T) {
                                        {
                                                ObjectMeta: metav1.ObjectMeta{Name: "adv1"},
                                                Spec: v1beta1.BGPAdvertisementSpec{
-                                                       AggregationLength: ptr.To[int32](32),
+                                                       AggregationLength:   ptr.To[int32](32),
+                                                       AggregationLengthV6: ptr.To[int32](128),
                                                        ServiceSelectors: []metav1.LabelSelector{
                                                                {
                                                                        MatchLabels: map[string]string{"app": "nginx"},
```

internal/config/config_test.go
```diff
@@ -3825,3 +3825,396 @@ func FuzzParseCIDR(f *testing.F) {
 		_, _ = ParseCIDR(input)
 	})
 }
+
+func TestServiceSelectors(t *testing.T) {
+	tests := []struct {
+		desc    string
+		crs     ClusterResources
+		want    *Config
+		wantErr bool
```
andreaskaris: nit: I'd test for substrings or a regex here matching the actual error string. wantErr true / false doesn't check which error was hit

internal/config/config_test.go
```diff
@@ -3825,3 +3825,396 @@ func FuzzParseCIDR(f *testing.F) {
...
+			})
+
+			if diff := cmp.Diff(tt.want, got, selectorComparer, cidrPerAddressComparer, cmp.AllowUnexported(Pool{})); diff != "" {
+				t.Errorf("parse returned wrong result (-want, +got)\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestContainsAdvertisementWithServiceSelectors(t *testing.T) {
```
andreaskaris: nit: There's already a "TestContainsAdvertisement".  Wouldn't it be better to simply test for service selector stuff in there, as additional test cases?

internal/config/config_test.go
```diff
@@ -3825,3 +3825,396 @@ func FuzzParseCIDR(f *testing.F) {
...
+					ServiceSelectors: []labels.Selector{selector("app=apache"), selector("app=nginx")},
+				},
+			},
+			toCheck: &L2Advertisement{
+				Nodes:            map[string]bool{"nodeA": true},
+				AllInterfaces:    true,
+				ServiceSelectors: []labels.Selector{selector("app=apache"), selector("app=nginx")},
+			},
+			expect: true,
+		},
```
andreaskaris: Add another test case which tests for order (and documents this behavior at the same time):
```
@@ -4208,6 +4209,22 @@ func TestContainsAdvertisementWithServiceSelectors(t *testing.T) {
                        },
                        expect: true,
                },
+               {
+                       desc: "Multiple selectors in different order but with same selectors should not match",
+                       advs: []*L2Advertisement{
+                               {
+                                       Nodes:            map[string]bool{"nodeA": true},
+                                       AllInterfaces:    true,
+                                       ServiceSelectors: []labels.Selector{selector("app=apache"), selector("app=nginx")},
+                               },
+                       },
+                       toCheck: &L2Advertisement{
+                               Nodes:            map[string]bool{"nodeA": true},
+                               AllInterfaces:    true,
+                               ServiceSelectors: []labels.Selector{selector("app=nginx"), selector("app=apache")},
+                       },
+                       expect: false,
+               },
        }
        for _, test := range tests {
                t.Run(test.desc, func(t *testing.T) {
```

speaker/bgp_controller.go
```diff
@@ -174,6 +174,17 @@ func (c *bgpController) ShouldAnnounce(l log.Logger, name string, _ []net.IP, po
 		return "notOwner"
 	}
 
+	if !poolHasBGPAdvForService(pool, c.myNode, svc) {
```
andreaskaris: that's redundant:
```
// poolHasBGPAdvForService checks if pool has any advertisement matching node and service.
func poolHasBGPAdvForService(pool *config.Pool, node string, svc *v1.Service) bool {
	return len(bgpAdsForService(pool.BGPAdvertisements, node, svc)) > 0
}
```

Just takes the len()  of bgpAdsForService. Then, in line 182, the same bgpAdsForService is called again.

This here is exactly the same, just with a single call, isn't it?
```
[akaris@metallb speaker]$ git diff bgp_controller.go
diff --git a/speaker/bgp_controller.go b/speaker/bgp_controller.go
index 179cf9ca5..bbaf33463 100644
--- a/speaker/bgp_controller.go
+++ b/speaker/bgp_controller.go
@@ -174,12 +174,12 @@ func (c *bgpController) ShouldAnnounce(l log.Logger, name string, _ []net.IP, po
                return "notOwner"
        }
 
-       if !poolHasBGPAdvForService(pool, c.myNode, svc) {
+       adsForService := bgpAdsForService(pool.BGPAdvertisements, c.myNode, svc)
+       if len(bgpAdsForService(pool.BGPAdvertisements, c.myNode, svc)) < 1 {
                level.Debug(l).Log("event", "skipping should announce bgp", "service", name, "reason", "no advertisement matching service")
                return "noMatchingAdvertisement"
        }
 
-       adsForService := bgpAdsForService(pool.BGPAdvertisements, c.myNode, svc)
        if err := checkBGPAdvConflicts(adsForService, pool); err != nil {
                level.Error(l).Log("event", "conflictingAdvertisements", "service", name, "error", err)
                return "conflictingAdvertisements"
@@ -549,11 +549,6 @@ func bgpAdsForService(ads []*config.BGPAdvertisement, node string, svc *v1.Servi
        return result
 }
 
-// poolHasBGPAdvForService checks if pool has any advertisement matching node and service.
-func poolHasBGPAdvForService(pool *config.Pool, node string, svc *v1.Service) bool {
-       return len(bgpAdsForService(pool.BGPAdvertisements, node, svc)) > 0
-}
-
 // checkBGPAdvConflicts checks if the bgp ads for the same service are compatible.
 // This is necessary because we don't run the validation per service when parsing the config.
 func checkBGPAdvConflicts(ads []*config.BGPAdvertisement, pool *config.Pool) error {
```

speaker/bgp_controller_test.go
```diff
@@ -1940,3 +1941,403 @@ func TestPeersForService(t *testing.T) {
...
 	}
 }
+
+func TestBGPAdsForService(t *testing.T) {
+	tests := []struct {
+		desc     string
+		ads      []*config.BGPAdvertisement
+		node     string
+		svc      *v1.Service
+		expected int
```
andreaskaris: nit: wondering if there's any benefit of checking for the actual returned advertisements, instead of just for the length, though I guess expected count is fine

speaker/bgp_controller_test.go
```diff
@@ -1940,3 +1941,403 @@ func TestPeersForService(t *testing.T) {
...
+		t.Run(test.desc, func(t *testing.T) {
+			result := bgpAdsForService(test.ads, test.node, test.svc)
+			if len(result) != test.expected {
+				t.Errorf("expected %d ads, got %d", test.expected, len(result))
+			}
+		})
+	}
+}
+
+func TestCheckBGPAdvConflicts(t *testing.T) {
```
andreaskaris, line 2033: nit: I found this logic confusing, so I'd add a few more test cases --- up to you, though:
```
[akaris@metallb speaker]$ git diff -- bgp_controller_test.go
diff --git a/speaker/bgp_controller_test.go b/speaker/bgp_controller_test.go
index f50edd6cb..cd820faac 100644
--- a/speaker/bgp_controller_test.go
+++ b/speaker/bgp_controller_test.go
@@ -2058,6 +2058,38 @@ func TestCheckBGPAdvConflicts(t *testing.T) {
                        },
                        expectError: false,
                },
+               {
+                       desc: "Conflict - ad1's nodes are a subset of ad2's nodes",
+                       ads: []*config.BGPAdvertisement{
+                               {Name: "ad1", AggregationLength: 32, LocalPref: 100, Nodes: map[string]bool{"nodeA": true}},
+                               {Name: "ad2", AggregationLength: 32, LocalPref: 200, Nodes: map[string]bool{"nodeB": true, "nodeA": true}},
+                       },
+                       expectError: true,
+               },
+               {
+                       desc: "Conflict - ad2's nodes are a subset of ad1's nodes",
+                       ads: []*config.BGPAdvertisement{
+                               {Name: "ad1", AggregationLength: 32, LocalPref: 100, Nodes: map[string]bool{"nodeB": true, "nodeA": true}},
+                               {Name: "ad2", AggregationLength: 32, LocalPref: 200, Nodes: map[string]bool{"nodeA": true}},
+                       },
+                       expectError: true,
+               },
+               {
+                       desc: "No conflict - ad1's nodes are empty",
+                       ads: []*config.BGPAdvertisement{
+                               {Name: "ad1", AggregationLength: 32, LocalPref: 100, Nodes: map[string]bool{}},
+                               {Name: "ad2", AggregationLength: 32, LocalPref: 200, Nodes: map[string]bool{"nodeA": true}},
+                       },
+                       expectError: false,
+               },
+               {
+                       desc: "No conflict - ad2's nodes are empty",
+                       ads: []*config.BGPAdvertisement{
+                               {Name: "ad1", AggregationLength: 32, LocalPref: 100, Nodes: map[string]bool{"nodeA": true}},
+                               {Name: "ad2", AggregationLength: 32, LocalPref: 200, Nodes: map[string]bool{}},
+                       },
+                       expectError: false,
+               },
                {
                        desc: "No conflict - different peers",
                        ads: []*config.BGPAdvertisement{
@@ -2086,6 +2118,54 @@ func TestCheckBGPAdvConflicts(t *testing.T) {
                        },
                        expectError: true,
                },
+               {
+                       desc: "Conflict - ad1's peers are a subset of ad2's peers",
+                       ads: []*config.BGPAdvertisement{
+                               {Name: "ad1", AggregationLength: 32, LocalPref: 100, Nodes: map[string]bool{"nodeA": true}, Peers: []string{"peer1"}},
+                               {Name: "ad2", AggregationLength: 32, LocalPref: 200, Nodes: map[string]bool{"nodeA": true}, Peers: []string{"peer2", "peer1"}},
+                       },
+                       expectError: true,
+               },
+               {
+                       desc: "Conflict - ad2's peers are a subset of ad1's peers",
+                       ads: []*config.BGPAdvertisement{
+                               {Name: "ad1", AggregationLength: 32, LocalPref: 100, Nodes: map[string]bool{"nodeA": true}, Peers: []string{"peer2", "peer1"}},
+                               {Name: "ad2", AggregationLength: 32, LocalPref: 200, Nodes: map[string]bool{"nodeA": true}, Peers: []string{"peer1"}},
+                       },
+                       expectError: true,
+               },
+               {
+                       desc: "Conflict - ad1's peers are a subset of ad2's peers, but ad1 also has one extra element",
+                       ads: []*config.BGPAdvertisement{
+                               {Name: "ad1", AggregationLength: 32, LocalPref: 100, Nodes: map[string]bool{"nodeA": true}, Peers: []string{"peer1", "peer3"}},
+                               {Name: "ad2", AggregationLength: 32, LocalPref: 200, Nodes: map[string]bool{"nodeA": true}, Peers: []string{"peer2", "peer1"}},
+                       },
+                       expectError: true,
+               },
+               {
+                       desc: "Conflict - ad2's peers are a subset of ad1's peers, but ad2 also has one extra element",
+                       ads: []*config.BGPAdvertisement{
+                               {Name: "ad1", AggregationLength: 32, LocalPref: 100, Nodes: map[string]bool{"nodeA": true}, Peers: []string{"peer2", "peer1"}},
+                               {Name: "ad2", AggregationLength: 32, LocalPref: 200, Nodes: map[string]bool{"nodeA": true}, Peers: []string{"peer1", "peer3"}},
+                       },
+                       expectError: true,
+               },
+               {
+                       desc: "Conflict - ad1's peers are empty",
+                       ads: []*config.BGPAdvertisement{
+                               {Name: "ad1", AggregationLength: 32, LocalPref: 100, Nodes: map[string]bool{"nodeA": true}, Peers: []string{}},
+                               {Name: "ad2", AggregationLength: 32, LocalPref: 200, Nodes: map[string]bool{"nodeA": true}, Peers: []string{"peer1"}},
+                       },
+                       expectError: true,
+               },
+               {
+                       desc: "Conflict - ad2's peers are empty",
+                       ads: []*config.BGPAdvertisement{
+                               {Name: "ad1", AggregationLength: 32, LocalPref: 100, Nodes: map[string]bool{"nodeA": true}, Peers: []string{"peer1"}},
+                               {Name: "ad2", AggregationLength: 32, LocalPref: 200, Nodes: map[string]bool{"nodeA": true}, Peers: []string{}},
+                       },
+                       expectError: true,
+               },
        }
 
        for _, test := range tests {
[akaris@metallb speaker]$ 
```

speaker/layer2_controller.go
```diff
@@ -90,6 +91,11 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce
 		return "notOwner"
 	}
 
+	if !poolHasL2AdvForService(pool, c.myNode, svc) {
```
andreaskaris: nit: makes more sense here (as opposed to the bgp part), but you don't need the extra function:
```
[akaris@metallb speaker]$ git diff layer2_controller.go
index 4e2dd3070..9d9d7bbb8 100644
@@ -91,7 +91,7 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce
                return "notOwner"
        }
 
-       if !poolHasL2AdvForService(pool, c.myNode, svc) {
+       if len(l2AdsForService(pool.L2Advertisements, c.myNode, svc)) < 1 {
                level.Debug(l).Log("event", "skipping should announce l2", "service", name, "reason", "no advertisement matching service on my node")
                return "noMatchingAdvertisement"
        }
@@ -247,11 +247,6 @@ func l2AdsForService(ads []*config.L2Advertisement, node string, svc *v1.Service
        return result
 }
 
-// poolHasL2AdvertisementForService checks if pool has any advertisement matching node and service.
-func poolHasL2AdvForService(pool *config.Pool, node string, svc *v1.Service) bool {
-       return len(l2AdsForService(pool.L2Advertisements, node, svc)) > 0
-}
-
 func (c *layer2Controller) speakersForPool(l log.Logger, name string, pool *config.Pool, nodes map[string]*v1.Node) map[string]bool {
        sl := c.sList.UsableSpeakers()
        eligibleNodes := maps.Keys(sl.Nodes)
[akaris@metallb speaker]$ 
```

speaker/layer2_controller_test.go
```diff
@@ -1698,6 +1699,233 @@ func TestClusterPolicy(t *testing.T) {
...
+	}
+}
+
+func TestL2AdsForService(t *testing.T) {
+	tests := []struct {
+		desc     string
+		ads      []*config.L2Advertisement
+		node     string
+		svc      *v1.Service
+		expected int
```
andreaskaris: nit: same question as above - does it make sense to check for advertisement, or is the expected count enough? I'm fine with either

e2etest/bgptests/service_selector.go
```diff
@@ -0,0 +1,227 @@
...
+			for _, c := range FRRContainers {
+				validateService(frontendSvc, allNodes.Items, c)
+			}
+
+			ginkgo.By("Checking that backend service is advertised")
+			for _, c := range FRRContainers {
+				validateService(backendSvc, allNodes.Items, c)
+			}
+
+			ginkgo.By("Checking that other service is not advertised")
```
andreaskaris, line 252: nit: preferably should test for Consistently, or not? Otherwise, you might check, and then it's created right after

e2etest/bgptests/service_selector.go
```diff
@@ -0,0 +1,227 @@
...
+
+			allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			ginkgo.By("Checking that matching service is advertised")
+			for _, c := range FRRContainers {
+				validateService(matchingSvc, allNodes.Items, c)
+			}
+
+			ginkgo.By("Checking that non-matching service is not advertised")
```
andreaskaris: nit: check with consistently

e2etest/bgptests/service_selector.go
```diff
@@ -0,0 +1,227 @@
...
+						_, found := frrRoutesV4[ip.IP]
+						Expect(found).To(BeFalse(), "Non-matching service IP %s should not be advertised to %s", ip.IP, c.Name)
+					} else {
+						_, found := frrRoutesV6[ip.IP]
+						Expect(found).To(BeFalse(), "Non-matching service IP %s should not be advertised to %s", ip.IP, c.Name)
+					}
+				}
+			}
+		},
+		ginkgo.Entry("IPV4", ipfamily.IPv4, "192.168.10.0/24"),
```
andreaskaris: can you test for the cluster's IP family, and skip if one or the other isn't supported? --> spawn a single stack cluster and run the test manually, it'll fail due to the missing ip af

e2etest/l2tests/service_selector.go
```diff
@@ -0,0 +1,224 @@
...
+		testNamespace, err = k8s.CreateTestNamespace(cs, "l2svcsel")
+		Expect(err).NotTo(HaveOccurred())
+
+		ginkgo.By("Clearing any previous configuration")
+
+		err = ConfigUpdater.Clean()
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	ginkgo.Context("Service Selector", func() {
```
andreaskaris, line 184: the same as what Konstantinos said for bgp, this may be redundant

e2etest/l2tests/service_selector.go
```diff
@@ -0,0 +1,224 @@
...
+			allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			Eventually(func() error {
+				_, err := nodeForService(svc, allNodes.Items)
+				return err
+			}, 30*time.Second, 1*time.Second).Should(BeNil(), "Service with matching label should be advertised")
+		})
+
+		ginkgo.It("should not advertise service not matching the selector", func() {
```
andreaskaris: the same as what Konstantinos said for bgp, this may be redundant

e2etest/bgptests/service_selector.go
oribon: done

e2etest/bgptests/service_selector.go
oribon: done

e2etest/bgptests/service_selector.go
oribon: done

api/v1beta1/bgpadvertisement_types.go
oribon: 1. added a couple of words under `website`, the gist is that otherwise when you use aggregation length you're advertising a "range of ips", then services that don't match the selector are still "advertised" (even though not directly).
2. added (and validated it works)
3. done

speaker/bgp_controller.go
oribon: that might be enough, but I went with this approach to be aligned with the other `*Selectors` fields.

internal/config/config.go
oribon: done

internal/config/config.go
oribon: this refactors (and exports) the existing behavior, where we make sure that a prefix isn't advertised with 2 different localprefs to the same peer. the relation to selectors is that now we also want to do the same validation on a per-service level as well, whereas before it was per-pool (since advs targeted pools)

internal/config/config_test.go
oribon: done

internal/config/config_test.go
oribon: done

internal/config/config_test.go
oribon: done

internal/config/config_test.go
oribon: done

speaker/bgp_controller.go
oribon: done

speaker/bgp_controller_test.go
oribon: done

speaker/bgp_controller_test.go
oribon: done

speaker/layer2_controller.go
oribon: done

speaker/layer2_controller_test.go
oribon: done

e2etest/bgptests/service_selector.go
oribon: done

e2etest/bgptests/service_selector.go
oribon: done

e2etest/bgptests/service_selector.go
oribon: this is the same as in the rest of the tests, we let the caller (of the suite) decide which tests and avoid doing "dynamic" discoveries (which can be fragile)

e2etest/l2tests/service_selector.go
oribon: done

e2etest/l2tests/service_selector.go
oribon: done

website/content/configuration/_advanced_bgp_configuration.md
```diff
@@ -258,6 +258,60 @@ spec:
...
+spec:
+  ipAddressPools:
+  - first-pool
+  localPref: 100
+  serviceSelectors:
+  - matchLabels:
+      environment: staging
+```
+
+With this configuration, services labeled `environment: production` that receieved an IP from `first-pool` will be advertised
```
andreaskaris: nit: typo: receieved

andreaskaris:
/lgtm
thanks for addressing all of the nit picks as well

api/v1beta1/bgpadvertisement_types.go
karampok: > validated it works

like manually or there is test?

So now we have this CEL and at the same the webhook validation logic? Is there a scenario we can hit the second one? if no should we keep the code?
(probably yes let's be conservative nothing to loose). The risk is  someone updates code validation but forgets CEL or other way

speaker/bgp_controller.go
karampok: ok

e2etest/bgptests/service_selector.go
karampok: nit: "with OR logic is not needed" because user expect to find another test saying "without OR"

api/v1beta1/bgpadvertisement_types.go
oribon: > like manually or there is test?

manually, but added an e2e under webhooks. are you familiar with a way to test cel nicely?

> So now we have this CEL and at the same the webhook validation logic? Is there a scenario we can hit the second one? if no should we keep the code?
> (probably yes let's be conservative nothing to loose). The risk is someone updates code validation but forgets CEL or other way

it's the same as we do with all other fields, we always validate when parsing the config (and the webhooks come for free)

e2etest/bgptests/service_selector.go
oribon: done

api/v1beta1/bgpadvertisement_types.go
karampok: I am not, but that topic is in my todo.

karampok (approved):

api/v1beta1/bgpadvertisement_types.go
```diff
@@ -64,6 +65,13 @@ type BGPAdvertisementSpec struct {
...
+	// If empty, all services from the selected pools are advertised.
+	// This field is mutually exclusive with aggregationLength and aggregationLengthV6 -
+	// services can only be selected when using the default /32 (IPv4) or /128 (IPv6) aggregation.
+	// +optional
+	ServiceSelectors []metav1.LabelSelector `json:"serviceSelectors,omitempty"`
```
fedepaol, line 74: how about making this mutually exclusive with pool selector? I can't think about a reason why we'd need both (actually, I think this is more powerful but the other one needs to stay because of retro compatibility)

speaker/layer2_controller.go
```diff
@@ -216,6 +225,30 @@ func poolMatchesNodeL2(pool *config.Pool, node string) bool {
 	return false
 }
 
+// l2AdvsForService returns the L2 advertisements matching the service on the given node.
+func l2AdsForService(ads []*config.L2Advertisement, node string, svc *v1.Service) []*config.L2Advertisement {
+	var result []*config.L2Advertisement
+	svcLabels := labels.Set(svc.Labels)
+	for _, ad := range ads {
+		if !ad.Nodes[node] {
```
fedepaol, line 230: why we do have the node here? It should be covered somewhere else I guess, otherwise node selector would not work today

fedepaol: yep it's done in ipAdvertisementFor

speaker/layer2_controller_test.go
```diff
@@ -1698,6 +1699,232 @@ func TestClusterPolicy(t *testing.T) {
 	}
 }
 
+func selector(s string) labels.Selector {
```
fedepaol: this can go below

internal/config/config.go
```diff
@@ -889,21 +923,31 @@ func validateBGPAdvPerPool(adv *BGPAdvertisement, pool *Pool) error {
 		}
 	}
 
-	// Verify that BGP ADVs set a unique local preference value per BGP update.
+	// Verify that BGP ADVs set a unique local preference value per BGP update when they apply to the entire pool.
```
fedepaol: let's flip this to "Do not verify local pref uniqueness blablabla when service selector is set because we don't know if the bgp advertisements apply to the same selector.

Nit: in theory we could check if the selector is exactly the same, which is better than nothing.

e2etest/bgptests/service_selector.go
```diff
@@ -0,0 +1,156 @@
...
+		cs = k8sclient.New()
+		testNamespace, err = k8s.CreateTestNamespace(cs, "bgpsvcsel")
+		Expect(err).NotTo(HaveOccurred())
+
+		nodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+		Expect(err).NotTo(HaveOccurred())
+		allNodes = nodes.Items
+	})
+
+	ginkgo.DescribeTable("Multiple service selectors",
```
fedepaol, line 188: let's add test cases where we have pool selectors and service selectors together (if we don't want to make them mutually exclusive). Let's test corner cases where a service gets a pool from an ip whose advertisement doesn't select it and vice versa. Same for l2advertisement

api/v1beta1/bgpadvertisement_types.go
fedepaol: ignore what I wrote, it makes sense as we have customization properties such as peer selector, local pref and community

speaker/layer2_controller.go
oribon: probably happened as a side-effect when brought from the bgp implementation, but imo the approach of fetching the relevant ads from outside and passing to `ipAdvertisementFor` is cleaner than filtering inside it, wdyt? (about keeping it this way and removing the filtering part)

speaker/layer2_controller_test.go
oribon: done

internal/config/config.go
oribon: updated the comment, re checking when the service selector is exactly the same felt more clunky than helpful (as I expect it to not cover most of the cases), but I'll implement it if you think otherwise

e2etest/bgptests/service_selector.go
oribon: added `Pool and service selectors together` cases

e2etest/bgptests/service_selector.go
```diff
@@ -0,0 +1,318 @@
...
+
+			ginkgo.By("Checking that svc-wrong-pool is NOT advertised (pool selector mismatch)")
+			Consistently(func() error {
+				for _, c := range FRRContainers {
+					frrRoutesV4, frrRoutesV6, err := frr.Routes(c)
+					if err != nil {
+						return err
+					}
+
+					for _, ip := range svcWrongPool.Status.LoadBalancer.Ingress {
```
fedepaol: this can be a local function

e2etest/l2tests/service_selector.go
```diff
@@ -0,0 +1,282 @@
...
+	var cs clientset.Interface
+	testNamespace := ""
+	var allNodes []corev1.Node
+
+	ginkgo.AfterEach(func() {
+		if ginkgo.CurrentSpecReport().Failed() {
+			k8s.DumpInfo(Reporter, ginkgo.CurrentSpecReport().LeafNodeText)
+		}
+
+		// Clean previous configuration.
```
fedepaol: no need

e2etest/l2tests/service_selector.go
```diff
@@ -0,0 +1,282 @@
...
+				},
+			}
+
+			err = ConfigUpdater.Update(resources)
+			Expect(err).NotTo(HaveOccurred())
+
+			ginkgo.By("Creating service with expose=true requesting pool-a - should be advertised")
+			svcPoolA, _ := service.CreateWithBackend(cs, testNamespace, "svc-pool-a", service.TrafficPolicyCluster,
+				func(s *corev1.Service) {
+					s.Labels = map[string]string{"expose": "true"}
```
fedepaol: we can have two functoins under tweat.go - WithLabels / WithAnnotations

speaker/layer2_controller.go
fedepaol: agree, let's remove the logic in the other side

e2etest/bgptests/service_selector.go
oribon: done

e2etest/l2tests/service_selector.go
oribon: done

e2etest/l2tests/service_selector.go
oribon: done

speaker/layer2_controller.go
oribon: done (in a separate commit, felt too messy to be included in the original one)

fedepaol:
LGTM, thanks!

PR #2927, opened 2026-01-23
refactor: fix FRR config file handling and use constants

andreaskaris:
Move configFileName and reloaderPidFileName to const block since they should be immutable defaults. Fix inverted logic in reloadConfig and generateAndReloadConfigFile where environment variable checks were incorrectly reassigning the global vars instead of using them as fallback values when env vars are not set.

/kind cleanup

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note

```

### DIFF
--- a/internal/bgp/frr/config.go
+++ b/internal/bgp/frr/config.go
@@ -22,9 +22,12 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 )
 
-var (
+const (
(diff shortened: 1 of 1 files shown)


andreaskaris:
@oribon  done and ready

PR #2928, opened 2026-01-23
Fix log message for invalidFileErr in docker test

andreaskaris:
When the frr-reload.py failed, it would log its error to stdout. The containerhandler would only catch stderr and thus would log an empty string. Fix this by reporting both stdout and stderr.

/kind cleanup

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note

```


andreaskaris:
@oribon  done and ready

Issue #2933, opened 2026-01-24
TestTwoAdvertisementsTwoSessions is flaky and suffers from failures and/or data races


andreaskaris:
### MetalLB Version

TestTwoAdvertisementsTwoSessions is flaky and suffers from failures and/or data races

### Deployment method

Charts

### Main CNI

TestTwoAdvertisementsTwoSessions is flaky and suffers from failures and/or data races

### Describe the bug

TestTwoAdvertisementsTwoSessions is flaky and suffers from failures and/or data races

### To Reproduce

TestTwoAdvertisementsTwoSessions is flaky and suffers from failures and/or data races

### Expected Behavior

TestTwoAdvertisementsTwoSessions is flaky and suffers from failures and/or data races

### Additional Context

TestTwoAdvertisementsTwoSessions is flaky and suffers from failures and/or data races

https://github.com/metallb/metallb/actions/runs/21305404353/job/61332069948?pr=2932

```
4m 45s
Run inv test
go: downloading sigs.k8s.io/controller-runtime v0.17.1-0.20240318095156-c7e1dc9b5302
go: downloading sigs.k8s.io/controller-runtime/tools/setup-envtest v0.0.0-20240318095156-c7e1dc9b5302
go: downloading github.com/go-logr/zapr v1.2.4
go: downloading github.com/go-logr/logr v1.2.4
go: downloading github.com/spf13/afero v1.6.0
go: downloading github.com/spf13/pflag v1.0.5
go: downloading go.uber.org/zap v1.26.0
go: downloading golang.org/x/text v0.12.0
go: downloading go.uber.org/multierr v1.10.0
/home/runner/work/metallb/metallb/dev-env/unittest/setup-envtest.sh /home/runner/work/metallb/metallb/dev-env/unittest
go: downloading k8s.io/apimachinery v0.34.1
go: downloading sigs.k8s.io/controller-runtime v0.22.3
go: downloading k8s.io/api v0.34.1
go: downloading sigs.k8s.io/yaml v1.6.0
go: downloading github.com/go-kit/log v0.2.1
go: downloading github.com/google/go-cmp v0.7.0
go: downloading k8s.io/component-base v0.34.1
go: downloading github.com/prometheus/client_golang v1.22.0
go: downloading github.com/mikioh/ipaddr v0.0.0-20190404000644-d465c8ab6721
go: downloading k8s.io/utils v0.0.0-20250604170112-4c0f3b243397
[...]

PR #2934, opened 2026-01-24
Use synctest.Test framework to unit test asynchronous functions

andreaskaris:
Use synctest.Test (https://go.dev/blog/testing-time) for tests in
    internal/bgp/frr. This specifically fixes flakiness for
    TestTwoAdvertisementsTwoSessions but will also avoid other flakes in
    those unit tests.

/kind bug

Fixes: https://github.com/metallb/metallb/issues/2933

**Note:** Use hide whitespace and you'll see that this is a relatively small commit.

Verification that the flake is gone:

```
[akaris@metallb metallb]$ cd internal/bgp/frr
[akaris@metallb frr]$  stress -p 1 go test -run TestTwoAdvertisementsTwoSessions   .
(...)
4m5s: 78 runs so far, 0 failures, 1 active
^C
[akaris@metallb frr]$  stress -p 1 go test .
5s: 0 runs so far, 0 failures, 1 active
(...)
6m40s: 19 runs so far, 0 failures, 1 active
```

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note

```

### DIFF
--- a/internal/bgp/frr/frr.go
+++ b/internal/bgp/frr/frr.go
@@ -441,8 +441,6 @@ func mockNewSessionManager(l log.Logger, logLevel logging.Level) *sessionManager
 
 	debouncer(reload, res.reloadConfig, debounceTimeout, failureTimeout, l)
 
-	reloadValidator(l, res.reloadConfig)
-
 	return res
 }
 
--- a/internal/bgp/frr/debounce_test.go
+++ b/internal/bgp/frr/debounce_test.go
@@ -5,6 +5,7 @@ package frr
 import (
 	"fmt"
 	"testing"
+	"testing/synctest"
 	"time"
 
 	"github.com/go-kit/log"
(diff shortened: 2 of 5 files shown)


internal/bgp/frr/frr.go
```diff
@@ -411,49 +414,72 @@ var debounceTimeout = 3 * time.Second
 var failureTimeout = time.Second * 5
 
 func NewSessionManager(l log.Logger, logLevel logging.Level) bgp.SessionManager {
+	ctx, cancel := context.WithCancel(context.Background())
```
karampok: Adding code changes without being used outside test never shoulds great.
Any reason why we cannot add a defer cancel to have a graceful shutdown.
(I know is not really needed, just it looks odd to see functions/methods never used outside the tests).

If only used in test does that mean only used int mockNewSessionManager?
if yes maybe here we can have `ctx,_ :=`

internal/bgp/frr/frr.go
```diff
@@ -411,49 +414,72 @@ var debounceTimeout = 3 * time.Second
...
-	reloadValidator(l, res.reloadConfig)
+	reloadValidator(res.ctx, l, res.reloadConfig)
 
 	return res
 }
 
-func reloadValidator(l log.Logger, reload chan<- reloadEvent) {
-	var tickerIntervals = 30 * time.Second
-	var prevReloadTimeStamp string
+func (sm *sessionManager) Close() {
```
karampok: Can we move that above so all methods are together? just visual

internal/bgp/frr/frr.go
```diff
@@ -411,49 +414,72 @@ var debounceTimeout = 3 * time.Second
...
+	if sm.cancelCtx != nil {
+		sm.cancelCtx()
+	}
+	if sm.reloadConfig != nil {
+		close(sm.reloadConfig)
```
karampok: Claude complains that this can panic (close on closed channels) and maybe some wait for both go routines to exit. I suppose not important but if one day we use that close in the code might be issue

andreaskaris: https://stackoverflow.com/questions/16105325/how-to-check-a-channel-is-closed-or-not-without-reading-it
I think there's no way in golang to figure out if a channel was closed or not, other than jumping through some esoteric hoops. As a side note, that code, as is minus the nil check, was suggested by Claude in the first place

I'll add a comment saying: `calling Close() on an already closed SessionManager causes a panic`

Edit: well, obviously could do sync.Once() ...

internal/bgp/frr/frr.go
andreaskaris: will do

internal/bgp/frr/frr.go
andreaskaris: we cannot add the defer directly in the SessionManager because it's async, so by the time the session manager returns, we're still doing business logic. Let me have another look at the code and check if it can be added in the caller of NewSessionManager, at the very end of the program logic.
My point though in making that change is that the unit tests actually do reveal a go function leak, so even though it doesn't bite us in prod, it's actually there -- granted, I'm not fixing that leak, either, so I'll check how and get back :-)

internal/bgp/frr/frr.go
andreaskaris: done

internal/bgp/frr/frr.go
andreaskaris: done

internal/bgp/frr/frr.go
andreaskaris: after looking at it again, it makes little sense to do pass the Close() up the stack in prod .. and the change looks cleaner imo than what was there before and helps with unit testing. So I'm inclined to keep it. I *think* The context is needed inside NewSessionManager because the reload() function waits for ctx.Done() - anyway, I'm afraid to leave it at nil 

Could pass a ctx.TODO() / ctx.Background() also, without the cancel .. I'm indifferent about that

note: I'm using Close() error because that implements the io.Closer interface

andreaskaris:
e2e-use-operator failed with https://github.com/metallb/metallb/actions/runs/21723893196/job/62661582040?pr=2934 --- wondering if that's due to flakes, rerunning tests

fedepaol:
Neat!

PR #2939, opened 2026-01-27
feat: Add IPv6 NDP Proxy support for Layer2 mode

changluyi:
/kind feature

## Summary

This PR adds IPv6 NDP (Neighbor Discovery Protocol) proxy support for MetalLB Layer2 mode, enabling proper IPv6 VIP advertisement in Layer 2 networks.

## Background & Use Case

As a committer on the Kube-OVN team, I encountered a scenario where:

1. **Underlay Network Setup**: A network interface is connected to an underlay subnet but has **no IP address configured** on it
2. **VIP Advertisement**: VIPs need to be advertised through this underlay interface
3. **IPv4 vs IPv6 Behavior Difference**:
   - **IPv4**: The kernel automatically proxies ARP for VIPs on interfaces without IPs
   - **IPv6**: The kernel does **NOT** automatically proxy NDP - this must be explicitly enabled via `proxy_ndp` sysctl and managed with neighbor entries

### How Traffic Works (Important Context)

You might wonder: if the underlay interface has no IP, wouldn't packets be dropped?

**No!** Here's how it works in Kube-OVN:

1. **Packet Arrival**: When a packet arrives at the underlay interface (even without IP)
2. **OVS Flow Table**: Open vSwitch flow tables capture and redirect the packet
3. **OVN Network Stack**: The packet is delivered to OVN's network stack for processing
4. **Kernel Response**: The kernel (in OVN namespace) responds with ARP/NDP replies

This is why IPv4 works automatically (kernel ARP proxy), but IPv6 needs explicit NDP proxy configuration.

This asymmetry causes IPv6 VIPs to fail in Layer2 mode when using underlay networks, while IPv4 works correctly.

## Solution

This PR implements NDP proxy management to match IPv4 ARP behavior:

- Automatically enables `proxy_ndp` sysctl per interface when needed
- Manages proxy neighbor entries using pure Go netlink library
- Cleanly removes proxy entries when VIPs are deactivated

## Changes

- **NDPProxyManager**: New component that manages IPv6 NDP proxy entries
  - Automatically enables/disables `proxy_ndp` sysctl per interface
  - Manages proxy neighbor entries using pure Go netlink library
  - Thread-safe with proper mutex protection
  - Path sanitization to prevent directory traversal

- **API**: Added `enableNDPProxy` field to `L2Advertisement`
  - Allows users to opt-in to NDP proxy functionality per advertisement

- **Implementation**: Pure Go using `github.com/vishvananda/netlink`
  - No external dependencies on `iproute2` or other tools
  - Compatible with distroless images
  - Follows MetalLB's existing patterns

## Test Plan

- [x] Unit tests for NDPProxyManager
- [x] Path sanitization tests
- [x] IPv4/IPv6 differentiation tests
- [x] Build verification

## Usage

```yaml
kind: L2Advertisement
metadata:
  name: example
spec:
  ipAddressPools:
  - ipv6-pool
  enableNDPProxy: true  # Enable NDP proxy for IPv6 VIPs
```

## Compatibility

- Works with existing IPv4 configurations (no impact)
- Requires `CAP_NET_ADMIN` capability (already in place for Layer2)
- Compatible with distroless images (no Alpine/iproute2 dependency)
- Enables IPv6 Layer2 on underlay networks (matching IPv4 behavior)

## Related

This brings IPv6 Layer2 behavior to parity with IPv4, which already works with ARP proxy on underlay networks without IPs.

```release-note
Add IPv6 NDP Proxy support for Layer2 mode
```

### DIFF
--- a/website/content/apis/_index.md
+++ b/website/content/apis/_index.md
@@ -309,6 +309,7 @@ _Appears in:_
 | `ipAddressPoolSelectors` _[LabelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#labelselector-v1-meta) array_ | A selector for the IPAddressPools which would get advertised via this advertisement.<br />If no IPAddressPool is selected by this or by the list, the advertisement is applied to all the IPAddressPools. |
 | `nodeSelectors` _[LabelSelector](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.27/#labelselector-v1-meta) array_ | NodeSelectors allows to limit the nodes to announce as next hops for the LoadBalancer IP. When empty, all the nodes having  are announced as next hops. |
 | `interfaces` _string array_ | A list of interfaces to announce from. The LB IP will be announced only from these interfaces.<br />If the field is not set, we advertise from all the interfaces on the host. |
+| `enableNDPProxy` _boolean_ | EnableNDPProxy enables kernel-level NDP proxying for IPv6 VIPs.<br />When enabled, the speaker will configure the kernel to proxy Neighbor Discovery<br />Protocol messages for the specified interfaces, allowing VIPs to be reachable<br />even when the announcing interface has no IP address configured.<br />This requires the speaker to have CAP_NET_ADMIN capability. |
 
 
 


oribon:
hey, thanks for submitting this
I haven't gone through the code (and I'm less familiar with this part), but iiuc the current ndp responder should have worked if your interfaces had a link-local ip, is it the case that they have no ip at all? trying to understand if this is a special case that can be avoided without adding (quite a bit of) extra complexity.

changluyi:
> hey, thanks for submitting this
> I haven't gone through the code (and I'm less familiar with this part), but iiuc the current ndp responder should have worked if your interfaces had a link-local ip, is it the case that they have no ip at all? trying to understand if this is a special case that can be avoided without adding (quite a bit of) extra complexity.

Yes, the interfaces have no IP at all (not even link-local). This is how Kube-OVN's underlay network works - the interface is purely a Layer 2 transport, and all IP processing happens in OVN's virtual network stack.
The interesting part: IPv4 already works this way. When an interface has no IP, the kernel automatically proxies ARP for VIPs, and MetalLB Layer2 works fine. We just need the same behavior for IPv6 - currently there's an asymmetry where IPv4 works but IPv6 doesn't in identical scenarios.

oribon:
> Yes, the interfaces have no IP at all (not even link-local). This is how Kube-OVN's underlay network works - the interface is purely a Layer 2 transport, and all IP processing happens in OVN's virtual network stack.

sorry if I'm missing something obvious, but why can't the interfaces have a link-local address? this currently looks like a (complex) workaround, which is also not very MetalLB-esque (managing sysctls, kernel networking, etc.). this is even before evaluating if using ndp_proxy is the right solution.

changluyi:
> > Yes, the interfaces have no IP at all (not even link-local). This is how Kube-OVN's underlay network works - the interface is purely a Layer 2 transport, and all IP processing happens in OVN's virtual network stack.
> 
> sorry if I'm missing something obvious, but why can't the interfaces have a link-local address? this currently looks like a (complex) workaround, which is also not very MetalLB-esque (managing sysctls, kernel networking, etc.). this is even before evaluating if using ndp_proxy is the right solution.

Thanks for reviewing!

Link-local addresses don't solve this because:

1. **VIPs are global IPv6 addresses** (e.g., `2001:db8::100`)
2. When a gateway sends NS asking "Who has 2001:db8::100?", the kernel won't reply even if the interface has a link-local address (`fe80::xxx`)
3. **Link-local and global addresses are different scopes** - the kernel won't proxy NDP between them

**In practice**, we've seen MetalLB fail with this error even when the interface has a link-local address:
```
write ip fe80::5054:ff:feb8:9b%br-provider->2341::10:180:124:1: 
sendmsg: network is unreachable
```

oribon:
> In practice, we've seen MetalLB fail with this error even when the interface has a link-local address:
> 
> write ip fe80::5054:ff:feb8:9b%br-provider->2341::10:180:124:1: 
> sendmsg: network is unreachable

I tested this scenario and confirmed that this is not a MetalLB issue but rather a routing one.
MetalLB gets the NDP request on br-provider from a global address (2341::10:180:124:1: ), but br-provider has only a link-local ip (and no route for that subnet) so it can't send back the response.
some alternatives could be: 
1. the NDP request could come from a link-local address instead of a global one.
2. add a route for the gateway's subnet via br-provider, e.g ip -6 route add 2341::/64 dev br-provider

is it feasible for kube-ovn to do one of those?

PR #2953, opened 2026-02-18
ConfigurationState: react to deletions

oribon:
ConfigReconciler and PoolReconciler now watch ConfigurationState and
re-report their conditions when the resource is recreated.
We explicitly don't act on the UPDATE path, because this is
handled by SSA.

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
Fix ConfigurationState to be updated when the resource is recreated.
```

### DIFF
--- a/e2etest/configurationstatetests/configurationstate.go
+++ b/e2etest/configurationstatetests/configurationstate.go
@@ -189,6 +189,25 @@ var _ = ginkgo.Describe("ConfigurationState", func() {
 		}, 60*time.Second, 5*time.Second).Should(Succeed())
 	})
 
+	ginkgo.It("should self-heal after deletion", func() {
+		stateName := "speaker-" + allNodes.Items[0].Name
+
+		ginkgo.By("Deleting the ConfigurationState resource")
+		toDelete := &metallbv1beta1.ConfigurationState{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      stateName,
+				Namespace: metallb.Namespace,
+			},
+		}
+		err := ConfigUpdater.Client().Delete(context.Background(), toDelete)
+		Expect(err).NotTo(HaveOccurred())
+
+		ginkgo.By("Verifying the ConfigurationState is recreated and returns to valid")
+		Eventually(func() error {
+			return stateMatches(stateName, validStatus)
+		}, 60*time.Second, 5*time.Second).Should(Succeed())
+	})
+
 })
 
 func allStatesExist(allNodes *corev1.NodeList) error {
@@ -236,6 +255,29 @@ func allStatesExist(allNodes *corev1.NodeList) error {
 	}
 	if diff := cmp.Diff(want, got.Items, opts...); diff != "" {
 		return fmt.Errorf("ConfigurationState list mismatch (-want +got):\n%s", diff)
+	}
+
+	componentFieldOwners := map[string]string{
+		"controller": "poolReconciler",
+		"speaker":    "configReconciler",
+	}
+	for _, item := range got.Items {
+		componentType := item.Labels["metallb.io/component-type"]
+		wantManager, ok := componentFieldOwners[componentType]
+		if !ok {
+			return fmt.Errorf("ConfigurationState %q has unknown component-type label %q", item.Name, componentType)
+		}
+
+		found := false
+		for _, mf := range item.ManagedFields {
+			if mf.Subresource == "status" && mf.Manager == wantManager {
+				found = true
+				break
+			}
+		}
+		if !found {
+			return fmt.Errorf("ConfigurationState %q (component=%s) missing SSA field owner %q in status managed fields", item.Name, componentType, wantManager)
+		}
 	}
 
 	return nil
--- a/internal/k8s/controllers/configuration_state_controller.go
+++ b/internal/k8s/controllers/configuration_state_controller.go
@@ -106,6 +106,31 @@ func NewConfigurationStateReconcilerPredicate(namespace, name string) predicate.
 	}
 }
 
+// NewConfigStateConditionReporterPredicate returns a predicate for controllers that write
+// conditions to a ConfigurationState CR. It fires only on Create events for the
+// target CR, so that conditions are re-reported after the resource is recreated.
+// Updates are handled by Server-Side Apply field ownership.
+func NewConfigStateConditionReporterPredicate(namespace, name string) predicate.Predicate {
+	matchesTarget := func(obj client.Object) bool {
+		return obj.GetNamespace() == namespace && obj.GetName() == name
+	}
+
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return matchesTarget(e.Object)
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
+			return false
+		},
+		DeleteFunc: func(e event.DeleteEvent) bool {
+			return false
+		},
+		GenericFunc: func(e event.GenericEvent) bool {
+			return false
+		},
+	}
+}
+
 func (r *ConfigurationStateReconciler) SetupWithManager(mgr ctrl.Manager) error {
 	p := NewConfigurationStateReconcilerPredicate(r.Namespace, r.ConfigStateName)
 
--- a/internal/k8s/controllers/config_controller.go
+++ b/internal/k8s/controllers/config_controller.go
@@ -33,6 +33,7 @@ import (
 	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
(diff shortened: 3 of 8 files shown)


oribon:
@karampok just to be sure, you're suggesting to remove the predicate thing with updates, because these are managed fields?
given that it's pretty straightforward, I'd keep the current mechanism (enforcing the ownership and values) + it pretty much aligns with the rest of our code. do you know what are the guidelines for this scenario?

internal/k8s/controllers/configuration_state_controller.go
```diff
@@ -106,6 +107,59 @@ func NewConfigurationStateReconcilerPredicate(namespace, name string) predicate.
 	}
 }
 
+// NewConfigurationStateWriterPredicate returns a predicate for controllers that write
```
karampok: If we keep the pattern I would prefer a name something like NewConfigStatetConditionReporterPredicate

internal/k8s/controllers/configuration_state_controller.go
```diff
@@ -106,6 +107,59 @@ func NewConfigurationStateReconcilerPredicate(namespace, name string) predicate.
...
+	oldCond := meta.FindStatusCondition(oldCS.Status.Conditions, conditionType)
+	newCond := meta.FindStatusCondition(newCS.Status.Conditions, conditionType)
+
+	if oldCond == nil && newCond == nil {
+		return false
+	}
+	if oldCond == nil || newCond == nil {
+		return true
+	}
+	return oldCond.Status != newCond.Status || oldCond.Reason != newCond.Reason || oldCond.Message != newCond.Message
```
karampok: Maybe some logic that hints fast what is the ignore field
```
oldCond.Timestamp = newCond.Timestamp =nil
reten oldCond == newCond
```
or api like the cmp.

internal/k8s/controllers/configuration_state_controller.go
karampok: this is still valid, Writer is a bit vague

e2etest/configurationstatetests/configurationstate.go
```diff
@@ -189,6 +189,39 @@ var _ = ginkgo.Describe("ConfigurationState", func() {
 		}, 60*time.Second, 5*time.Second).Should(Succeed())
 	})
 
+	ginkgo.It("should have SSA field owners for each reconciler", func() {
```
karampok: wondering if each It adds significant time and whether we  should keep single It
(given we cannot run test in paralle)

karampok (approved):
thanks

internal/k8s/controllers/configuration_state_controller.go
```diff
@@ -106,6 +106,31 @@ func NewConfigurationStateReconcilerPredicate(namespace, name string) predicate.
...
+func NewConfigStateConditionReporterPredicate(namespace, name string) predicate.Predicate {
+	matchesTarget := func(obj client.Object) bool {
+		return obj.GetNamespace() == namespace && obj.GetName() == name
+	}
+
+	return predicate.Funcs{
+		CreateFunc: func(e event.CreateEvent) bool {
+			return matchesTarget(e.Object)
+		},
+		UpdateFunc: func(e event.UpdateEvent) bool {
```
andreaskaris, line 122: IMHO, we have an antipattern here (not in your commit, but in the controllers' design).
We have 2 controllers writing the same resource, and this should not happen, when the only controller updating anything in the configurationstate resource should be the configuration_state_controller (or the config_controller, but not both)
 See configuration_state_controller.go:80 which also updates the resource.

The result is that we get a reconcile event in both controllers, and both will update the resource potentially.

I added some debug statements to your original PR, and get something like:
```
1. delete configurationstate

|-2. configcontroller runs: "NewConfigurationStateWriterPredicate":"DELETE" --> ignored
| concurrent
|-3. configurationstate controller runs and recreates resource

|-4. configurationstate_controller runs (noop, or update .Status.ErrorSummary if it raced with 5.b. and errors were present) (let's assume this is noop)
| - concurrent
|-5. configcontroller runs: {"NewConfigurationStateWriterPredicate":"CREATE" --> run reportCondition  --> updates the resource


|-7. configcontroller "NewConfigurationStateWriterPredicate":"UPDATE"  --> noop
|concurrent
|-8. configurationstate_controller runs and updates "result" field to valid  .. **does this change the managed fields for the entire resource???**

|-9. configurationstate_controller runs and noop
| concurrent
|- 10. configcontroller runs: {"CHANGED":true,"NewConfigurationStateWriterPredicate":"UPDATE", --> run reportCondition --> does in theory not update the resource as SSA noop .. **but does it set the manager field again?**

For some reason, this is run again:
|-11. configurationstate_controller runs and noop
|
|-12. configcontroller runs: {"CHANGED":false,"NewConfigurationStateWriterPredicate"
```

So setting updatefunc here to false is likely the right step, but only because we have an antipattern in the first place

andreaskaris: I think the easiest fix:

**q: can't ConfigReconciler and PoolReconciler compute the .Status.ErrorSummary directly?**

all of this here:
```
	result := metallbv1beta1.ConfigurationResultUnknown
	if len(configStatus.Status.Conditions) > 0 {
		result = metallbv1beta1.ConfigurationResultValid
	}

	var errorMessages []string
	for _, cond := range configStatus.Status.Conditions {
		if cond.Status == metav1.ConditionFalse && cond.Reason == ErrorTypeConfiguration {
			result = metallbv1beta1.ConfigurationResultInvalid
			errorMessages = append(errorMessages, cond.Message)
		}
	}

	configStatus.Status.Result = result
	configStatus.Status.ErrorSummary = strings.Join(errorMessages, "\n")
	if err := r.Client.Status().Update(ctx, &configStatus); err != nil {
		level.Error(r.Logger).Log("controller", r, "error", "failed to update status", "result", result, "error", err)
		return ctrl.Result{}, err
	}
```
should be moved from the configuration_state_controller.go to the reportCondition --> that will stop this weird dance between updates, and it can be calculated on the fly in `reportCondition` anyway; it's not necessary to have that in `configuration_state_controller.go` which should only handle resource creation, and we can get rid of the Update()

andreaskaris: There's also only ever a single condition ... so what I mean is doing something like this:
```
func (r *ConfigReconciler) reportCondition(ctx context.Context, conditionErr error) error {
	if r.ConfigStateName == "" {
		return nil
	}

	condition := metav1.Condition{
		Type:    ConditionTypeConfigReconcilerValid,
		Status:  metav1.ConditionTrue,
		Reason:  ErrorTypeNone,
		Message: "",
		// Always set to now (tracks last reconciliation, not last status change).
		LastTransitionTime: metav1.Now(),
	}

	switch {
	case errors.Is(conditionErr, ErrConfiguration):
		condition.Status = metav1.ConditionFalse
		condition.Reason = ErrorTypeConfiguration
		condition.Message = conditionErr.Error()
	case conditionErr != nil:
		condition.Status = metav1.ConditionFalse
		condition.Reason = ErrorTypeUnknown
		condition.Message = conditionErr.Error()
	}

	configStatus := &metallbv1beta1.ConfigurationState{
		TypeMeta: metav1.TypeMeta{
			APIVersion: "metallb.io/v1beta1",
			Kind:       "ConfigurationState",
		},
		ObjectMeta: metav1.ObjectMeta{
			Name:      r.ConfigStateName,
			Namespace: r.Namespace,
		},
		Status: metallbv1beta1.ConfigurationStateStatus{
			Conditions: []metav1.Condition{condition},
		},
	}

	level.Error(r.Logger).Log("reportCondition", "reportCondition", "namespace", r.Namespace, "configstatename", r.ConfigStateName)
	if err := r.Status().Patch(ctx, configStatus, client.Apply, client.FieldOwner("configReconciler"), client.ForceOwnership); err != nil {
		return fmt.Errorf("patch %s/%s: %w", r.Namespace, r.ConfigStateName, err)
	}

	return nil
}
```

and removing the stuff in configuration_state_controller.go .. and doing the same for the poolController

karampok: @andreaskaris  i am not sure what is the antipattern?

> We have 2 controllers writing the same resource, 

what is same resource? (is same CR, but config/pool controller write only each one condition)

Also the updateFunction is always false, so not trigger takes place now?

karampok: > There's also only ever a single condition


  - configReconcilerValid — written by ConfigReconciler
  - poolReconcilerValid — written by PoolReconciler

and the ConfigurationStateReconciler can only aggregate and write the global Result/ErrorSummary

andreaskaris: @karampok From what I understand, CustomResources should ideally only ever be managed (=written) by a single controller; especially when using SSA. The .Update() statement from one controller + the SSA from another controller don't coexist nicely together.
Right now, there are 2 of them managing the same resource at the same time: configreconciler and configurationstatereconciler    (respectively poolreconciler and configurationstatereconciler).

If I read things correctly ... The configstate resource can be created by a patch in configreconciler or poolreconciler (ssa does create and update) ; and the aggregation for the Result and ErrorSummary can happen directly in the configreconciler or poolreconciler. So there's no need for the configurationstatereconciler.

Or am I missing something?

edit: working on a patch atm to test this

karampok: https://github.com/metallb/metallb/blob/main/design/crd-configstate.md but I think the SSA exists specifally management component (e.g. operator) to maintain a global CR where subcomponents (e.g. operands) to write its own safely their condition, and the management component to decide global status (Degraded status)

andreaskaris: @karampok @oribon https://github.com/metallb/metallb/pull/2969 uses @oribon's original commits and shows that the `ConfigurationStateReconciler` isn't needed; the only 2 things that it does is:
a) create a new resource
b) aggregate conditions into `Result` + `ErrorSummary`

a) can be taken care of by configreconciler and poolreconciler
b) made no sense, because: there's only ever a single condition on each ConfigurationState; even if there were multiple: conditions are exclusively managed by ConfigReconciler or PoolReconciler --> the ConfigReconciler and PoolReconciler can set `Result` + `ErrorSummary` directly, no reason to retrieve and update

By removing the ConfigurationStateReconciler, ownership is clear, and reconciliation can be brought down to a single run in most cases (whereas current code will always ping-pong between updates from 2 controllers, or if we disable updates via the predicates we give up tamper protection)

oribon:
thanks for the discussion, I'll send this to merge (as it is relevant for both implementations I guess) and we'll continue it over https://github.com/metallb/metallb/pull/2969

PR #2973, opened 2026-03-07
fix: remove deprecated metallb.universe.tf managed annotations on upgrade

### LINKED ISSUE #2642: Deprecated MetalLB managed annotations not removed from service on upgrade
### MetalLB Version

0.14.9

### Deployment method

Charts

### Main CNI

flannel

### Kubernetes Version

1.31.4

### Cluster Distribution

k3s

### Describe the bug

Annotations starting with "metallb.universe.tf" are deprecated as announced.

However, if a service was deployed using an older version, it contains metallb managed annotation `metallb.universe.tf/ip-allocated-from-pool`.

### To Reproduce

1. Deploy a service with metallb < 0.14.9 and assign an IP using `metallb.universe.tf/loadBalancerIPs`
2. Upgrade metallb to 0.14.9
3. Re-apply the same service after changing the deprecated annotation to `metallb.io/loadBalancerIPs`

### Expected Behavior

MetalLB will complain about use of deprecated annotations in kubernetes events - however that annotation was set by MetalLB.

{68C91FDB-6B72-4AA5-A8AD-2ABE068D59D9}

Do note that service gets the new annotation `metallb.io/ip-allocated-from-pool`, but it does not remove the old annotation - which is what I'd expect.

### Additional Context

```
# Source: haproxy/templates/service.yaml
apiVersion: v1
kind: Service
metadata:
  name: haproxy
  namespace: default
  labels:
    helm.sh/chart: haproxy-1.23.0
    app.kubernetes.io/name: haproxy
    app.kubernetes.io/instance: haproxy
    app.kubernetes.io/version: "3.0.4"
    app.kubernetes.io/managed-by: Helm
  annotations:
    metallb.io/loadBalancerIPs: "192.168.0.50"
spec:
  type: LoadBalancer
  selector:
    app.kubernetes.io/name: haproxy
    app.kubernetes.io/instance: haproxy
  externalTrafficPolicy: Local
  ports:
  - name: http
    protocol: TCP
    port: 80
    targetPort: http
  - name: https
    protocol: TCP
    port: 443
    targetPort: https
```

Onyx2406:
## Summary
- Delete the deprecated `metallb.universe.tf/ip-allocated-from-pool` annotation when setting the new `metallb.io/ip-allocated-from-pool` annotation
- Also clean up the deprecated annotation in `clearServiceState`

## Motivation
Fixes #2642 — Services deployed with older MetalLB versions retain the deprecated `metallb.universe.tf/ip-allocated-from-pool` annotation after upgrading. The controller sets the new `metallb.io` annotation but never removes the old `metallb.universe.tf` one, leaving stale annotations on services.

## Root Cause
`convergeBalancer` (line 215) sets `AnnotationIPAllocateFromPool` but doesn't delete `DeprecatedAnnotationIPAllocateFromPool`. `clearServiceState` (line 248) only deletes the new annotation.

## Changes
- `controller/service.go`: Add `delete(svc.Annotations, DeprecatedAnnotationIPAllocateFromPool)` in both `convergeBalancer` and `clearServiceState`

## Test plan
- [x] When MetalLB assigns an IP, the deprecated annotation is removed if present
- [x] When service state is cleared, both annotations are removed
- [x] No change in behavior for services without deprecated annotations

### DIFF
--- a/controller/service.go
+++ b/controller/service.go
@@ -213,6 +213,7 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 		svc.Annotations = make(map[string]string)
 	}
 	svc.Annotations[AnnotationIPAllocateFromPool] = pool
+	delete(svc.Annotations, DeprecatedAnnotationIPAllocateFromPool)
 
 	return nil
 }
@@ -246,6 +247,7 @@ func serviceFamilyChanged(
 func (c *controller) clearServiceState(key string, svc *v1.Service) {
 	c.ips.Unassign(key)
 	delete(svc.Annotations, AnnotationIPAllocateFromPool)
+	delete(svc.Annotations, DeprecatedAnnotationIPAllocateFromPool)
 	svc.Status.LoadBalancer = v1.LoadBalancerStatus{}
 }
 


fedepaol:
thanks for the PR (and sorry for the delay).
Couple of asks:
- please sign the commit
- please add unit tests

PR #2974, opened 2026-03-09
TEMPORARY FIX: replace kube-rbac-proxy registry

oribon:
gcr.io is no longer available for the kube-rbac-proxy image, we replace it with registry.k8s.io.
This fix is temporary (and unblocks CI), the real fix is removing the dependency on kube-rbac-proxy as was suggested a while ago.

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
/kind regression

**Release note**:

```release-note
NONE
```

### DIFF
--- a/config/prometheus-frr-k8s/kustomization.yaml
+++ b/config/prometheus-frr-k8s/kustomization.yaml
@@ -24,3 +24,6 @@ patches:
       kind: ServiceMonitor
       name: frr-k8s-metrics-monitor
 namespace: metallb-system
+images:
+  - name: gcr.io/kubebuilder/kube-rbac-proxy
+    newName: registry.k8s.io/kubebuilder/kube-rbac-proxy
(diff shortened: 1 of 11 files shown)


config/prometheus-frr-k8s/kustomization.yaml
```diff
@@ -24,3 +24,6 @@ patches:
       kind: ServiceMonitor
       name: frr-k8s-metrics-monitor
 namespace: metallb-system
+images:
```
fedepaol, line 27: we should file an issue to remind ourselves to remove this after fixing frrk8s too

tasks.py
```diff
@@ -558,12 +558,23 @@ def dev_env(
 
     frr_k8s_ns = "frr-k8s-system"
     if bgp_type == "frr-k8s-external":
+        with tempfile.NamedTemporaryFile(mode="w", suffix=".yaml", delete=False) as f:
```
fedepaol, line 561: same here

PR #2975, opened 2026-03-10
e2etests: add nodeSelector with bgpAdv update

ajaggapa:
**Is this a BUG FIX or a FEATURE ?**: New e2e test for bgpAdvertisement with nodeSelector update

>
> /kind bug
 /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**What this PR does / why we need it**: Migrate tests from eco-gotests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etest/bgptests/node_selector.go
+++ b/e2etest/bgptests/node_selector.go
@@ -263,6 +263,91 @@ var _ = ginkgo.Describe("BGP Node Selector", func() {
 		ginkgo.Entry("IPV6", ipfamily.IPv6, "fc00:f853:0ccd:e799::/116"),
 	)
 
+	ginkgo.DescribeTable("Two advertisements - updating one nodeSelector to non-matching withdraws its service without affecting the other",
+		func(pairingIPFamily ipfamily.Family, addresses []string) {
+			var allNodes *corev1.NodeList
+			allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			resources := config.Resources{
+				Pools: []metallbv1beta1.IPAddressPool{
+					{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: "first-pool",
+						},
+						Spec: metallbv1beta1.IPAddressPoolSpec{
+							Addresses: []string{addresses[0]},
+						},
+					}, {
+						ObjectMeta: metav1.ObjectMeta{
+							Name: "second-pool",
+						},
+						Spec: metallbv1beta1.IPAddressPoolSpec{
+							Addresses: []string{addresses[1]},
+						},
+					},
+				},
+				Peers: metallb.PeersForContainers(FRRContainers, pairingIPFamily),
+				BGPAdvs: []metallbv1beta1.BGPAdvertisement{
+					{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: "first-adv",
+						},
+						Spec: metallbv1beta1.BGPAdvertisementSpec{
+							NodeSelectors:   k8s.SelectorsForNodes(allNodes.Items),
+							IPAddressPools: []string{"first-pool"},
+						},
+					}, {
+						ObjectMeta: metav1.ObjectMeta{
+							Name: "second-adv",
+						},
+						Spec: metallbv1beta1.BGPAdvertisementSpec{
+							NodeSelectors:   k8s.SelectorsForNodes(allNodes.Items),
+							IPAddressPools: []string{"second-pool"},
+						},
+					},
+				},
+			}
+			for _, c := range FRRContainers {
+				err := frrcontainer.PairWithNodes(cs, c, pairingIPFamily)
+				Expect(err).NotTo(HaveOccurred())
+			}
+
+			err = ConfigUpdater.Update(resources)
+			Expect(err).NotTo(HaveOccurred())
+
+			firstService, _ := testservice.CreateWithBackend(cs, testNamespace, "first-lb", testservice.WithSpecificPool("first-pool"))
+			defer testservice.Delete(cs, firstService)
+			secondService, _ := testservice.CreateWithBackend(cs, testNamespace, "second-lb", testservice.WithSpecificPool("second-pool"))
+			defer testservice.Delete(cs, secondService)
+
+			ginkgo.By("Checking first service announced from all nodes")
+			for _, c := range FRRContainers {
+				validateService(firstService, allNodes.Items, c)
+			}
+			ginkgo.By("Checking second service announced from all nodes")
+			for _, c := range FRRContainers {
+				validateService(secondService, allNodes.Items, c)
+			}
+
+			ginkgo.By("Updating second BGP advertisement with non-existing node selector")
+			resources.BGPAdvs[1].Spec.NodeSelectors = []metav1.LabelSelector{
+				{MatchLabels: map[string]string{"non-existing-label": "true"}},
+			}
+			err = ConfigUpdater.Update(resources)
+			Expect(err).NotTo(HaveOccurred())
+
+			ginkgo.By("Validating first service still advertised from all nodes")
+			for _, c := range FRRContainers {
+				validateService(firstService, allNodes.Items, c)
+			}
+			ginkgo.By("Validating second service no longer advertised from all nodes")
+			checkServiceNotOnNodes(secondService, allNodes.Items, pairingIPFamily)
+		},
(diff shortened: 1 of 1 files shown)


oribon:
iiuc the main value of this test is verifying that updating the adv to a "bad nodeselector" withdraws the service. if that's the case, not sure I understand why we need 2 nodes, 2 services (and a new test), can we just extend the existing `One service, one advertisement with node selector, labeling node` with something like:
```go
ginkgo.By("Updating advertisement nodeSelector to non-matching label")
resources.BGPAdvs[0].Spec.NodeSelectors = []metav1.LabelSelector{
	{MatchLabels: map[string]string{"non-existing-label": "true"}},
}
err = ConfigUpdater.Update(resources)

ginkgo.By("Validating service IP no longer advertised from any node")
checkServiceNotOnNodes(svc, allNodes.Items, pairingIPFamily)
```

ajaggapa:
> iiuc the main value of this test is verifying that updating the adv to a "bad nodeselector" withdraws the service. if that's the case, not sure I understand why we need 2 nodes, 2 services (and a new test), can we just extend the existing `One service, one advertisement with node selector, labeling node` with something like:
> 
> ```go
> ginkgo.By("Updating advertisement nodeSelector to non-matching label")
> resources.BGPAdvs[0].Spec.NodeSelectors = []metav1.LabelSelector{
> 	{MatchLabels: map[string]string{"non-existing-label": "true"}},
> }
> err = ConfigUpdater.Update(resources)
> Expect(err).NotTo(HaveOccurred())
> 
> ginkgo.By("Validating service IP no longer advertised from any node")
> checkServiceNotOnNodes(svc, allNodes.Items, pairingIPFamily)
> ```

@oribon  We also want to make sure that the other bgpAdvertisement does not get disrupted when the bgpAdvertisement under test is updated.

ajaggapa:
@oribon 

Adding to the point above, Adding more validations or extending the current test will skip subsequent validations even when one of it fails early.

oribon:
> We also want to make sure that the other bgpAdvertisement does not get disrupted when the bgpAdvertisement under test is updated.

I think the principle of "update resources, verify existing is intact" shows up multiple times in several tests, not sure we need to have a dedicated test just for bgpAds.

> Adding to the point above, Adding more validations or extending the current test will skip subsequent validations even when one of it fails early.

I guess this applies to any multi-step test we have, and imo if the test has a natural flow (and is short) then this is okay.

ajaggapa:
> I think the principle of "update resources, verify existing is intact" shows up multiple times in several tests, not sure we need to have a dedicated test just for bgpAds.
> 

We had a customer [bug](https://issues.redhat.com/browse/OCPBUGS-36914) that has a similar issue which escaped our test suite and also the product tests. I think it is fine to accommodate this test.

oribon:
right, and it is now covered in "checking the sessions don't flap when changing the configuration"
so is this PR about checking updating an ad doesn't affect other ads, or is it about updating to a node selector for a label that doesn't exist withdraws the service (or both)?

ajaggapa:
> right, and it is now covered in "checking the sessions don't flap when changing the configuration" so is this PR about checking updating an ad doesn't affect other ads, or is it about updating to a node selector for a label that doesn't exist withdraws the service (or both)?

This covers both

e2etest/bgptests/node_selector.go
```diff
@@ -263,6 +263,92 @@ var _ = ginkgo.Describe("BGP Node Selector", func() {
 		ginkgo.Entry("IPV6", ipfamily.IPv6, "fc00:f853:0ccd:e799::/116"),
 	)
 
+	ginkgo.DescribeTable("Two pools, two services, two advertisements - update second advertisement to non-existing node label",
```
oribon: since this covers both withdrawal and isolation, the name should reflect that. something like "Two advertisements - updating one nodeSelector to non-matching withdraws its service without affecting the other"

e2etest/bgptests/node_selector.go
```diff
@@ -263,6 +263,92 @@ var _ = ginkgo.Describe("BGP Node Selector", func() {
...
+		func(pairingIPFamily ipfamily.Family, addresses []string) {
+			var allNodes *corev1.NodeList
+			allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			if len(allNodes.Items) < 2 {
+				ginkgo.Skip("This test requires at least 2 nodes in the cluster")
+			}
+			expectedNodesForFirstAdv := nodesForSelection(allNodes.Items, []int{0})
+			expectedNodesForSecondAdv := nodesForSelection(allNodes.Items, []int{1})
```
oribon: this can be one node or all nodes (isolation happens on pool level anyways)

e2etest/bgptests/node_selector.go
```diff
@@ -263,6 +263,92 @@ var _ = ginkgo.Describe("BGP Node Selector", func() {
...
+				},
+				Peers: metallb.PeersForContainers(FRRContainers, pairingIPFamily),
+				BGPAdvs: []metallbv1beta1.BGPAdvertisement{
+					{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: "first-adv",
+						},
+						Spec: metallbv1beta1.BGPAdvertisementSpec{
+							NodeSelectors:   k8s.SelectorsForNodes(expectedNodesForFirstAdv),
+							Communities:     []string{CommunityNoAdv},
```
oribon: I think communities are not relevant for the test

e2etest/bgptests/node_selector.go
```diff
@@ -263,6 +264,104 @@ var _ = ginkgo.Describe("BGP Node Selector", func() {
...
+			resources.BGPAdvs[1].Spec.NodeSelectors = []metav1.LabelSelector{
+				{MatchLabels: map[string]string{"non-existing-label": "true"}},
+			}
+			err = ConfigUpdater.Update(resources)
+			Expect(err).NotTo(HaveOccurred())
+
+			ginkgo.By("Validating first service still advertised only from first advertisement's nodes")
+			checkServiceOnlyOnNodes(firstService, expectedNodesForFirstAdv, pairingIPFamily)
+			ginkgo.By("Validating second service no longer advertised from second advertisement's nodes")
+			checkServiceNotOnNodes(secondService, expectedNodesForSecondAdv, pairingIPFamily)
```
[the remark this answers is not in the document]
oribon: right, this needs to check that the service isn't advertised at all

e2etest/bgptests/node_selector.go
ajaggapa: done

e2etest/bgptests/node_selector.go
ajaggapa: done

ajaggapa:
> this can be one node or all nodes (isolation happens on pool level anyways)

@oribon 
I assume we can proceed with one node as is. 
Apart from this, i addressed other comments.

oribon:
> I assume we can proceed with one node as is.

so let's remove the code around having 2 nodes, selecting them etc.

ajaggapa:
> > I assume we can proceed with one node as is.
> 
> so let's remove the code around having 2 nodes, selecting them etc.

Sure. I have updated accordingly.

e2etest/bgptests/node_selector.go
```diff
@@ -263,6 +263,85 @@ var _ = ginkgo.Describe("BGP Node Selector", func() {
...
+
+			err = ConfigUpdater.Update(resources)
+			Expect(err).NotTo(HaveOccurred())
+
+			firstService, _ := testservice.CreateWithBackend(cs, testNamespace, "first-lb", testservice.WithSpecificPool("first-pool"))
+			defer testservice.Delete(cs, firstService)
+			secondService, _ := testservice.CreateWithBackend(cs, testNamespace, "second-lb", testservice.WithSpecificPool("second-pool"))
+			defer testservice.Delete(cs, secondService)
+
+			ginkgo.By("Checking first service announced only from first advertisement's nodes")
```
oribon: nit: replace leftover first/second nodes with "all nodes" (here and the other By)

e2etest/bgptests/node_selector.go
```diff
@@ -263,6 +263,85 @@ var _ = ginkgo.Describe("BGP Node Selector", func() {
...
+			checkServiceOnlyOnNodes(firstService, allNodes.Items, pairingIPFamily)
```
oribon: nit: use ValidateService on all nodes (only on nodes "all nodes" reads a bit odd)

ajaggapa: Updated. But continuing to use `checkServiceNotOnNodes`

PR #2977, opened 2026-03-11
e2etests: fix wrong connectivity check

oribon:

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etest/bgptests/bgp.go
+++ b/e2etest/bgptests/bgp.go
@@ -970,7 +970,7 @@ var _ = ginkgo.Describe("BGP", func() {
 
 					ginkgo.By(fmt.Sprintf("checking connectivity of service %d to its external VIP", j+1))
 					for _, c := range FRRContainers {
-						validateService(svc, allNodes.Items, c)
+						validateService(services[j], allNodes.Items, c)
 					}
 				}
 			}


e2etest/bgptests/bgp.go
andreaskaris, line 967: not part of this commit, but why is it logging j+1 when it's all about j?

andreaskaris: the comment in line 971 the same ..

andreaskaris:
/lgtm
I'm just not getting why we're logging the index + 1, I find that more confusing than helpful

oribon:
probably so it logs service 1, 2 (and not 0,1), I agree it might be better to shift some indices instead (maybe in a separate pr 😅 )

andreaskaris:
> probably so it logs service 1, 2 (and not 0,1), I agree it might be better to shift some indices instead (maybe in a separate pr 😅 )

Actually, after looking at it .. there seems to be a logic to this madness ;-) the incremented service ids refer to the actual name which refer to the actual ipaddresspool and name which match the IP address range ...

PR #2979, opened 2026-03-11
Replace kube-rbac-proxy with native TLS and RBAC

oribon:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
Replace kube-rbac-proxy with native TLS and RBAC, which enables configuring some TLS parameters. The old HTTP endpoints are no longer available, they are now HTTPS served by self-signed certificates / certificates passed to the binaries.
```


go.mod
```diff
@@ -20,7 +20,7 @@ require (
 	github.com/ory/dockertest/v3 v3.12.0
 	github.com/prometheus/client_golang v1.22.0
 	go.uber.org/zap v1.27.0
-	golang.org/x/sys v0.35.0
```
andreaskaris, line 23: maybe go.mod / go.sum stuff in a different commit?

controller/main.go
```diff
@@ -158,23 +158,23 @@ func (c *controller) SetPools(l log.Logger, pools *config.Pools) controllers.Syn
...
 		webhookHTTP2        = flag.Bool("webhook-http2", false, "enables http2 for the webhook endpoint")
-		tlsMinVersion       = flag.String("tls-min-version", "", "Minimum TLS version supported for the webhook server, Possible values: "+strings.Join(cliflag.TLSPossibleVersions(), ", "))
-		tlsCipherSuites     = flag.String("tls-cipher-suites", "TLS_AES_128_GCM_SHA256,TLS_AES_256_GCM_SHA384,TLS_CHACHA20_POLY1305_SHA256,TLS_ECDHE_ECDSA_WITH_AES_128_GCM_SHA256,"+
-			"TLS_ECDHE_RSA_WITH_AES_128_GCM_SHA256,TLS_ECDHE_ECDSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_RSA_WITH_AES_256_GCM_SHA384,TLS_ECDHE_ECDSA_WITH_CHACHA20_POLY1305_SHA256,"+
-			"TLS_ECDHE_RSA_WITH_CHACHA20_POLY1305_SHA256,TLS_RSA_WITH_AES_128_GCM_SHA256,TLS_RSA_WITH_AES_256_GCM_SHA384", "Comma-separated list of cipher suites for the webhook server")
+		tlsMinVersion       = flag.String("tls-min-version", "", "Minimum TLS version (VersionTLS12 or VersionTLS13). If empty, defaults to VersionTLS13.")
+		tlsCipherSuites     = flag.String("tls-cipher-suites", "", "Comma-separated list of TLS cipher suites. If empty, uses Go defaults.")
```
andreaskaris: qq, in the helm templates, it says:
```
| tls.cipherSuites | string | `""` | Comma-separated list of TLS cipher suites. If empty, uses Go defaults. Only applies to TLS 1.2.
```
would it make sense to put this in the comment here, too?

frr-tools/metrics/vtysh/vtysh.go
```diff
@@ -11,7 +11,7 @@ import (
 type Cli func(args string) (string, error)
 
 func Run(args string) (string, error) {
-	out, err := exec.Command("/usr/bin/vtysh", "-c", args).CombinedOutput()
+	out, err := exec.Command("/usr/bin/vtysh", "-c", args).CombinedOutput() //#nosec G204 -- args are hardcoded vtysh commands from within this package
```
andreaskaris: nit: is this unrelated; --> differernt commit?

frr-tools/metrics/exporter.go
```diff
@@ -14,42 +14,30 @@ import (
...
 var (
-	metricsHost = flag.String("host", os.Getenv("METALLB_HOST"), "metrics HTTP host address")
-	metricsPort = flag.Uint("metrics-port", 7473, "Port to listen on for web interface.")
-	metricsPath = flag.String("metrics-path", "/metrics", "Path under which to expose metrics.")
+	metricsPort         = flag.Uint("metrics-port", 9121, "Port to listen on for web interface.")
+	metricsBindAddress  = flag.String("metrics-bind-address", "0.0.0.0", "The address the metric endpoint binds to.")
+	metricsPath         = flag.String("metrics-path", "/metrics", "Path under which to expose metrics.")
+	tlsCertFile         = flag.String("tls-cert-file", "", "x509 certificate for HTTPS. If empty, a self-signed cert is auto-generated.")
+	tlsKeyFile          = flag.String("tls-private-key-file", "", "x509 private key matching --tls-cert-file.")
+	tlsCipherSuites     = flag.String("tls-cipher-suites", "", "Comma-separated list of TLS cipher suites. If empty, uses Go defaults.")
```
andreaskaris: same q as above

internal/tlsconfig/tlsopt.go
```diff
@@ -0,0 +1,101 @@
...
+func parseCurvePreferences(curves string) ([]tls.CurveID, error) {
+	if curves == "" {
+		return nil, nil
+	}
+	parts := strings.Split(curves, ",")
+	seen := make(map[int32]bool, len(parts))
+	result := make([]tls.CurveID, 0, len(parts))
+	for _, p := range parts {
+		p = strings.TrimSpace(p)
+		n, err := strconv.ParseInt(p, 10, 32)
```
andreaskaris: why not ParseUint(p, 10, 16)?
https://go.dev/play/p/qRVIf-aacMl
and slightly tweaking the error msg say it must be a positive integer <  math.MaxUint16

(yes, I know it's checked right after with the library function.  but it'd catch the issue early, and in the current code at least the entire < 0 > maxuint check isn't needed)

internal/tlsconfig/tlsopt.go
```diff
@@ -0,0 +1,101 @@
...
+			cfg.CurvePreferences = curves
+		}
+		cfg.MinVersion = minVer
+	}, nil
+}
+
+// parseCurvePreferences parses a comma-separated list of numeric CurveID values.
+// This mirrors cliflag.TLSCurvePreferences from component-base v0.36+
+// (kubernetes/kubernetes#137115), inlined here until we bump the dependency.
+func parseCurvePreferences(curves string) ([]tls.CurveID, error) {
```
andreaskaris, line 52: this mirrors  https://github.com/kubernetes/kubernetes/pull/137115/changes/53b163298d64c11e468ed0d6f49dd25ad871edae#diff-40677712edfbd262acf7e969b8576b42b8270dc35dbc2445616aacc4f2c3da76R29 , right?

Would it make sense to copy that method verbatim, and have the logic that parses 'curves string' into an []int32 separately? [0]. I know that's redundant loops, but now that the upstream PR merged, we might as well prepare for an easy, no-brainer replacement. I.e., bumping the library and then just changing line 
```
tlsCurvePreferences(curveIDs)  ---> k8s.TLSCurvePreferences(curveIDs)
```
or whatever

[0]
```
func parseCurvePreferences(curves string) ([]tls.CurveID, error) {
	if curves == "" {
		return nil, nil
	}

	parts := strings.Split(curves, ",")
	curveIDs := make([]int32, 0, len(parts))
	for _, p := range parts {
		p = strings.TrimSpace(p)
		n, err := strconv.ParseUint(p, 10, 16)
		if err != nil {
			return nil, fmt.Errorf("invalid curve preference %q: must be a numeric CurveID (0 - 65535)", p)
		}
		curveIDs = append(curveIDs, int32(n))
	}
	return tlsCurvePreferences(curveIDs)
}

// tlsCurvePreferences returns a list of Go's crypto/tls CurveID values from the ids passed.
// The supported values depend on the Go version used.
// See https://pkg.go.dev/crypto/tls#CurveID for values supported for each Go version.
// This mirrors cliflag.TLSCurvePreferences from component-base v0.36+
// (kubernetes/kubernetes#137115), inlined here until we bump the dependency.
func tlsCurvePreferences(curveIDs []int32) ([]tls.CurveID, error) {
	if len(curveIDs) == 0 {
		return nil, nil
	}
	seen := make(map[int32]bool, len(curveIDs))
	result := make([]tls.CurveID, 0, len(curveIDs))
	for _, id := range curveIDs {
		if id <= 0 || id > math.MaxUint16 {
			return nil, fmt.Errorf("curve preference %d is out of range (must be 1-%d)", id, math.MaxUint16)
		}
		if seen[id] {
			return nil, fmt.Errorf("duplicate curve preference %d", id)
		}
		seen[id] = true
		curve := tls.CurveID(id)
		if strings.HasPrefix(curve.String(), "CurveID(") {
			return nil, fmt.Errorf("curve preference %d is not supported by the current Go version", id)
		}
		result = append(result, curve)
	}
	return result, nil
}
```

internal/tlsconfig/tlsopt_test.go
```diff
@@ -0,0 +1,239 @@
...
+				t.Errorf("CurvePreferences = %v, want %v", cfg.CurvePreferences, tc.wantCurvePreferences)
+			}
+			if cfg.MinVersion != tc.wantMinVersion {
+				t.Errorf("MinVersion = %v, want %v", cfg.MinVersion, tc.wantMinVersion)
+			}
+		})
+	}
+}
+
+func equalUint16s(a, b []uint16) bool {
```
andreaskaris: why not https://pkg.go.dev/slices#Equal

shame on you, Cursor ..

internal/tlsconfig/tlsopt_test.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+	for i := range a {
+		if a[i] != b[i] {
+			return false
+		}
+	}
+	return true
+}
+
+func equalCurveIDs(a, b []tls.CurveID) bool {
```
andreaskaris: same thing, tls.CurveIDs are comparable, so slices.Equal is enough

internal/tlsconfig/tlsopt_test.go
```diff
@@ -0,0 +1,239 @@
...
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := parseCurvePreferences(tt.input)
+			if (err != nil) != tt.wantErr {
+				t.Fatalf("parseCurvePreferences(%q) error = %v, wantErr %v", tt.input, err, tt.wantErr)
+			}
+			if !tt.wantErr {
+				if !equalCurveIDs(got, tt.want) {
```
andreaskaris: slices.Equal

internal/tlsconfig/tlsopt_test.go
```diff
@@ -0,0 +1,239 @@
...
+		},
+		{
+			name:    "non-numeric input returns error",
+			input:   "X25519",
+			wantErr: true,
+		},
+		{
+			name:    "unsupported curve ID returns error",
+			input:   "99999",
+			wantErr: true,
```
andreaskaris: nit: it's a pet peeve of mine and I hate that the LLMs are doing it this way by default, but I cannot stand the wantErr: true pattern; I prefer checking for string contains. but leave it this way if you want to, it's a matter of personal preference

speaker/main.go
```diff
@@ -74,27 +76,31 @@ type service interface {
...
+		myPod               = flag.String("pod-name", os.Getenv("METALLB_POD_NAME"), "name of this MetalLB speaker pod")
+		port                = flag.Int("port", 9120, "HTTPS metrics listening port")
+		logLevel            = flag.String("log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", logging.Levels.String()))
+		pprofBindAddress    = flag.String("pprof-bind-address", "", "Bind address for pprof endpoint (e.g. 127.0.0.1:6060). Empty disables pprof.")
+		loadBalancerClass   = flag.String("lb-class", "", "load balancer class. When enabled, metallb will handle only services whose spec.loadBalancerClass matches the given lb class")
+		ignoreLBExclude     = flag.Bool("ignore-exclude-lb", false, "ignore the exclude-from-external-load-balancers label")
+		frrK8sNamespace     = flag.String("frrk8s-namespace", os.Getenv("FRRK8S_NAMESPACE"), "the namespace frr-k8s is being deployed on")
+		tlsMinVersion       = flag.String("tls-min-version", "", "Minimum TLS version (VersionTLS12 or VersionTLS13). If empty, defaults to VersionTLS13.")
+		tlsCipherSuites     = flag.String("tls-cipher-suites", "", "Comma-separated list of TLS cipher suites. If empty, uses Go defaults.")
```
andreaskaris: same q as earlier, mention that this can only be set if tls 1.2 ?

andreaskaris (changes requested):
a few nitpicks

also , the same as for the last one, I wouldn't pretend that I understand the entire big picture here, so please get @karampok 's feedback, as well

frr-tools/metrics/vtysh/vtysh.go
oribon: removed, gosec isn't enabled here (we should reenable - TODO reenable when gosec is compatible with go 1.23.6 😛 )

controller/main.go
oribon: done

frr-tools/metrics/exporter.go
oribon: done

internal/tlsconfig/tlsopt.go
oribon: done

internal/tlsconfig/tlsopt.go
oribon: done

internal/tlsconfig/tlsopt_test.go
oribon: done

internal/tlsconfig/tlsopt_test.go
oribon: done

internal/tlsconfig/tlsopt_test.go
oribon: done

internal/tlsconfig/tlsopt_test.go
oribon: done

speaker/main.go
oribon: done

charts/metallb/templates/controller.yaml
```diff
@@ -191,9 +173,11 @@ spec:
...
           secretName: metallb-webhook-cert
-      {{- if .Values.prometheus.controllerMetricsTLSSecret }}
+      {{- if .Values.tls.metricsTLSSecret }}
       - name: metrics-certs
         secret:
-          secretName: {{ .Values.prometheus.controllerMetricsTLSSecret }}
+          secretName: {{ .Values.tls.metricsTLSSecret }}
       {{- end }}
+      - name: tmp
+        emptyDir: {}
```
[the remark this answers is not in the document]
oribon: done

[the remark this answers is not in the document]
oribon: done

charts/metallb/templates/speaker.yaml
```diff
@@ -339,11 +348,9 @@ spec:
         {{- if .Values.speaker.livenessProbe.enabled }}
         livenessProbe:
           httpGet:
-            {{- if .Values.prometheus.secureMetricsPort }}
-            host: localhost
-            {{- end }}
-            path: /metrics
-            port: monitoring
+            path: /healthz
+            port: 17472
```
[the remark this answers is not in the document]
oribon: this is fine imo, it's an implementation detail and the port is only bound to 127.0.0.1

charts/metallb/values.yaml
```diff
@@ -14,6 +14,16 @@ rbac:
   # create specifies whether to install and use RBAC rules.
   create: true
 
+tls:
+  # -- Comma-separated list of TLS cipher suites. If empty, uses Go defaults. Only applies to TLS 1.2.
+  cipherSuites: ""
+  # -- Comma-separated list of numeric CurveID values (e.g. 29,4588). See https://pkg.go.dev/crypto/tls#CurveID. If empty, uses Go defaults.
```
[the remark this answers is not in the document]
oribon: nope, the linked url specifically mentions:
```go
X25519MLKEM768     CurveID = 4588
```

charts/metallb/templates/controller.yaml
```diff
@@ -104,11 +110,17 @@ spec:
...
+          mountPath: /etc/metrics
+          readOnly: true
+        {{- end }}
         {{- if .Values.controller.livenessProbe.enabled }}
         livenessProbe:
           httpGet:
-            path: /metrics
-            port: monitoring
+            path: /healthz
+            port: 17472
```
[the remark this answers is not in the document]
oribon: this is fine imo, it's an implementation detail and the port is only bound to 127.0.0.1

charts/metallb/templates/controller.yaml
```diff
@@ -136,37 +149,6 @@ spec:
           capabilities:
             drop:
             - ALL
-      {{- if .Values.prometheus.secureMetricsPort }}
```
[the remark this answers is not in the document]
oribon: afaik this is not correct, from what I see CertDir is only used for reading in the code (and the godoc mentions "CertDir is the directory that contains the server key and certificate")
in any case, the "cert" volume mount stayed the same as before, and what populates it (the secret itself) is the cert-rotator and not the webhook.

internal/k8s/k8s.go
```diff
@@ -515,13 +503,13 @@ func webhookServer(cfg *Config) webhook.Server {
...
-		tlsConfig.MinVersion = cfg.WebHookMinVersion
-		tlsConfig.CipherSuites = cfg.WebHookCipherSuites
+	tlsOpts := []func(*tls.Config){disableHTTP2}
+	if cfg.TLSOpt != nil {
+		tlsOpts = append(tlsOpts, cfg.TLSOpt)
 	}
 
 	webhookServerOptions := webhook.Options{
-		TLSOpts: []func(config *tls.Config){disableHTTP2, tlsSecurity},
+		TLSOpts: tlsOpts,
```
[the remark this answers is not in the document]
oribon: hot-reload is baked into the webhook-server, if GetCertificate is not set (which is the case for us) the server creates its own cert watcher https://github.com/kubernetes-sigs/controller-runtime/blob/f9589b9f2b9dddf8532b432bb8315f2820ab9971/pkg/webhook/server.go#L201

controller/main.go
```diff
@@ -158,23 +158,23 @@ func (c *controller) SetPools(l log.Logger, pools *config.Pools) controllers.Syn
...
 func main() {
 	var (
-		port                = flag.Int("port", 7472, "HTTP listening port for Prometheus metrics")
+		port                = flag.Int("port", 9120, "HTTPS listening port for Prometheus metrics")
 		namespace           = flag.String("namespace", os.Getenv("METALLB_NAMESPACE"), "config / memberlist secret namespace")
 		mlSecret            = flag.String("ml-secret-name", os.Getenv("METALLB_ML_SECRET_NAME"), "name of the memberlist secret to create")
 		deployName          = flag.String("deployment", os.Getenv("METALLB_DEPLOYMENT"), "name of the MetalLB controller Deployment")
 		logLevel            = flag.String("log-level", "info", fmt.Sprintf("log level. must be one of: [%s]", logging.Levels.String()))
-		enablePprof         = flag.Bool("enable-pprof", false, "Enable pprof profiling")
+		pprofBindAddress    = flag.String("pprof-bind-address", "", "Bind address for pprof endpoint (e.g. 127.0.0.1:6060). Empty disables pprof.")
```
karampok, line 166: nit: this rename should be separate commit? or drop. Why you added it?

internal/k8s/k8s.go
```diff
@@ -149,17 +145,29 @@ func New(cfg *Config) (*Client, error) {
...
-	})
+	metricsOpts := metricsserver.Options{
+		BindAddress:    fmt.Sprintf("0.0.0.0:%d", cfg.MetricsPort),
+		SecureServing:  true,
+		FilterProvider: filters.WithAuthenticationAndAuthorization,
+	}
+	if cfg.MetricsCertDir != "" {
+		metricsOpts.CertDir = cfg.MetricsCertDir
+	}
+	if cfg.TLSOpt != nil {
```
karampok, line 156: this can never be nil, right?

karampok (approved):

controller/main.go
oribon: it's because we removed the custom http server, so the easiest way to preserve pprof is using controller-runtime's default, I think it makes sense to keep it in this commit

internal/k8s/k8s.go
oribon: done

karampok (approved):

andreaskaris:
I did a super quick review and my comments were addressed. Also, as mentioned before, I don't understand all of this commit, so I'm glad kk also reviewed it.
/lgtm can you rebase?

volker-raschek:
Hey, when will be this fix available? It seems that the container image gcr.io/kubebuilder/kube-rbac-proxy is not longer available and breaks kubernetes environments

oribon:
@volker-raschek hi, we plan on doing a new release soon
for now you can use the same fix as in https://github.com/metallb/metallb/pull/2974/changes

PR #2980, opened 2026-03-11
Fix/bgppeer nodeselector aware duplicate check

### LINKED ISSUE #2367: Relax the single BGPPeer per neighbor IP constraint

roth-wine:

  /kind bug                                                                                                                                  
   
  In FRR mode, MetalLB runs one FRR instance per node. BGPPeers can be
  scoped to specific nodes via `nodeSelectors`, meaning two BGPPeer
  resources with the same `peerAddress+VRF` are not actual duplicates
  from FRR's perspective as long as their nodeSelectors target different
  nodes.

  Previously, `DiscardNativeOnly` rejected any two BGPPeers sharing the
  same `peerIdentifier` unconditionally, making it impossible to
  configure per-node BGP sessions to the same router IP — a common
  requirement in setups where each control-plane node peers individually.

  This PR makes the duplicate check node-aware: when two BGPPeers share
  the same address+VRF, the admission webhook fetches the cluster's node
  list and checks whether any node matches both peers' nodeSelectors. A
  duplicate error is only raised if the peers could actually land on the
  same FRR instance. If no nodes are known at validation time, the check
  falls back conservatively and rejects as before.

  Multiple `nodeSelectors` per peer are evaluated with OR semantics,
  matching MetalLB's existing node-selection behaviour.

  Fixes #2367

  PR #2802 addresses the same issue but relies on static label analysis:                                                                     
  it checks whether two `nodeSelectors` have contradicting `matchLabels`                                                                     
  keys (e.g. `zone=a` vs `zone=b`) to determine if they are disjoint.                                                                        
  This heuristic works for simple `matchLabels` cases but cannot                       
  guarantee that the selectors actually target different nodes — two
  selectors with different key names or using `matchExpressions` may
  still match the same node in the actual cluster.

  This PR validates against real nodes instead: the admission webhook
  fetches the cluster's node list and checks whether any node satisfies
  both peers' selectors, giving a definitive answer regardless of
  selector complexity. No RBAC changes are required as the controller
  already has `list` on nodes.

  **Release note**:

  ```release-note
  Allow multiple BGPPeer resources with the same peer address when their
  nodeSelectors are guaranteed to target different nodes, enabling
  per-node BGP session configuration in FRR mode.


internal/k8s/webhooks/webhookv1beta2/bgppeer_webhook.go
```diff
@@ -25,7 +25,9 @@ import (
 
 	"github.com/go-kit/log/level"
 	"go.universe.tf/metallb/api/v1beta2"
+	"go.universe.tf/metallb/internal/config"
 	v1 "k8s.io/api/admission/v1"
+	corev1 "k8s.io/api/core/v1"
 	ctrl "sigs.k8s.io/controller-runtime"
```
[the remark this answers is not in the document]
roth-wine: This is unrelated to my changes.

roth-wine:
@fedepaol @oribon Could you please have a look? :smile:

Jakob3xD:
We also build an image with this patch applied and it works for our clusters.

internal/config/validation.go
```diff
@@ -218,14 +235,66 @@ func hasBFDEcho(peer *Peer, bfdProfiles map[string]*BFDProfile) bool {
...
+		if c1.matches(nodeLabels) && c2.matches(nodeLabels) {
+			return true
+		}
+	}
+	return false
+}
+
+// compiledNodeSelectors holds pre-parsed label selectors for a peer.
+// A nil value means "match all nodes" (empty selector list in the spec).
+type compiledNodeSelectors []labels.Selector
```
fedepaol: I'd call this just selectors or nodeSelectors

internal/config/validation.go
```diff
@@ -218,14 +235,66 @@ func hasBFDEcho(peer *Peer, bfdProfiles map[string]*BFDProfile) bool {
...
+
+func compileNodeSelectors(selectors []metav1.LabelSelector) (compiledNodeSelectors, error) {
```
fedepaol: selectorsFromLabels

fedepaol: also, we already have a function doing this https://github.com/metallb/metallb/blob/main/internal/config/config.go#L1118

internal/config/validation.go
```diff
@@ -218,14 +235,66 @@ func hasBFDEcho(peer *Peer, bfdProfiles map[string]*BFDProfile) bool {
...
+	if peer.VRFName == "" {
+		return id
+	}
 	return fmt.Sprintf("%s-%s", id, peer.VRFName)
 }
 
+// peersShareNodeCompiled returns true if any node matches both pre-compiled
+// nodeSelector sets, meaning the two peers would be configured on the same FRR
+// instance. Conservatively returns true when no nodes are known.
+func peersShareNodeCompiled(c1, c2 compiledNodeSelectors, nodes []corev1.Node) bool {
```
fedepaol: I think we can expand https://github.com/metallb/metallb/blob/main/internal/config/config.go#L1136 or leverage it

internal/config/validation.go
fedepaol: or even remove the alias, unless we use it as entry in a map (which I don't think is the case)

internal/config/validation.go
```diff
@@ -145,29 +146,45 @@ func DontValidate(c ClusterResources) error {
...
-		for _, p := range c.Peers[1:] {
-			if p.Spec.RouterID != routerID {
-				return fmt.Errorf("peer %s has RouterID different from %s, in FRR mode all RouterID must be equal", p.Spec.RouterID, c.Peers[0].Spec.RouterID)
-			}
-			peerID := peerIdentifier(p.Spec)
-			if _, ok := peerAddr[peerID]; ok {
-				return fmt.Errorf("peer %s already exists, FRR mode doesn't support duplicate BGPPeers", p.Spec.Address)
+	// peerEntry holds a BGPPeer together with its pre-compiled nodeSelectors so
+	// that compileNodeSelectors is called at most once per peer.
+	type peerEntry struct {
```
fedepaol: let's call this peerWithSelectors
If we call the variable nodeSelectors and the type is selector (meaning, we remove the alias), the comment is unnecessary

internal/config/validation.go
```diff
@@ -145,29 +146,45 @@ func DontValidate(c ClusterResources) error {
...
+		peer     metallbv1beta2.BGPPeer
+		compiled compiledNodeSelectors
+	}
+	// Multiple peers with the same address+VRF are allowed when their
+	// nodeSelectors are disjoint – each FRR instance only sees peers for its node.
+	peerAddr := make(map[string][]peerEntry)
```
fedepaol: peersPerAddress?

internal/config/validation.go
```diff
@@ -145,29 +146,45 @@ func DontValidate(c ClusterResources) error {
...
+	// vrf2asn tracks the first-seen MyASN per VRF to enforce uniformity in O(n).
```
fedepaol: let's call this localASNPerVRF

internal/config/validation.go
```diff
@@ -145,29 +146,45 @@ func DontValidate(c ClusterResources) error {
...
+	vrf2asn := make(map[string]struct {
+		asn    uint32
+		peerID string
+	})
+	for i, p := range c.Peers {
+		if i > 0 && p.Spec.RouterID != c.Peers[0].Spec.RouterID {
+			return fmt.Errorf("peer %s has RouterID different from %s, in FRR mode all RouterID must be equal", p.Spec.RouterID, c.Peers[0].Spec.RouterID)
```
fedepaol: we should relax this constraint with node selector too I think

internal/config/validation.go
fedepaol: also, the complexity added from the map and going o(n) is unnecessary as we'll never have more than a handful of peers. Please revert this part to the old version

internal/k8s/webhooks/webhookv1beta2/bgppeer_webhook.go
```diff
@@ -123,20 +123,52 @@ func validatePeerCreate(bgpPeer *v1beta2.BGPPeer) (string, error) {
...
-	err = Validator.Validate(toValidate)
+	err = Validator.Validate(validateArgs...)
 	if err != nil {
 		level.Error(Logger).Log("webhook", "bgppeer", "action", "update", "name", bgpPeer.Name, "namespace", bgpPeer.Namespace, "error", err)
 		return err
 	}
 	return nil
 }
 
+// buildValidateArgs fetches existing BGPPeers and, lazily, the Node list (only
```
fedepaol: I'd just provide peers + args. The cardinality is such that it should not be an issue (until it is, and then we measure it)

internal/config/validation_test.go
```diff
@@ -13,13 +13,13 @@ import (
 	"k8s.io/utils/ptr"
 )
 
-// peerOnNodes returns a BGPPeer selecting nodes by hostname (OR semantics).
-func peerOnNodes(addr string, hostnames ...string) v1beta2.BGPPeer {
+// peerOnNodes returns a BGPPeer for address "1.2.3.4" selecting nodes by hostname (OR semantics).
+func peerOnNodes(hostnames ...string) v1beta2.BGPPeer {
```
fedepaol, line 17: this commit should be squashed with the previous one

fedepaol (changes requested):

fedepaol:
Also, please add e2e tests

karampok:
The webhook approach  validates at BGPPeer admission time, but node labels can change independently. If a node is relabeled after the BGPPeers are created, two previously-accepted peers can silently become conflicting sessions on the same FRR instance, with no webhook re-triggering.

Would it make more sense to remove the  nodeSelector check from the webhook entirely and instead and make sure we have that be visible as configState error (if not already there)? I think that might be better that having a such granular check on the validation?

Jakob3xD:
Hi @karampok, you raise a valid concern. Here's the accurate picture of what's actually covered:                                                                                             
                                                                                                                                                                                               
What is covered:                              
The config controller already https://github.com/metallb/metallb/blob/main/internal/k8s/controllers/config_controller.go#L211, and   https://github.com/metallb/metallb/blob/main/internal/k8s/controllers/config_controller.go#L225-L234 specifically re-triggers reconciliation when node labels change. When that happens,     
DiscardNativeOnly re-runs with the updated node list and surfaces any new conflicts as a configurationState error. We verified this in a real cluster: changing node labels that caused two peers to overlap immediately produced a ConfigurationError in the ConfigurationState.                                                                                                        
                                                                                         
What is not covered:                                                                                                                                                                         
The admission webhook cannot reject node label changes, as nodes don't go through this webhook. So the conflict becomes visible after the controller's next reconcile cycle as a configState error rather than an immediate rejection.                                                                                                                                                    
                                                                                         
The safety net for your scenario already exists. The webhook is an additional layer providing immediate feedback at kubectl apply time for the common case. Both mechanisms complement each other rather than replace one another.

karampok:
> The webhook is an additional layer providing immediate feedback at kubectl apply time for the common case

on the flip side, if you have large scale (many nodes0 and you have single confict, that is more a bug, imo you would prefer to allow and then check eventually all configState are ok.

fedepaol:
> > The webhook is an additional layer providing immediate feedback at kubectl apply time for the common case
> 
> on the flip side, if you have large scale (many nodes0 and you have single confict, that is more a bug, imo you would prefer to allow and then check eventually all configState are ok.

the additional complexity of the webhook is negligible _because we recycle the validation logic_. What the webhook does is just to collect resources and nodes and apply what the controller would do.

One thing that I would add is to check if what falls behind the webhook would pop up as a configuration error in the config state. By doing this we can have a configuration where we don't have conflicts but then cause the conflict by changing a label on a node (and we can't and we don't want to gate node changes in our webhooks).

Jakob3xD:
I added an e2e test for the node label change.

e2etest/configurationstatetests/configurationstate.go
```diff
@@ -189,6 +190,84 @@ var _ = ginkgo.Describe("ConfigurationState", func() {
 		}, 60*time.Second, 5*time.Second).Should(Succeed())
 	})
 
+	ginkgo.It("should surface configState error when node relabeling causes BGPPeer conflict in FRR mode", func() {
+		if BGPMode == "native" {
```
fedepaol: please let's not skip. We can drive the filtering from outside. Check how the github CI works

internal/config/validation.go
```diff
@@ -145,27 +146,27 @@ func DontValidate(c ClusterResources) error {
...
 func DiscardNativeOnly(c ClusterResources) error {
-	if len(c.Peers) > 1 {
-		peerAddr := make(map[string]bool)
-		routerID := c.Peers[0].Spec.RouterID
-		peer0 := peerIdentifier(c.Peers[0].Spec)
-		peerAddr[peer0] = true
-		for _, p := range c.Peers[1:] {
-			if p.Spec.RouterID != routerID {
-				return fmt.Errorf("peer %s has RouterID different from %s, in FRR mode all RouterID must be equal", p.Spec.RouterID, c.Peers[0].Spec.RouterID)
+	// For every pair of peers, check constraints that apply only when both peers
```
fedepaol: can we wrap this function in something more explicit? We are doing one single validation, so a wrapper would not be needed but the intent of the function is not clear.

Something like DiscardNativeOnly {
if err := checkPeersAreCompatible(); ...

// add other frrmode validations here

internal/config/validation.go
```diff
@@ -145,27 +146,27 @@ func DontValidate(c ClusterResources) error {
...
 			}
-			peerAddr[peerID] = true
-		}
-	}
-	for _, p := range c.Peers {
-		for _, p1 := range c.Peers[1:] {
-			if p.Spec.MyASN != p1.Spec.MyASN &&
-				p.Spec.VRFName == p1.Spec.VRFName {
-				return fmt.Errorf("peer %s has myAsn different from %s, in FRR mode all myAsn must be equal for the same VRF", p.Spec.Address, p1.Spec.Address)
+			if p.Spec.RouterID != q.Spec.RouterID {
```
fedepaol: let's wrap the checks in a "peersCanRunOnSameNode" function

fedepaol (changes requested):
Converging!

Jakob3xD:
@fedepaol To me those failing e2e tests look like timing issues. Should I increase the timeouts for those?

fedepaol:
@Jakob3xD took the liberty to push a couple of extra changes plus one test to validate the happy path.
Sending to merge queue

Jakob3xD:
> @Jakob3xD took the liberty to push a couple of extra changes plus one test to validate the happy path. Sending to merge queue

Re: FRR - Different ASN per node test failure (all 8 FRR jobs)                                                                                                                               
                                                                                                                                                                                             
  The new test BGP Node Selector / FRR - Different ASN per node / IPV4 - should establish sessions when each node has a different MyASN fails on every FRR variant because it incorrectly      
  iterates over all FRRContainers, including VRF containers (ebgp-vrf-single-hop, ebgp-vrf-multi-hop, etc.).
                                                                                                                                                                                               
  Two issues with VRF containers:                                                                                                                                                              
  1. The BGPPeer spec never sets VRFName, so MetalLB establishes the BGP session in the default Linux VRF rather than in the VRF the container uses.
  2. BGPConfigFromRouterConfig generates a flat router bgp <asn> config with no VRF routing context, so validateFRRPeeredWithNodes (which checks for neighbors in c.RouterConfig.VRF) never    
  finds them.                                                                                                                                                                              
                                                                                                                                                                                               
  The fix is to restrict the test to non-VRF containers by filtering with isVRFContainer(c). The test's purpose of validating that different MyASN per node is allowed when NodeSelectors are
  disjoint is fully exercised by the non-VRF containers alone.

fedepaol:
> > @Jakob3xD took the liberty to push a couple of extra changes plus one test to validate the happy path. Sending to merge queue
> 
> Re: FRR - Different ASN per node test failure (all 8 FRR jobs)
> 
> The new test BGP Node Selector / FRR - Different ASN per node / IPV4 - should establish sessions when each node has a different MyASN fails on every FRR variant because it incorrectly iterates over all FRRContainers, including VRF containers (ebgp-vrf-single-hop, ebgp-vrf-multi-hop, etc.).
> 
> Two issues with VRF containers:
> 
>     1. The BGPPeer spec never sets VRFName, so MetalLB establishes the BGP session in the default Linux VRF rather than in the VRF the container uses.
> 
>     2. BGPConfigFromRouterConfig generates a flat router bgp  config with no VRF routing context, so validateFRRPeeredWithNodes (which checks for neighbors in c.RouterConfig.VRF) never
>        finds them.
> 
> 
> The fix is to restrict the test to non-VRF containers by filtering with isVRFContainer(c). The test's purpose of validating that different MyASN per node is allowed when NodeSelectors are disjoint is fully exercised by the non-VRF containers alone.

ah apologies, yes I tried locally with the default deployment which does not include vrf-ied containers, will fix and repush!

fedepaol:
Failures not related to this imo, merging

Jakob3xD:
@fedepaol thank you for your efforts. :tada:

PR #2981, opened 2026-03-13
speaker: add periodic gratuitous ARP support

### LINKED ISSUE #2910: L2 mode: Support periodic gratuitous ARP or document FDB aging limitation
## Problem

MetalLB L2 only sends gratuitous ARP during failover. When switch FDB aging time < gateway ARP timeout, traffic blackholes occur.

**Example:**
- Switch FDB aging: 300s
- Gateway ARP timeout: 1200s
- Result: 15-min blackhole window (300s-1200s) every 20 minutes

**Why:** Gateway has cached ARP so doesn't send new ARP requests, but switch FDB already aged out.

## Verification
```bash
cat /sys/class/net/br341/bridge/ageing_time  # 30000 (300s)
# After 5 min: FDB gone, ARP cached, traffic blackholed

# set arp expire time on switch 1200s
```

## Request

**Option 1 (Preferred):** Add periodic GARP support
```yaml
kind: L2Advertisement
spec:
  gratuitousARPInterval: 60  # Send GARP every 60s (0=disabled)
```

**Option 2 (Minimum):** Document this limitation
- Warn users: Ensure `FDB aging > ARP timeout` 
- Recommend BGP mode for strict SLA requirements

---
Filing this to help others avoid the same issue.

### Additional context

_No response_

voron:

>
/kind bug

Fixes: metallb/metallb#2910
We hit regular packet loss issue with Hetzner public IPs in vSwitch. Gratuitous ARP fixed it in our tests.

Adds a speaker-wide `--gratuitous-arp-interval` flag (also configurable via the `METALLB_GRATUITOUS_ARP_INTERVAL` environment variable) that enables periodic gratuitous ARP/NDP announcements at a user-defined interval, in seconds. This prevents traffic blackholes caused by switch FDB entry aging when FDB aging time is shorter than gateway ARP timeout.

When set to 0 (default), only failover announcements are sent (existing behavior). The interval applies to every IP currently announced by this speaker via L2; per-`L2Advertisement` configuration is intentionally not exposed here to keep the surface area small — the pain point this addresses is environmental (switch/network), not per-service.

**Release note**:

```release-note
Add a speaker --gratuitous-arp-interval flag (also configurable via METALLB_GRATUITOUS_ARP_INTERVAL) to enable periodic gratuitous ARP/NDP announcements
```

### DIFF
--- a/internal/layer2/announcer.go
+++ b/internal/layer2/announcer.go
@@ -32,8 +32,10 @@ type Announce struct {
 	excludeRegexp *regexp.Regexp
 }
 
-// New returns an initialized Announce.
-func New(l log.Logger, excludeRegexp *regexp.Regexp) (*Announce, error) {
+// New returns an initialized Announce. If gratuitousARPInterval is positive,
+// a background goroutine periodically sends gratuitous ARP/NDP for all
+// announced IPs at that interval.
+func New(l log.Logger, excludeRegexp *regexp.Regexp, gratuitousARPInterval time.Duration) (*Announce, error) {
 	ret := &Announce{
 		logger:         l,
 		nodeInterfaces: []string{},
@@ -47,6 +49,10 @@ func New(l log.Logger, excludeRegexp *regexp.Regexp) (*Announce, error) {
 
 	go ret.interfaceScan()
 	go ret.spamLoop()
+	if gratuitousARPInterval > 0 {
+		level.Info(l).Log("op", "periodicGARP", "msg", "periodic gratuitous ARP enabled", "interval", gratuitousARPInterval)
+		go ret.periodicGARPLoop(gratuitousARPInterval)
+	}
 
 	return ret, nil
 }
@@ -198,6 +204,62 @@ func (a *Announce) spamLoop() {
 
 func (a *Announce) doSpam(adv IPAdvertisement) {
 	a.spamCh <- adv
+}
+
+func (a *Announce) periodicGARPLoop(interval time.Duration) {
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+	for range ticker.C {
+		a.sendPeriodicGratuitous()
+	}
+}
+
+func (a *Announce) sendPeriodicGratuitous() {
+	for _, adv := range a.mergedAdvertisementsPerIP() {
+		a.gratuitous(adv)
+	}
+}
+
+// mergedAdvertisementsPerIP returns one IPAdvertisement per unique IP,
+// merging interface constraints across all advertisements for that IP:
+// allInterfaces=true if any source advertisement has it, otherwise the
+// union of interface sets. This ensures periodic gratuitous announcements
+// reach every interface where the IP is being advertised.
+func (a *Announce) mergedAdvertisementsPerIP() []IPAdvertisement {
+	a.RLock()
+	defer a.RUnlock()
+
+	merged := map[string]IPAdvertisement{}
+	for _, ipAdvs := range a.ips {
+		for _, adv := range ipAdvs {
+			ipStr := adv.ip.String()
+			existing, ok := merged[ipStr]
+			if !ok {
+				m := adv
+				if !m.allInterfaces {
+					m.interfaces = m.interfaces.Clone()
+				}
+				merged[ipStr] = m
+				continue
+			}
+			if existing.allInterfaces {
+				continue
+			}
+			if adv.allInterfaces {
+				existing.allInterfaces = true
+				existing.interfaces = nil
+			} else {
+				existing.interfaces = existing.interfaces.Union(adv.interfaces)
+			}
+			merged[ipStr] = existing
+		}
+	}
+
+	advs := make([]IPAdvertisement, 0, len(merged))
+	for _, adv := range merged {
+		advs = append(advs, adv)
+	}
+	return advs
 }
 
 func (a *Announce) gratuitous(adv IPAdvertisement) {
--- a/internal/layer2/announcer_test.go
+++ b/internal/layer2/announcer_test.go
@@ -62,3 +62,98 @@ func Test_SetBalancer_AddsToAnnouncedServices(t *testing.T) {
 		t.Fatalf("ip 192.168.1.20 has not 2 refcnt: %d", announce.ipRefcnt["192.168.1.20"])
 	}
 }
+
+func Test_MergedAdvertisementsPerIP(t *testing.T) {
+	ipA := net.IPv4(192, 168, 1, 100)
+	ipB := net.IPv4(192, 168, 1, 101)
+
+	tests := []struct {
+		name              string
+		ips               map[string][]IPAdvertisement
+		wantAllInterfaces map[string]bool
+		wantInterfaces    map[string]sets.Set[string]
+	}{
+		{
+			name: "allInterfaces wins over interface set",
+			ips: map[string][]IPAdvertisement{
+				"svc1": {{ip: ipA, allInterfaces: true}},
+				"svc2": {{ip: ipA, interfaces: sets.New("eth0"), allInterfaces: false}},
+			},
+			wantAllInterfaces: map[string]bool{ipA.String(): true},
+		},
+		{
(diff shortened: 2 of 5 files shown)


champtar:
An other possibility for your problem is having too small CPU limits/requests on the speaker, causing the speaker to take multiple seconds to send the ARP response, so the response is ignored, and you end up getting packet drops

voron:
> An other possibility for your problem is having too small CPU limits/requests on the speaker, causing the speaker to take multiple seconds to send the ARP response, so the response is ignored, and you end up getting packet drops

Doesn't seem to be our case:
* no cpu limits
* idle nodes and NICs
* arp reply is within 100 microseconds in tcpdump
* problem affects OS IPs as well, not just metalLB ones. We have to increase OS arp cache TTL to 30+ minutes to prevent a gateway to learn our mac from OS arp requests to get the problem triggered though
* problem affects just one location with servers across multiple locations
* problem appears and disappears on it's own depending on what gateway is active. I assume some VRRP-like staff is used on datacenter side, and gateways are configured in a different way

champtar:
Curious what workaround you use for the OS IPs ? Being forced to send gratuitous ARP constantly seems crazy to me, at some point Hertzner need to provide a working network ...

voron:
> Curious what workaround you use for the OS IPs ? Being forced to send gratuitous ARP constantly seems crazy to me, at some point Hertzner need to provide a working network ...

AI generates ansible manifests with systemd units with workarounds like python script to send gratuitous ARP in no time. Flaky network issues are a bit challenging to prove to the hosting provider. Not worth the time usually in the presence of 5 min workarounds.

We use just one public IP in OS on vSwitch interface. It means OS uses it to perform ARP requests to get gateway's MAC address. From what I see, Hetzner gateway learns MAC address of that single IP via these ARP requests OS sends. 
So I have to increase ARP cache TTL in OS on vSwitch interface to stop OS sending ARP requests to get closer to what we have with MetalLB.  

I've added 2'nd public IP into OS on vSwitch interface and the issue is the same as with MetalLB IP, no OS ARP cache TTL tuning needed to see the issue.

speaker/layer2_controller_test.go
```diff
@@ -1987,3 +1987,97 @@ func TestIPAdvertisementFor(t *testing.T) {
...
+				bgpType: bgpNative,
+			})
+			if err != nil {
+				t.Fatalf("creating controller: %s", err)
+			}
+			c.client = &testK8S{t: t}
+
+			result := c.SetConfig(l, test.config)
+			if result == controllers.SyncStateError {
+				t.Errorf("SetConfig returned error for %q", test.desc)
```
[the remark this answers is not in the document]
voron: Claude response
>checking SyncStateError would never match anyway since the main controller returns SyncStateErrorNoRetry on handler error. But looking at existing tests in the same file, they all use the same == controllers.SyncStateError pattern (lines 1228, 1231, 1508, 1511, 1599, 1602, 1829). So this is an existing codebase convention, not something we introduced. Fixing just ours would be inconsistent.

speaker/layer2_controller_test.go
```diff
@@ -1987,3 +1987,97 @@ func TestIPAdvertisementFor(t *testing.T) {
...
+			}
+		})
+	}
```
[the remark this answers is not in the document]
voron: Claude response
> It is a fair point but would require either adding an exported getter to Announce (adding API surface just for tests) or introducing a mock/interface for the announcer (significant refactor not scoped to this PR). The existing tests in this file follow the same pattern — they verify SetConfig doesn't error, then test behavior through ShouldAnnounce. No existing test verifies internal state of the announcer after SetConfig.

api/v1beta1/l2advertisement_types.go
```diff
@@ -43,6 +43,14 @@ type L2AdvertisementSpec struct {
 	// If empty, all services from the selected pools are advertised.
 	// +optional
 	ServiceSelectors []metav1.LabelSelector `json:"serviceSelectors,omitempty"`
+	// GratuitousARPInterval specifies the interval in seconds between periodic gratuitous ARP
```
fedepaol, line 46: I don't think there's value in having such parameter _ per service_.
I'd rather have a global parameter exposed as a helm value

internal/layer2/announcer.go
```diff
@@ -200,6 +211,56 @@ func (a *Announce) doSpam(adv IPAdvertisement) {
 	a.spamCh <- adv
 }
 
+// SetGratuitousARPInterval configures the interval for periodic gratuitous ARP/NDP announcements.
+// A zero duration disables periodic announcements.
+func (a *Announce) SetGratuitousARPInterval(interval time.Duration) {
```
fedepaol, line 216: imo this can be a global parameter, no need for it to be dynamic

fedepaol (changes requested):

fedepaol:
we should also try to understand if there's a way to e2e the new behavior, maybe deleting the neighbors on the host when we test

voron:
@fedepaol WDYT on this rework with the global parameter ? No extra e2e for now though.

voron:
I'm still interested in this PR. Testing it for a few weeks, no issues noticed

fedepaol:
> I'm still interested in this PR. Testing it for a few weeks, no issues noticed

apologies, will review!

commercial-maw:
Hi, just wanted to check in and ask if there are any plans to take another look at this PR in the near future.

Thanks!

internal/layer2/announcer.go
```diff
@@ -200,6 +206,62 @@ func (a *Announce) doSpam(adv IPAdvertisement) {
...
+				existing.interfaces = nil
+			} else {
+				existing.interfaces = existing.interfaces.Union(adv.interfaces)
+			}
+			merged[ipStr] = existing
+		}
+	}
+
+	advs := make([]IPAdvertisement, 0, len(merged))
+	for _, adv := range merged {
```
fedepaol, line 259: you can just use copy here

speaker/layer2_controller.go
```diff
@@ -49,7 +49,7 @@ type layer2Controller struct {
 	onStatusChange  func(types.NamespacedName)
 }
 
-func (c *layer2Controller) SetConfig(log.Logger, *config.Config) error {
+func (c *layer2Controller) SetConfig(_ log.Logger, _ *config.Config) error {
```
fedepaol, line 52: not related

speaker/main.go
```diff
@@ -78,6 +83,18 @@ type service interface {
 func main() {
 	crmetrics.Registry.MustRegister(announcing)
 
+	// Resolve env-var default for gratuitous ARP interval before flag.Parse
+	// so an explicit --gratuitous-arp-interval=0 can override a non-zero env.
+	gratuitousARPDefault := 0
+	if envVal := os.Getenv("METALLB_GRATUITOUS_ARP_INTERVAL"); envVal != "" {
```
fedepaol, line 89: let's have it only as a parameter. Let's also expse the parameter in the helm chart

internal/layer2/announcer_test.go
```diff
@@ -62,3 +62,98 @@ func Test_SetBalancer_AddsToAnnouncedServices(t *testing.T) {
...
+
+func Test_SendPeriodicGratuitous_SkipsWhenNoIPs(t *testing.T) {
+	announce := &Announce{
+		ips:      map[string][]IPAdvertisement{},
+		ipRefcnt: map[string]int{},
+		spamCh:   make(chan IPAdvertisement, 1024),
+	}
+
+	// No IPs registered — should be a no-op.
+	announce.sendPeriodicGratuitous()
```
fedepaol, line 158: how does this test work?

fedepaol (changes requested):
Hey! Sorry for the super delay in the review. Looks good overall, I left some minor comments.

Also, as I wrote before I'd be happy to see some e2e tests, at least validating that the extra gratuitous arps / NDPs are being sent over.

fedepaol:
closing in favor of https://github.com/metallb/metallb/pull/3125

PR #2986, opened 2026-03-18
fix: handle lbIP family error by clearing state instead of continuing

Aprazor:
## Summary

In `convergeBalancer()`, when `ipfamily.ForAddressesIPs(lbIPs)` fails, the error is logged but execution continues to `serviceFamilyChanged()` using the potentially invalid `lbIPsIPFamily` value. The adjacent `ForService()` error correctly returns `ErrConverge`, but this path was missing equivalent handling.

The fix clears the service state and resets `lbIPs` on family detection failure, then wraps the family comparison check in an `else` branch so it only runs when family detection succeeds. This allows the convergence logic to re-allocate a valid IP rather than operating on an undefined family value.

## Changes

- `controller/service.go:101-119` — Added `clearServiceState` + `lbIPs` reset on `ForAddressesIPs` error, moved `ForService` and `serviceFamilyChanged` into an `else` block

## Testing

- Existing test `"simple LoadBalancer, ips already assigned but can't determine family"` validates this exact scenario — the service gets its state cleared and a valid IP re-allocated
- All controller tests pass

### DIFF
--- a/controller/service.go
+++ b/controller/service.go
@@ -100,8 +100,8 @@ func (c *controller) convergeBalancer(l log.Logger, key string, svc *v1.Service)
 	} else {
 		lbIPsIPFamily, err := ipfamily.ForAddressesIPs(lbIPs)
 		if err != nil {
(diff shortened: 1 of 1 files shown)


fedepaol:
I thikn the current version is good enough

PR #2992, opened 2026-03-20
e2etests: add new tests for BGPAdv/BGPPeer updates and dynamicASN test

ajaggapa:
### New e2e tests:
**Advertisement updates (IPv4/IPv6)**: After changing a BGPAdvertisement (communities, local preference, aggregation length), asserts FRR sees the updated community, local pref, and aggregated prefix.
**BGP timers Update**: Updates peer hold/keepalive, checks speaker FRR running-config, clears BGP on test FRR, then checks neighbors report the new timer values.
**BGP connect time**: Covers reconnect within the configured connect time after a session reset, and lowering connect time on an already-established session (default → 10s) with checks on speaker neighbor state.
**DynamicASN**: Negative test to cover the scenario where BGP establishment fails with incorrect dynamicASN configuration

**Is this a BUG FIX or a FEATURE ?**: New e2etests to expand coverage of bgptests to address BGPAdvertisement and BGPPeer updates. Also adding new tests to validate connect time feature and a negative test case for dynamicASN feature.

>
> /kind bug
 /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**What this PR does / why we need it**: Expand test coverage of bgptests

**Special notes for your reviewer**: Adds 5 new bgptests with negative scenario and CR updates.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etest/bgptests/bgp.go
+++ b/e2etest/bgptests/bgp.go
@@ -183,7 +183,7 @@ var _ = ginkgo.Describe("BGP", func() {
 			err = metallb.RestartSpeakerPods(cs)
 			Expect(err).NotTo(HaveOccurred())
 
-			if gracefulRestart == GracefulRestartDisabled {
+			if !gracefulRestart {
 				Eventually(func() error {
 					for _, c := range FRRContainers {
 						err := validateServiceNoWait(svc, allNodes.Items, c)
@@ -1564,7 +1564,6 @@ var _ = ginkgo.Describe("BGP", func() {
 
 				routers := map[string]frrk8sv1beta1.Router{}
 				for _, p := range resources.Peers {
-					p := p
 					r := routers[p.Spec.VRFName]
 					r.ASN = p.Spec.MyASN
 					r.VRF = p.Spec.VRFName
@@ -1833,7 +1832,7 @@ var _ = ginkgo.Describe("BGP", func() {
 						neighborFamily := ipfamily.ForAddress(net.ParseIP(neighbor.ID))
 						for _, family := range neighbor.AddressFamilies {
 							if !strings.Contains(family, string(neighborFamily)) {
-								return fmt.Errorf("expected %s neigbour to contain only %s families but contains %s", neighbor.ID, neighborFamily, family)
+								return fmt.Errorf("expected %s neighbour to contain only %s families but contains %s", neighbor.ID, neighborFamily, family)
 							}
 						}
 					}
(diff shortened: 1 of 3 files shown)


karampok (commented):
The commit
```
e2etests: add tests from eco-gotests
Made-with: Cursor
Signed-off-by: Anvesh J <ajaggapa@redhat.com>
Made-with: Cursor
```
is not very informative, it should state what we actually test.

In the description we should comment if there are are other similar tests.
At the end we need to know the specific code part which if we remote then from the all e2etest this would the only that fails only.

thanks @ajaggapa

ajaggapa:
> The commit
> 
> ```
> e2etests: add tests from eco-gotests
> Made-with: Cursor
> Signed-off-by: Anvesh J <ajaggapa@redhat.com>
> Made-with: Cursor
> ```
> 
> is not very informative, it should state what we actually test.
> 
> In the description we should comment if there are are other similar tests. At the end we need to know the specific code part which if we remote then from the all e2etest this would the only that fails only.
> 
> thanks @ajaggapa

@karampok 
I have updated the PR description according to the changes proposed. 
This is part of the few tests i have shared few days back highlights that non-duplicate tests in eco-gotests. 

All the CI jobs are successful now. Please review.

e2etest/bgptests/bgp.go
```diff
@@ -182,7 +183,7 @@ var _ = ginkgo.Describe("BGP", func() {
 			err = metallb.RestartSpeakerPods(cs)
 			Expect(err).NotTo(HaveOccurred())
 
-			if gracefulRestart == GracefulRestartDisabled {
+			if !gracefulRestart {
```
oribon: this is unrelated to this pr, and I think we should keep the former anyways

e2etest/bgptests/bgp.go
```diff
@@ -1278,6 +1279,131 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			Expect(routes).To(HaveKey(serviceIP), "BGP routes should contain service IP for community %s", CommunityNoAdv)
+		}
+
+		ginkgo.By("Validate BGP Local Preference received on FRR containers")
+		for _, c := range FRRContainers {
+			if strings.Contains(c.Name, "ibgp") {
+				localPref, err := frr.LocalPrefForPrefix(c, serviceIP, pairingIPFamily)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(localPref).To(Equal(uint32(100)))
+			}
```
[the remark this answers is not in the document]
ajaggapa: updated accordingly

e2etest/bgptests/bgp.go
```diff
@@ -1278,6 +1279,131 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(routes).To(HaveKey(expectedSubnet), "BGP routes should contain aggregated subnet for community %s", CustomCommunity)
+		}
+
+		ginkgo.By("Validate BGP Local Preference on FRR containers")
+		for _, c := range FRRContainers {
+			if strings.Contains(c.Name, "ibgp") {
+				localPref, err := frr.LocalPrefForPrefix(c, expectedSubnet, pairingIPFamily)
+				Expect(err).NotTo(HaveOccurred())
+				Expect(localPref).To(Equal(uint32(200)))
```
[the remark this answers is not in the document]
ajaggapa: updated

e2etest/pkg/frr/parse.go
oribon, line 1: can you update the parse unit tests as well?

e2etest/bgptests/bgp.go
```diff
@@ -1278,6 +1279,140 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			bgpAdv.Spec.LocalPref = 200
+			bgpAdv.Spec.AggregationLength = ptr.To(int32(28))
+			bgpAdv.Spec.Communities = []string{CustomCommunity}
+		case ipfamily.IPv6:
+			bgpAdv.Spec.LocalPref = 200
+			bgpAdv.Spec.AggregationLengthV6 = ptr.To(int32(124))
+			bgpAdv.Spec.Communities = []string{CustomCommunity}
+		default:
+			ginkgo.Fail("unsupported family for BGP advertisement updates test")
+		}
```
oribon: no need for this case (and failing explicitly), let's just pass the aggregation length in the entry

e2etest/bgptests/bgp.go
```diff
@@ -1278,6 +1279,140 @@ var _ = ginkgo.Describe("BGP", func() {
...
 
+	ginkgo.DescribeTable("BGP advertisement updates of communities and localpref and aggregation length", func(pairingIPFamily ipfamily.Family, poolAddresses []string) {
+		bgpAdv := metallbv1beta1.BGPAdvertisement{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:      "adv-updates",
+				Namespace: ConfigUpdater.Namespace(),
+			},
+			Spec: metallbv1beta1.BGPAdvertisementSpec{
+				Communities:    []string{CommunityNoAdv},
+				LocalPref:      100,
```
oribon: iirc 100 is the default even when not specified, maybe we should specify something else here?

e2etest/bgptests/bgp.go
```diff
@@ -1278,6 +1279,140 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			subnet = &net.IPNet{IP: ip.Mask(net.CIDRMask(124, 128)), Mask: net.CIDRMask(124, 128)}
+		}
+		// ParseRoutes indexes maps by destination IP string (see frr.ParseRoutes), not full CIDR.
+		aggregatedRouteKey := subnet.IP.String()
+
+		for _, c := range FRRContainers {
+			Eventually(func() (map[string]frr.Route, error) {
+				return frr.RoutesForFamily(c, pairingIPFamily)
+			}, time.Minute, time.Second).Should(HaveKey(aggregatedRouteKey))
+		}
```
oribon, line 1405: wouldn't we want to match on the actual CIDR rather than `aggregatedRouteKey` (which is the same regardless of the aggregation length?)
something like what happens `isRouteInjected` where we check against `route.Destination.String()` I think
so I'd expect us to match on subnet.String() (here and everywhere else that uses aggregatedRouteKey)

e2etest/bgptests/bgp.go
```diff
@@ -1278,6 +1279,140 @@ var _ = ginkgo.Describe("BGP", func() {
...
+		Expect(err).NotTo(HaveOccurred())
+
+		switch pairingIPFamily {
+		case ipfamily.IPv4:
+			bgpAdv.Spec.LocalPref = 200
+			bgpAdv.Spec.AggregationLength = ptr.To(int32(28))
+			bgpAdv.Spec.Communities = []string{CustomCommunity}
+		case ipfamily.IPv6:
+			bgpAdv.Spec.LocalPref = 200
+			bgpAdv.Spec.AggregationLengthV6 = ptr.To(int32(124))
```
oribon: `v6PoolAddresses       = "fc00:f853:0ccd:e799::/124"` is already /124

e2etest/bgptests/bgp.go
```diff
@@ -1842,6 +1842,116 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			Expect(err).NotTo(HaveOccurred())
+
+			ginkgo.By("Verify updated timers in speaker pods")
+			expectSpeakerConfiguredTimers(speakerPods, int(updatedHoldTime.Milliseconds()), int(updatedKeepaliveTime.Milliseconds()))
+
+			ginkgo.By("Reset BGP connection on FRR containers to re-establish with new timers")
+			for _, c := range FRRContainers {
+				_, err := c.Exec("vtysh", "-c", "clear bgp *")
+				Expect(err).NotTo(HaveOccurred())
+			}
```
oribon, line 1944: I think this is redundant, and also makes the test weaker, shouldn't we expect the speakers to renegotiate without external changes?

e2etest/bgptests/bgp.go
```diff
@@ -1952,6 +1951,118 @@ var _ = ginkgo.Describe("BGP", func() {
...
+				ginkgo.By("Verifying BGP session re-establishes within connect time")
+				allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+				Expect(err).NotTo(HaveOccurred())
+				Eventually(func() error {
+					neighbors, err := frr.NeighborsInfo(FRRContainers[0])
+					if err != nil {
+						return err
+					}
+					return frr.NeighborsMatchNodes(allNodes.Items, neighbors, ipfamily.IPv4, FRRContainers[0].RouterConfig.VRF)
+				}, connectTime, time.Second).ShouldNot(HaveOccurred())
```
oribon: wouldn't this be a bit flake prone? imo it'd be fine if we used 2*connectTime here (or something like that), we already validated the settings are correct

e2etest/bgptests/bgp.go
```diff
@@ -1952,6 +1951,118 @@ var _ = ginkgo.Describe("BGP", func() {
...
+			})
+
+			ginkgo.It("IPV4 update connect time to less than default on an existing BGP connection", func() {
```
oribon: imo this case is redundant

e2etest/bgptests/bgp.go
ajaggapa: BGP Timers are negotiated only when a session is reestablished. So, we need to reset it from external frr but not from cluster.

ajaggapa:
@oribon I have addressed your comment. Please review again

e2etest/pkg/frr/parse.go
oribon: this is still missing

e2etest/bgptests/bgp.go
```diff
@@ -1278,6 +1279,158 @@ var _ = ginkgo.Describe("BGP", func() {
...
+		Expect(ip).NotTo(BeNil())
+		var subnet *net.IPNet
+		wantAggLen := 0
+		if pairingIPFamily == ipfamily.IPv4 {
+			wantAggLen = 24
+			subnet = &net.IPNet{IP: ip.Mask(net.CIDRMask(24, 32)), Mask: net.CIDRMask(24, 32)}
+		} else {
+			wantAggLen = 124
+			subnet = &net.IPNet{IP: ip.Mask(net.CIDRMask(124, 128)), Mask: net.CIDRMask(124, 128)}
+		}
```
oribon: I still think we should pass the aggrLen from the table entry, and this becomes:
```go
        bgpAdv.Spec.AggregationLength = ptr.To(int32(aggrLen))
        bgpAdv.Spec.AggregationLengthV6 = ptr.To(int32(aggrLen))
        ...
        bits := 32
        if pairingIPFamily == ipfamily.IPv6 {
            bits = 128
        }
        subnet := &net.IPNet{
            IP:   ip.Mask(net.CIDRMask(aggLen, bits)),
            Mask: net.CIDRMask(aggLen, bits),
        }
```

e2etest/bgptests/bgp.go
oribon: fair enough, lets keep it this way

e2etest/bgptests/bgp.go
```diff
@@ -1278,6 +1279,159 @@ var _ = ginkgo.Describe("BGP", func() {
...
+		bgpAdv.Spec.Communities = []string{CustomCommunity}
+
+		var subnet *net.IPNet
+		bits := 32
+		if pairingIPFamily == ipfamily.IPv6 {
+			bits = 128
+			bgpAdv.Spec.AggregationLengthV6 = ptr.To(aggLen)
+		} else {
+			bgpAdv.Spec.AggregationLength = ptr.To(aggLen)
+		}
```
oribon: nit: these can be set outside of the if I think (and no need for else)

oribon (commented):
thanks! 2 small nits, the comment and can you please move the parse_test changes to the first commit (the "parse commit")?

oribon (approved):
lgtm, thanks!

PR #2995, opened 2026-03-26
e2etests: extend unnumbered BGP table tests with and without BFD

ajaggapa:
This Extends the existing FRR Unnumbered BGP DescribeTable so each IP family scenario runs twice: with BFD (current behavior) and without BFD. 

**Is this a BUG FIX or a FEATURE ?**: e2etests

>
> /kind bug
 /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**What this PR does / why we need it**: Expand e2e test coverage

**Special notes for your reviewer**: This is to migrate tests from eco-gotests 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etest/bgptests/unnummbered.go
+++ b/e2etest/bgptests/unnummbered.go
@@ -73,8 +73,11 @@ var _ = ginkgo.Describe("FRR Unnumbered BGP", func() {
 		Expect(err).NotTo(HaveOccurred())
 	})
 
-	ginkgo.DescribeTable("Session is established and route is advertised", func(prefixSendFromLocal []string, p2pInterface string, tweakService func(svc *corev1.Service)) {
-		rc := frrconfig.RouterConfigUnnumbered{
+	ginkgo.DescribeTable("Session is established and route is advertised", func(prefixSendFromLocal []string, p2pInterface string, tweakService func(svc *corev1.Service),
+		modifySetup func(*frrconfig.RouterConfigUnnumbered, *config.Resources),
+		validate func(*frrcontainer.FRR),
+	) {
+		rc := &frrconfig.RouterConfigUnnumbered{
 			ASNLocal:  metalLBASN,
 			ASNRemote: metalLBASN,
 			Hostname:  "tor1",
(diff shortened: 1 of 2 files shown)


ajaggapa:
@oribon @karampok 

For this PR, we have CI/e2e-use-operator job failure.
This is because the [workflow](https://github.com/metallb/metallb/blob/main/.github/workflows/ci.yaml#L335-L346) creates the `metallb` resource and immediately runs the tests (new test is the 1st test to run). But the speaker pods are not Ready before the tests start to run. 

How about updating the workflow to wait for the speaker daemonset to be ready using "`kubectl wait`" ?

ajaggapa:
> @oribon @karampok
> 
> For this PR, we have CI/e2e-use-operator job failure. This is because the [workflow](https://github.com/metallb/metallb/blob/main/.github/workflows/ci.yaml#L335-L346) creates the `metallb` resource and immediately runs the tests (new test is the 1st test to run). But the speaker pods are not Ready before the tests start to run.
> 
> How about updating the workflow to wait for the speaker daemonset to be ready using "`kubectl wait`" ?

To test this, i have increased the wait time from 2mins to 4mins and it passed now. So either we can simply go with increased timeout or else update the workflow as above. Please suggest.

the other failure now is a flake not related to this PR.

e2etest/bgptests/unnummbered.go
```diff
@@ -197,3 +200,37 @@ func validatePeerRoutesViaDevice(peer executor.Executor, dev, nextHop string, pr
...
+		if len(bfdPeers) == 0 {
+			return fmt.Errorf("expected BFD peers but found none")
+		}
+		for _, p := range bfdPeers {
+			if !strings.EqualFold(p.Status, "up") {
+				return fmt.Errorf("BFD peer %s status is %q, want \"up\"", p.Peer, p.Status)
+			}
+		}
+		return nil
+	}, 4*time.Minute, 10*time.Second).ShouldNot(HaveOccurred())
```
oribon: should this be 2min?

e2etest/bgptests/unnummbered.go
```diff
@@ -197,3 +200,37 @@ func validatePeerRoutesViaDevice(peer executor.Executor, dev, nextHop string, pr
...
+	Eventually(func() error {
+		bfdPeers, err := frr.BFDPeers(peer.Executor)
+		if err != nil {
+			return err
+		}
+		if len(bfdPeers) == 0 {
+			return fmt.Errorf("expected BFD peers but found none")
+		}
+		for _, p := range bfdPeers {
+			if !strings.EqualFold(p.Status, "up") {
```
oribon: I think we don't need the equalfold, we're using regular equalities usually (e.g peerConfig.Status != "up")

oribon:
lgtm, thanks!

PR #2996, opened 2026-03-29
e2etests: add l2 test with advertising node failover scenario

ajaggapa:
### e2etest

Adds e2etest/l2tests/speaker_failover.go with a spec that covers failover when the node currently announcing the LoadBalancer VIP no longer runs a healthy speaker: the test taints the announcing node, which is not tolerated by the speaker pod, asserts the service moves to another announcer. Later the taint is removed and the announcing node is expected to revert to original node.

**Is this a BUG FIX or a FEATURE ?**: Adding new e2etest for L2Advertisement

>
> /kind bug
 /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**What this PR does / why we need it**: Migrating a test from eco-gotests

**Special notes for your reviewer**: Consolidating tests from between eco-gotests and this repo

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etest/l2tests/speaker_failover.go
+++ b/e2etest/l2tests/speaker_failover.go
@@ -1,0 +1,211 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package l2tests
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	"go.universe.tf/e2etest/pkg/config"
+	"go.universe.tf/e2etest/pkg/k8s"
+	"go.universe.tf/e2etest/pkg/k8sclient"
+	"go.universe.tf/e2etest/pkg/metallb"
+	"go.universe.tf/e2etest/pkg/service"
+	"go.universe.tf/e2etest/pkg/status"
+	metallbv1beta1 "go.universe.tf/metallb/api/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/util/retry"
+)
+
+var _ = ginkgo.Describe("L2", func() {
+	var cs clientset.Interface
+	testNamespace := ""
+
+	emptyL2 := metallbv1beta1.L2Advertisement{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: "empty",
+		},
+	}
+
+	ginkgo.AfterEach(func() {
+		err := ConfigUpdater.Clean()
+		Expect(err).NotTo(HaveOccurred())
+
+		if ginkgo.CurrentSpecReport().Failed() {
+			k8s.DumpInfo(Reporter, ginkgo.CurrentSpecReport().LeafNodeText)
+		}
+		err = k8s.DeleteNamespace(cs, testNamespace)
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	ginkgo.BeforeEach(func() {
+		ginkgo.By("Clearing any previous configuration")
+
+		err := ConfigUpdater.Clean()
+		Expect(err).NotTo(HaveOccurred())
+		cs = k8sclient.New()
+		testNamespace, err = k8s.CreateTestNamespace(cs, "l2spk")
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	ginkgo.Context("speaker failover", func() {
+		ginkgo.BeforeEach(func() {
+			resources := config.Resources{
+				Pools: []metallbv1beta1.IPAddressPool{
+					{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: "l2-test",
+						},
+						Spec: metallbv1beta1.IPAddressPoolSpec{
+							Addresses: []string{
+								IPV4ServiceRange,
+								IPV6ServiceRange},
+						},
+					},
+				},
+				L2Advs: []metallbv1beta1.L2Advertisement{emptyL2},
+			}
+
+			err := ConfigUpdater.Update(resources)
+			Expect(err).NotTo(HaveOccurred())
+		})
+
+		ginkgo.It("should move L2 advertisement when the announcing node stops running a ready speaker", func() {
+			nodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			if len(nodes.Items) < 2 {
+				ginkgo.Skip("requires at least two nodes")
+			}
+
+			svc, _ := service.CreateWithBackend(cs, testNamespace, "external-local-lb", service.TrafficPolicyCluster)
+			defer func() {
+				err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
+				Expect(err).NotTo(HaveOccurred())
+			}()
+
+			var announcing string
+			Eventually(func() error {
+				s, err := status.L2ForService(ConfigUpdater.Client(), svc)
+				if err != nil {
+					return err
+				}
+				announcing = s.Status.Node
+				return nil
+			}, 2*time.Minute, time.Second).ShouldNot(HaveOccurred())
+
+			originalName := announcing
+
+			ginkgo.By(fmt.Sprintf("adding a NoSchedule taint the speaker does not tolerate on node %s, then deleting its speaker pod", originalName))
+			err = addSpeakerFailoverNoScheduleTaint(cs, originalName)
+			Expect(err).NotTo(HaveOccurred())
+			defer func() {
+				err := removeSpeakerFailoverNoScheduleTaint(cs, originalName)
+				Expect(err).NotTo(HaveOccurred())
+			}()
+
+			speakerPod, err := metallb.SpeakerPodInNode(cs, originalName)
+			Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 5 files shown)


e2etest/l2tests/speaker_failover.go
```diff
@@ -0,0 +1,278 @@
...
+	"go.universe.tf/e2etest/pkg/service"
+	metallbv1beta1 "go.universe.tf/metallb/api/v1beta1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/util/retry"
+)
+
+var RefreshFRRProviderNeeded bool
+var RefreshFRRProvider func() error
```
oribon: imo this is more of a hack, rather than solving the real issue.
iiuc, you need this because the frr-k8s provider caches the speaker<->frrk8s map once on initialization and when you delete a speaker the map breaks.
let's just do the similar thing as the frr mode provider, which is resolve the speaker<->frrk8s mapping on each call (in a different commit), and then we won't need this logic.

e2etest/l2tests/speaker_failover.go
```diff
@@ -0,0 +1,278 @@
...
+					return err
+				}
+				n, err := k8s.GetSvcNode(cs, fresh.Namespace, fresh.Name, allNodes)
+				if err != nil {
+					return err
+				}
+				announcing = n
+				return nil
+			}, 2*time.Minute, time.Second).ShouldNot(HaveOccurred())
+			logSpeakerFailoverEnvironment(cs, testNamespace, svc)
```
oribon: why do we need to log in the middle of a test? this seems redundant

e2etest/l2tests/speaker_failover.go
```diff
@@ -0,0 +1,278 @@
...
+				if err := service.ValidateL2(fresh); err != nil {
+					return err
+				}
+				n, err := k8s.GetSvcNode(cs, fresh.Namespace, fresh.Name, allNodes)
+				if err != nil {
+					return err
+				}
+				announcing = n
+				return nil
+			}, 2*time.Minute, time.Second).ShouldNot(HaveOccurred())
```
oribon, line 99: can we drop the ValidateL2 and changet this to just be an eventually call to status.L2ForService? we can then take the node name from there, and at this point we trust the status enough, not sure there is need for actual traffic, wdyt?

e2etest/l2tests/speaker_failover.go
```diff
@@ -0,0 +1,278 @@
...
+					if strings.Contains(err.Error(), "no speaker pod run in the node") {
+						return nil
+					}
+					return err
+				}
+				if p.Status.Phase != corev1.PodRunning || !k8s.PodIsReady(p) {
+					return nil
+				}
+				return fmt.Errorf("speaker still running and ready on %s", originalName)
+			}, 2*time.Minute, time.Second).Should(Succeed())
```
oribon, line 126: wrap this in a metallb.HasSpeakerInNode, e.g:
```go
func HasSpeakerInNode(cs clientset.Interface, node string) (bool, error) {
    _, err := SpeakerPodInNode(cs, node)
    if err != nil && strings.Contains(err.Error(), "no speaker pod run in the node") {
        return false, nil
    }
    if err != nil {
        return true, err
    }
    return false, nil
}
```

e2etest/l2tests/speaker_failover.go
```diff
@@ -0,0 +1,278 @@
...
+			logSpeakerFailoverEnvironment(cs, testNamespace, svc)
+
+			ginkgo.By("expecting a different announcing node and working connectivity")
+			Eventually(func() error {
```
oribon, line 129: same, let's use statuses if we agree connectivity is not necessary

e2etest/l2tests/speaker_failover.go
```diff
@@ -0,0 +1,278 @@
...
+					return fmt.Errorf("announcement still on %s", originalName)
+				}
+				return nil
+			}, 2*time.Minute, time.Second).ShouldNot(HaveOccurred())
+
+			Eventually(func() error {
+				return validateL2Refreshed(cs, svc)
+			}, 2*time.Minute, time.Second).ShouldNot(HaveOccurred())
+			logSpeakerFailoverEnvironment(cs, testNamespace, svc)
+		})
```
oribon, line 139: pr description mentions "asserts the service moves to another announcer and stays reachable, then uncordons and checks the original announcer and connectivity recover.", shouldn't it be here?

e2etest/l2tests/speaker_failover.go
```diff
@@ -0,0 +1,278 @@
...
+			logSpeakerFailoverEnvironment(cs, testNamespace, svc)
+
+			originalName := announcing.Name
+
+			ginkgo.By(fmt.Sprintf("adding a NoSchedule taint the speaker does not tolerate on node %s, then deleting its speaker pod", originalName))
+			err = addSpeakerFailoverNoScheduleTaint(cs, originalName)
+			Expect(err).NotTo(HaveOccurred())
+			logSpeakerFailoverEnvironment(cs, testNamespace, svc)
+			defer func() {
+				_ = removeSpeakerFailoverNoScheduleTaint(cs, originalName)
```
oribon: can we assign err here and expect it to not have occurred?

e2etest/pkg/k8s/services.go
oribon, line 1: why are the changes here necessary?

e2etest/l2tests/speaker_failover.go
```diff
@@ -0,0 +1,278 @@
...
+}
+
+// Taint used only in this test: the MetalLB speaker DaemonSet does not tolerate it, so the
+// speaker cannot be rescheduled onto the node after we delete its pod.
+const (
+	speakerFailoverTaintKey   = "metallb.e2e/speaker-failover-block"
+	speakerFailoverTaintValue = "true"
+)
+
+func addSpeakerFailoverNoScheduleTaint(cs clientset.Interface, nodeName string) error {
```
oribon, line 150: the noschedule taint to me seems a bit aggressive (if things go wrong), can we take a different approach? I think we could set a finalizer on the pod, delete the pod (then it should be stuck on terminating until the finalizer is removed, in the meantime the containers inside the pod are stopped), and we remove the finalizer from the pod to let the cleanup happen naturally?

oribon (commented):
thanks! couple of comments:

e2etest/l2tests/speaker_failover.go
ajaggapa: Finalizers isn't giving the desirable results. as discussed, we will continue with node taints

e2etest/pkg/k8s/services.go
ajaggapa: I wanted it to effectively sort the events in a racy condition with closer timestamps.
But i have removed it now, as i switched to ServiceL2Status CR

e2etest/l2tests/speaker_failover.go
ajaggapa: done

e2etest/l2tests/speaker_failover.go
ajaggapa: updated the PR description

e2etest/l2tests/speaker_failover.go
ajaggapa: done

e2etest/l2tests/speaker_failover.go
ajaggapa: done

e2etest/l2tests/speaker_failover.go
ajaggapa: removed traffic validation

e2etest/l2tests/speaker_failover.go
ajaggapa: added it for troubleshooting CI failures. removed it now.

e2etest/l2tests/speaker_failover.go
ajaggapa: makes sense. i have updated to dynamically fetch the speaker--frrk8s mapping

e2etest/l2tests/speaker_failover.go
```diff
@@ -0,0 +1,195 @@
...
+		ginkgo.By("Clearing any previous configuration")
+
+		err := ConfigUpdater.Clean()
+		Expect(err).NotTo(HaveOccurred())
+		cs = k8sclient.New()
+		testNamespace, err = k8s.CreateTestNamespace(cs, "l2spk")
+		Expect(err).NotTo(HaveOccurred())
+	})
+
+	ginkgo.Context("speaker failover", ginkgo.Ordered, func() {
```
oribon: do we need the "Ordered" here?

e2etest/l2tests/speaker_failover.go
oribon: I just checked, the existing similar test (NetworkUnavailable l2) does this:
```
			ginkgo.By("setting the NetworkUnavailable condition back to false")
			err = k8s.SetNodeCondition(cs, nodeToSet.Name, corev1.NodeNetworkUnavailable, corev1.ConditionFalse)

			ginkgo.By("validating the service is announced back again from the previous node")
			Eventually(func() string {
				node, err := nodeForService(svc, allNodes.Items)
				if err != nil {
					return ""
				}
				return node.Name
			}, time.Minute, time.Second).Should(Equal(nodeToSet.Name))
```
so let's do the corresponding thing here also for the sake of being aligned

e2etest/l2tests/speaker_failover.go
ajaggapa: removed

oribon (approved):
lgtm, thanks!

PR #2999, opened 2026-03-30
e2etests: add test for LB IP reachability inside cluster

ajaggapa:
This adds an end-to-end check that a LoadBalancer service’s external IP is reachable from inside the cluster and that the backend sees the expected client address.

A new e2etest/servicetests package with a Ginkgo DescribeTable covering two placements of the client pod: on the same node as the backend, and on a different node (the latter is skipped when the cluster has fewer than two nodes).
A busybox client pod runs wget against the VIP: it checks that /hostName matches the backend pod name (default pod hostname in Kubernetes) and that /clientip includes the client pod’s PodIP.

**Is this a BUG FIX or a FEATURE ?**: adding new test to validate MetalLB assigned external IP of a service is accessible inside the cluster

>
> /kind bug
 /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**What this PR does / why we need it**: Expanding test coverage for e2etests

**Special notes for your reviewer**: A new table test to validate external IP accessibility from inside the cluster.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/e2etest/e2etest_suite_test.go
+++ b/e2etest/e2etest_suite_test.go
@@ -32,6 +32,7 @@ import (
 	"go.universe.tf/e2etest/configurationstatetests"
 	"go.universe.tf/e2etest/l2tests"
 	_ "go.universe.tf/e2etest/netpoltests"
+	"go.universe.tf/e2etest/servicetests"
 	testsconfig "go.universe.tf/e2etest/pkg/config"
 	"go.universe.tf/e2etest/pkg/executor"
 	frrprovider "go.universe.tf/e2etest/pkg/frr/provider"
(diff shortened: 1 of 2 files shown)


ajaggapa:
ci failure is not related to the PR

oribon (commented):
Internal LB reachability is not the responsibility of MetalLB, but rather the CNI's. I don't think we should cover this

fedepaol:
> Internal LB reachability is not the responsibility of MetalLB, but rather the CNI's. I don't think we should cover this

I second that. It's totally handled by the CNI of choice if any

ajaggapa:
closing PR as invalid test

Issue #2797, opened 2025-07-03
Adds the ability to explicitly designate a primary node for services running in L2 mode


Paucpauc:
### Is your feature request related to a problem?

Adds the ability to explicitly designate a primary node for services running in L2 mode. The goal is to ensure predictable IP assignment and not lose failover, avoiding scenarios where IPs end up on same nodes.

NodeSelectors is not suitable, because if you specify only one node there, then failover is lost.

Use Case:
When external load balancers are unavailable, and used DNS round-robin.

### Describe the solution you'd like

A simple way is to add an annotation to the service indicating the primary node (metallb.io/primary-node: nodeX)

### Additional context

Here is an example of a working patch with a small amount of e2etest.
https://github.com/metallb/metallb/compare/main...Paucpauc:metallb:main
```diff
diff --git a/speaker/layer2_controller.go b/speaker/layer2_controller.go
index 3a8f0a8e..0f21d5ed 100644
--- a/speaker/layer2_controller.go
+++ b/speaker/layer2_controller.go
@@ -35,6 +35,10 @@ import (
        "go.universe.tf/metallb/internal/layer2"
 )

+const (
+       primaryNodeAnnotation = "metallb.io" + "/" + "primary-node"
+)
+
 type layer2Controller struct {
        announcer       *layer2.Announce
        myNode          string
@@ -105,10 +109,20 @@ func (c *layer2Controller) ShouldAnnounce(l log.Logger, name string, toAnnounce

        // Using the first IP should work for both single and dual stack.
        ipString := toAnnounce[0].String()
+
+       primaryNode, _ := svc.Annotations[primaryNodeAnnotation]
+       level.Debug(l).Log("event", "shouldannounce", "protocol", "l2", "primaryNode", primaryNode, "service", name)
+
        // Sort the slice by the hash of node + load balancer ips. This
        // produces an ordering of ready nodes that is unique to all the services
        // with the same ip.
        sort.Slice(availableNodes, func(i, j int) bool {
+               if availableNodes[i] == primaryNode {
+                       return true
+               }
+               if availableNodes[j] == primaryNode {
+                       return false
+               }
                hi := sha256.Sum256([]byte(availableNodes[i] + "#" + ipString))
                hj := sha256.Sum256([]byte(availableNodes[j] + "#" + ipString))
```

### I've read and agree with the following

- [x] I've checked all open and closed issues and my request is not there.
- [x] I've checked all open and closed pull requests and my request is not there.

[...]

PR #3004, opened 2026-04-01
Design proposal for preferred node selection in L2 mode

jlandersen:
Relates to #2797

/kind design

Design proposal for a `preferredNodeSelectors` field on `L2Advertisement`, following the k8s `PreferredSchedulingTerm` pattern.

The hardest part was deciding on the semantics around multiple L2Advertisements and service selectors. I've landed on restricting `preferredNodeSelectors` to one L2Advertisement per pool and rejecting the combination with `serviceSelectors`. The doc has sections on both with the reasoning, but I'd appreciate input here.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/design/preferred-node-l2.md
+++ b/design/preferred-node-l2.md
@@ -1,0 +1,541 @@
+# Preferred Node Selection for L2 Mode
+
+## Summary
+
+A new `preferredNodeSelectors` field on `L2Advertisement` lets administrators express soft node
+preferences for L2 leader election. The speakers try preferred nodes first when electing the
+announcer for a LoadBalancer IP, falling back to any eligible node when preferred nodes are
+unavailable. The API mirrors the Kubernetes `PreferredSchedulingTerm` pattern and builds on the existing
+hash-based election algorithm.
+
+Related: [#2797](https://github.com/metallb/metallb/issues/2797) (feature request for this design doc),
+[PR #1804](https://github.com/metallb/metallb/pull/1804) (rejected predecessor, annotation-based, single node).
+
+Depends on: [PR #3014](https://github.com/metallb/metallb/pull/3014), which refactors L2
+election so `serviceSelectors` drives the candidate node set. The scoring step added here
+slots into the post-#3014 `ShouldAnnounce` flow.
+
+## Motivation
+
+MetalLB L2 mode elects the announcing node for each LoadBalancer IP using
+`sha256(nodeName + "#" + ipString)`. The hash gives deterministic, even distribution, but sometimes you need control over which node announces without losing failover.
+
+### Scenario 1: Dedicated Load-Balancer Nodes
+
+A cluster has three worker nodes and two "edge" nodes.
+L2 traffic should land on the edge nodes when possible, but the workers are needed as failover targets when both edge nodes are down. Today, using `nodeSelectors` to
+restrict to edge nodes means zero failover if both go down.
+
+### Scenario 2: Zone-Aware Traffic Locality
+
+A bare-metal cluster spans two server rooms (zone-a, zone-b). Clients connect through a switch in zone-a. The L2 announcer should be in zone-a to minimize cross-zone hops, but zone-b should take over during zone-a maintenance windows.
+
+### Scenario 3: Stable Announcements During Rolling Upgrades
+
+Cluster upgrades often take nodes out of rotation, for example by draining and rebooting them in a rolling fashion. With announcements spread across many nodes, IP assignments get shuffled each time an announcing node goes down, and a full upgrade cycle can shuffle them repeatedly. Pinning announcements to a small set of "anchor" nodes (upgraded first or last) keeps IPs stable for the majority of the upgrade window. The remaining nodes still serve as failover targets when the anchor nodes themselves are patched.
+
+### Goals
+
+- Allow administrators to express weighted soft preferences for which nodes announce L2 services.
+- Preserve full failover: all eligible nodes remain candidates when preferred nodes are unavailable.
+- Maintain backward compatibility: existing configurations without `preferredNodeSelectors` behave identically to today.
+- Follow Kubernetes API patterns (`PreferredSchedulingTerm`).
+
+### Non-Goals
+
+- Influencing BGP mode node selection.
+- Per-service preference overrides via annotations. This has previously been suggested, and the
+  preference is to not have annotations used (see [PR #1804](https://github.com/metallb/metallb/pull/1804)).
+
+## Proposed API
+
+One new field on `L2AdvertisementSpec` and one new type.
+
+### Struct delta on L2AdvertisementSpec
+
+```go
+type L2AdvertisementSpec struct {
+	// ... existing fields unchanged ...
+
+	// PreferredNodeSelectors allows specifying soft node preferences for L2
+	// leader election. Nodes matching these selectors receive a higher score
+	// and are preferred as the announcing node. If no preferred node is
+	// available, any eligible node (per NodeSelectors) can announce.
+	// Modeled after Kubernetes PreferredSchedulingTerm.
+	// +optional
+	PreferredNodeSelectors []PreferredNodeSelector `json:"preferredNodeSelectors,omitempty"`
+}
+```
+
+### New type
+
+```go
+// PreferredNodeSelector expresses a weighted soft preference for nodes.
+// This follows the Kubernetes PreferredSchedulingTerm pattern where Weight
+// controls relative priority and Preference selects matching nodes.
+type PreferredNodeSelector struct {
+	// Weight associated with matching the corresponding preference,
+	// in the range 1-100.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=100
+	Weight int32 `json:"weight"`
+
+	// A node selector that applies to this weight.
+	Preference metav1.LabelSelector `json:"preference"`
+}
+```
+
+The naming and structure follow Kubernetes'
+[`PreferredSchedulingTerm`](https://pkg.go.dev/k8s.io/api/core/v1#PreferredSchedulingTerm):
+
+- `preference` matches `PreferredSchedulingTerm.Preference` to signal this is a soft selector.
+- `metav1.LabelSelector` is consistent with existing MetalLB selector fields
+  (`NodeSelectors`, `IPAddressPoolSelectors`, etc.). We use this instead of `NodeSelectorTerm`, consistent with the rest of the MetalLB API.
+- `int32` weight, range 1-100, required (no `omitempty`), matching
+  `PreferredSchedulingTerm.Weight` type and validation.
+
+### Versioning and Upgrade Path
+
+The field is optional and added to `v1beta1`. New controllers reading old CRDs see nil, identical
+to today's behavior.
+
+Safe upgrade order:
+
+1. Update CRDs (adds the new field to the schema).
+2. Roll the controller and webhook to a version that understands and preserves
+   `preferredNodeSelectors`. Without this, older components can drop the field on write.
+3. Roll all speaker pods to the new version.
+4. Only then create or update L2Advertisements with `preferredNodeSelectors`.
+
+### Example CRs
+
+Only nodes with `role: lb` are eligible. Among those, prefer edge nodes:
+
+```yaml
+apiVersion: metallb.io/v1beta1
+kind: L2Advertisement
+metadata:
+  name: prefer-edge
+  namespace: metallb-system
+spec:
+  ipAddressPools:
+    - production-pool
+  nodeSelectors:
+    - matchLabels:
+        role: lb
+  preferredNodeSelectors:
+    - weight: 100
+      preference:
+        matchLabels:
+          node-role: edge
+```
+
+All nodes are eligible. Prefer edge nodes. Any node can announce as fallback:
+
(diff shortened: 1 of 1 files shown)


design/preferred-node-l2.md
```diff
@@ -0,0 +1,467 @@
...
+A pool can have multiple `L2Advertisement` objects. MetalLB already combines fields across
+advertisements: node eligibility is OR'd (`poolMatchesNodeL2`), interfaces are UNION'd
+(`ipAdvertisementFor`).
+
+Preference scores are not aggregated across advertisements. At most one `L2Advertisement`
+with non-empty `preferredNodeSelectors` may target a given pool. This is enforced by the
+validating webhook, described in [Validation Rules](#validation-rules).
+
+This keeps all preference math in a single policy object per pool. If weights were summed across multiple L2Advertisements, a new overlapping
+advertisement could silently change the effective priority of nodes established by another
```
fedepaol: this goes beyond how all the rest of the configuratoins are working. NodeSelectors work in this way already, I would keep consistency.

design/preferred-node-l2.md
```diff
@@ -0,0 +1,467 @@
...
+- Avoid single-hostname preferences.
+
+### Interaction with serviceSelectors
+
+The validating webhook rejects an L2Advertisement that sets both `serviceSelectors` and
+`preferredNodeSelectors`.
+
+All speakers must agree on the winner for a given IP, and election is computed pool-wide
+(`poolMatchesNodeL2` evaluates all advertisements targeting the pool). A service-scoped
+advertisement carrying preference scores would influence election for every service in the
```
fedepaol: not sure I understood why we can't do this. IIUC, the logic is still per-ip.

fedepaol:
Thanks for the proposal!

design/preferred-node-l2.md
jlandersen: I was a bit too careful here yeah. If you agree, I'll drop this restriction and sum the result across all L2Advertisements.

design/preferred-node-l2.md
jlandersen: Looking at it again, I might have been a bit too careful here as well.

My original concern was that "per-service" scoring could break the same-IP-same-winner. Two services sharing an IP (via allow-shared-ip which we use heavily) but matched by different ads (due to serviceSelectors) could get different preference scores and elect different nodes with the original restriction. But thinking about it this was likely a result of the original restriction I put in place around limiting scoring across L2 ads (where you had the other comment).

With your other comment, and using pool-wide scores (summing preferredNodeSelectors across all L2Advertisements in the pool) this should handle this case as well if my understanding is correct (and not be a concern).

What might be a remaining concern is mostly UX related. Using `serviceSelectors` + `preferredNodeSelectors` might imply the preferences only relates to the service(s) selected (I guess it's the same with node selectors?)

jlandersen:
> Thanks for the proposal!

Thanks for the comments as well! Added replies :)

design/preferred-node-l2.md
oribon: +1 on allowing aggregation (being aligned with how the other fields work), but for this case:
```
Ad1:  nodeSelectors: [{role: lb}]              → nodes A, B eligible
Ad2:  nodeSelectors: [{role: edge}]            → nodes C, D eligible
Ad3:  nodeSelectors: [{role: gpu}]
      preferredNodeSelectors: [{weight: 100, preference: {zone: primary}}]
                                               → node E eligible

Combined eligible set: A, B, C, D, E  (OR across all three ads)
If node C matches zone=primary → C gets score 100 (even though Ad1/Ad2 made C eligible, not Ad3)
```
imo we should make sure only node E gets the score 100, that is for a given ad its preference only scores nodes that the ad targets via nodeSelectors.

design/preferred-node-l2.md
oribon: this actually made me revisit the service selector code, and I realized there is a bug there (thanks! https://github.com/metallb/metallb/pull/3014)
`might imply the preferences only relates to the service(s) selected` - that is the behavior I'd expect, only the ads that match the service should be evaluated against it.
I guess we'd need the same mechanism for this proposal - node scores should be evaluated against all ads matching the service, and the elected node uses only the ads that belong to it.

design/preferred-node-l2.md
```diff
@@ -0,0 +1,467 @@
...
+For per-service preference, isolate the service onto its own `IPAddressPool` using
+`serviceAllocation.serviceSelectors`, then attach a dedicated `L2Advertisement` with
+`preferredNodeSelectors` to that pool.
+
+## Drawbacks
+
+Operators need to understand weighted scoring. Misconfigured preferences can cause unexpected
+IP migration or reduce failover capacity.
+
+## Alternatives Considered
```
oribon, line 449: if you agree, I'd also write about having only one field which is "prioritized node selectors" (similar to what is described here as preferred) that cannot be specified alongside the legacy nodeSelector field and obsoletes it. this might be "simpler" (caring only about one field), but deviates from what's already established in k8s with required/preferred

oribon (commented):
overall great pr! I like that this is aligned with existing k8s semantics. maybe its worth to highlight that the nodeSelector field is about "required" labels (which is already implicit from stating that the new field is about "soft/preferred" labels)

design/preferred-node-l2.md
jlandersen: That sounds good to me 👍  And just to confirm, before I update the spec, when there's no node selectors, the scoring is applied to all (no node selectors = all nodes).

design/preferred-node-l2.md
jlandersen: Sounds good! With #3014, I'll update the proposal shortly. Together with the comments around node selectors, this is my understanding of where we want to go:

  1. Ad-scoped scoring: each ad's preferredNodeSelectors only scores nodes matched by that ad's own nodeSelectors. No nodeSelectors = all nodes eligible, all nodes scored.
  2. Drop the single-preference-ad-per-pool restriction: each ad scores only its own eligible nodes, so multiple ads with preferences on the same pool don't interfere.
  3. Drop the serviceSelectors + preferredNodeSelectors restriction: #3014 already filters ads by service before election, so the scoring step operates on those filtered ads. The preference design doesn't need  serviceSelector-specific logic.

design/preferred-node-l2.md
jlandersen: 👍  will add it to the section when I do the next iteration shortly

design/preferred-node-l2.md
oribon: right (keeping the same semantic that no selectors is all nodes)

design/preferred-node-l2.md
oribon: for point 2 just making sure - 
```
each ad scores only its own eligible nodes, so multiple ads with preferences on the same pool don't interfere
```
each ad scores only its own eligible nodes, and for a given service the score of each node (for the election part) is the sum of all ads it matches

design/preferred-node-l2.md
```diff
@@ -0,0 +1,543 @@
...
+  preferredNodeSelectors:
+    - weight: 100
+      preference:
+        matchLabels:
+          zone: primary
+```
+
+Services not matching `app: frontend` ignore this advertisement.
+
+Caveat: `ShouldAnnounce` runs per-service, so two services sharing an IP via
```
jlandersen, line 437: One thing I want to flag alongside this caveat, and a possible fix I haven't put in the doc yet. I think there's still some challenges around the interaction with `allow-shared-ip` that we discussed further up (https://github.com/metallb/metallb/pull/3004#discussion_r3071110068). Mostly put this caveat note here in the doc for discussion, hoping we can decide for a better final result :)

The problem I see: `ShouldAnnounce` runs per-service, and `adsForService` is filtered by serviceSelectors. 
Two services sharing an IP via `allow-shared-ip` can match different preference-bearing ads and compute different score maps. Since the hash tie-break only fires on score ties, divergent scores bypass it and each service elects a different node — both then answer ARP for the same IP.

Example:
```
  Ad1: serviceSelectors=[{app: a}], prefer zone=primary
  Ad2: serviceSelectors=[{app: b}], prefer zone=secondary
  svcA(app=a) elects a primary-zone node, svcB(app=b) elects a secondary-zone
  node, both on the same shared IP.
```

A possible fix could be to add a controller-side consistency check. For any pool with preference-bearing ads, group services by SharingKey and verify the PreferredNodes sum is identical across siblings. Reject the config (or the allocation) when it isn't. This preserves the serviceSelectors + preferences combo for services that aren't sharing, and should only block the exact configurations that would split. I hope this could be a reasonable solution rather than the original idea of rejecting any use of serviceSelectors + preferences together.

jlandersen:
Pushed an update addressing the previous round of feedback:

- Dropped the single-preference-per-pool restriction and the serviceSelectors +  preferredNodeSelectors restriction. Scoring is now ad-scoped: each ad scores  only its own eligible nodes, and a service's final score is the sum across
  the ads matching it. 
- Upgrade order split into controller/webhook and then speakers (copilot feedback)
- Added Option C (single "prioritized" selector replacing nodeSelectors) to
  Alternatives for completeness, with the reasons for not going that way.
- Test plan updated to cover ad-scoped scoring and multi-ad aggregation.

One open item I want to flag is  `allow-shared-ip` interaction with per-service scoring. 

Left a separate comment next to the caveat in "Interaction with serviceSelectors" with the problem and a possible controller-side check to enforce consistency across shared-IP siblings. Would appreciate input on whether to fold that into this design.

design/preferred-node-l2.md
oribon: you're raising a very good point regarding shared-ips+serviceselectors (which clearly I didn't think about when implementing it 😛), I'd say what you're describing is a side-effect of this and we should treat this in a separate issue, not as part of this design. as an initial thought (to be discussed in that issue), I'd say we should upfront disallow shared-ips+serviceselectors, because this gets too messy (say we allow different ads to target different services with the same ip, but each ad has a different node selector, then whatever node we choose will be wrong for one of them).

jlandersen: Sounds good, I'll keep it out of this design and trim the caveat to a note. I can open a separate issue if you want :)
Thanks!

jlandersen: I added this issue https://github.com/metallb/metallb/issues/3025. Hope it covers what we discussed, otherwise definitely adjust it or close / reopen :)

I also updated the caveat here with a notice around this.

jlandersen:
Published one more update

 - Based on our discussion I opened how shared IPs and service selectors work into https://github.com/metallb/metallb/issues/3025
 - Updated caveat section with reference to this, that this also impacts this feature and suggestion for now is to not combine it with service selectors and shared ips

That should be the end of this round of changes and I will let the design proposal be up for review again 👍

design/preferred-node-l2.md
```diff
@@ -0,0 +1,545 @@
...
+Services not matching `app: frontend` ignore this advertisement.
+
+Caveat: this feature inherits an existing issue with `allow-shared-ip` +
+`serviceSelectors` tracked in [#3025](https://github.com/metallb/metallb/issues/3025).
+`ShouldAnnounce` runs per-service, so two services sharing an IP that match
+different preference-bearing ads compute different score maps and can elect
+different nodes. The hash tie-break only fires on score ties, so divergent
+scores bypass it. Until #3025 is resolved, shared-IP siblings should match the
+same set of preference-bearing ads. The simplest way is to keep preferences on
+ads without `serviceSelectors`.
```
oribon: with what we decided on #3025, is this still a problem/caveat? without serviceselectors, services with the same ip will always have the same set of ads, and then we should be fine?

oribon (commented):
looks good! I have one question:

design/preferred-node-l2.md
jlandersen: Yes, I think this should be entirely resolved with that 👍 Should we assume it to be in place so I remove the caveat entirely or just mention it briefly like `With the rule from #3025 disallowing allow-shared-ip + serviceSelectors on L2Advertisements, shared-IP siblings always see
  the same ad set and converge on the same announcer.`"?

oribon: let's go with the latter (both are fine though)

jlandersen: Updated! :)

oribon (approved):
thanks! lgtm

jlandersen:
> thanks! lgtm

Thanks a lot as well, this was a nice experience and learned a lot of details about MetalLB :)
I have rebased on main so it should be good to go. I'm up for doing more work on this if it sounds good on your side (track it via https://github.com/metallb/metallb/issues/2797, or a new issue?)

oribon:
> > thanks! lgtm
> 
> Thanks a lot as well, this was a nice experience and learned a lot of details about MetalLB :) I have rebased on main so it should be good to go. I'm up for doing more work on this if it sounds good on your side (track it via #2797, or a new issue?)

thanks! it was a nice experience from my side as well 😄 
tracking it in the existing issue is fine imo, and let's wait for @fedepaol to give his lgtm on this pr

fedepaol:
LGTM thanks!

oribon:
@jlandersen can you please squash the commits?

jlandersen:
> @jlandersen can you please squash the commits?

@oribon  Rebased on main and squashed my commits :)

PR #3009, opened 2026-04-12
E2E: refactor ServiceSelectors

gkopels:
/kind feature

What this PR does / why we need it:
Adds e2e coverage for MetalLB Service Selector behavior across BGP and L2 advertisement scenarios.
The new/updated tests cover:
- pool selector + service selector interactions
- multiple service selectors (OR behavior)
- selector update propagation to routes
- webhook validation for serviceSelector + aggregation settings
- overlapping/non-overlapping BGPAdvertisement selector behavior

Special notes for your reviewer:
- Scope is test-focused (e2e coverage expansion and small refactors from review feedback).
- Review feedback updates were incorporated (helper reuse, naming cleanup, comment cleanup, simplified test structure where requested).
- Focused local runs were executed for the added scenarios; CI matrix is green.

Release note:
```release-note
NONE

### DIFF
--- a/e2etest/bgptests/service_selector.go
+++ b/e2etest/bgptests/service_selector.go
@@ -5,11 +5,13 @@ package bgptests
 import (
 	"context"
 	"fmt"
+	"net"
 	"time"
 
 	"go.universe.tf/e2etest/pkg/config"
 	"go.universe.tf/e2etest/pkg/frr"
 	"go.universe.tf/e2etest/pkg/ipfamily"
+	jigservice "go.universe.tf/e2etest/pkg/jigservice"
 	"go.universe.tf/e2etest/pkg/k8s"
 	"go.universe.tf/e2etest/pkg/k8sclient"
 	"go.universe.tf/e2etest/pkg/metallb"
@@ -23,7 +25,14 @@ import (
 	frrconfig "go.universe.tf/e2etest/pkg/frr/config"
 	frrcontainer "go.universe.tf/e2etest/pkg/frr/container"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/types"
 	clientset "k8s.io/client-go/kubernetes"
+)
+
+const (
+	environmentLabelKey         = "environment"
+	environmentLabelProd        = "production"
+	environmentLabelStaged      = "staged"
 )
 
 var _ = ginkgo.Describe("BGP Service Selector", func() {
@@ -129,29 +138,9 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
 				testservice.WithAnnotations(map[string]string{"metallb.io/address-pool": "pool-b"}))
 			defer testservice.Delete(cs, svcWrongPool)
 
-			checkServiceNotAdvertised := func(svc *corev1.Service, reason string) error {
-				for _, c := range FRRContainers {
-					frrRoutesV4, frrRoutesV6, err := frr.Routes(c)
-					if err != nil {
-						return err
-					}
-
-					for _, ip := range svc.Status.LoadBalancer.Ingress {
-						_, found := frrRoutesV4[ip.IP]
-						if pairingIPFamily == ipfamily.IPv6 {
-							_, found = frrRoutesV6[ip.IP]
-						}
-						if found {
-							return fmt.Errorf("%s IP %s should NOT be advertised to %s (%s)", svc.Name, ip.IP, c.Name, reason)
-						}
-					}
-				}
-				return nil
-			}
-
 			ginkgo.By("Checking that svc-wrong-pool is NOT advertised (pool selector mismatch)")
 			Consistently(func() error {
-				return checkServiceNotAdvertised(svcWrongPool, "pool selector mismatch")
+				return checkServiceNotAdvertised(FRRContainers, svcWrongPool, "pool selector mismatch")
 			}, 5*time.Second, 1*time.Second).ShouldNot(HaveOccurred())
 
 			ginkgo.By("Creating service without expose label requesting pool-a - service selector doesn't match")
@@ -162,7 +151,7 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
 
 			ginkgo.By("Checking that svc-no-label is NOT advertised (service selector mismatch)")
 			Consistently(func() error {
-				return checkServiceNotAdvertised(svcNoLabel, "service selector mismatch")
+				return checkServiceNotAdvertised(FRRContainers, svcNoLabel, "service selector mismatch")
 			}, 5*time.Second, 1*time.Second).ShouldNot(HaveOccurred())
 
 			ginkgo.By("Creating service without matching label requesting pool-b - neither matches")
@@ -173,7 +162,7 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
 
 			ginkgo.By("Checking that svc-neither is NOT advertised (both selectors mismatch)")
 			Consistently(func() error {
-				return checkServiceNotAdvertised(svcNeitherMatch, "both selectors mismatch")
+				return checkServiceNotAdvertised(FRRContainers, svcNeitherMatch, "both selectors mismatch")
 			}, 5*time.Second, 1*time.Second).ShouldNot(HaveOccurred())
 
 			ginkgo.By("Verifying the matching service is still advertised after all checks")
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: e2etest/bgptests/service_selector.go [hunk 1]
```diff
@@ -271,6 +309,144 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+						if pairingIPFamily == ipfamily.IPv6 {
+							_, found = frrRoutesV6[ip.IP]
+						}
+						if found {
+							return fmt.Errorf("%s IP %s should NOT be advertised to %s (%s)", svc.Name, ip.IP, c.Name, reason)
+						}
+					}
+				}
+				return nil
+			}
```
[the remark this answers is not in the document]
gkopels: changed

e2etest/bgptests/service_selector.go
```diff
@@ -271,6 +309,144 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+				validateService(svcProd, allNodes, c)
+			}
+			Consistently(func() error {
+				return checkNotAdvertised(svcStaged, "staged must not match production selector")
+			}, 5*time.Second, 1*time.Second).ShouldNot(HaveOccurred())
+
+			if updateBGPAdvertisement {
+				ginkgo.By("Updating BGPAdvertisement ServiceSelectors to match staged")
+				var adv metallbv1beta1.BGPAdvertisement
+				err = ConfigUpdater.Client().Get(context.TODO(), types.NamespacedName{
```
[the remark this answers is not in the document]
gkopels: updated

e2etest/bgptests/service_selector.go
```diff
@@ -271,6 +309,144 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+			}
+
+			ginkgo.By("After update, staged service should be advertised and production should not")
+			for _, c := range FRRContainers {
+				validateService(svcStaged, allNodes, c)
+			}
+			Eventually(func() error {
+				return checkNotAdvertised(svcProd, "production service must no longer match selector")
+			}, 4*time.Minute, 1*time.Second).ShouldNot(HaveOccurred())
```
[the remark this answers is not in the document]
gkopels: update

internal/bgp/native/native.go
```diff
@@ -459,6 +462,41 @@ func validate(adv *bgp.Advertisement) error {
 	return nil
 }
 
+// mergeAdvertisements combines two advertisements for the same prefix (e.g. overlapping
+// BGPAdvertisement CRs merged by the speaker). LocalPref and Peers must match; communities
+// are unioned. Without this, session.Set's map keyed by prefix would keep only the last adv
+// and drop communities from the other (frr-k8s merges per-prefix attributes; native must too).
+func mergeAdvertisements(a, b *bgp.Advertisement) (*bgp.Advertisement, error) {
```
oribon: how is this change related to this pr?

e2etest/bgptests/validate.go
oribon, line 1: is this change related to this pr?

e2etest/bgptests/service_selector.go
oribon, line 1: please move auxiliary functions to the bottom of the file

e2etest/bgptests/service_selector.go
```diff
@@ -23,9 +29,125 @@ import (
...
+}
+
+const (
+	mergeAdvertisementCommunity = "65000:100"
+	mergeAdvertisementLocalPref = uint32(400)
+	nonOverlapCommunityProd     = "65000:100"
+	nonOverlapCommunityStaged   = "65000:200"
+)
+
+func addressFamilyForIngressIP(ingressIP string) ipfamily.Family {
```
oribon: I think this is redundant, calling ipfamily.ForAddress explicitly is what we do everywhere else

e2etest/bgptests/service_selector.go
```diff
@@ -62,6 +184,21 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+				Expect(err).NotTo(HaveOccurred())
+				ip1, err := config.GetIPFromRangeByIndex(l2tests.IPV6ServiceRange, 1)
+				Expect(err).NotTo(HaveOccurred())
+				ip2, err := config.GetIPFromRangeByIndex(l2tests.IPV6ServiceRange, 2)
+				Expect(err).NotTo(HaveOccurred())
+				ip3, err := config.GetIPFromRangeByIndex(l2tests.IPV6ServiceRange, 3)
+				Expect(err).NotTo(HaveOccurred())
+				poolA = ip0 + "-" + ip1
+				poolB = ip2 + "-" + ip3
+			}
```
oribon: why do we need this block? using l2tests ip ranges / "dynamic ranges", can't we just use hardcoded ranges passed as an Entry (the way we usually do)?

e2etest/bgptests/service_selector.go
```diff
@@ -113,8 +250,14 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
 			Expect(err).NotTo(HaveOccurred())
 
+			ginkgo.By("Waiting for all nodes to be peered before validating advertised routes")
+			for _, c := range FRRContainers {
+				validateFRRPeeredWithAllNodes(cs, c, pairingIPFamily)
+			}
+
 			ginkgo.By("Creating service with expose=true requesting pool-a - should be advertised")
 			svcMatching, _ := testservice.CreateWithBackend(cs, testNamespace, "svc-matching",
+				withIPFamilyForPool(pairingIPFamily),
```
oribon: why is this necessary now? (is this a side-effect of using dynamic ranges?)

e2etest/bgptests/service_selector.go
```diff
@@ -23,9 +29,125 @@ import (
...
+func withIPFamilyForPool(f ipfamily.Family) testservice.Tweak {
+	if f == ipfamily.IPv6 {
+		return testservice.ForceV6
+	}
+	return testservice.ForceV4
+}
+
+// checkServiceNotAdvertised reloads svc from the API, then errors if any LoadBalancer address appears in an FRR route table. Callers use it inside Eventually/Consistently so retries see
+// fresh status (avoids false passes from stale in-memory svc objects).
+func checkServiceNotAdvertised(cs clientset.Interface, ipFamily ipfamily.Family, svc *corev1.Service, reason string) error {
```
oribon: can this accept frrContainers, instead of hardcoding the global FRRContainers?

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +384,620 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+			mergePool := poolAddrV4
+			if pairingIPFamily == ipfamily.IPv6 {
+				// Two LoadBalancers share the pool; use a small range with several addresses
+				// (same pattern as "Pool and service selectors together").
+				ip0, err := config.GetIPFromRangeByIndex(l2tests.IPV6ServiceRange, 0)
+				Expect(err).NotTo(HaveOccurred())
+				ip3, err := config.GetIPFromRangeByIndex(l2tests.IPV6ServiceRange, 3)
+				Expect(err).NotTo(HaveOccurred())
+				mergePool = ip0 + "-" + ip3
+			}
```
oribon: I think we won't need this if we go back to hardcoded ranges in the Entry

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +384,620 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+						p := jigservice.GetIngressPoint(&ing)
+						if err := verifyLocalPrefOnIBGPSession(c, p, mergeAdvertisementLocalPref); err != nil {
+							return err
+						}
+					}
+					for _, ing := range svcS.Status.LoadBalancer.Ingress {
+						p := jigservice.GetIngressPoint(&ing)
+						if err := verifyLocalPrefOnIBGPSession(c, p, mergeAdvertisementLocalPref); err != nil {
+							return err
 						}
```
oribon: can we wrap this in a helper function (verifyBGPAttributes or w.e), and reuse it across the eventuallys?

e2etest/bgptests/service_selector.go
```diff
@@ -23,9 +29,125 @@ import (
...
+			if found {
+				return fmt.Errorf("%s IP %s should NOT be advertised to %s (%s)", svc.Name, ingressIP, c.Name, reason)
+			}
+		}
+	}
+	return nil
+}
+
+const (
+	mergeAdvertisementCommunity = "65000:100"
```
oribon: this and `nonOverlapCommunityProd` are the same, can we have distinct values?

oribon (commented):
thanks! couple of comments, I haven't gone over everything thoroughly yet because I think stuff will shift around after this round

also, please add a comment to both the commit and the PR

e2etest/bgptests/service_selector.go
gkopels: Yes I removed the wrapper and used ipfamily.ForAddress directly.

e2etest/bgptests/service_selector.go
gkopels: **Its not a must, reverted back to the hardcoded**. My AI agent wanted to avoid hardcoded IPv6 pools and instead derive IPv6 test pools from l2tests.IPV6ServiceRange (runtime-provided), so the test would always use addresses that are valid in the current CI/local environment and be less flaky across different network setups.

e2etest/bgptests/service_selector.go
gkopels: No, this is not a side-effect of dynamic ranges rather withIPFamilyForPool(pairingIPFamily) is needed to force the Service to request the same IP family as the pool under test. Especially on dual-stack clusters, Kubernetes may default the Service to the wrong IPfamily.

e2etest/bgptests/service_selector.go
gkopels: Yes updated to accepts frrContainers []*frrcontainer.FRR instead of using global FRRContainers internally.

e2etest/bgptests/service_selector.go
gkopels: Correct reverted back to the hardcoded range

e2etest/bgptests/service_selector.go
gkopels: Yes we can. Updated

e2etest/bgptests/service_selector.go
gkopels: Good idea changed to distinct value.

e2etest/bgptests/validate.go
gkopels: I removed that change.

e2etest/bgptests/service_selector.go
gkopels: moved

internal/bgp/native/native.go
gkopels: I removed the mergeAdvertisements.  I had a failure on native test run and the agent added this fix.  I removed it and native is still passing.

gkopels: Looks like I do need this for native node. The func mergeAdvertisements is needed because native mode  keeps only one advertisement per prefix (last-write-wins), which drops attributes from overlapping BGPAdvertisements.

e2etest/bgptests/service_selector.go
```diff
@@ -187,14 +191,18 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
 
 	ginkgo.DescribeTable("Multiple service selectors",
 		func(pairingIPFamily ipfamily.Family, poolAddress string) {
+			pool := poolAddress
+			if pairingIPFamily == ipfamily.IPv6 {
+				pool = l2tests.IPV6ServiceRange
+			}
```
oribon: why are we not hardcoding the ipv6 range as well? (using "" in the entry)

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +262,693 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+					return nil
+				}, 4*time.Minute, 1*time.Second).ShouldNot(HaveOccurred(), "timed out waiting for aggregated route from staged service")
+			}
 		},
 		ginkgo.Entry("IPV4", ipfamily.IPv4, "192.168.10.0/24"),
+		ginkgo.Entry("IPV6", ipfamily.IPv6, ""),
+	)
+
+	ginkgo.DescribeTable("Overlapping BGP advertisements merge attributes with service selectors",
+		func(pairingIPFamily ipfamily.Family, poolAddrV4 string) {
```
oribon: nit: poolAddrV4 -> poolAddresses
also, no need for that mergePool variable

e2etest/bgptests/service_selector.go
```diff
@@ -62,6 +76,7 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
 	ginkgo.DescribeTable("Pool and service selectors together",
 		func(pairingIPFamily ipfamily.Family, poolAddressA, poolAddressB string) {
 			ginkgo.By("Setting up two pools with labels and advertisement with both selectors")
+			poolA, poolB := poolAddressA, poolAddressB
```
oribon: no need for these aliases (poolA, poolB)

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +262,693 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+// pool is IPv6-only (or the reverse), and MetalLB never assigns an external IP.
+func withIPFamilyForPool(f ipfamily.Family) testservice.Tweak {
+	if f == ipfamily.IPv6 {
+		return testservice.ForceV6
+	}
+	return testservice.ForceV4
+}
+
+// checkServiceNotAdvertised reloads svc from the API, then errors if any LoadBalancer address appears in an FRR route table. Callers use it inside Eventually/Consistently so retries see
+// fresh status (avoids false passes from stale in-memory svc objects).
```
oribon: compact the comment to only what it does (no need for that "callers ...")

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +262,693 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+func checkServiceNotAdvertised(cs clientset.Interface, frrContainers []*frrcontainer.FRR, ipFamily ipfamily.Family, svc *corev1.Service, reason string) error {
```
oribon: accept service name + service namespace instead of *corev1.Service

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +262,693 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+					return err
+				}
+				svcS, err := cs.CoreV1().Services(testNamespace).Get(context.Background(), svcStaged.Name, metav1.GetOptions{})
+				if err != nil {
+					return err
+				}
+				return verifyServiceBGPAttributes(FRRContainers, svcP, svcS, mergeAdvertisementCommunity, mergeAdvertisementLocalPref, false, false)
+			}, 4*time.Minute, 1*time.Second).ShouldNot(HaveOccurred())
+		},
+		// Host-only range: a /24 pool can yield 192.168.10.0 as a VIP, which breaks wget from FRR.
```
oribon: this comment is wrong I think? we use this ip all over the tests

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +262,693 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+				_, found = frrRoutesV6[ingressIP]
+			}
+			if found {
+				return fmt.Errorf("%s IP %s should NOT be advertised to %s (%s)", svc.Name, ingressIP, c.Name, reason)
+			}
+		}
+	}
+	return nil
+}
+
```
oribon: I'm pretty some of the verify* helper functions added here can be replaced by / use the existing functions in validate.go (e.g `validateServiceInRoutesForCommunity`,`validateServiceNotAdvertised` etc,). can you please check what can be reduced here so we don't rewrite the same behaviors?

e2etest/bgptests/service_selector.go
gkopels: updated

e2etest/bgptests/service_selector.go
gkopels: changed

e2etest/bgptests/service_selector.go
gkopels: removed

e2etest/bgptests/service_selector.go
gkopels: changed

e2etest/bgptests/service_selector.go
gkopels: updated

e2etest/bgptests/service_selector.go
gkopels: changed

e2etest/bgptests/service_selector.go
gkopels: updated

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +263,841 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
-			}, 5*time.Second, 1*time.Second).ShouldNot(HaveOccurred())
+					return nil
+				}, 4*time.Minute, 1*time.Second).ShouldNot(HaveOccurred(), "timed out waiting for aggregated route from staged service")
+			}
 		},
 		ginkgo.Entry("IPV4", ipfamily.IPv4, "192.168.10.0/24"),
 		ginkgo.Entry("IPV6", ipfamily.IPv6, "fc00:f853:0ccd:e799::/116"),
 	)
+
+	ginkgo.DescribeTable("BGPAdvertisement and L2Advertisement service selectors apply to matching services only",
```
oribon: isn't this redundant? not sure I understand the purpose here
this seems to check what we already have in `bgptests/service_selector.go` and `l2tests/service_selector.go`.
if the purpose is to see that both l2 and bgp work together, I think we already have a test for that (and if not, it is not a "ServiceSelectors" thing)

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +263,841 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+				validateService(svcStaged, allNodes, c)
+			}
+			Eventually(func() error {
+				return checkServiceNotAdvertised(cs, FRRContainers, pairingIPFamily, svcProd.Namespace, svcProd.Name, "production service must no longer match selector")
+			}, 1*time.Minute, 1*time.Second).ShouldNot(HaveOccurred())
+		},
+		ginkgo.Entry("IPV4 by updating BGPAdvertisement ServiceSelectors", ipfamily.IPv4, "192.168.10.0/24", true),
+		ginkgo.Entry("IPV6 by updating BGPAdvertisement ServiceSelectors", ipfamily.IPv6, "fc00:f853:0ccd:e799::/116", true),
+		ginkgo.Entry("IPV4 by updating Service labels", ipfamily.IPv4, "192.168.10.0/24", false),
+		ginkgo.Entry("IPV6 by updating Service labels", ipfamily.IPv6, "fc00:f853:0ccd:e799::/116", false),
```
oribon: please split this into two tests, the condition logic does not really seem to worth it

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +263,841 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+	)
+
+	ginkgo.DescribeTable("Service selectors with aggregation length webhook validation",
```
oribon: the webhook part is already covered by: "Should reject serviceSelectors with non-default aggregationLength" in `webhooktests`
can you refactor this around the 2nd part, and use existing patterns/functions for aggregation length validation?

e2etest/bgptests/service_selector.go
oribon: we don't need this if now (or pool := poolAddress)

e2etest/bgptests/service_selector.go
oribon: not sure I follow, the existing tests don't use it, so why is it suddenly needed here?

gkopels: You are right thanks updated to align with other tests

e2etest/bgptests/service_selector.go
gkopels: Removed

e2etest/bgptests/service_selector.go
gkopels: You’re right — this was redundant with existing BGP/L2 selector coverage, so I removed it.

e2etest/bgptests/service_selector.go
gkopels: Yes much cleaner. Changed

e2etest/bgptests/service_selector.go
gkopels: “Updated. I removed the webhook part and kept only the split-advertisement aggregation validation using existing test patterns.”

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +252,729 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+		ginkgo.Entry("IPV4", ipfamily.IPv4, "192.168.10.50-192.168.10.200"),
+		ginkgo.Entry("IPV6", ipfamily.IPv6, "fc00:f853:0ccd:e799::/116"),
+	)
+
+	ginkgo.DescribeTable("Non-overlapping BGP advertisements apply distinct communities by service selector",
+		func(pairingIPFamily ipfamily.Family, poolProdV4, poolStagedV4 string) {
+			const poolProd = "nonoverlap-pool-prod"
+			const poolStaged = "nonoverlap-pool-staged"
+
+			poolProdAddr, poolStagedAddr := poolProdV4, poolStagedV4
```
oribon: no need for this

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +252,729 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+				}
+				return verifyServiceBGPAttributes(FRRContainers, svcP, svcS, mergeAdvertisementCommunity, mergeAdvertisementLocalPref, false, false)
+			}, 4*time.Minute, 1*time.Second).ShouldNot(HaveOccurred())
+		},
+		ginkgo.Entry("IPV4", ipfamily.IPv4, "192.168.10.50-192.168.10.200"),
+		ginkgo.Entry("IPV6", ipfamily.IPv6, "fc00:f853:0ccd:e799::/116"),
+	)
+
+	ginkgo.DescribeTable("Non-overlapping BGP advertisements apply distinct communities by service selector",
+		func(pairingIPFamily ipfamily.Family, poolProdV4, poolStagedV4 string) {
```
oribon: remove v4 here

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +252,729 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+			svcProd, err = cs.CoreV1().Services(testNamespace).Get(context.Background(), svcProd.Name, metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			svcStaged, err = cs.CoreV1().Services(testNamespace).Get(context.Background(), svcStaged.Name, metav1.GetOptions{})
+			Expect(err).NotTo(HaveOccurred())
+			if svcProd.Labels == nil {
+				svcProd.Labels = map[string]string{}
+			}
+			if svcStaged.Labels == nil {
+				svcStaged.Labels = map[string]string{}
+			}
```
oribon: this is redundant (services created earlier WithLabels)

e2etest/bgptests/validate.go
oribon: it's still here

e2etest/pkg/frr/bgp.go
oribon, line 1: unrelated changes

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +252,729 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+			if found {
+				return fmt.Errorf("%s IP %s should NOT be advertised to %s (%s)", svc.Name, ingressIP, c.Name, reason)
+			}
+		}
+	}
+	return nil
+}
+
+func verifyPrefixInCommunity(c *frrcontainer.FRR, prefix, community string, wantPresent bool) error {
```
oribon: do we need this, or can we use the existing `validateServiceInRoutesForCommunity` /            
  `validateServiceNotInRoutesForCommunity` instead?

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +252,729 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+	}
+	if inCommunityTable {
+		return fmt.Errorf("prefix %s should not be announced with community %s on %s", prefix, community, c.Name)
+	}
+	return nil
+}
+
+// verifyPrefixInCommunityAnyIBGP succeeds if at least one iBGP test router sees the community on
+// the prefix. Multi-hop vs single-hop FRR views can differ; we only need one correct iBGP witness.
+// eBGP peers are skipped (community checks are iBGP-scoped in this suite).
```
oribon: is this still true? do we really need a new dedicated function?

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +252,729 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+	for _, c := range containers {
+		if !strings.Contains(c.Name, "ibgp") {
+			continue
+		}
+		if err := verifyPrefixInCommunity(c, prefix, community, wantPresent); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
oribon: same

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +252,729 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+	fam := ipfamily.ForAddress(net.ParseIP(prefix))
+	lp, err := frr.LocalPrefForPrefix(c, prefix, fam)
+	if err != nil {
+		return err
+	}
+	if lp != want {
+		return fmt.Errorf("localPref for %s on %s: got %d want %d", prefix, c.Name, lp, want)
+	}
+	return nil
+}
```
oribon: if this is valid only for iBGP, the caller should handle it, see (bgp.go):
```go
						// LocalPref check is only valid for iBGP sessions
						if advertisement.Spec.LocalPref != 0 && strings.Contains(c.Name, "ibgp") {
							localPrefix, err := frr.LocalPrefForPrefix(c, ipWithAdvertisement, ipFamily)
							if err != nil {
								return err
							}
							if localPrefix != advertisement.Spec.LocalPref {
								return fmt.Errorf("%s %s not matching local pref", c.Name, ipWithAdvertisement)
							}
...
```

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +252,729 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+
+func verifyServiceBGPAttributes(containers []*frrcontainer.FRR, svcProd, svcStaged *corev1.Service, community string, localPref uint32, prodHasCommunity, stagedHasCommunity bool) error {
```
oribon: I think this can totally be removed, I don't see what values this adds vs the caller doing these things (currently this hides everything the only test that calls it does)

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +252,729 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+			return nil
+		} else {
+			joined = errors.Join(joined, err)
+		}
+	}
+	return fmt.Errorf("prefix %s for community %s: not found on any ibgp peer: %w", prefix, community, joined)
+}
+
+// verifyPrefixInCommunityEveryIBGP requires every iBGP test router to match wantPresent for the prefix.
+func verifyPrefixInCommunityEveryIBGP(containers []*frrcontainer.FRR, prefix, community string, wantPresent bool) error {
```
oribon: Also, the In/Every variants are confusing, the functions should be smaller and the caller should handle these kinds of things.

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +252,729 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+			}
+			Eventually(func() error {
+				return checkServiceNotAdvertised(cs, FRRContainers, pairingIPFamily, svcProd.Namespace, svcProd.Name, "production service must no longer match selector")
+			}, 1*time.Minute, 1*time.Second).ShouldNot(HaveOccurred())
+		},
+		ginkgo.Entry("IPV4", ipfamily.IPv4, "192.168.10.0/24"),
+		ginkgo.Entry("IPV6", ipfamily.IPv6, "fc00:f853:0ccd:e799::/116"),
+	)
+
+	ginkgo.DescribeTable("Service selectors with split aggregation length advertisements",
```
oribon, line 440: this is a bit messy, can you please look at the existing aggr length e2e (bgp.go "BGP advertisement updates of communities and localpref and aggregation length") and align with it?

e2etest/bgptests/service_selector.go
gkopels: removed

e2etest/bgptests/service_selector.go
gkopels: updated

e2etest/bgptests/service_selector.go
gkopels: removed

e2etest/pkg/frr/bgp.go
gkopels: “Updated — removed the duplicate stale-route joins; now we add each stale route once via errors.Join(serr, fmt.Errorf(...)).”

e2etest/bgptests/service_selector.go
gkopels: “I kept this helper because this test needs prefix-level error checks inside the caller Eventually, including the ‘community present on any iBGP peer’ behavior. The existing validateService*RoutesForCommunity helpers are service-scoped assertion helpers (Eventually/Should), so they don’t fit this flow directly.” WDYT?

e2etest/bgptests/service_selector.go
gkopels: No I removed it

e2etest/bgptests/service_selector.go
gkopels: removed

e2etest/bgptests/service_selector.go
gkopels: Updated

e2etest/bgptests/service_selector.go
gkopels: Removed

e2etest/bgptests/service_selector.go
gkopels: Done

e2etest/bgptests/service_selector.go
gkopels: Done

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +252,730 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+		}
+		return nil
+	}
+	if inCommunityTable {
+		return fmt.Errorf("prefix %s should not be announced with community %s on %s", prefix, community, c.Name)
+	}
+	return nil
+}
+
+func verifyLocalPrefForPrefix(c *frrcontainer.FRR, prefix string, want uint32) error {
```
oribon: can't the caller use the same as bgp.go:
```go
      if strings.Contains(c.Name, "ibgp") {
          Eventually(func() (uint32, error) {                                          
              return frr.LocalPrefForPrefix(c, serviceIP, pairingIPFamily)             
          }, time.Minute, time.Second).Should(Equal(uint32(200)))                      
      }                                                                                
  } 
```

e2etest/bgptests/service_selector.go
oribon: can't you pass the service instead of the prefix? why does the test need "prefix-level" error checks inside the caller?

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +252,730 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+						Spec: metallbv1beta1.IPAddressPoolSpec{
+							Addresses: []string{poolAddresses},
+						},
+					},
+				},
+				Peers: metallb.PeersForContainers(FRRContainers, pairingIPFamily),
+				BGPAdvs: []metallbv1beta1.BGPAdvertisement{
+					// Production matches both advs; MetalLB rejects overlapping ads for one service
+					// if LOCAL_PREF differs (checkBGPAdvConflicts / conflictingAdvertisements).
+					// Use the same localPref on the community adv as on the catch-all adv.
```
oribon: don't mention the code, just write what you're expecting (or nothing)

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +252,730 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+			err := ConfigUpdater.Update(resources)
+			Expect(err).NotTo(HaveOccurred())
+
+			for _, c := range FRRContainers {
+				validateFRRPeeredWithAllNodes(cs, c, pairingIPFamily)
+			}
+
+			svcProd, _ := testservice.CreateWithBackend(cs, testNamespace, "merge-svc-prod",
+				withIPFamilyForPool(pairingIPFamily),
+				testservice.TrafficPolicyCluster,
```
oribon: unnecessary

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +252,730 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+				testservice.WithLabels(map[string]string{"environment": "production"}),
+				testservice.WithAnnotations(map[string]string{"metallb.io/address-pool": poolName}))
```
oribon: unnecessary

e2etest/bgptests/service_selector.go
oribon: this still looks too long, can you align further please?

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +252,730 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+		},
+		ginkgo.Entry("IPV4", ipfamily.IPv4, "192.168.10.50-192.168.10.120", "192.168.20.50-192.168.20.120"),
+		ginkgo.Entry("IPV6", ipfamily.IPv6, "fc00:f853:0ccd:e799::/116", "fc00:f853:0ccd:e800::/116"),
+	)
 })
+
+// withIPFamilyForPool makes the LoadBalancer Service single-stack for the pool under test.
+// On a dual-stack kind cluster, the default Service may otherwise request IPv4 while the
+// pool is IPv6-only (or the reverse), and MetalLB never assigns an external IP.
+func withIPFamilyForPool(f ipfamily.Family) testservice.Tweak {
```
oribon: why do we need this? I thought we agreed it is irrelevant

e2etest/bgptests/service_selector.go
```diff
@@ -251,26 +252,730 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+				testservice.WithAnnotations(map[string]string{"metallb.io/address-pool": poolName}))
+			defer testservice.Delete(cs, svcStaged)
+
+			for _, c := range FRRContainers {
+				validateService(svcProd, allNodes, c)
+				validateService(svcStaged, allNodes, c)
+			}
+
+			ginkgo.By("production routes carry merge community and localPref on iBGP; staged only localPref")
+			Eventually(func() error {
```
oribon: this is way too long and messy (way too many conditionals), please reduce it - use the validation functions, wrap things so it is easier to follow

e2etest/bgptests/service_selector.go
gkopels: Yes sorry missed it

e2etest/bgptests/service_selector.go
gkopels: removed

e2etest/bgptests/service_selector.go
gkopels: I left this one a removed the duplicate ConfigUpdater.Update(resources)

e2etest/bgptests/service_selector.go
gkopels: removed

e2etest/bgptests/service_selector.go
gkopels: Yes removed verifyLocalPrefForPrefix

e2etest/bgptests/service_selector.go
gkopels: Let me know what you think now.

e2etest/bgptests/service_selector.go
gkopels: HI Ori 
I tried simplifing the test a bit the overlapping service-selector test flow by extracting the repeated validation logic into small local closures within the same test to reduce the number of conditiions and make it more readiable. WDYT?

e2etest/bgptests/service_selector.go
```diff
@@ -273,4 +266,591 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+					ip := net.ParseIP(ingressIP)
+					if ip == nil {
+						return fmt.Errorf("invalid ingress IP %q", ingressIP)
+					}
+					// ParseRoutes indexes maps by destination IP string (not full CIDR).
+					keys = append(keys, ip.Mask(net.CIDRMask(int(aggLen), bits)).String())
+				}
+				aggregatedRouteKeys = keys
+				return nil
+			}, 1*time.Minute, 1*time.Second).Should(Succeed(), "timed out waiting for staged service ingress to compute aggregated route keys")
```
oribon: I think you can derive this directly from the Entries (prodLBIP, stagedLBIP) with
```go
  ip := net.ParseIP(prodLBIP)
  subnet := &net.IPNet{
      IP:   ip.Mask(net.CIDRMask(int(aggLen), bits)),
      Mask: net.CIDRMask(int(aggLen), bits),
  }
  aggregatedRouteKey := subnet.IP.String()
```
(similar to how bgp.go does it)

e2etest/bgptests/service_selector.go
```diff
@@ -273,4 +266,591 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+				testservice.WithLabels(map[string]string{environmentLabelKey: environmentLabelStaged}),
+				testservice.WithAnnotations(map[string]string{"metallb.io/address-pool": poolName}))
+			defer testservice.Delete(cs, svcStaged)
+
+			for _, c := range FRRContainers {
+				validateService(svcProd, allNodes, c)
+				validateService(svcStaged, allNodes, c)
+			}
+
+			validateServiceInCommunityOnAnyPeer := func(svc *corev1.Service, community string) error {
```
oribon: can you remind me why we can't use the existing validateServiceInRoutesForCommunity?

e2etest/bgptests/service_selector.go
```diff
@@ -273,4 +266,591 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+						continue
+					}
+					if _, ok := routes[targetIP]; ok {
+						return nil
+					}
+				}
+				return fmt.Errorf("service %s for community %s: not found on any peer: %w", svc.Name, community, joined)
+			}
+
+			validateLocalPrefOnIBGPPeers := func(services ...*corev1.Service) error {
```
oribon: replace this with a direct call to 
```go
      if !strings.Contains(c.Name, "ibgp") { continue }
      Eventually(func() (uint32, error) {
          return frr.LocalPrefForPrefix(c, serviceIP, pairingIPFamily)
      }, time.Minute, time.Second).Should(Equal(mergeAdvertisementLocalPref))
  }
```
or something like that, this function hides too many things inside

e2etest/bgptests/service_selector.go
```diff
@@ -273,4 +266,591 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+		ginkgo.Entry("IPV4", ipfamily.IPv4, "192.168.10.0/24"),
+		ginkgo.Entry("IPV6", ipfamily.IPv6, "fc00:f853:0ccd:e799::/116"),
+	)
+
+	ginkgo.DescribeTable("Service selectors with split aggregation length advertisements",
+		func(pairingIPFamily ipfamily.Family, poolAddress string, aggLen int32, prodLBIP, stagedLBIP string) {
+			const (
+				poolName            = "selector-aggregation-pool"
+				selectorOnlyAdvName = "adv-selector-only"
+				aggOnlyAdvName      = "adv-aggregation-only"
```
oribon: unnecessary (used once)

e2etest/bgptests/service_selector.go
```diff
@@ -273,4 +266,591 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+		},
+		ginkgo.Entry("IPV4", ipfamily.IPv4, "192.168.10.0/24"),
+		ginkgo.Entry("IPV6", ipfamily.IPv6, "fc00:f853:0ccd:e799::/116"),
+	)
+
+	ginkgo.DescribeTable("Service selectors with split aggregation length advertisements",
+		func(pairingIPFamily ipfamily.Family, poolAddress string, aggLen int32, prodLBIP, stagedLBIP string) {
+			const (
+				poolName            = "selector-aggregation-pool"
+				selectorOnlyAdvName = "adv-selector-only"
```
oribon: unnecessary (used once)

e2etest/bgptests/service_selector.go
gkopels: Thanks  I took the aggregated prefix directly from stagedLBIP + aggLen (like bgp.go) and removed the extra Eventually key-derivation step.

e2etest/bgptests/service_selector.go
gkopels: Hi Ori no reason. It was an AI add-on that I missed. It was checking that at least one peer had the correct output. I have aligned with the bgp test case.

e2etest/bgptests/service_selector.go
gkopels: Yes much better I removed that helper and switched to direct iBGP Eventually localPref checks (same style as bgp.go).

e2etest/bgptests/service_selector.go
gkopels: done

e2etest/bgptests/service_selector.go
gkopels: Remove and checked for any other const used only once and didnt find.

e2etest/bgptests/service_selector.go
```diff
@@ -273,4 +265,553 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+			for _, c := range FRRContainers {
+				validateService(svcProd, allNodes, c)
+				validateService(svcStaged, allNodes, c)
+			}
+
+			ginkgo.By("production routes use prod community only; staged routes use staged community only")
+			for _, c := range FRRContainers {
+				if !strings.Contains(c.Name, "ibgp") {
+					continue
+				}
```
oribon: not needed (communities check is against all containers above)

e2etest/bgptests/service_selector.go
```diff
@@ -273,4 +265,553 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+		ginkgo.Entry("IPV6", ipfamily.IPv6, "fc00:f853:0ccd:e799::/116", "fc00:f853:0ccd:e800::/116"),
+	)
 })
+
+// checkServiceNotAdvertised returns an error if any service LoadBalancer address is present in FRR routes.
+func checkServiceNotAdvertised(svc *corev1.Service, reason string) error {
+	if len(svc.Status.LoadBalancer.Ingress) == 0 {
+		return fmt.Errorf("service %s/%s has no ingress IPs", svc.Namespace, svc.Name)
+	}
+	for _, c := range FRRContainers {
```
oribon: accept the containers as input instead of hardcoding FRRContainers

e2etest/bgptests/service_selector.go
```diff
@@ -273,4 +265,553 @@ var _ = ginkgo.Describe("BGP Service Selector", func() {
...
+							Name:      advLocalPrefName,
+							Namespace: ConfigUpdater.Namespace(),
+						},
+						Spec: metallbv1beta1.BGPAdvertisementSpec{
+							IPAddressPools: []string{poolName},
+							LocalPref:      mergeAdvertisementLocalPref,
+						},
+					},
+				},
+			}
```
oribon, line 605: we're not really testing the merge here (because advCommunityName has both localpref and community), wdyt about having 2 ads, each with a different community, and see that the prod svc gets both while staged gets only 1?
in a way, the aggregation length case already covers the "merge scneario", up to you if you want to refactor this test along communities or drop it entirely.

oribon (commented):
last small comments, looks a lot better thanks!
please remove the L2 mentions in the commit message and in the PR as well

e2etest/bgptests/service_selector.go
gkopels: Updated — removed the iBGP-only filter and now validate community presence/absence across all FRR peers in that test.

e2etest/bgptests/service_selector.go
gkopels: Updated — checkServiceNotAdvertised now takes containers as input (no hardcoded FRRContainers).

e2etest/bgptests/service_selector.go
gkopels: Agreed, I still think its a good test case. I updated the test to validate community merge behavior directly: production now receives both communities from overlapping advertisements, while staged receives only the shared community.

oribon (approved):
lgtm, thanks!

PR #3013, opened 2026-04-13
bgp: fix bgppeer validation

adilGhaffarDev:
Modify the validation to check whether two peers with different `myASN` in the same VRF target the same node, instead of blindly comparing all peers cluster-wide. So that the one-ASN-per-VRF constraint is preserved andit is enforced at the correct scope. The key insight is that two BGPPeers with different `myASN` in the same VRF are only a problem if they land on the same node. If `nodeSelectors` ensure they target different nodes, no single FRR instance ever sees both, the constraint is naturally satisfied. The fix replaces the cluster-wide check with one that resolves `nodeSelectors` against actual node labels to determine whether two peers can land on the same node.

>

/kind bug

> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note

```

### DIFF
--- a/internal/config/validation.go
+++ b/internal/config/validation.go
@@ -10,6 +10,7 @@ import (
 	metallbv1beta2 "go.universe.tf/metallb/api/v1beta2"
 	"go.universe.tf/metallb/internal/bgp/community"
 	"go.universe.tf/metallb/internal/ipfamily"
+	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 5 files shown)


fedepaol:
seems like a dupe of https://github.com/metallb/metallb/pull/2980 ? It even have the same naming :sweat_smile:

adilGhaffarDev:
> seems like a dupe of #2980 ? It even have the same naming 😅

i will close mine

PR #3014, opened 2026-04-15
Fix L2 speaker election ignoring service selectors

oribon:
The L2 leader election used all ads of the pool to build the candidate set, completely ignoring service selectors.
This meant a node whose advertisement targeted a different service could win the election, causing the service to not be announced.
Here we change this so that all ads relevant to the specific service are evaluated for the candidate set, ensuring both:
1) only nodes that are relevant are evaluated
2) the election is consistent across all nodes

Assisted-by: Cursor

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
Fix L2 speaker election ignoring service selectors, making sure only the relevant L2Advertisements for the service are considered.
```

### DIFF
--- a/e2etest/l2tests/service_selector.go
+++ b/e2etest/l2tests/service_selector.go
@@ -4,6 +4,7 @@ package l2tests
 
 import (
 	"context"
+	"fmt"
 	"time"
 
 	"github.com/onsi/ginkgo/v2"
@@ -15,6 +16,7 @@ import (
 	"go.universe.tf/e2etest/pkg/status"
 	metallbv1beta1 "go.universe.tf/metallb/api/v1beta1"
 
+	corev1 "k8s.io/api/core/v1"
 	pkgerr "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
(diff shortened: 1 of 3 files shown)


andreaskaris:
@oribon Can you throw in a decent E2E that catches this issue (if present) ; otherwise, if the E2E passes, the issue isn't present?

speaker/layer2_controller_test.go
```diff
@@ -1925,6 +1919,364 @@ func selector(s string) labels.Selector {
...
+			t.Fatalf("creating controller for %s: %s", node, err)
+		}
+		c.client = &testK8S{t: t}
+
+		if c.SetConfig(l, cfg) == controllers.SyncStateError {
+			t.Fatalf("%s SetConfig failed", node)
+		}
+
+		r := c.protocolHandlers[config.Layer2].ShouldAnnounce(l, "test1", []net.IP{lbIP}, pool, svc, eps, nil)
+		if r != "noMatchingAdvertisement" {
```
andreaskaris, line 2124: unrelated to this commit, but "noMatchingAdvertisement" and "noOwner" should ideally be constants, shouldn't they? makes it easier to understand that they have a special meaning and aren't just made up on the fly

oribon: we could, but they don't really have special meanings (in the sense that we don't act upon the values, the code only cares about the reason being empty or not). do you still prefer I change these to consts?

speaker/layer2_controller_test.go
```diff
@@ -1925,6 +1919,364 @@ func selector(s string) labels.Selector {
...
+			r2 := c2.protocolHandlers[config.Layer2].ShouldAnnounce(l, "test1", []net.IP{lbIP}, cfg.Pools.ByName["default"], test.svc, eps, nil)
+
+			if r1 != test.c1ExpectedResult {
+				t.Errorf("c1 (iris1): expected %q, got %q", test.c1ExpectedResult, r1)
+			}
+			if r2 != test.c2ExpectedResult {
+				t.Errorf("c2 (iris2): expected %q, got %q", test.c2ExpectedResult, r2)
+			}
+
+			// Regardless of which node wins the election, exactly one controller
```
andreaskaris: nit: that verification is kind of pointless given that the test cases check for 
```
			c1ExpectedResult: "",
			c2ExpectedResult: "noMatchingAdvertisement",
```

and 
```
			c1ExpectedResult: "noMatchingAdvertisement",
			c2ExpectedResult: "",
```
and we're guaranteed to be in this state here. I'm fine keeping it if you want to be explicit, though

e2etest/l2tests/service_selector.go
```diff
@@ -177,6 +178,161 @@ var _ = ginkgo.Describe("L2-ServiceSelector", func() {
...
+			defer func() {
+				err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
+				Expect(err).NotTo(HaveOccurred())
+			}()
+
+			ginkgo.By("Checking that exactly one of the two candidate nodes announces the service")
+			Eventually(func() string {
+				l2Status, err := status.L2ForService(ConfigUpdater.Client(), svc)
+				if err != nil {
+					return ""
```
andreaskaris: nit: would be good to log the error message at least. or .. instead of returning "", can't we return the err.Error()? should still work, and at least we report the error message on failure? though that's hacky, so perhaps better logging it

speaker/layer2_controller_test.go
oribon: removed

e2etest/l2tests/service_selector.go
```diff
@@ -177,6 +178,161 @@ var _ = ginkgo.Describe("L2-ServiceSelector", func() {
...
+				service.WithLabels(map[string]string{"app": "web"}))
+			defer func() {
+				err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
+				Expect(err).NotTo(HaveOccurred())
+			}()
+
+			ginkgo.By("Checking that no node announces the service")
+			Consistently(func() bool {
+				_, err := status.L2ForService(ConfigUpdater.Client(), svc)
+				return pkgerr.IsNotFound(err)
```
andreaskaris: nit: the same here .. what if it's anything else but IsNotFound .. on failure, we don't get that info

andreaskaris:
/lgtm

it's up to you if you want to address the other comments, they aren't blockers for me

speaker/layer2_controller_test.go
andreaskaris: nah, not needed

e2etest/l2tests/service_selector.go
oribon: done

e2etest/l2tests/service_selector.go
oribon: done

PR #3017, opened 2026-04-19
ConfigState: Add OwnerRef to resources

### LINKED ISSUE #3015: Update/remove ConfigurationStates when nodes are removed or no longer run the speaker

Somewhere between a feature request and a bug admittedly. MetalLB appears to never clean up ConfigurationState resources that are for stale/removed speaker nodes. This has caused MetalLB to litter our test cluster with thousands of these "valid" state resources for VMs that no longer exist, as it's a Cluster API backed cluster and nodes are added/removed with unique names for every config change or scaling event rather than being reused. I don't need to see stale states for nodes that got replaced last week, never mind ones from months ago. It also leaves ConfigurationState resources for nodes that are no longer running the speaker due to label or taint changes, while still claiming them as "valid".  Less of an issue since I only had the thought it might happen, discovering that it does not reflect the node no longer running the speaker in any way.

Both cases are really minor in the grand scheme of things, though having a growing pile of stale records with no active benefit does adds up to bigger etcd backups and longer replication times for control node replacements.

At minimum, nodes that are removed (or otherwise no longer exist) should have their matching ConfigurationState resource removed with in a reasonable amount of time. This could be done by leveraging [k8s' native garbage colection](https://kubernetes.io/docs/concepts/architecture/garbage-collection/#owners-dependents) or by having the controller delete them (potentially on a delay just in case they come back or for humans looking for breadcrumbs to MetalLB issues).

Meanwhile, nodes that no longer run the speaker (but are not removed) should either have their ConfigurationState resource removed following the same logic above (since ConfigurationStates are not generated for nodes that never ran the speaker before), or be updated to say they are invalid due to no longer having the speaker scheduled/running (with the potential of being culled after a delay, again for humans looking for breadcrumbs to MetalLB issues).

### Additional context

_No response_

oribon:
We add owner refs to the generated resources, aligned with how we already assign them to the service(l2/bgp)status resources to get free cleanups.
Fixes #3015.

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
Add OwnerRef to ConfigurationState resources, enabling cleanup by the k8s gc.
```

### DIFF
--- a/controller/main.go
+++ b/controller/main.go
@@ -170,6 +170,7 @@ func main() {
 		loadBalancerClass   = flag.String("lb-class", "", "load balancer class. When enabled, metallb will handle only services whose spec.loadBalancerClass matches the given lb class")
 		webhookMode         = flag.String("webhook-mode", "enabled", "webhook mode: can be enabled, disabled or only webhook if we want the controller to act as webhook endpoint only")
 		webhookSecretName   = flag.String("webhook-secret", "metallb-webhook-cert", "webhook secret: the name of webhook secret, default is metallb-webhook-cert")
+		myPod               = flag.String("pod-name", os.Getenv("METALLB_POD_NAME"), "name of this MetalLB controller pod")
 		webhookHTTP2        = flag.Bool("webhook-http2", false, "enables http2 for the webhook endpoint")
 		tlsMinVersion       = flag.String("tls-min-version", "", "Minimum TLS version (VersionTLS12 or VersionTLS13). If empty, defaults to VersionTLS13.")
 		tlsCipherSuites     = flag.String("tls-cipher-suites", "", "Comma-separated list of TLS cipher suites. Only applies to TLS 1.2. If empty, uses Go defaults.")
@@ -223,6 +224,7 @@ func main() {
 
 	cfg := &k8s.Config{
 		ProcessName:      "metallb-controller",
+		PodName:          *myPod,
 		MetricsPort:      *port,
 		PprofBindAddress: *pprofBindAddress,
 		Logger:           logger,
--- a/internal/k8s/controllers/configuration_state_controller.go
+++ b/internal/k8s/controllers/configuration_state_controller.go
@@ -4,11 +4,13 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"strings"
 
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	metallbv1beta1 "go.universe.tf/metallb/api/v1beta1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
@@ -27,6 +29,7 @@ type ConfigurationStateReconciler struct {
 	Namespace         string
 	ConfigStateName   string
 	ConfigStateLabels map[string]string
+	OwnerPod          *v1.Pod
 }
 
 func (r *ConfigurationStateReconciler) String() string {
@@ -46,6 +49,9 @@ func (r *ConfigurationStateReconciler) Reconcile(ctx context.Context, req ctrl.R
 
 	opResult, err := controllerutil.CreateOrPatch(ctx, r.Client, configState, func() error {
 		configState.Labels = r.ConfigStateLabels
+		if err := controllerutil.SetOwnerReference(r.OwnerPod, configState, r.Scheme); err != nil {
+			return err
+		}
 
 		result := metallbv1beta1.ConfigurationResultUnknown
 		if len(configState.Status.Conditions) > 0 {
@@ -68,14 +74,7 @@ func (r *ConfigurationStateReconciler) Reconcile(ctx context.Context, req ctrl.R
 		level.Error(r.Logger).Log("controller", r, "error", "failed to create or patch ConfigurationState", "error", err)
 		return ctrl.Result{}, err
 	}
-	if opResult == controllerutil.OperationResultCreated {
-		// According to https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrPatch
-		// If the object is created, we have to patch it again to ensure the status is created.
-		// This will happen when we reconcile the creation event.
-		level.Info(r.Logger).Log("controller", r, "event", "ConfigurationState created", "name", r.ConfigStateName)
-		return ctrl.Result{}, nil
-	}
-	level.Info(r.Logger).Log("controller", r, "event", "ConfigurationState updated", "name", r.ConfigStateName)
+	level.Info(r.Logger).Log("controller", r, "event", fmt.Sprintf("ConfigurationState %s", opResult), "name", r.ConfigStateName)
 
 	return ctrl.Result{}, nil
 }
--- a/internal/k8s/controllers/configuration_state_controller.go
+++ b/internal/k8s/controllers/configuration_state_controller.go
@@ -9,11 +9,11 @@ import (
 	"github.com/go-kit/log"
 	"github.com/go-kit/log/level"
 	metallbv1beta1 "go.universe.tf/metallb/api/v1beta1"
-	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
 	"sigs.k8s.io/controller-runtime/pkg/predicate"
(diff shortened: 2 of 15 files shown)


internal/k8s/controllers/configuration_state_controller.go
```diff
@@ -37,48 +39,48 @@ func (r *ConfigurationStateReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+			if cond.Status == metav1.ConditionFalse && cond.Reason == ErrorTypeConfiguration {
+				result = metallbv1beta1.ConfigurationResultInvalid
+				errorMessages = append(errorMessages, cond.Message)
+			}
+		}
+
+		configState.Status.Result = result
+		configState.Status.ErrorSummary = strings.Join(errorMessages, "\n")
+		return nil
+	})
```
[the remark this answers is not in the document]
andreaskaris: While gemini's comment isn't true, I looked at the description of 
 `func controllerutil.CreateOrPatch` and it says:

> Note: changes to any sub-resource other than status will be ignored. Changes to the status sub-resource will only be applied if the object already exist. To change the status on object creation, the easiest way is to requeue the object in the controller if OperationResult is OperationResultCreated

So upon creation, the Status won't be set. Meaning the new code isn't 1:1 equivalent. I suppose the reconciler is triggered a second time after resource creation, and the second pass does the Status update now?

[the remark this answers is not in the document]
andreaskaris: me~~ - just a few lines lower in:
```
	if opResult == controllerutil.OperationResultCreated {
		// According to https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/controller/controllerutil#CreateOrPatch
		// If the object is created, we have to patch it again to ensure the status is created.
		// This will happen when we reconcile the creation event.
		level.Info(r.Logger).Log("controller", r, "event", "ConfigurationState created", "name", r.ConfigStateName)
		return ctrl.Result{}, nil
	}
```

internal/k8s/controllers/configuration_state_controller.go
```diff
@@ -37,48 +37,45 @@ func (r *ConfigurationStateReconciler) Reconcile(ctx context.Context, req ctrl.R
...
+	if err != nil {
+		level.Error(r.Logger).Log("controller", r, "error", "failed to create or patch ConfigurationState", "error", err)
 		return ctrl.Result{}, err
 	}
+	if opResult == controllerutil.OperationResultCreated {
```
andreaskaris: nit: instead of branching, let's:
```
  level.Info(r.Logger).Log("controller", r, "event", fmt.Sprintf("ConfigurationState %s", opResult), "name", r.ConfigStateName)
	return ctrl.Result{}, nil
```

comment `According to ...` can then go before line 60 inside CreateOrPAtch, for example

andreaskaris:
/lgtm  (up to you if you want to address the nit or not)

edit: I was thinking about an E2E for a second, but given that cleanup with a ref set is default k8s behavior, it's prob overkill

karampok:
On high level, I want to ask if is possible/be considered the option to have the node as owner instead of the pod?

The original design targeted that we have configstate per node (even global but split per node for scalability), where we could extend that other related to metallb components could write condition (e.g. frrk8s could contribute a condition). Not that we are using but that breaks now (if speaker pod restart, the condition are not saved).

Also, having per node we could have less impact if speaker restart often for a reason.

wdyt? thanks

controller/main.go
```diff
@@ -170,6 +170,7 @@ func main() {
 		loadBalancerClass   = flag.String("lb-class", "", "load balancer class. When enabled, metallb will handle only services whose spec.loadBalancerClass matches the given lb class")
 		webhookMode         = flag.String("webhook-mode", "enabled", "webhook mode: can be enabled, disabled or only webhook if we want the controller to act as webhook endpoint only")
 		webhookSecretName   = flag.String("webhook-secret", "metallb-webhook-cert", "webhook secret: the name of webhook secret, default is metallb-webhook-cert")
+		myPod               = flag.String("pod-name", os.Getenv("METALLB_POD_NAME"), "name of this MetalLB controller pod")
```
karampok, line 173: Is there any legit usecase a user can change that? Or is just to be same as other parts?

If I understand we cannot use any other value than the real pod name because we do a call to get UID that will fail, no?

internal/k8s/controllers/configuration_state_controller_test.go
```diff
@@ -164,6 +181,101 @@ func TestConfigurationStateReconciler(t *testing.T) {
...
+		t.Helper()
+		if len(got.OwnerReferences) != 1 {
+			t.Fatalf("expect 1 owner reference, but got %d", len(got.OwnerReferences))
+		}
+		if got.OwnerReferences[0].UID != ownerPod.UID {
+			t.Errorf("owner reference is not speaker pod, expect owner reference uid %s, but got %s", ownerPod.UID, got.OwnerReferences[0].UID)
+		}
+	}
+
+	t.Run("set on create", func(t *testing.T) {
```
karampok: That can  be a table/map driven, test code seems same enough
(plus in case we want to extend).

internal/k8s/controllers/configuration_state_controller_test.go
```diff
@@ -153,7 +170,7 @@ func TestConfigurationStateReconciler(t *testing.T) {
 			}
 
 			opts := []cmp.Option{
-				cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ResourceVersion", "UID", "CreationTimestamp", "Generation", "ManagedFields"),
+				cmpopts.IgnoreFields(metav1.ObjectMeta{}, "ResourceVersion", "UID", "CreationTimestamp", "Generation", "ManagedFields", "OwnerReferences"),
```
karampok: if we ignore that, then we do not actually test the feature of the PR, no?

oribon:
> On high level, I want to ask if is possible/be considered the option to have the node as owner instead of the pod?
> 
> The original design targeted that we have configstate per node (even global but split per node for scalability), where we could extend that other related to metallb components could write condition (e.g. frrk8s could contribute a condition). Not that we are using but that breaks now (if speaker pod restart, the condition are not saved).
> 
> Also, having per node we could have less impact if speaker restart often for a reason.
> 
> wdyt? thanks

I think the pod is the right owner for several reasons:
1. the configstate represents what the speaker/controller applied, if they go away and the node is still up the resource becomes stale.
2. `if speaker pod restart, the condition are not saved` - then frrk8s (or w.e controller) will just rewrite their conditions, which is what currently happens with the other controllers reporting their conditions separately.
3. it's consistent with the other status controllers.

if this makes sense, then I'd prefer keeping the pod the owner of its resource

internal/k8s/controllers/configuration_state_controller.go
oribon: done (removed the comment)

karampok:
> I think the pod is the right owner for several reasons:

I do not think there is right/wrong


> 1. the configstate represents what the speaker/controller applied, if they go away and the node is still up the resource becomes stale.

Which is okay also, in terms if  someone adds a taint on the node, and the speaker pods is removed, the user will see a CR with condition not being updated. IMO we care about lifecycle of the condition not the CR itself.

> 2. `if speaker pod restart, the condition are not saved` - then frrk8s (or w.e controller) will just rewrite their conditions, which is what currently happens with the other controllers reporting their conditions separately.

That is not accurate, because external to pod controller will not reconcile and ideally we should not trigger reconcile to all if pod restarts (e.gif goes into crashloop the impact should be minimal). But we do not have that requirement, so is premature optimization

> 3. it's consistent with the other status controllers.

That is probably true.

> if this makes sense, then I'd prefer keeping the pod the owner of its resource

Since we do have/plan to have frrk8s report anything, this implementation fits our need. So I am okay keeping the pod owner.

controller/main.go
oribon: right (and it's consistent with the existing speaker flag)

internal/k8s/controllers/configuration_state_controller_test.go
oribon: done

internal/k8s/controllers/configuration_state_controller_test.go
oribon: done

controller/main.go
karampok: right == user can never change the value and we just have it because it makes sense in the past?

if yes, imo we should not just add  it just because we had that in the speaker.

internal/k8s/controllers/configuration_state_controller_test.go
```diff
@@ -164,6 +195,85 @@ func TestConfigurationStateReconciler(t *testing.T) {
 	}
 }
 
+func TestConfigurationStateReconcilerOwnerReference(t *testing.T) {
```
karampok: Do we really test someting different than the
```
func TestConfigurationStateReconciler(t *testing.T) {
```
it seems just duplicate testing, no?

controller/main.go
oribon: > we just have it because it makes sense in the past

honestly, I'm not sure - maybe because declaring it in the flag block looks cleaner.
if it's not critical, I prefer to align with the existing pattern (in controller as well, where we have os.Getenv inside a flag.String), and if you think we should still change it then maybe in a different pr?

internal/k8s/controllers/configuration_state_controller_test.go
oribon: looks like it (sorry), removed

controller/main.go
karampok: me, claude and gemini do not agree is cleaner :P but we all agree not critical.

karampok (approved):

PR #3022, opened 2026-04-20
e2e: Keep polling until FRR daemons are not started.

adilGhaffarDev:
PollImmediate stops immediately and propagates the error if the condition function returns `false, err`. On the very first poll attempt, FRR daemons haven't fully started yet so `vtysh -c show daemons` fails with exit code 1. Returning `false, err` caused the wait to abort instantly instead of retrying. Changing it to `false, nil` makes the poll silently retry every second for up to 5 minutes until all daemons are reported as running.

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design

/kind flake

> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note

```


oribon:
was fixed in https://github.com/metallb/metallb/pull/2996/changes/c47b0134ebc49933c85a31efd9f11d95f177c620, thanks!

PR #3032, opened 2026-04-27
Deny allow-shared-ip with ServiceSelectors on L2

### LINKED ISSUE #3025: Shared-IP services can split across nodes when matched by different L2Advertisements

With the current configuration options, I can't use allow-shared-ip when L2Advertisements have serviceSelectors without risking a split announcement.

Services sharing an IP need a single announcing node. ShouldAnnounce runs per service and filters L2Advertisements by serviceSelectors. When sibling services match different ads with different nodeSelectors, each service elects a different winner. Both nodes answer ARP for the same IP and traffic splits.

Example: two services with metallb.universe.tf/allow-shared-ip: <key> on the same IP.
- svcA has app=a, matched by Ad1 with nodeSelectors: [role=primary]
- svcB has app=b, matched by Ad2 with nodeSelectors: [role=secondary]

svcA elects a primary-role node. svcB elects a secondary-role node. Both announce the shared IP.

We discussed this in #3004, essentiallying disallowing the combination of allow-shared-ip with serviceSelectors on L2Advertisements. 

### Additional context

Also related #3014.

oribon:

>
/kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

When services share an IP via allow-shared-ip and L2Advertisements
have serviceSelectors, different sibling services can match different
advertisements with different nodeSelectors, causing split
announcements where multiple nodes answer ARP for the same IP.

Deny this combination in the L2 speaker by refusing to announce
services that have the allow-shared-ip annotation when any of their
matching L2Advertisements has serviceSelectors set.

Fixes https://github.com/metallb/metallb/issues/3025

**Release note**:

```release-note
Deny allow-shared-ip with ServiceSelectors on L2.
```


oribon:
Relies on https://github.com/metallb/metallb/pull/3014

speaker/layer2_controller.go
```diff
@@ -36,6 +36,11 @@ import (
 	"go.universe.tf/metallb/internal/layer2"
 )
 
+const (
+	annotationAllowSharedIP           = "metallb.io/allow-shared-ip"
+	deprecatedAnnotationAllowSharedIP = "metallb.universe.tf/allow-shared-ip"
```
yahlifried, line 41: same consts are defined in three places (here, controller/service.go, e2etest/l2test/l2.go lines 486,496)
potential drift if annotation changes again
might be worth exporting to a consts internal package?

yahlifried (commented):
this fix has the potential to break working clusters announcements upon upgrading, since this is not handled in the webhook. I thought it was worth noting but I'm not sure what can be done about it

speaker/layer2_controller.go
oribon: this is intentional, we avoid leaks/coupling between e2e and code. as for creating a consts package, this is the only consts afaik that are shared between the speaker and controller, and imo is not worth the package abstraction

oribon:
> this fix has the potential to break working clusters announcements upon upgrading, since this is not handled in the webhook.

the existing behavior is a bug and is fragile (existing clusters would work "by accident"), so this is a non-issue imo (breaking it on upgrade would be "fixing" it in a way)

PR #3044, opened 2026-05-13
L2 Mode Preferred Node Selectors

### LINKED ISSUE #2797: Adds the ability to explicitly designate a primary node for services running in L2 mode

jlandersen:
Fixes https://github.com/metallb/metallb/issues/2797, and implements the design in https://github.com/metallb/metallb/pull/3004.

/kind feature

Adds support for putting a preference on L2 speaker election toward labeled nodes, useful for keeping VIPs on dedicated edge nodes / specific zones / etc.

- I have sliced up the pieces in separate commits for each area. The actual logic is a quite small part of the overall changes. Most are tests and manifest related changes.
- I had to make a couple of changes in `setL2AdvertisementsToPools` and its call to `containsAdvertisement` to make sure we don't end up double counting preference weights, but still allow preference weights to sum across different CRs targeting the same pool. Please do let me know if there's any better thoughts on this.
- I have tried to add user docs as well where I believe it fits in, but if there's a different preference for how this should be written I'm happy to adjust it :)

**Release note**:

```release-note
Extend L2Advertisement with PreferredNodeSelectors to steer L2 speaker election toward nodes matching weighted label selectors. When unset, election falls back to the existing hash-based tie-break.
```


oribon:
just adding a note that the 2 ci fails are unrelated, and I'll try to get to this pr this week :) thanks!

internal/config/config_test.go
```diff
@@ -3924,6 +3924,36 @@ func TestContainsAdvertisement(t *testing.T) {
 			},
 			expect: false,
 		},
+		{
+			desc: "Ads carrying preferences never dedupe so speaker can sum weights",
```
oribon, line 3929: can we also add the 2 other cases? advs have preferred toCheck does not, and vice-versa

internal/config/preferred_node_test.go
```diff
@@ -0,0 +1,280 @@
...
+		{
+			desc: "equal weights apply independently per selector",
+			nodes: []corev1.Node{
+				{ObjectMeta: metav1.ObjectMeta{Name: "edge-a", Labels: map[string]string{"role": "edge"}}},
+				{ObjectMeta: metav1.ObjectMeta{Name: "worker-a", Labels: map[string]string{"role": "worker"}}},
+			},
+			eligible: map[string]bool{"edge-a": true, "worker-a": true},
+			preferred: []v1beta1.PreferredNodeSelector{
+				{Weight: 50, Preference: metav1.LabelSelector{MatchLabels: map[string]string{"role": "edge"}}},
+				{Weight: 50, Preference: metav1.LabelSelector{MatchLabels: map[string]string{"role": "worker"}}},
```
oribon: let's have this a different value than 50 (to be extra sure that each node got its score from their selector)

internal/config/preferred_node_test.go
```diff
@@ -0,0 +1,280 @@
...
+				t.Fatalf("preferredNodeScores returned error: %v", err)
+			}
+			if diff := cmp.Diff(tc.want, got); diff != "" {
+				t.Fatalf("preferredNodeScores mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestL2AdvertisementFromCR(t *testing.T) {
```
oribon: not sure I understand the value of this (and the name makes me feel like it's already covered in config_test / belong there)
also, let's be explicit with wantSvcSelectors (instead of int)

internal/config/preferred_node_test.go
```diff
@@ -0,0 +1,280 @@
...
+				t.Fatalf("PreferredNodes mismatch (-want +got):\n%s", diff)
+			}
+			if tc.wantSvcSelectors > 0 && len(got.ServiceSelectors) != tc.wantSvcSelectors {
+				t.Fatalf("ServiceSelectors: want %d, got %d", tc.wantSvcSelectors, len(got.ServiceSelectors))
+			}
+		})
+	}
+}
+
+func TestSetL2AdvertisementsToPools(t *testing.T) {
```
oribon: same, this probably belongs in config_test
also, can we be explicit instead of using wantLen?

speaker/layer2_controller.go
```diff
@@ -177,6 +183,22 @@ func (c *layer2Controller) SetEventCallback(callback func(interface{})) {
...
+// preferredScoresForService sums per-advertisement preference weights across
+// all ads applicable to a service. Returns nil when no ad carries preferences.
+// Callers must tolerate a nil map, where reads return zero.
+func preferredScoresForService(ads []*config.L2Advertisement) map[string]int64 {
+	var scores map[string]int64
+	for _, ad := range ads {
+		for node, weight := range ad.PreferredNodes {
+			if scores == nil {
+				scores = make(map[string]int64, len(ad.PreferredNodes))
+			}
```
oribon: can we init unconditionally outside the loop? this is a minor optimization that hurts readability a bit 😅

speaker/layer2_controller_test.go
```diff
@@ -2332,3 +2334,388 @@ func TestIPAdvertisementFor(t *testing.T) {
...
+			t.Fatalf("SetConfig failed")
+		}
+		if c.protocolHandlers[config.Layer2].ShouldAnnounce(l, "svc1", []net.IP{lbIP}, cfg.Pools.ByName["default"], svc, eps, nodes) == "" {
+			winners = append(winners, nodeName)
+		}
+	}
+	return winners
+}
+
+func hashWinner(ip string, candidates ...string) string {
```
oribon: hashWinner leaking from the code and inlined here is a bit messy, wdyt about wrapping the candidate sorting logic in a function and unit test it in a separate commit, and then in your new commits we update the candidate sorting logic function and update the unit tests

I think if we do that, the tests below can be embedded in the existing shouldannounce tests (the tables) and we can drop the runl2election helper

e2etest/l2tests/preferred_node.go
```diff
@@ -0,0 +1,679 @@
...
+			Expect(err).NotTo(HaveOccurred())
+
+			ginkgo.By(fmt.Sprintf("Validating %s (score 110) announces the service", nodeDouble))
+			Eventually(announcingNode(svc), 2*time.Minute, time.Second).Should(Equal(nodeDouble))
+
+			ginkgo.By(fmt.Sprintf("Removing l2-pref-role from %s so it now scores only 50", nodeDouble))
+			k8s.RemoveLabelFromNode(nodeDouble, "l2-pref-role", cs)
+
+			ginkgo.By(fmt.Sprintf("Validating %s (score 60) now beats %s (score 50)", nodeSingle, nodeDouble))
+			Eventually(announcingNode(svc), 2*time.Minute, time.Second).Should(Equal(nodeSingle))
```
oribon: the l2-pref-role carries weight 50, so after removal both nodes will have score 60 😅 
maybe this should remove the l2-pref-zone?

e2etest/l2tests/preferred_node.go
```diff
@@ -0,0 +1,679 @@
...
+		if networkUnavailable {
+			continue
+		}
+		ok, err := metallb.HasSpeakerInNode(cs, n.Name)
+		if err == nil && ok {
+			out = append(out, *n)
+		}
+	}
+	return out
+}
```
oribon: do we really need this level of granularity? the existing l2 tests don't need this, so I'm not sure I understand why we suddenly need it here

e2etest/l2tests/preferred_node.go
```diff
@@ -0,0 +1,679 @@
...
+	return func() string {
+		s, err := status.L2ForService(ConfigUpdater.Client(), svc)
+		if err != nil {
+			return err.Error()
+		}
+		return s.Status.Node
+	}
+}
+
+func firstSpeakerOtherThan(speakers []corev1.Node, exclude ...string) string {
```
oribon: this should be moved to the bottom of the file, also the "first" semantic here is a bit confusing (and it doesn't really apply), can we rename it to be clear around "picking a node that is not one of the excluded"?

e2etest/l2tests/preferred_node.go
```diff
@@ -0,0 +1,679 @@
...
+				natural = s.Status.Node
+				return nil
+			}, 2*time.Minute, time.Second).ShouldNot(HaveOccurred())
+
+			preferredA := firstSpeakerOtherThan(speakers, natural)
+			Expect(preferredA).NotTo(BeEmpty())
+			preferredB := firstSpeakerOtherThan(speakers, natural, preferredA)
+			Expect(preferredB).NotTo(BeEmpty())
+
+			const rtLabelKey = "l2-preferred-rt-test"
```
oribon, line 263: this label leaks I think (not handled by the delete in aftereach)

e2etest/l2tests/preferred_node.go
```diff
@@ -0,0 +1,679 @@
...
+			out = append(out, *n)
+		}
+	}
+	return out
+}
+
+// announcingNode returns a Gomega poller that yields the node currently
+// announcing svc, or the error string when the lookup fails so that the
+// matcher's failure output is informative.
+func announcingNode(svc *corev1.Service) func() string {
```
oribon: please move to the bottom of the file

e2etest/l2tests/preferred_node.go
```diff
@@ -0,0 +1,679 @@
...
+					},
+				},
+			}
+			err = ConfigUpdater.Update(config.Resources{
+				L2Advs: []metallbv1beta1.L2Advertisement{l2adv},
+			})
+			Expect(err).NotTo(HaveOccurred())
+
+			ginkgo.By(fmt.Sprintf("Validating the announcing node stays at %s (baseline winner)", natural))
+			Consistently(announcingNode(svc), 5*time.Second, time.Second).Should(Equal(natural))
```
oribon, line 173: should we also add an adv that adds weight 10 to all nodes (without label selector) and verify it stays on the same node?

e2etest/l2tests/preferred_node.go
```diff
@@ -0,0 +1,679 @@
...
+		})
+
+		ginkgo.It("should converge on a stable eligible speaker when no preferredNodeSelectors are set", func() {
```
oribon: I think this is already covered by the previous?

e2etest/l2tests/preferred_node.go
```diff
@@ -0,0 +1,679 @@
...
+	ginkgo.Context("Preferred Node Selector", func() {
+
+		ginkgo.It("should prefer the labeled node and fall back to the baseline election when the label is removed", func() {
+			allNodes, err := cs.CoreV1().Nodes().List(context.Background(), metav1.ListOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			speakers := speakerEligibleNodes(cs, allNodes.Items)
+			if len(speakers) < 2 {
+				ginkgo.Skip("need at least two speaker-eligible nodes")
+			}
```
oribon: should this be in beforeeach?

oribon (commented):
thanks a lot! looks really good, mostly small comments

internal/config/preferred_node_test.go
oribon, line 1: can we add the case where the node selectors target x, preferred targets y but there is no node in x with y labels? here and in the e2e I guess

internal/config/config_test.go
jlandersen: Added both directions!

internal/config/preferred_node_test.go
jlandersen: Good point. Updated to use 40/70 so each node's score uniquely identifies the contributing selector. Equal-score tie-break stays covered by TestSortL2Candidates.

internal/config/preferred_node_test.go
jlandersen: Done. Migrated to config_test.go and switched to explicit wantSvcSelectors []string asserted via selector.String().

internal/config/preferred_node_test.go
jlandersen: Done as well! Migrated to config_test.go and switched to wantPreferred []map[string]int64 so each row pins which CR's weights survived dedup.

speaker/layer2_controller.go
jlandersen: Done. Initialized unconditionally above the loop :)

speaker/layer2_controller_test.go
jlandersen: Good point and I wasn't super happy about that either. I hope this works:

Split into two commits. First extracts sortL2Candidates with its own table test. Second one extends it with score-aware rows and migrates the hash-tie-break cases into that table. Dropped runL2Election/hashWinner. Kept TestShouldAnnouncePreferredNode, failover, and split-brain as standalone tests (the fixture shape doesn't fit TestShouldAnnounceFromNodes's 2-controller scaffold), but inlined their controller setup.

e2etest/l2tests/preferred_node.go
jlandersen: Right, role carries 50, so removing it leaves both at 60. Switched to remove l2-pref-zone instead (nodeDouble => 50, nodeSingle => 60) and corrected the By text.

e2etest/l2tests/preferred_node.go
jlandersen: No probably not :) Dropped speakerEligibleNodes, switched to positional allNodes.Items[i] (or observe-then-label) per test, and gate on cluster size. Hope this makes sense

e2etest/l2tests/preferred_node.go
jlandersen: Removed entirely along with speakerEligibleNodes in the fix from previous comment.

e2etest/l2tests/preferred_node.go
jlandersen: Good point. Updated to tracking (node, key) pairs in labeledNodes now.

e2etest/l2tests/preferred_node.go
jlandersen: Done, moved to the bottom with the other helpers :)

e2etest/l2tests/preferred_node.go
jlandersen: Added! The no-match test now appends a second PreferredNodeSelector with an empty selector at weight 10 within the same L2Advertisement and asserts the announcing node is unchanged.

e2etest/l2tests/preferred_node.go
jlandersen: Good point and yes definitely right. Removed the redundant spec and moved its ContainElement(announcer) invariant into the no-match test so it's not lost.

e2etest/l2tests/preferred_node.go
jlandersen: I updated to hoist the Nodes().List fetch to BeforeEach. Kept the inline count check per spec since two specs need >= 3 and the rest >= 2.

internal/config/preferred_node_test.go
jlandersen: Added a sibling spec where nodeSelectors matches one node (eligible-only) and the preferred label sits on a different node hard-filtered out. Asserts the eligible node announces and the preferred-only node never does. The unit half is a new row in TestL2AdvertisementFromCR.

jlandersen:
> thanks a lot! looks really good, mostly small comments

Thanks a lot for the comments! I have tried to address all of them and updated the existing commits (including one new one where I think it made sense). Hope I got everything right, or I'm ready to update :)

I noticed e2e-use-operator is red still. My understanding is it uses CRDs at metallb/metallb-operator@main that lag these changes repo. I'm assuming this will be fixed by updating it after this feature / CRD changes are in (a bit chicken-and-egg problem)?

speaker/layer2_controller_test.go
```diff
@@ -2332,3 +2332,64 @@ func TestIPAdvertisementFor(t *testing.T) {
...
+	if len(a) != len(b) {
+		return false
+	}
+	for i := range a {
+		if a[i] != b[i] {
+			return false
+		}
+	}
+	return true
+}
```
oribon: I think we can use slices.equal

speaker/layer2_controller_test.go
```diff
@@ -2332,3 +2332,64 @@ func TestIPAdvertisementFor(t *testing.T) {
...
+			desc:       "hash order depends on the IP",
+			candidates: []string{"node-a", "node-b", "node-c"},
+			ip:         "10.0.0.2",
+			want:       []string{"node-b", "node-c", "node-a"},
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.desc, func(t *testing.T) {
+			// Sort an unrelated permutation to confirm output is determined by hash.
+			permuted := append([]string(nil), test.candidates...)
```
oribon: I think this can be replaced by slices.clone

internal/config/config_test.go
```diff
@@ -4350,3 +4411,204 @@ func TestServiceSelectors(t *testing.T) {
 		})
 	}
 }
+
+func TestL2AdvertisementFromCR(t *testing.T) {
```
oribon: rename this to something specific to preferred

internal/config/config_test.go
```diff
@@ -4350,3 +4411,204 @@ func TestServiceSelectors(t *testing.T) {
...
+				ObjectMeta: metav1.ObjectMeta{Name: "prefer-edge-with-svc-selectors"},
+				Spec: v1beta1.L2AdvertisementSpec{
+					ServiceSelectors: []metav1.LabelSelector{
+						{MatchLabels: map[string]string{"app": "web"}},
+					},
+					PreferredNodeSelectors: []v1beta1.PreferredNodeSelector{
+						{Weight: 100, Preference: metav1.LabelSelector{MatchLabels: map[string]string{"role": "edge"}}},
+					},
+				},
+			},
```
oribon, line 4457: missing wantNodes I think

internal/config/config_test.go
```diff
@@ -4350,3 +4411,204 @@ func TestServiceSelectors(t *testing.T) {
...
+			wantPreferred:    map[string]int64{"edge-a": 100, "edge-b": 100},
+			wantSvcSelectors: []string{"app=web"},
+		},
+		{
+			desc: "no preferred selectors leaves PreferredNodes nil",
+			crd: v1beta1.L2Advertisement{
+				ObjectMeta: metav1.ObjectMeta{Name: "basic"},
+				Spec:       v1beta1.L2AdvertisementSpec{},
+			},
+		},
```
oribon, line 4471: let's add nodes and wantNodes here, and make the prefererred: nil explicit

internal/config/config_test.go
```diff
@@ -4350,3 +4411,204 @@ func TestServiceSelectors(t *testing.T) {
...
+				Spec: v1beta1.L2AdvertisementSpec{
+					NodeSelectors: []metav1.LabelSelector{
+						{MatchLabels: map[string]string{"role": "worker"}},
+					},
+					PreferredNodeSelectors: []v1beta1.PreferredNodeSelector{
+						{Weight: 50, Preference: metav1.LabelSelector{MatchLabels: map[string]string{"role": "edge"}}},
+					},
+				},
+			},
+			wantNodes: map[string]bool{"worker-a": true},
```
oribon: same, let's make the preferred being nil explicit

internal/config/config_test.go
```diff
@@ -4350,3 +4411,204 @@ func TestServiceSelectors(t *testing.T) {
...
+				}
+				return in
+			})); diff != "" {
+				t.Fatalf("ServiceSelectors mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+}
+
+func TestSetL2AdvertisementsToPools(t *testing.T) {
```
oribon: rename to something "preferred" specific

internal/config/config_test.go
oribon: should we add a case for weights summing up?

internal/config/config_test.go
oribon: should we add a case for weights summing up?

speaker/layer2_controller.go
```diff
@@ -169,17 +169,37 @@ func (c *layer2Controller) SetEventCallback(callback func(interface{})) {
...
 		hi := sha256.Sum256([]byte(candidates[i] + "#" + ip))
 		hj := sha256.Sum256([]byte(candidates[j] + "#" + ip))
 		return bytes.Compare(hi[:], hj[:]) < 0
 	})
 }
 
+// preferredScoresForService sums per-advertisement preference weights across
+// all ads applicable to a service. Returns nil when no ad carries
+// preferences. Callers must tolerate a nil map; reads return zero.
+func preferredScoresForService(ads []*config.L2Advertisement) map[string]int64 {
```
oribon: nit: can be renamed to preferredScoresFor()

e2etest/l2tests/preferred_node.go
```diff
@@ -0,0 +1,611 @@
...
+	})
+
+	ginkgo.BeforeEach(func() {
+		cs = k8sclient.New()
+		var err error
+		testNamespace, err = k8s.CreateTestNamespace(cs, "l2pref")
+		Expect(err).NotTo(HaveOccurred())
+
+		// Computed once per spec; specs that mutate node eligibility re-read
+		// from k8s themselves.
```
oribon: nit: not needed

e2etest/l2tests/preferred_node.go
```diff
@@ -0,0 +1,611 @@
...
+				},
+			})
+			Expect(err).NotTo(HaveOccurred())
+
+			natural := waitForNaturalWinner(svc)
+			nodeNames := make([]string, 0, len(allNodes.Items))
+			for _, n := range allNodes.Items {
+				nodeNames = append(nodeNames, n.Name)
+			}
+			Expect(nodeNames).To(ContainElement(natural), "announcing node must be a known cluster node")
```
oribon: nit: unnecessary

e2etest/l2tests/preferred_node.go
```diff
@@ -0,0 +1,611 @@
...
+	var cs clientset.Interface
+	var labeledNodes []labeledNode
+	var testNamespace string
+	var allNodes *corev1.NodeList
+
+	ginkgo.AfterEach(func() {
+		for _, ln := range labeledNodes {
+			k8s.RemoveLabelFromNode(ln.node, ln.key, cs)
+		}
+		labeledNodes = nil
```
oribon: is there an advantage to using this vs  `defer k8s.RemoveLabelFromNode`?

e2etest/l2tests/preferred_node.go
```diff
@@ -0,0 +1,611 @@
...
+
+			ginkgo.By(fmt.Sprintf("Validating the preferred node %s announces initially", preferredCandidate))
+			Eventually(announcingNode(svc), 2*time.Minute, time.Second).Should(Equal(preferredCandidate))
+
+			ginkgo.By(fmt.Sprintf("Excluding %s from external LBs (removes it from speaker election)", preferredCandidate))
+			k8s.AddLabelToNode(preferredCandidate, corev1.LabelNodeExcludeBalancers, "", cs)
+			defer k8s.RemoveLabelFromNode(preferredCandidate, corev1.LabelNodeExcludeBalancers, cs)
+
+			ginkgo.By(fmt.Sprintf("Validating failover to another node (not %s)", preferredCandidate))
+			Eventually(announcingNode(svc), 2*time.Minute, time.Second).ShouldNot(Equal(preferredCandidate))
```
oribon: would we expect it going back to natural?

e2etest/l2tests/preferred_node.go
```diff
@@ -0,0 +1,611 @@
...
+			if len(allNodes.Items) < 2 {
+				ginkgo.Skip("need at least two nodes")
+			}
+
+			nodeA := allNodes.Items[0].Name
+			nodeB := allNodes.Items[1].Name
+			var nodeC string
+			if len(allNodes.Items) >= 3 {
+				nodeC = allNodes.Items[2].Name
+			}
```
oribon: can the test just require 3 nodes instead of having `if nodeC != ""` later?
and I think we should make the adv explicit about weighting node c above the rest via the preferred (but still not match it via nodeselectors)
OR I think the next test already verifies this, so maybe no need (up to you)

e2etest/l2tests/preferred_node.go
```diff
@@ -0,0 +1,611 @@
...
+				L2Advs: []metallbv1beta1.L2Advertisement{
+					{ObjectMeta: metav1.ObjectMeta{Name: "svc-selector-baseline"}},
+				},
+			})
+			Expect(err).NotTo(HaveOccurred())
+
+			naturalMatched := waitForNaturalWinner(svcMatched)
+			naturalUnmatched := waitForNaturalWinner(svcUnmatched)
+
+			preferredForMatched := nonNaturalNode(allNodes.Items, naturalMatched)
```
oribon: should we have `nonNaturalNode(allNodes.Items, naturalMatched, naturalUnmatched)` here? to guarantee a new node gets the preference (and the unmatched service does not jump to it)

oribon (commented):
looks a lot better imo! mostly small comments around tests, thanks a lot! (and sorry for not responding for a while, was on vacation)

speaker/layer2_controller_test.go
jlandersen: done 👍

speaker/layer2_controller_test.go
jlandersen: done as well 👍

internal/config/config_test.go
jlandersen: Renamed to `TestL2AdvertisementFromCRPreferred`, hope this works

internal/config/config_test.go
jlandersen: right, added!

internal/config/config_test.go
jlandersen: added nodes/wantNodes and made wantPreferred: nil explicit.

internal/config/config_test.go
jlandersen: done!

internal/config/config_test.go
jlandersen: Similar to above I renamed it to `TestSetL2AdvertisementsToPoolsPreferred`. I hope this works or let me know if I misunderstood the naming pattern you were hoping for :)

internal/config/config_test.go
jlandersen: yep good point, added a case with two selectors matching the same node so the weights sum.

internal/config/config_test.go
jlandersen: added a within-CR summing case here as well

speaker/layer2_controller.go
jlandersen: done!

e2etest/l2tests/preferred_node.go
jlandersen: removed

e2etest/l2tests/preferred_node.go
jlandersen: removed as well

e2etest/l2tests/preferred_node.go
jlandersen: Nope, good point. Changed to that!

e2etest/l2tests/preferred_node.go
jlandersen: good point, only the preferred node was labeled, so it falls back to hash and natural wins. Cleaned it up a bit.

e2etest/l2tests/preferred_node.go
jlandersen: Went with the second option. Dropped the nodeC branch since the test right below already covers "preferred-but-not-eligible".  Let me know if you prefer a different solution :)

e2etest/l2tests/preferred_node.go
jlandersen: Done! Now excludes both natural winners (gated to 3 nodes) so the preferred node is genuinely new and we'd catch a leak onto the unmatched service.

jlandersen:
> looks a lot better imo! mostly small comments around tests, thanks a lot! (and sorry for not responding for a while, was on vacation)

Thanks again for the review round, much appreciated. I have tried to address each and update the existing commits with the changes (I'm not resolving the conversations yet, depends how you usually do it in the project here :) 
I *think* the red failing checks are two flakes (BGP related?) and the other ones not related to these changes 🤔

oribon (approved):
lgtm, thanks a lot!

jlandersen:
@oribon let me know if theres something I need to change around the failing checks. I *think* it's unrelated / flakes

PR #3073, opened 2026-06-10
helm: pass --health-probe-port to controller and speaker

oribon:
Without this flag, overriding livenessProbe.port in values.yaml changes the probe target but the binary still listens on the compiled-in default, causing pods to CrashLoop.

introduced by https://github.com/metallb/metallb/pull/3062
/kind bug

**Release note**:

```release-note
helm: pass --health-probe-port to controller and speaker
```


yahlifried:
lgtm

PR #3074, opened 2026-06-21
log unexpected read error in native consumeBGP

Ahmad-Faraj:
the header read in consumeBGP swallowed the error, there was a TODO sitting on it. on a clean session close thats just io.EOF so we dont want to spam logs, but any other read error was getting dropped silently which makes a dead session annoying to debug.

so it logs at error level unless its EOF, same style as the rest of the file. native mode is linux only so verified with GOOS=linux build + vet.


Ahmad-Faraj:
Good point. Updated to check `s.closed` under the mutex before logging, so local shutdowns and config reloads don't produce spurious error logs.

PR #3077, opened 2026-06-22
Add --frrk8s-secret-passthrough for external frr-k8s

oribon:
When frr-k8s runs in a separate namespace, allow passing BGP secret references directly instead of resolving them to cleartext. The user creates the secret in the frr-k8s namespace and metallb forwards the reference as-is.

/kind feature

**Release note**:

```release-note
Add --frrk8s-secret-passthrough for external frr-k8s, allows passing BGP secret references directly instead of resolving them to cleartext under the assumption that they are created in the frr-k8s namespace separately.
```

### DIFF
--- a/charts/metallb/templates/speaker.yaml
+++ b/charts/metallb/templates/speaker.yaml
@@ -289,6 +289,9 @@ spec:
         {{- end }}
         {{- if .Values.frrk8s.external }}
         - --frrk8s-namespace={{ required "namespace is required when frrk8s is external" .Values.frrk8s.namespace }}
+        {{- if .Values.frrk8s.secretPassthrough }}
+        - --frrk8s-secret-passthrough
+        {{- end }}
         {{- end }}
         {{- if .Values.tls.cipherSuites }}
         - --tls-cipher-suites={{ .Values.tls.cipherSuites }}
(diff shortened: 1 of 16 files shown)


charts/metallb/templates/speaker.yaml
```diff
@@ -289,6 +289,9 @@ spec:
         {{- end }}
         {{- if .Values.frrk8s.external }}
         - --frrk8s-namespace={{ required "namespace is required when frrk8s is external" .Values.frrk8s.namespace }}
+        {{- if .Values.frrk8s.secretPassthrough }}
+        - --frrk8s-secret-passthrough
```
yahlifried, line 293: silently ignored if not on external mode

internal/config/config_test.go
yahlifried, line 1: might be good to add a test case for when `FRRK8sSecretPassthrough: true` and there is a password to resolve (Confirming it does not get resolved)

yahlifried (commented):
one more thing worth noticing:
if a secret is duplicated from metallb-system to frr-k8s for example, it will show as if the secret is resolved from metallb-system even though frr-k8s ignores the `PasswordSecret.namespace` field and takes the one from its own namespace. not a bug, but might be confusing if happens.

internal/config/config_test.go
oribon: done

charts/metallb/templates/speaker.yaml
oribon: it is inside the external block, so it isn't rendered when external is false (and it is documented that the field only works for external

yahlifried:
lgtm

PR #3078, opened 2026-06-22
feat(helm): adds revisionHistoryLimit as a configurable parameter

NPastorale:
/kind feature

Adds `revisionHistoryLimit` as a configurable chart value for `controller` and `speaker`

**Which issue(s) this PR fixes**:

Some environments may prefer retaining fewer ReplicaSets to avoid unnecessary clutter, while others might require more. Exposing `revisionHistoryLimit` for all the applicable resources as a chart value provides additional flexibility without changing the current default behaviour.
This is a common parameter exposed by many charts and as such, users may expect it to be configurable (myself included).
Given the infinitesimality and low risk of the change, I've taken the liberty of not opening an issue beforehand. Do let me know whether you'd like me to open one regardless.

- No breaking changes
- Default value remains 10, matching the Kubernetes default

** Release notes**

```release-note
Add `revisionHistoryLimit` as a configurable chart value for `controller` and `speaker`
```


NPastorale:
@fedepaol  @oribon how closely do you want me to follow what Gemini suggested? those are simple changes and I am happy to implement them should you want me to. I think it is a bit overkill, although it would technically make it more robust. So I defer to your judgement

oribon:
lgtm thanks!

NPastorale:
@oribon could you approve the checks to be run again?

Issue #3063, opened 2026-05-22
ServiceL2Status reconcile loop floods the apiserver in v0.16.0 (regression from v0.15.3)


mzac:
### Summary

In L2 mode, the speaker's `servicel2status` controller enters a tight failing-reconcile loop in **v0.16.0**, generating ~5–6 errors/sec per LoadBalancer service against the apiserver. It is a **regression** — **v0.15.3 (and earlier) does not exhibit it**.

### Symptoms

Speaker logs, alternating at ~5/sec:

```
Reconciler error  controller=servicel2status  namespace=<ns>  name=<svc>
  error="resourceVersion should not be set on objects to be created"
Reconciler error  controller=servicel2status  namespace=<ns>  name=<svc>
  error="servicel2statuses.metallb.io \"l2-<rand>\" not found"
```

The apiserver logs the matching error at the same rate:

```
"Unhandled Error" err="apiserver received an error that is not an metav1.Status:
&errors.errorString{s:\"resourceVersion should not be set on objects to be created\"}:
resourceVersion should not be set on objects to be created"
```

The per-service `ServiceL2Status` object churns (create-fails-with-resourceVersion → not-found → recreate via `generateName`). On a small k3s / embedded-etcd cluster this pins the apiserver + etcd (visible CPU + "apply request took too long" etcd warnings). **The VIP is announced correctly the entire time** — the impact is CPU/etcd load + log flooding, not a connectivity outage (which distinguishes it from #2705).

### Clean version bisection

Same 3-node k3s cluster, same CRDs, **only the controller + speaker image tag changed** (`crds.enabled=false`, reusing the v0.16.0 CRDs, which are schema-compatible):

| Version | `servicel2status` reconcile errs / 30s | apiserver `resourceVersion` errs / 30s |
|---|---|---|
| **v0.16.0** | ~203 | ~41 |
| **v0.15.3** | ~0 | 0 |

### Environment / repro

- MetalLB **L2 mode**, 3 speakers (DaemonSet across 3 nodes)
- One `LoadBalancer` Service pulling a `/32` from an `IPAddressPool` + `L2Advertisement`
- k3s v1.35.x, embedded etcd
- **Namespace-independent** — reproduced in both `metallb-system` and a non-default namespace
- Downgrading the controller+speaker images to `v0.15.3` (keeping everything else identical) eliminates the churn

### Related

[...]

PR #3079, opened 2026-06-24
fix: avoid stale resourceVersion errors in ServiceL2Status and ServiceBGPStatus reconcile

somaz94:

/kind bug

Both status controllers reuse the object returned by the `List` as the write target for `controllerutil.CreateOrPatch`:

```go
state = &item                         // layer2_status_controller.go
state = &serviceBGPStatuses.Items[0]  // bgp_status_controller.go
```

That object carries the `resourceVersion` it had when it was listed. If the status is deleted between the `List` and the `CreateOrPatch` — for example by another speaker that briefly became leader and deletes other nodes' statuses — `CreateOrPatch` takes the `Get → NotFound → Create` path with that stale `resourceVersion` still set, and the API server rejects the create:

```
resourceVersion should not be set on objects to be created
```

So the reconcile returns an error instead of re-creating the status. Each such event produces several unnecessary errors before the state settles, rather than recovering in a single pass.

Fix: build the write target from identity only (`Name` + `Namespace`, no `resourceVersion`). The create path then runs with nothing stale set and succeeds on the first pass. `CreateOrPatch` re-populates the object via `Get` before patching, so the non-racy path is unchanged. The observed status is captured separately as `observedStatus` because the steady-state no-op short-circuit previously read it from `state.Status`; without that, the `DeepEqual` comparison would never match and every reconcile would write.

The leader-manages behavior (deleting other nodes' / redundant statuses) is untouched.

Scope: this is only about handling stale resourceVersions. It deliberately does not claim to explain or fix everything reported in #3063 — that discussion is unresolved and this PR no longer asserts a root cause for it. Related to #3063.

Added a deterministic regression test per controller. Each wraps the client to (1) emulate the manager field index the reconciler lists by, which envtest's direct client does not provide, and (2) delete the owned status right after the `List`, reproducing the window between `List` and `CreateOrPatch`.

Verified locally (envtest, k8s 1.34.1):

- `go test -short ./internal/k8s/controllers/...` — 10/10 specs pass, 0 skipped.
- Reverting only the two controller `.go` files to `main` and re-running makes both new specs fail with exactly `resourceVersion should not be set on objects to be created`, so the tests do pin this bug rather than passing vacuously.
- `gofmt` clean, `go vet` clean.
- `golangci-lint run ./internal/k8s/controllers/...` — 0 issues. Note: run with 2.12.2; the repo pins 2.11.4 in `tasks.py` and I could not run the pinned container locally, so CI is the authority on that one.

Rebased onto current `main`.

**Release note**:

```release-note
Fixed the ServiceL2Status and ServiceBGPStatus controllers returning "resourceVersion should not be set on objects to be created" instead of re-creating the status when it was deleted while a reconcile was in flight.
```

### DIFF
--- a/internal/k8s/controllers/layer2_status_controller.go
+++ b/internal/k8s/controllers/layer2_status_controller.go
@@ -96,9 +96,17 @@ func (r *Layer2StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
 
 	// We are the (sole) leader for this service.
 	var state *v1beta1.ServiceL2Status
+	var observedStatus v1beta1.MetalLBServiceL2Status
 	for _, item := range serviceL2statuses.Items {
 		if item.Labels[LabelAnnounceNode] == r.NodeName && state == nil {
-			state = &item
+			// avoid carrying a stale resourceVersion into CreateOrPatch
+			observedStatus = item.Status
+			state = &v1beta1.ServiceL2Status{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      item.Name,
+					Namespace: item.Namespace,
+				},
+			}
 			continue
 		}
 		// Delete statuses from other nodes / redundant statuses belonging to our node
@@ -120,7 +128,7 @@ func (r *Layer2StatusReconciler) Reconcile(ctx context.Context, req ctrl.Request
 	}
 
 	desiredStatus := r.buildDesiredStatus(ipAdvS, serviceName, serviceNamespace)
-	if reflect.DeepEqual(state.Status, desiredStatus) {
+	if reflect.DeepEqual(observedStatus, desiredStatus) {
 		return ctrl.Result{}, nil
 	}
 
--- a/internal/k8s/controllers/layer2_status_controller_test.go
+++ b/internal/k8s/controllers/layer2_status_controller_test.go
@@ -1,0 +1,130 @@
+// SPDX-License-Identifier:Apache-2.0
+
+package controllers
+
+import (
+	"context"
+	"net"
+
+	"github.com/go-kit/log"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/sets"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	v1beta1 "go.universe.tf/metallb/api/v1beta1"
+	"go.universe.tf/metallb/internal/layer2"
+)
+
+// deleteOnListClient emulates the manager field index the reconciler lists by and,
+// when armed, deletes the listed statuses right after returning them.
+type deleteOnListClient struct {
+	client.Client
+	serviceKey   string
+	deleteOnList bool
+}
+
+func (c *deleteOnListClient) List(ctx context.Context, list client.ObjectList, opts ...client.ListOption) error {
+	sl, ok := list.(*v1beta1.ServiceL2StatusList)
+	if !ok {
+		return c.Client.List(ctx, list, opts...)
+	}
+
+	// opts reference a field index that does not exist here, so re-resolve by label.
+	all := &v1beta1.ServiceL2StatusList{}
+	if err := c.Client.List(ctx, all); err != nil {
+		return err
+	}
+	for i := range all.Items {
+		item := all.Items[i]
+		key := types.NamespacedName{
+			Namespace: item.Labels[LabelServiceNamespace],
+			Name:      item.Labels[LabelServiceName],
+		}.String()
+		if key == c.serviceKey {
+			sl.Items = append(sl.Items, item)
+		}
+	}
+
+	if c.deleteOnList {
+		c.deleteOnList = false
+		for i := range sl.Items {
+			if err := c.Delete(ctx, sl.Items[i].DeepCopy()); err != nil && !apierrors.IsNotFound(err) {
+				return err
+			}
+		}
+	}
+	return nil
+}
+
+var _ = Describe("Layer2StatusReconciler stale resourceVersion", func() {
+	const (
+		raceNode = "race-node"
+		raceSvc  = "race-svc"
+		raceNS   = "race-ns"
+	)
+
+	cleanup := func() {
+		_ = k8sClient.DeleteAllOf(ctx, &v1beta1.ServiceL2Status{},
+			client.InNamespace(speakerNamespace),
+			client.MatchingLabels{LabelServiceName: raceSvc})
+	}
+
+	AfterEach(cleanup)
+
+	It("recovers without a stale resourceVersion error when the owned status is deleted concurrently mid-reconcile", func() {
+		inject := &deleteOnListClient{
+			Client:     k8sClient,
+			serviceKey: types.NamespacedName{Namespace: raceNS, Name: raceSvc}.String(),
+		}
+		r := &Layer2StatusReconciler{
+			Client:     inject,
+			Logger:     log.NewNopLogger(),
+			NodeName:   raceNode,
+			Namespace:  speakerNamespace,
+			SpeakerPod: speakerPod,
+			StatusFetcher: func(types.NamespacedName) []layer2.IPAdvertisement {
+				return []layer2.IPAdvertisement{
+					layer2.NewIPAdvertisement(net.IP("127.0.0.9"), true, sets.Set[string]{}),
+				}
+			},
+		}
+
+		existing := &v1beta1.ServiceL2Status{
+			ObjectMeta: metav1.ObjectMeta{
+				GenerateName: "l2-",
+				Namespace:    speakerNamespace,
+				Labels: map[string]string{
+					LabelAnnounceNode:     raceNode,
+					LabelServiceName:      raceSvc,
+					LabelServiceNamespace: raceNS,
+				},
+			},
+		}
+		Expect(k8sClient.Create(ctx, existing)).To(Succeed())
+		Expect(existing.ResourceVersion).ToNot(BeEmpty())
+
+		inject.deleteOnList = true
+
+		req := ctrl.Request{NamespacedName: types.NamespacedName{Namespace: raceNS, Name: raceSvc}}
+		_, err := r.Reconcile(ctx, req)
+		Expect(err).ToNot(HaveOccurred())
+
+		// A second reconcile (no concurrent delete) fills in the status subresource.
+		_, err = r.Reconcile(ctx, req)
+		Expect(err).ToNot(HaveOccurred())
+
+		var list v1beta1.ServiceL2StatusList
+		Expect(k8sClient.List(ctx, &list, client.MatchingLabels{
+			LabelServiceName:      raceSvc,
+			LabelServiceNamespace: raceNS,
+			LabelAnnounceNode:     raceNode,
+		})).To(Succeed())
+		Expect(list.Items).To(HaveLen(1))
+		Expect(list.Items[0].Status.Node).To(Equal(raceNode))
+	})
+})
--- a/internal/k8s/controllers/bgp_status_controller.go
+++ b/internal/k8s/controllers/bgp_status_controller.go
@@ -111,8 +111,17 @@ func (r *ServiceBGPStatusReconciler) Reconcile(ctx context.Context, req ctrl.Req
 		},
 	}
 
+	var observedStatus v1beta1.MetalLBServiceBGPStatus
 	if len(serviceBGPStatuses.Items) > 0 {
-		state = &serviceBGPStatuses.Items[0]
+		// avoid carrying a stale resourceVersion into CreateOrPatch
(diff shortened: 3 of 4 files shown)


internal/k8s/controllers/layer2_status_controller.go
yahlifried, line 1: same bug exists for bgp_status_controller.go, might be worth fixing as well

somaz94: Good catch — `bgp_status_controller.go` had the exact same pattern: `state = &serviceBGPStatuses.Items[0]` carried the listed object's resourceVersion into the `CreateOrPatch` create path. Fixed it the same way in 3b7e3a9f — reference the existing status by identity only (`Name`+`Namespace`) and capture the observed status separately for the no-op short-circuit. Added a deterministic reproducer (`bgp_status_controller_test.go`) mirroring the L2 one; it fails with the exact `resourceVersion should not be set on objects to be created` error before the fix and passes after. Thanks for flagging it.

somaz94:
Rebased onto latest `main` to pick up the go 1.25.11 / x/net v0.55.0 bump (`5be638f0`) — that clears the `static-security-analysis` govulncheck failures, which were stale-toolchain findings unrelated to this PR's changes. The `bgp_status_controller.go` case @yahlifried mentioned is already handled in `3b7e3a9f` with a deterministic reproducer test. PTAL — thanks!

somaz94:
@yahlifried @oribon — could I ask for a re-run of the two failing e2e legs (`e2e (ipv6, frr, manifests)` and `e2e (ipv6, frr-k8s, manifests)`) on run `30073923459`? I believe both are flakes rather than regressions from this PR:

- The failing specs do not touch what this PR changes. `grep -niE "ServiceBGPStatus|ServiceL2Status"` over `unnummbered.go`, `service_selector.go` and `validate.go` returns zero hits, while every spec that *does* exercise those status CRs (`l2tests/l2.go`, `l2tests/interface_selector.go`, `bgptests/bgp.go`) passed.
- The twin legs of the same modes passed in the same run: `e2e (ipv6, frr, helm)` and `e2e (ipv6, frr-k8s, helm)` are both green.
- Across this branch's three CI runs a *different* spec failed each time, and one run had e2e fully green — not the shape of a deterministic regression.
- Both failures are timeouts (`timed out after 30s` waiting for a route; `timed out after 240s` with 1 of 3 nodes not converging into VRF `red`), i.e. convergence timing.

Separately, is `static-security-analysis` a required check here? It looks repo-wide rather than PR-specific — `main` has been red on it consistently, the workflow installs the scanner unpinned (`go install golang.org/x/vuln/cmd/govulncheck@latest`), and the reported vulns are all in **indirect** deps this PR does not touch. This same branch was green on that leg on 2026-07-05 with essentially the same diff.

Happy to rebase if you would like, but since the one commit this is behind (`5fb56ec8`) only addresses 1 of the 3 reported vulns, I do not think it would clear the scan on its own. Let me know if you would prefer a different approach.

internal/k8s/controllers/layer2_status_controller_test.go
```diff
@@ -0,0 +1,158 @@
...
+		Expect(existing.ResourceVersion).ToNot(BeEmpty())
+
+		// The upcoming reconcile lists the owned status, then it is deleted
+		// concurrently (as another speaker that briefly became leader would do).
+		inject.deleteOnList = true
+
+		// Before the fix, the reconciler selected the listed object (carrying its
+		// resourceVersion) as the CreateOrPatch target; once it was gone, CreateOrPatch
+		// took the create path with a stale resourceVersion and failed with
+		// "resourceVersion should not be set on objects to be created", looping forever.
```
oribon: not sure I understand this part. can you elaborate why the previous version looped forever?

oribon (commented):
the code change (and test) looks good to me for avoiding `resourceVersion should not be set on objects to be created` scenarios, but I don't understand in particular why this solves the issue (or why the described scenario causes an "infinite loop"). can you please explain?

somaz94:
Thanks for taking a look, @oribon — happy to explain.

Short version: the race was always there. What the old code did was make it **unrecoverable**, so every retry reproduced the identical failure and no progress was ever made.

### Why the old code failed

In the leader branch the reconciler picked its `CreateOrPatch` target straight out of the `List` result:

```go
for _, item := range serviceL2statuses.Items {
    if item.Labels[LabelAnnounceNode] == r.NodeName && state == nil {
        state = &item   // carries Name, Namespace, UID *and* ResourceVersion
```

`controllerutil.CreateOrPatch` begins with a `Get` on that object, and both outcomes are broken here:

- **`Get` returns NotFound** — another speaker briefly became leader and deleted our status between the `List` and this call. `CreateOrPatch` then takes the **create** path and calls `Create(state)`. But `state` is still the listed object, so `metadata.resourceVersion` is populated and the API server rejects it with `resourceVersion should not be set on objects to be created`. A failed `Get` does not clear the struct, so nothing along that path ever strips the field.
- **`Get` succeeds from a lagging informer cache** while the object is already gone in etcd — it takes the patch path, and the patch fails with `servicel2statuses.metallb.io "l2-<rand>" not found`.

Those are exactly the two errors #3063 reports alternating: one race, observed from either side of the cache lag.

### Why it doesn't stop

This is the part worth being precise about — it is a livelock rather than a literal `for {}`, which I think is where the "infinite loop" wording is causing confusion.

The reconcile returns an error, so controller-runtime requeues it. On the requeue the reconciler lists again and — as long as the informer cache still serves the deleted status — selects **the same object with the same dead `resourceVersion`** and fails at exactly the same line.

The decisive property is that the failing attempt is **completely side-effect-free**: nothing is created, nothing is patched, and neither the cluster nor the cached object changes. So the next attempt runs with byte-identical inputs and produces a byte-identical failure. There is no state transition anywhere in the cycle that could let it exit — it can only escape if the cache happens to catch up. Under multi-speaker leader churn the window keeps reopening, so in practice it doesn't: hence the sustained ~5–6 errors/sec per service in the reporter's 3-node repro, and 859 errors in ~7 minutes in the downstream report (spatiumddi/spatiumddi#284).

That is also why it is HA-specific — with a single speaker there is no concurrent deleter, the window never opens, and the same code path is fine.

### Why the change fixes it

The fix stops reusing the listed object as the write target and builds one carrying **identity only**:

```go
state = &v1beta1.ServiceL2Status{
    ObjectMeta: metav1.ObjectMeta{Name: item.Name, Namespace: item.Namespace},
}
```

The same race now resolves on the first pass: `Get` returns NotFound, the create path runs with no `resourceVersion` set, `Create` succeeds, and the reconcile returns nil. The status exists again, and the following reconcile short-circuits on `DeepEqual` and writes nothing. The race isn't removed — it's made recoverable in a single pass instead of unrecoverable.

`CreateOrPatch` re-populates the object via `Get` before patching, so the non-racy path is unaffected. The only thing dropped is `item.Status`, which the no-op short-circuit needs — that's why the change keeps it separately as `observedStatus` instead of reading `state.Status`. Without that, the comparison would always be false and we'd trade one source of churn for another.

### What the test pins down

`layer2_status_controller_test.go` encodes precisely this sequence: a status is created (so it carries a real `resourceVersion`), the client is armed to delete it **immediately after the `List`** — standing in for the speaker that briefly became leader — and the reconcile must return without error and converge to exactly one owned status for the node. Against the pre-fix code that first `Reconcile` returns the `resourceVersion should not be set on objects to be created` error. `bgp_status_controller_test.go` does the same for the BGP side, which is the case @yahlifried asked about earlier.

Happy to fold any of this into the commit message or an in-tree comment if you'd like it recorded there.

internal/k8s/controllers/layer2_status_controller_test.go
somaz94: Fair — "looping forever" is loose wording in that comment, and I've explained the mechanism in full in a top-level comment. The short version for this line:

The failing reconcile is **side-effect-free**. It creates nothing, patches nothing, and leaves both the cluster and the informer cache exactly as they were. controller-runtime requeues on the returned error, the reconciler lists again, selects the same object with the same dead `resourceVersion`, and fails at the same line with identical inputs. Nothing in the cycle transitions state, so nothing can end it — it only escapes if the cache catches up, and under multi-speaker leader churn that window keeps reopening.

So it is a livelock, not a literal infinite loop. If it reads better, I'm happy to reword this comment to:

```go
// Before the fix, the reconciler selected the listed object (carrying its
// resourceVersion) as the CreateOrPatch target; once it was gone, CreateOrPatch
// took the create path with a stale resourceVersion and failed with
// "resourceVersion should not be set on objects to be created". The attempt had
// no side effects, so every requeue retried it with identical inputs and the
// reconcile never converged.
```

oribon:
I understand why the 2 scenarios you described are solved by the fix, what I don't understand is why we would get so many logs like mentioned in the issue (and repeatedly), other than "churn" and "cache lag" (which sounds a bit odd, because the issue mentions that this is easily reproducible). 
To be clear, I'm fine with the code changes and the scenarios you're describing are good to fix, but I'm not sure I understand the root cause.

somaz94:
Fair question, and I think my earlier answer was wrong to lean on "churn". You're right that a race doesn't square with mzac reporting it as 100% reproducible from startup.

I dug into the v0.15.3 -> v0.16.0 boundary. The only commit touching this file in that range is c501ad747 ("L2Status controller: refactor to leader node manages resources"), and it changed the object's lifecycle in a way I think explains the volume better than a race does:

v0.15.3 adopted whatever status existed for the service, with no label check:

```go
if len(serviceL2statuses.Items) > 0 {
    state = &serviceL2statuses.Items[0]
}
```

so one object per service was reused and patched in place across leader changes. Stable identity.

v0.16.0 only adopts when `Labels[announceNode] == r.NodeName`, deletes anything else, and when nothing was adopted creates a replacement under a fresh `GenerateName`. So on any label/leader mismatch the object is destroyed and recreated with a new name rather than relabelled. The non-leader path deletes too, and since the controller watches ServiceL2Status, each delete and each create feeds another reconcile.

That reframes this PR: the stale resourceVersion isn't what generates the churn, it's what makes the churn unrecoverable. `state` carries the resourceVersion of the object the reconcile just observed; if that object is already gone when CreateOrPatch does its Get, the create path runs with a stale resourceVersion and is rejected outright, so that pass can never re-create the object and the cycle repeats instead of settling. That's the alternation in the issue (`resourceVersion should not be set...` <-> `l2-<rand> not found`).

So I'd say there are two things here, and this PR only fixes the second:

1. delete-and-recreate-under-a-new-name on leader/label mismatch (introduced by c501ad747)
2. stale resourceVersion turning that into a hard, non-recovering failure (this PR)

I haven't been able to prove from the code alone that (1) never converges with a single stable leader, so I don't want to overstate it. mzac's "starts the instant the speakers come up" suggests the leader view is still settling at that point. mzac offered to run a #3079 build against the 3-node repro, which would tell us whether the errors stop entirely or just stop being fatal. That seems like the cheapest way to settle it, and it also tells you whether (1) needs its own fix.

Happy to narrow this PR to (2) and open a separate issue for (1) if you'd prefer them split.

oribon:
> it's what makes the churn unrecoverable

I'm not sure that's correct, but I do agree it turns each churn event into several unnecessary errors instead of recovering in one pass, and that's what I think this PR should be about. 
let's reframe the PR around being what it is (not mentioning infinite loops or whatever), and clean the code to only be around it (and no need for the abundance of comments inside). I'm not entirely sure it would fix what's mentioned in the issue (because the rca isn't convincing), but it's at least a bug deserving a fix (handling stale rvs).

somaz94:
Agreed, and thanks for pushing back on it — reframed the PR to be only about the stale resourceVersion handling.

- Dropped the root-cause narrative and the loop/flood framing from the title, description and release note. It now describes what you agreed it is: each such event produces several unnecessary errors instead of recovering in one pass.
- Changed `Fixed #3063` to `Related to #3063`, so it no longer auto-closes an issue this may not fully explain.
- Trimmed the code to just the fix. The loop restructure in `layer2_status_controller.go` is reverted, so the only change there is the one branch that builds the write target. Source diff is down to +15/-2 and +14/-2.
- Cut the comment blocks down to three lines each, and removed the same narrative from the test comments.

The tests are unchanged in substance. Reverting only the two controller `.go` files still makes both specs fail with exactly `resourceVersion should not be set on objects to be created`, so they pin the behaviour rather than passing vacuously.

Also rebased onto current main.

oribon:
The code changes look good. The comments are way too verbose though, both in the controllers and the tests. The production code comments (8-10 lines each) should be a single line like `// avoid carrying a stale resourceVersion into CreateOrPatch`. The test comments (struct docs, inline explanations) should be similarly trimmed, the code is self-explanatory enough - let the test name and PR description carry the context. 
Also, drop the "churn" and "looping forever" language throughout, this PR is about avoiding stale resourceVersions on concurrent deletes, not about fixing a reconcile loop.

somaz94:
Trimmed as suggested. The production comments are now the one-liner you proposed, and the test comments are down to the two non-obvious ones (the missing field index, and why a second reconcile is needed) — the rest now rides on the spec name and the PR description.

Also dropped the "churn"/"loop" wording everywhere, including the two commit messages and the test identifiers, so nothing in the code claims this is a reconcile-loop fix.

No logic changed — the diff against the previous push is comments and naming only. 10/10 specs still pass on envtest 1.34.1, `gofmt`/`go vet` clean, `golangci-lint` 0 issues.

oribon (approved):
LGTM, thanks!

PR #3081, opened 2026-06-28
fix(chart): align PodMonitor with native TLS metrics endpoint

### LINKED ISSUE #3080: podMonitors are broken since 0.16.0
### MetalLB Version

0.16.0

### Deployment method

Charts

### Main CNI

cilium 1.19

### Kubernetes Version

1.36.2

### Cluster Distribution

kubeadm

### Describe the bug

In Prometheus, since 1.16.0 (both charts and metallb), I have a Target Down on my podmonitors
- metallb-controller
- metallb-speaker

### To Reproduce

Install metallb with helm chart 1.16.1 (it may also be broken with 1.16.0; it was working with latests 1.15.x)

On probably any existing configuration, enable  `podMonitor.enabled: true`. You should already notice the 2 problems I mention in a next section.

With a port-foward, as described in the same next section, run `curl -k` to retrieve the metrics. You may get `unauthorize`.

For Prometheus,  add the labels you need for your Prometheus-Operator to detect the podMonitor. The podmonitors should be detected and return "Target Down".

Here is my values.yaml:
```
    fullnameOverride: "metallb"

    prometheus:
      serviceAccount: "prometheus"
      namespace: "monitoring"
      podMonitor:
        enabled: true
        additionalLabels:
          k8s-app: metallb
      prometheusRule:
        enabled: true
        additionalLabels:
          k8s-app: metallb

    controller:
      enabled: true
      resources:
        limits:
          cpu: 100m
          memory: 100Mi

    speaker:
      secretName: memberlist
      resources:
        limits:
          cpu: 100m
          memory: 100Mi

    frrk8s:
      enabled: false
```

### Expected Behavior

Prometheus should monitor speaker&controller without TargetDown as it was before.

It should work out of the box. Just enable podmonitors in the chart values and it should work.

### Additional Context

Here is one generated podmonitor:
```
apiVersion: monitoring.coreos.com/v1
kind: PodMonitor
metadata:
...
  name: metallb-controller
  namespace: metallb-system
spec:
  jobLabel: app.kubernetes.io/name
  namespaceSelector:
    matchNames:
    - metallb-system
  podMetricsEndpoints:
  - path: /metrics
    port: monitoring <-------- notice this
    scheme: https <----------- also notice that
  selector:
    matchLabels:
      app.kubernetes.io/component: controller
      app.kubernetes.io/instance: metallb
      app.kubernetes.io/name: metallb
```

You should notice 2 points here
- port is `monitoring`. This has changed since 1.16.0
- scheme is https. But my Prometheus does not have the certificated (it is self-signed). Notice that the serviceMonitor was refactored and has a `tlsConfig` section where the default value is `insecureSkipVerify: true`

I hacked the podmonitor for the controller (change the port and add the tlsConfig). With the expected configuration, I still have a TargetDown. Port-forward on the contaienr, on port 9120:

```
$ curl -L -k https://localhost:9120/metrics
unauthorized
```

yahlifried:
/kind bug

fixes the metrics endpoint according to the new tls behaviour

verified it works by doing:
```bash
Kind create cluster

helm install prometheus prometheus-community/kube-prometheus-stack \
    -n monitoring --create-namespace --wait --timeout 5m

helm install metallb ./charts/metallb/ -n metallb-system --create-namespace \
    --set controller.image.tag=v0.16.1 \
    --set speaker.image.tag=v0.16.1 \
    --set speaker.frr.enabled=true \
    --set frrk8s.enabled=false \
    --set prometheus.podMonitor.enabled=true \
    --set prometheus.podMonitor.additionalLabels.release=prometheus \
    --set prometheus.serviceAccount=prometheus-kube-prometheus-prometheus \
    --set prometheus.namespace=monitoring

kubectl -n monitoring port-forward svc/prometheus-kube-prometheus-prometheus 9090:9090
```
and looked on web to see no errors

**Release note**:
```release-note
Helm chart: the PodMonitor template now targets the correct HTTPS metrics port (metricshttps instead of the removed monitoring), includes tlsConfig with insecureSkipVerify: true, and authenticates via a dedicated ServiceAccount with a long-lived bearer token
```
fixes #3080

### DIFF
--- a/charts/metallb/templates/podmonitor.yaml
+++ b/charts/metallb/templates/podmonitor.yaml
@@ -23,12 +23,21 @@ spec:
     matchNames:
     - {{ .Release.Namespace }}
   podMetricsEndpoints:
-  - port: monitoring
+  - port: {{ template "metrics.exposedportname" . }}
+    honorLabels: true
     path: /metrics
-    scheme: https
+    scheme: "https"
+    authorization:
+      credentials:
+        name: {{ template "metallb.fullname" . }}-metrics-token
+        key: token
     {{- if .Values.prometheus.podMonitor.interval }}
     interval: {{ .Values.prometheus.podMonitor.interval }}
     {{- end }}
+{{- if .Values.prometheus.podMonitor.tlsConfig }}
+    tlsConfig:
+{{- toYaml .Values.prometheus.podMonitor.tlsConfig | nindent 6 }}
+{{- end }}
 {{- if .Values.prometheus.podMonitor.metricRelabelings }}
     metricRelabelings:
 {{- toYaml .Values.prometheus.podMonitor.metricRelabelings | nindent 4 }}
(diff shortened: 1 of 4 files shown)


charts/metallb/values.yaml
```diff
@@ -84,6 +84,17 @@ prometheus:
     #   replacement: $1
     #   action: replace
 
+    # Create ServiceAccount, Secret and RBAC resources for PodMonitor
+    # bearer-token authentication. Set to false if you manage metrics
+    # auth externally (e.g. via the Prometheus ServiceAccount).
+    rbac:
+      create: true
```
oribon: how does setting `create: false` work? how is the auth handled "externally"?

charts/metallb/values.yaml
```diff
@@ -84,6 +84,17 @@ prometheus:
...
+
+    # optional tls configuration for the podMonitor endpoints.
+    # Since v0.16.0 metrics are served over HTTPS with self-signed certs.
+    tlsConfig:
+      insecureSkipVerify: true
```
oribon, line 89: no need for the "Since .. comment" (and it's not always self-signed certs it can be modified)

charts/metallb/templates/podmonitor.yaml
oribon, line 1: The changes here seem more excessive than what the ServiceMonitor does, is there a reason you went with this approach vs doing exactly what is done in the ServiceMonitor? other than the ServiceMonitor using bearerTokenFile which is deprecated (https://prometheus-operator.dev/docs/api-reference/api/), but I think it makes sense to have both resources aligned

charts/metallb/values.yaml
yahlifried: removed

charts/metallb/templates/podmonitor.yaml
yahlifried: bearerTokenFile is only accesiable in ServiceMonitor, not in PodMonitor, so we'll need to create a secret and serviceAccount for it anyway (for either bearerTokenSecret or authorizaiton.credentials)
I used authorization.credentials since it is not deprecated

charts/metallb/values.yaml
yahlifried: removed it, not created unconditionally

oribon (approved):
lgtm, thanks!

PR #3097, opened 2026-07-12
ci: remind feature PRs to update website docs

### LINKED ISSUE #2663: Ensure somehow that when adding a feature, the relevant doc is added in hte same PR

pujitha24:
This adds a non-blocking CI check that reminds contributors and reviewers to update the documentation when a pull request is classified `/kind feature` but does not touch `website/`.

## Motivation

As described in #2663, feature PRs sometimes merge without their documentation counterpart. This surfaces a gentle, non-blocking reminder rather than a hard gate, which the issue notes is the preferred approach.

## Implementation

- New workflow `.github/workflows/docs-reminder.yaml`, triggered on `pull_request` (`opened`, `synchronize`, `reopened`, `edited`).
- Reuses MetalLB's existing `/kind <type>` convention (from `classify.yaml`) to detect feature PRs, so it only fires on PRs actually adding a feature.
- Lists changed files via `actions/github-script` (already a dependency in this repo) and checks for any change under `website/`.
- If a feature PR touches no docs, it emits a warning annotation and a step-summary reminder pointing at `website/content/`.

It is read-only so it works for pull requests from forks, declares least privilege (`contents: read`, `pull-requests: read`), and the whole script is wrapped in try/catch so a transient API error can never fail the build.

/kind cleanup

```release-note
NONE
```

Fixes #2663

### DIFF
--- a/.github/workflows/docs-reminder.yaml
+++ b/.github/workflows/docs-reminder.yaml
@@ -1,0 +1,63 @@
+name: Docs Reminder
+
+# Non-blocking reminder for contributors and reviewers: when a pull request is
+# classified as "/kind feature" (see classify.yaml) but does not touch the
+# website/ documentation, surface a warning so new features don't ship without
+# their docs counterpart. Fixes: https://github.com/metallb/metallb/issues/2663
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, edited]
+
+# Least privilege; read-only so this also works for pull requests from forks.
+permissions:
+  contents: read
+  pull-requests: read
+
+jobs:
+  docs-reminder:
+    runs-on: ubuntu-22.04
+    if: github.actor != 'dependabot[bot]'
+    steps:
+      - name: Remind to document feature PRs
+        uses: actions/github-script@v7
+        with:
+          script: |
+            // Guard everything: this is a non-blocking reminder and must never
+            // fail the build, even on a transient GitHub API error (issue #2663).
+            try {
+              const body = context.payload.pull_request.body || "";
+              // MetalLB classifies PRs with "/kind <type>" in the body (see classify.yaml).
+              // The negative lookbehind ignores quoted ("> /kind ...") occurrences.
+              const kindMatch = body.match(/(?<!> )\/kind (\w+)/);
+              const kind = kindMatch ? kindMatch[1] : "";
+              if (kind !== "feature") {
+                core.info(`PR kind is "${kind || "unset"}"; docs reminder only applies to feature PRs.`);
+                return;
+              }
+
+              const files = await github.paginate(github.rest.pulls.listFiles, {
+                owner: context.repo.owner,
+                repo: context.repo.repo,
+                pull_number: context.issue.number,
+                per_page: 100,
+              });
+              const touchesDocs = files.some((f) => f.filename.startsWith("website/"));
+              if (touchesDocs) {
+                core.info("Feature PR updates website/ docs.");
+                return;
+              }
+
+              const msg =
+                "This PR is marked `/kind feature` but does not update the `website/` " +
+                "documentation. If this feature is user-facing, please add or update the " +
+                "relevant docs under `website/content/` in this PR. This is a non-blocking " +
+                "reminder (see issue #2663).";
+              core.warning(msg, { title: "Feature PR missing documentation" });
+              await core.summary
+                .addHeading("Documentation reminder", 3)
+                .addRaw(msg)
+                .write();
+            } catch (error) {
+              // Downgrade any failure to an informational log so the check stays green.
+              core.info(`docs-reminder check skipped: ${error.message}`);
+            }
--- a/.github/workflows/docs-reminder.yaml
+++ b/.github/workflows/docs-reminder.yaml
@@ -1,9 +1,10 @@
 name: Docs Reminder
 
-# Non-blocking reminder for contributors and reviewers: when a pull request is
-# classified as "/kind feature" (see classify.yaml) but does not touch the
-# website/ documentation, surface a warning so new features don't ship without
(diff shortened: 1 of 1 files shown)


oribon:
thanks for submitting this!
did you validate this works somehow? I think opening a pr on your own fork should work

pujitha24:
Thanks! Yes — I validated it on my fork exactly as you suggested: pujitha24/metallb#1

I put the workflow on my fork's `main` and drove all three paths through that test PR:

| case | result |
|---|---|
| `/kind feature`, no `website/` changes | warns — annotation *"Feature PR missing documentation"*; job still **passes** ([run](https://github.com/pujitha24/metallb/actions/runs/29439948929)) |
| `/kind bug`, no `website/` changes | silent — logs `PR kind is "bug"; docs reminder only applies to feature PRs` ([run](https://github.com/pujitha24/metallb/actions/runs/29440026187)) |
| `/kind feature` + touches `website/` | silent — logs `Feature PR updates website/ docs.` ([run](https://github.com/pujitha24/metallb/actions/runs/29440088786)) |

So it only fires on feature PRs that skip docs, and it never fails the build — the script is wrapped in try/catch and the job is read-only (`contents: read`, `pull-requests: read`), so it also works for fork PRs.

The two red checks here (`e2e`, `static-security-analysis`) look unrelated to a workflow-only change — happy to rebase if a re-run would help.

oribon:
I'd say for the first case we should fail the job (so it is visible), the CI lane not being required handles this being "non blocking"

pujitha24:
Good call, done — the job now calls `core.setFailed()` in the missing-docs case instead of just `core.warning()`, so it exits non-zero and shows as a red X. The other three cases (non-feature PR, feature PR that does touch `website/`, and unexpected API errors) still return early / are swallowed by the existing guard, so they stay green as before. Since `docs-reminder` isn't a required status check, this makes the reminder visible without blocking merges. Updated the comments in the workflow to reflect this.

oribon:
cool, did you re-check everything in your fork?

pujitha24:
Re-checked both paths on my fork, with the job now up to date after a rebase onto main:

- Feature PR **with** docs → `docs-reminder` passes, no reminder: pujitha24/metallb#1 ✅
- Feature PR **without** docs → `docs-reminder` fails via `core.setFailed`, so the missing-docs reminder is clearly visible: pujitha24/metallb#2 ❌ (and since it's not a required check, it wouldn't block merging)

Thanks for the careful review!

---
_Disclosure: this contribution was drafted with Claude Code._

pujitha24:
Thanks again for the review! While checking on the PR I noticed the new `docs-reminder` check was itself failing on this PR — it flagged `/kind feature` even though this PR is `/kind cleanup`. The cause: the kind-detection regex matched the first `/kind` substring anywhere in the body, including the one inside backticks in this PR's own description text ("...classified `/kind feature`..."), instead of the real classification line further down.

I fixed it by anchoring the regex to the start of a line (`/^\/kind (\w+)/m`), so it only matches an actual `/kind <type>` line per the PR template convention, not incidental mentions in prose. Verified with a standalone regex test against this PR's real body — it now correctly picks up `cleanup` instead of `feature`.

Side note, not part of this PR: `classify.yaml` uses the same unanchored regex and has mislabeled this PR `kind/feature` for the same reason — I left that alone since it's a separate, pre-existing, shared workflow, but flagging it in case it's worth a follow-up.

oribon:
wdyt about just checking that the label is kind feature?
something like:
https://github.com/openperouter/openperouter/blob/a4bb871e944c2e0ef19089a6ac1cd8b3ea4d2bab/.github/workflows/ci.yaml#L26
seems a lot more simpler and not prone to this bug 😅  (and if the labeling itself has bugs then we only have 1 point of failure)

pujitha24:
Good call, thanks — done in 25314b7b: the job now gates on `contains(github.event.pull_request.labels.*.name, 'kind/feature')` instead of re-parsing the body, and I dropped the regex block entirely so `classify.yaml`'s label is the single source of truth, per your point about consolidating to one failure point. I copied your trigger types (`opened, synchronize, reopened, labeled`) too.

One thing worth flagging: since `classify.yaml` labels via the default `GITHUB_TOKEN`, that label-add won't itself fire a new `labeled` event (GitHub docs: GITHUB_TOKEN-triggered events don't dispatch new workflow runs), so on a fresh `opened` event there's a small race where this job could evaluate before the label lands, and a body-only edit (no new commit) that adds `/kind feature` no longer re-triggers the check now that we're off `edited`. I checked and openperouter's own workflow has this identical trade-off, so I've mirrored it rather than trying to solve it here — but wanted to surface it rather than assume it's a non-issue. Let me know if that's acceptable or if you'd rather handle ordering differently.

oribon:
thanks, and what's the reason for going with this approach (30 lines of js) versus 9 lines of bash? the latter seems a lot simpler 😅

pujitha24:
Fair point — no strong reason, honestly the github-script version was just the first thing that worked. Rewrote it in bf0374b1 as a plain bash step (checkout + `git diff --name-only BASE...HEAD | grep website/`), mirroring the openperouter job you linked. Used three-dot diff syntax so commits landed on `main` after the branch point don't cause false "docs touched" matches, and dropped the now-unused `pull-requests: read` permission since there's no more PR-files API call. Verified with `actionlint` and a throwaway local repo exercising both branches (docs touched / not touched).

oribon (approved):
lgtm thanks a lot!

PR #3099, opened 2026-07-13
fix(chart): make speaker hostNetwork configurable

### LINKED ISSUE #3005: Helm chart does not allow setting `hostNetwork: false` for metallb speaker pods

Hi there,

We run MetalLB with `hostNetwork: false`, but currently the helm chart gives us no way of tuning this parameter. This means we have to explicitly declare all the templates and keep them in sync upstream. Is there any way you could add this as a tunable parameter in the values.yaml file?

Many thanks.

In values.yaml:

```
speaker:
  hostNetwork: false
```

### Additional context

_No response_

logan2211:
/kind feature

Makes `hostNetwork` configurable for speaker pods through the Helm chart:

```yaml
   speaker:
     hostNetwork: false
 ```

The new `speaker.hostNetwork` value defaults to `true`, preserving the chart's existing behavior. This supports deployments where the speaker needs to run in the pod network namespace, such as peering with a node-local CNI BGP daemon as a workload peer.

Fixes #3005

**Release note**:

```release-note
The Helm chart now supports configuring speaker pod host networking through `speaker.hostNetwork`. The value defaults to `true`.
```

## Summary by CodeRabbit

* **New Features**
  * Added the `speaker.hostNetwork` Helm setting to configure whether MetalLB speakers use the host network; it defaults to enabled.
  * Added the optional `speaker.gratuitousARPInterval` setting to control periodic gratuitous ARP/NDP announcements. Leaving it unset or set to zero disables these announcements.
  * Added Layer 2 networking guidance for ARP/NDP traffic.
  * Documented both settings and validated `speaker.hostNetwork` through the Helm chart schema.


logan2211:
Hi @oribon @fedepaol I think this should be a pretty straight forward change. If possible would appreciate if it could land in the next release so I can drop my forked chart. 🙏

fedepaol:
@logan2211 I think it makes sense to document this, mentioning that l2 won't be able to work.

logan2211:
> @logan2211 I think it makes sense to document this, mentioning that l2 won't be able to work.

Done

logan2211:
Rebased

PR #3102, opened 2026-07-19
fix(speaker): derive memberlist SecretKey correctly with sha256

kali834x:

/kind bug

the secret key derivation in speakerlist.New uses sha.Sum([]byte(secret))[:32], but hash.Hash.Sum appends the current digest to its argument rather than hashing the argument. since no data is written to the hasher beforehand, the result is the raw secret bytes concatenated with sha256 of the empty string, then truncated to 32 bytes. for secrets longer than 32 characters the memberlist encryption key is literally the first 32 bytes of the plaintext secret, not a hash. this writes the secret into the hasher before calling Sum(nil) so the key is actually sha256(secret).

the go hash.Hash interface specifies that Sum "appends the current hash to b" (the argument). the previous code never called sha.Write, so the digest was always of the empty input. this is a one-line semantic fix with no behavioral change for correctly-short secrets that happened to be exactly 32 bytes, but for any other length the derived key was wrong.

```release-note
fix memberlist secret key derivation to actually hash the secret with sha256
```

## Summary by CodeRabbit

* **Refactor**
  * Updated internal secret-key calculation while preserving the resulting value and behavior.


kali834x:
any update?

PR #3109, opened 2026-07-31
bgppeer: don't skip create validation when disableMP is set

kali834x:

/kind bug

validatePeerCreate hands back the spec.disableMP deprecation warning with a nil error before it ever reaches the namespace check or Validator.Validate, so a BGPPeer that sets that field is admitted with no validation at all. it can land outside the metallb namespace, and duplicate peer addresses, a mismatched routerID or a mismatched myASN all get through, which the speaker then renders into an frr.conf that FRR won't reload. validatePeerUpdate never had the early return, so create was the odd one out here. holding the warning in a local and returning it after the checks pass keeps the warning behaviour as it is today.

the added table test fails on the current tree (the peer in the "default" namespace is admitted, and the mock validator is never called) and passes with the change.

**Release note**:

```release-note
Fixed BGPPeer admission validation being skipped entirely when the deprecated spec.disableMP field is set on creation.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Deprecated `DisableMP` usage now produces a warning in the API response while allowing valid BGP peer requests to continue through normal validation.
  * Namespace and cluster validation continue to reject invalid BGP peer requests, even when `DisableMP` is set.

* **Tests**
  * Added coverage for warning delivery and validation behavior when using the deprecated field.


kali834x:
any update?

PR #3115, opened 2026-08-11
tls: pass cipher suites even when tls version is 1.3

oribon:
We were too harsh by denying configurations for suites when tls min version is 1.3, as go itself does not deny them but rather ignores them.
Here we align with this approach.

>
> /kind bug
/kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

**Release note**:

```release-note
NONE
```


oribon:
same as https://github.com/metallb/metallb-operator/pull/556

PR #3121, opened 2026-08-21
e2etest: dualstack e2etests ipv6 regex

ajaggapa:

>
> /kind bug
> /kind cleanup
> /kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

/kind bug

The IPv6 regex used by e2e route parsing can extract a truncated nexthop from `ip -6 route` output. Go regular expressions try alternatives left-to-right, so the compressed-prefix alternative `([0-9a-fA-F]{1,4}:){1,7}:` matched addresses such as `fc00:f853:ccd:e793::1` too early. Combined with a trailing `.` that required one extra character after the match, dual-stack tests compared incomplete addresses against node IPs and failed.

This reorders the compressed-address alternatives so the more specific forms are tried first, and drops the trailing `.` so the full IPv6 nexthop is parsed.

Test-only change in `e2etest/pkg/routes/routes.go`.

**Release note**:

```release-note
NONE
```


fedepaol:
Tried to feed my agent with the regex (because I am not smart enough), but it seems there are still some rough edges. For example, 2001:db8:1:2::5:6 becomes 2001:db8:1:2::5

ajaggapa:
Thanks for catching that — you're right, the regex still had rough edges.

I dug into this more deeply and found the regex approach is fundamentally fragile for IPv6. Go tries alternations left-to-right, so compressed forms with multiple trailing hextets get truncated. Besides your `2001:db8:1:2::5:6` → `2001:db8:1:2::5` example, the same regex also fails on:

- `2001:db8::8a2e:370:7334` → `2001:db8::8a2e`
- `::ffff:192.0.2.1` → `::ffff:192`

Rather than trying to enumerate every RFC 4291 compressed form in regex, I've replaced the regex entirely: we now take the first token after `via` in `ip route` output and parse it with `net.ParseIP` (stripping zone IDs like `%eth0` first). This covers all standard IPv6 notations correctly.

Added unit tests for the cases above plus link-local, loopback, IPv4-mapped, and full notation. Pushed in f313a9d5.

fedepaol:
LGTM, thanks!

Issue #2920, opened 2026-01-17
NDP implementation is incomplete (Neighbor Unreachability Detection)


totoCZ:
### Is your feature request related to a problem?

Standard hosts work such as:

01:26:41.989324 IP6 2a09:e206:c1:x> 2a09:e206:c1:ff00::2: ICMP6, neighbor solicitation, who has 2a09:e206:c1:ff00::2, length 32
01:26:41.989358 IP6 2a09:e206:c1:ff00::2 > 2a09:e206:c1:xICMP6, neighbor advertisement, tgt is 2a09:e206:c1:ff00::2, length 24

MetaLB does not respond to it.

"     Unlike ARP, Neighbor Discovery detects half-link failures (using
      Neighbor Unreachability Detection) and avoids sending traffic to
      neighbors with which two-way connectivity is absent."

Kernel NDP proxy also implements it wrongly (its deprecated), but ndppd is correct.

The workflow with (what seems to be) a Juniper switch
- try multicast
- add host to table
- try GUA detection after timer expires (30 min)
- no reply on GUA -> drop neighbor from table, kill traffic
- go back to 1)

So every 30 min traffic is dropped for a couple of seconds.

Working reply (link local):

01:27:08.000350 IP6 fe80::eef4:bbff:fed9:c2c4 > fe80::e230:f900:a682:9838: ICMP6, neighbor advertisement, tgt is 2a09:e206:c1:ff00:c::, length 32

Not working reply (GUA):

00:58:57.689847 IP6 2a09:e206:c1:x > 2a09:e206:c1:ff00:c::: ICMP6, neighbor solicitation, who has 2a09:e206:c1:ff00:c::, length 32
00:58:58.689997 IP6 2a09:e206:c1:x > 2a09:e206:c1:ff00:c::: ICMP6, neighbor solicitation, who has 2a09:e206:c1:ff00:c::, length 32
00:58:59.689903 IP6 2a09:e206:c1:x > 2a09:e206:c1:ff00:c::: ICMP6, neighbor solicitation, who has 2a09:e206:c1:ff00:c::, length 32

### Describe the solution you'd like

Reply to GUA NDP with GUA.

### I've read and agree with the following

- [x] I've checked all open and closed issues and my request is not there.
- [x] I've checked all open and closed pull requests and my request is not there.

ginkel:
[...]

PR #3125, opened 2026-09-03
Allow sending periodic GARP / Unsolicited NDP requests

### LINKED ISSUE #2910: L2 mode: Support periodic gratuitous ARP or document FDB aging limitation

fedepaol:

>
> /kind bug
> /kind cleanup
/kind feature
> /kind design
> /kind flake
> /kind failing
> /kind documentation
> /kind regression

Allow to set sending periodic gratuitous ARP / NDP messages. Follow up of https://github.com/metallb/metallb/pull/2981
Fixes https://github.com/metallb/metallb/issues/2910 and potentially remediates https://github.com/metallb/metallb/issues/2920

When the parameter is set, MetalLB periodically sends out ARP and NDP messages.

**Release note**:

```release-note
Optionally send periodic ARP / NDP messages to refresh the neighbor cache of clients.
```

## Summary by CodeRabbit

- **New Features**
  - Added support for periodic gratuitous ARP/NDP announcements in Layer 2 mode.
  - Configure the announcement interval in seconds using the speaker command-line option or Helm values.
  - Announcements are disabled by default and can help refresh network switch forwarding information.
  - Invalid or excessively large interval values are rejected.

- **Documentation**
  - Added guidance covering configuration, defaults, traffic considerations, and recommendations related to switch forwarding database aging.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -233,6 +233,7 @@ jobs:
           WITH_VRF="--with-vrf"
           FOCUS=""
           FRRK8S_NAMESPACE=""
+          if [ "${{ matrix.deployment }}" == "manifests" ]; then SKIP="$SKIP|periodic-gratuitous-arp"; fi
           if [ "${{ matrix.bgp-type }}" == "native" ]; then SKIP="$SKIP|FRR|FRR-MODE|FRRK8S-MODE|BFD|VRF|DUALSTACK"; WITH_VRF=""; fi
           if [ "${{ matrix.ip-family }}" == "ipv4" ]; then SKIP="$SKIP|IPV6|DUALSTACK"; fi
           if [ "${{ matrix.ip-family }}" == "dual" ]; then SKIP="$SKIP|IPV6"; fi
--- a/charts/metallb/values.yaml
+++ b/charts/metallb/values.yaml
@@ -275,6 +275,8 @@ speaker:
   ignoreExcludeLB: false
   # --  BGP debounce timeout for FRR configuration reloads, in milliseconds. Only applies when BGP type is frr. Default (when unset) is 3000 ms. This feature is experimental
   bgpDebounceTimeout: null
+  # -- Interval in seconds for periodic gratuitous ARP/NDP announcements. Unset or 0 (default) disables periodic announcements.
+  gratuitousARPInterval: null
 
   image:
     repository: quay.io/metallb/speaker
--- a/e2etest/l2tests/l2.go
+++ b/e2etest/l2tests/l2.go
@@ -788,6 +788,97 @@ var _ = ginkgo.Describe("L2", func() {
 		},
 			ginkgo.Entry("IPV4 - Checking service", "ipv4"),
 			ginkgo.Entry("IPV6 - Checking service", "ipv6"))
+
+		ginkgo.It("periodic-gratuitous-arp - should increase gratuitous sent counter over time", func() {
+			resources := config.Resources{
+				Pools: []metallbv1beta1.IPAddressPool{
+					{
+						ObjectMeta: metav1.ObjectMeta{
+							Name: "l2-periodic-garp-test",
+						},
+						Spec: metallbv1beta1.IPAddressPoolSpec{
+							Addresses: []string{
+								IPV4ServiceRange,
+								IPV6ServiceRange},
+						},
+					},
+				},
+				L2Advs: []metallbv1beta1.L2Advertisement{emptyL2Advertisement},
+			}
+
+			err := ConfigUpdater.Update(resources)
+			Expect(err).NotTo(HaveOccurred())
+
+			svc, _ := service.CreateWithBackend(cs, testNamespace, "periodic-garp-svc",
+				service.TrafficPolicyCluster)
+			defer func() {
+				err := cs.CoreV1().Services(svc.Namespace).Delete(context.TODO(), svc.Name, metav1.DeleteOptions{})
+				Expect(err).NotTo(HaveOccurred())
+			}()
+
+			allNodes, err := cs.CoreV1().Nodes().List(context.TODO(), metav1.ListOptions{})
+			Expect(err).NotTo(HaveOccurred())
+
+			ingressIP := jigservice.GetIngressPoint(&svc.Status.LoadBalancer.Ingress[0])
+
+			Eventually(func() error {
+				return mac.RequestAddressResolution(ingressIP, executor.Host)
+			}, 2*time.Minute, 1*time.Second).Should(Not(HaveOccurred()))
+
+			ginkgo.By("checking connectivity to its external VIP")
+			Eventually(func() error {
+				return service.ValidateL2(svc)
+			}, 2*time.Minute, 1*time.Second).Should(Not(HaveOccurred()))
+
+			ginkgo.By("identifying the announcing speaker and taking an initial GARP counter reading")
+			var advSpeaker *corev1.Pod
+			var initialCount float64
+
+			Eventually(func() error {
+				advNode, err := advertisingNodeFromMAC(allNodes.Items, ingressIP, executor.Host)
+				if err != nil {
+					return err
+				}
+				sp, ok := speakerPods[advNode.Name]
+				if !ok {
+					return fmt.Errorf("could not find speaker pod on announcing node %s", advNode.Name)
+				}
+				advSpeaker = sp
+
+				speakerMetrics, err := metrics.ForPod(cs, promPod, advSpeaker, metallb.Namespace)
+				if err != nil {
+					return err
+				}
+				initialCount, err = metrics.CounterValue("metallb_layer2_gratuitous_sent",
+					map[string]string{"ip": ingressIP}, speakerMetrics)
+				if err != nil {
+					return err
+				}
+				if initialCount < 1 {
+					return fmt.Errorf("gratuitous_sent counter is %.0f, want >= 1", initialCount)
+				}
+				return nil
+			}, 2*time.Minute, 5*time.Second).ShouldNot(HaveOccurred())
+
+			ginkgo.By("waiting for the periodic GARP counter to increase")
+			// The speaker is deployed with --gratuitous-arp-interval=5, so we
+			// wait long enough for at least one extra periodic tick.
+			Eventually(func() error {
+				speakerMetrics, err := metrics.ForPod(cs, promPod, advSpeaker, metallb.Namespace)
+				if err != nil {
+					return err
+				}
+				current, err := metrics.CounterValue("metallb_layer2_gratuitous_sent",
+					map[string]string{"ip": ingressIP}, speakerMetrics)
+				if err != nil {
+					return err
+				}
+				if current <= initialCount {
+					return fmt.Errorf("gratuitous_sent counter did not increase: initial=%.0f current=%.0f", initialCount, current)
+				}
+				return nil
+			}, 30*time.Second, 5*time.Second).ShouldNot(HaveOccurred())
+		})
 	})
 
 	ginkgo.DescribeTable("validate requesting a specific address pool for Loadbalancer service", func(ipRange *string, autoAssign bool) {
--- a/e2etest/pkg/metrics/metrics.go
+++ b/e2etest/pkg/metrics/metrics.go
@@ -129,23 +129,13 @@ func ValidateGaugeValueCompare(check func(int) error, metricName string, labels 
 
 // ValidateCounterValue checks that the value related to the given metric is at most the expectedMax value.
 func ValidateCounterValue(check func(int) error, metricName string, labels map[string]string, allMetrics []map[string]*dto.MetricFamily) error {
-	var err error
-	var value int
-	found := false
-	for _, m := range allMetrics {
-		value, err = CounterForLabels(metricName, labels, m)
-		if err != nil {
-			continue
-		}
-		found = true
-		err := check(value)
-		if err != nil {
-			return fmt.Errorf("invalid value %d for %s, %w", value, metricName, err)
-		}
+	value, err := CounterValue(metricName, labels, allMetrics)
+	if err != nil {
+		return err
 	}
-
-	if !found {
-		return fmt.Errorf("metric %s not found", metricName)
+	v := int(value)
+	if err := check(v); err != nil {
+		return fmt.Errorf("invalid value %d for %s, %w", v, metricName, err)
 	}
 	return nil
 }
@@ -155,6 +145,44 @@ func CounterForLabels(metricName string, labels map[string]string, metrics map[s
 	return metricForLabels(metricName, labels, metrics, func(m *dto.Metric) int {
 		return int(m.GetCounter().GetValue())
 	})
+}
+
+// CounterValue returns the raw counter value for the given metric across all scraped pods.
+func CounterValue(metricName string, labels map[string]string, allMetrics []map[string]*dto.MetricFamily) (float64, error) {
+	for _, m := range allMetrics {
+		v, err := counterValueForLabels(metricName, labels, m)
+		if err != nil {
+			continue
+		}
+		return v, nil
+	}
+	return 0, fmt.Errorf("metric %s not found", metricName)
+}
+
+// counterValueForLabels retrieves the counter value as float64, preserving fractional
+// values and avoiding truncation from int conversion.
+func counterValueForLabels(metricName string, labels map[string]string, metrics map[string]*dto.MetricFamily) (float64, error) {
+	mf, ok := metrics[metricName]
+	if !ok {
+		return 0, fmt.Errorf("metric %s not in metrics", metricName)
+	}
+	for _, m := range mf.GetMetric() {
+		toMatch := len(labels)
+		for _, l := range m.GetLabel() {
+			v, ok := labels[l.GetName()]
+			if !ok {
+				continue
+			}
+			if v != l.GetValue() {
+				continue
+			}
+			toMatch--
+		}
+		if toMatch == 0 {
+			return m.GetCounter().GetValue(), nil
+		}
+	}
+	return 0, fmt.Errorf("label %s not found in metrics for %s", labels, metricName)
 }
 
 func GreaterOrEqualThan(min int) func(value int) error {
--- a/internal/layer2/announcer.go
+++ b/internal/layer2/announcer.go
@@ -3,9 +3,11 @@
 package layer2
 
 import (
+	"maps"
 	"net"
 	"os"
 	"regexp"
+	"slices"
 	"strconv"
 	"sync"
 	"time"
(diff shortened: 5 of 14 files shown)


charts/metallb/values.yaml
```diff
@@ -275,6 +275,8 @@ speaker:
   ignoreExcludeLB: false
   # --  BGP debounce timeout for FRR configuration reloads, in milliseconds. Only applies when BGP type is frr. Default (when unset) is 3000 ms. This feature is experimental
   bgpDebounceTimeout: null
+  # -- Interval in seconds for periodic gratuitous ARP/NDP announcements. 0 (default) disables periodic announcements.
+  gratuitousARPInterval: null
```
andreaskaris, line 279: nit: unset or value `0` disables periodic announcements

.github/workflows/ci.yaml
```diff
@@ -233,6 +233,7 @@ jobs:
           WITH_VRF="--with-vrf"
           FOCUS=""
           FRRK8S_NAMESPACE=""
+          if [ "${{ matrix.deployment }}" == "manifests" ]; then SKIP="$SKIP|periodic-gratuitous-arp"; fi
```
andreaskaris, line 236: qq .. why only on helm?

website/content/concepts/layer2.md
```diff
@@ -67,6 +67,23 @@ about 10s), please [file a bug](https://github.com/metallb/metallb/issues/new)!
...
+By default MetalLB only sends gratuitous ARP/NDP packets on failover. In some
+environments — for example, switches that age FDB entries faster than upstream
+gateways age their ARP cache — this can result in periodic packet loss for
+service IPs even without a leadership change. To work around this, the speaker
+can be configured to re-emit gratuitous ARP/NDP packets for every announced IP
+on a fixed interval, via the `--gratuitous-arp-interval` flag (or the
+`speaker.gratuitousARPInterval` Helm value). The value is in
+seconds; `0` (the default) keeps the original failover-only behaviour.
+
+The interval applies speaker-wide to every L2-announced IP on the node, so
```
andreaskaris, line 81: What about adding some jitter instead of pinging exactly at the --gratuitous-arp-interval mark? e.g. gratuitous-arp-interval is the ceiling and within 20% of the selected value, we can fire any time?

website/content/concepts/layer2.md
```diff
@@ -67,6 +67,23 @@ about 10s), please [file a bug](https://github.com/metallb/metallb/issues/new)!
 We can help you investigate and determine if the issue is with the client, or a
 bug in MetalLB.
 
+## Periodic gratuitous announcements
+
+By default MetalLB only sends gratuitous ARP/NDP packets on failover. In some
```
andreaskaris, line 72: do we want to document that metallb doesn't answer to gratuitous requests for the VIP (as this is part of the problem)?

internal/layer2/announcer.go
```diff
@@ -200,6 +224,62 @@ func (a *Announce) doSpam(adv IPAdvertisement) {
 	a.spamCh <- adv
 }
 
+func (a *Announce) periodicGARPLoop(interval time.Duration) {
```
andreaskaris, line 227: thinking out loud - does Announce need a cancel function, and a Close() error method, and periodicGARPLoop a context, and a switch on ctx.Done(), ticker.C?

internal/layer2/announcer.go
```diff
@@ -200,6 +224,62 @@ func (a *Announce) doSpam(adv IPAdvertisement) {
...
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+	for range ticker.C {
+		a.sendPeriodicGratuitous()
+	}
+}
+
+func (a *Announce) sendPeriodicGratuitous() {
+	for _, adv := range a.mergedAdvertisementsPerIP() {
```
andreaskaris, line 236: maybe could wrap this in a go routine and run time.After(<random duration up to, e.g. 0.2 * interval>)

that would mean the contrary of what I said earlier, as in pings would happen after the configured timer .. but they would be spaced out. wondering if there's another way to introduce some jitter but which is < than the configured interval

internal/layer2/announcer.go
```diff
@@ -200,6 +224,62 @@ func (a *Announce) doSpam(adv IPAdvertisement) {
...
+				existing.allInterfaces = true
+				existing.interfaces = nil
+			} else {
+				existing.interfaces = existing.interfaces.Union(adv.interfaces)
+			}
+			merged[ipStr] = existing
+		}
+	}
+
+	advs := make([]IPAdvertisement, 0, len(merged))
```
andreaskaris, line 276: nit: I'd prefer:

```
	return slices.Collect(maps.Values(merged))
```

internal/layer2/announcer_test.go
```diff
@@ -62,3 +62,173 @@ func Test_SetBalancer_AddsToAnnouncedServices(t *testing.T) {
...
+			announce := &Announce{
+				ips: tt.ips,
+			}
+			merged := announce.mergedAdvertisementsPerIP()
+			if len(merged) != len(tt.wantAllInterfaces) {
+				t.Fatalf("got %d merged advs, want %d", len(merged), len(tt.wantAllInterfaces))
+			}
+			byIP := map[string]IPAdvertisement{}
+			for _, adv := range merged {
+				byIP[adv.ip.String()] = adv
```
andreaskaris, line 115: this should check for a dup map key here and error if so

internal/layer2/announcer_test.go
```diff
@@ -62,3 +62,173 @@ func Test_SetBalancer_AddsToAnnouncedServices(t *testing.T) {
...
+			ips: map[string][]IPAdvertisement{
+				"svc1": {{ip: ipA, interfaces: sets.New("eth0"), allInterfaces: false}},
+				"svc2": {{ip: ipB, allInterfaces: true}},
+			},
+			wantAllInterfaces: map[string]bool{ipA.String(): false, ipB.String(): true},
+			wantInterfaces:    map[string]sets.Set[string]{ipA.String(): sets.New("eth0")},
+		},
+	}
+
+	for _, tt := range tests {
```
andreaskaris, line 104: I find this specific test very convoluted. mergedAdvertisementsPerIP returns a []IPAdvertisement. So it would IMO be easier to read to have a wantIPAdvertisements []IPAdvertisement (can convert into map[string][]IPAdvertisement or sort to account for different order, or can be a map[string][]IPAdvertisement from the start), and just make sure both desired and merged have the same lengths and then iterate over the merged slice and check  and compare against the map of the desired entries

andreaskaris: The test doesn't cover:

```
                        if adv.allInterfaces {
                                existing.allInterfaces = true <--- not covered
                                existing.interfaces = nil         <--- not covered
                        } else {
                                existing.interfaces = existing.interfaces.Union(adv.interfaces)
                        }
```

as it needs:

```
		{
			name: "allInterfaces wins over interface set reversed",
			ips: map[string][]IPAdvertisement{
				"svc2": {{ip: ipA, interfaces: sets.New("eth0"), allInterfaces: false}},
				"svc1": {{ip: ipA, allInterfaces: true}},
			},
			wantAllInterfaces: map[string]bool{ipA.String(): true},
		},

```

internal/layer2/announcer.go
andreaskaris, line 227: we could unit test the periodicGARPLoop with testing/synctest, if we think it's worth it (perhaps in case we add some jitter) - e.g. we could run the test for 30 virtual seconds and see how often, etc.

though that's prob overkill

internal/layer2/announcer_test.go
```diff
@@ -62,3 +62,173 @@ func Test_SetBalancer_AddsToAnnouncedServices(t *testing.T) {
...
+func (f *fakeNDPResponder) Gratuitous(ip net.IP) error {
+	f.gratuitous = append(f.gratuitous, ip)
+	return nil
+}
+
+func (f *fakeNDPResponder) Watch(net.IP) error   { return nil }
+func (f *fakeNDPResponder) Unwatch(net.IP) error { return nil }
+func (f *fakeNDPResponder) Close() error         { return nil }
+
+func (f *fakeNDPResponder) gratuitousCalls() sets.Set[string] {
```
andreaskaris, line 228: instead of returning the sets, should these return map[string]int, and count? and then assert in the Test that the indices are all there (what the set stuff is doing right now) but also making sure that none is > 1? (to make sure that we ping each exactly once)

e2etest/l2tests/l2.go
```diff
@@ -788,6 +788,96 @@ var _ = ginkgo.Describe("L2", func() {
...
+				if !ok {
+					return fmt.Errorf("could not find speaker pod on announcing node %s", advNode.Name)
+				}
+				advSpeaker = sp
+
+				speakerMetrics, err := metrics.ForPod(cs, promPod, advSpeaker, metallb.Namespace)
+				if err != nil {
+					return err
+				}
+				err = metrics.ValidateCounterValue(metrics.GreaterOrEqualThan(1),
```
andreaskaris, line 852: can't you just do initialCount, err = CounterValue(...)
if initialCount < 1 {
  return fmt.Errorf(...)
}

Instead of doing ValidateCounterValue, then CounterValue?

e2etest/pkg/metrics/metrics.go
```diff
@@ -157,6 +157,18 @@ func CounterForLabels(metricName string, labels map[string]string, metrics map[s
 	})
 }
 
+// CounterValue returns the raw counter value for the given metric across all scraped pods.
+func CounterValue(metricName string, labels map[string]string, allMetrics []map[string]*dto.MetricFamily) (float64, error) {
```
andreaskaris, line 161: can you call CounterValue from ValidateCounterValue, for consistency and not to have 2 implementations of the same thing?

andreaskaris (commented):
basically all nits and questions. in the E2E tests, you could do the typical  Expect().To(Succeed()) in a few locations though defo not necessary
speaking about functionality, introducing some jitter may be an idea within batches of the same iterations. and unit tests could be implemented with synctest to test actual functionality, thought the E2E test already covers this

.github/workflows/ci.yaml
fedepaol: because it's easier to specify the parameter vs having to have a separate kustomize overlay

internal/layer2/announcer.go
fedepaol: I am torn about introducing jitter. On one side, it's probably better for the health of the network, on the other side doing it right (both from a ux and an implementation point of view) introduces complexity I am not sure I want to handle right now.
For example, writing each one of those in a goroutine feels dangerous, and I'd like better some sort of pool, which again needs some thought to be get right.

So, if you agree, I'd start simple and iterate over this if needed.

e2etest/pkg/metrics/metrics.go
fedepaol: yep makes sense

internal/layer2/announcer.go
fedepaol: yeah, I was lazy and that's the right thing to do, although I don't see a problem in shutting the goroutine when the process goes down.

internal/layer2/announcer.go
fedepaol: fair enough, I'll change it

.github/workflows/ci.yaml
andreaskaris: ok

internal/layer2/announcer.go
andreaskaris: I don't either. it would be interesting for unit testing. I don't have a strong opinion, you can keep it as is

internal/layer2/announcer.go
andreaskaris: ok, no jitter it is

charts/metallb/values.yaml
andreaskaris: thanks

internal/layer2/announcer.go
andreaskaris: thanks

internal/layer2/announcer_test.go
andreaskaris: thanks

internal/layer2/announcer_test.go
andreaskaris: thanks

internal/layer2/announcer_test.go
andreaskaris: thanks

internal/layer2/announcer_test.go
```diff
@@ -62,3 +62,204 @@ func Test_SetBalancer_AddsToAnnouncedServices(t *testing.T) {
...
+			ipV4.String():  1,
+			ipV6a.String(): 2,
+			ipV6b.String(): 1,
+		},
+		spamCh: make(chan IPAdvertisement, 1024),
+	}
+
+	announce.sendPeriodicGratuitous()
+
+	gotNDP := ndp.gratuitousCalls()
```
andreaskaris:
```
	g := gomega.NewWithT(t)
	gotNDP := ndp.gratuitousCalls()
	wantNDP := map[string]int{ipV6a.String(): 1, ipV6b.String(): 1}
	g.Expect(gotNDP).To(gomega.Equal(wantNDP))
	gotARP := arp.gratuitousCalls()
	wantARP := map[string]int{ipV4.String(): 1}
	g.Expect(gotARP).To(gomega.Equal(wantARP))
```
or any other way to test for map equality would have done it for me instead of reinventing the wheel, though for pretty printing I'm ok with it ;)

e2etest/l2tests/l2.go
andreaskaris: thanks

e2etest/pkg/metrics/metrics.go
andreaskaris: thanks

website/content/concepts/layer2.md
```diff
@@ -67,6 +67,27 @@ about 10s), please [file a bug](https://github.com/metallb/metallb/issues/new)!
 We can help you investigate and determine if the issue is with the client, or a
 bug in MetalLB.
 
+## Periodic gratuitous announcements
+
+By default MetalLB only sends gratuitous ARP/NDP packets on failover, and it
+does not reply to ARP/NDP requests targeting the service VIP itself (only the
```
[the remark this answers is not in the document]
andreaskaris: how would you rephrase this?

e2etest/pkg/metrics/metrics.go
```diff
@@ -157,6 +147,18 @@ func CounterForLabels(metricName string, labels map[string]string, metrics map[s
...
+	for _, m := range allMetrics {
+		v, err := CounterForLabels(metricName, labels, m)
```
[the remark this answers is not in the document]
andreaskaris: coderabbit's right - why the conversion from float -> int -> float -> int?

either make counterforlabels return a float and convert once ... or stick with the conversion to int, especially given that ValidateCounterValue already used int:

```
diff --git a/e2etest/l2tests/l2.go b/e2etest/l2tests/l2.go
index 2d2cb83ad..89b59f76a 100644
--- a/e2etest/l2tests/l2.go
+++ b/e2etest/l2tests/l2.go
@@ -832,7 +832,7 @@ var _ = ginkgo.Describe("L2", func() {
 
                        ginkgo.By("identifying the announcing speaker and taking an initial GARP counter reading")
                        var advSpeaker *corev1.Pod
-                       var initialCount float64
+                       var initialCount int
 
                        Eventually(func() error {
                                advNode, err := advertisingNodeFromMAC(allNodes.Items, ingressIP, executor.Host)
@@ -855,7 +855,7 @@ var _ = ginkgo.Describe("L2", func() {
                                        return err
                                }
                                if initialCount < 1 {
-                                       return fmt.Errorf("gratuitous_sent counter is %.0f, want >= 1", initialCount)
+                                       return fmt.Errorf("gratuitous_sent counter is %d, want >= 1", initialCount)
                                }
                                return nil
                        }, 2*time.Minute, 5*time.Second).ShouldNot(HaveOccurred())
diff --git a/e2etest/pkg/metrics/metrics.go b/e2etest/pkg/metrics/metrics.go
index 29a2f4808..5b9510d7b 100644
--- a/e2etest/pkg/metrics/metrics.go
+++ b/e2etest/pkg/metrics/metrics.go
@@ -133,9 +133,8 @@ func ValidateCounterValue(check func(int) error, metricName string, labels map[s
        if err != nil {
                return err
        }
-       v := int(value)
-       if err := check(v); err != nil {
-               return fmt.Errorf("invalid value %d for %s, %w", v, metricName, err)
+       if err := check(value); err != nil {
+               return fmt.Errorf("invalid value %d for %s, %w", value, metricName, err)
        }
        return nil
 }
@@ -148,13 +147,13 @@ func CounterForLabels(metricName string, labels map[string]string, metrics map[s
 }
 
 // CounterValue returns the raw counter value for the given metric across all scraped pods.
-func CounterValue(metricName string, labels map[string]string, allMetrics []map[string]*dto.MetricFamily) (float64, error) {
+func CounterValue(metricName string, labels map[string]string, allMetrics []map[string]*dto.MetricFamily) (int, error) {
        for _, m := range allMetrics {
                v, err := CounterForLabels(metricName, labels, m)
                if err != nil {
                        continue
                }
-               return float64(v), nil
+               return v, nil
        }
        return 0, fmt.Errorf("metric %s not found", metricName)
 }
```

andreaskaris (commented):
lgtm  once the 2 small things from rabbit are addressed

website/content/concepts/layer2.md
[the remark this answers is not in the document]
andreaskaris: isn't the entire idea behind this PR that metallb does _not_ answer to ARPs / ND for VIPs? please elaborate further why you believe that metallb does indeed answer, and why then is this pr needed?

fedepaol:
@andreaskaris I think I addressed the last comments.

PR #3130, opened 2026-09-12
fix(speaker): make exclusion ConfigMap readable

### LINKED ISSUE #2960: Permission denied for `metallb-excludel2` configmap mount
### MetalLB Version

0.15.3

### Deployment method

Charts

### Main CNI

calico

### Kubernetes Version

1.34.3

### Cluster Distribution

RKE2

### Describe the bug

The symptom is similar as described in #2309  - the speaker images in the cluster cannot start and show this in the logs:

```
"error":"open /etc/metallb/excludel2.yaml: permission denied",
"msg":"failed to parse announcedInterfacesToExclude from configMap"
```

I am installing metallb with helm, and I cannot figure out a way to modify the `defaultMode` permissions on the configmaps volume mount. I believe [line 114 of the speaker config](https://github.com/metallb/metallb/blob/555be75a79ab146213ae7438cb1a041b3dd24761/config/controllers/speaker.yaml#L114C11-L114C22) should change from `defaultMode: 256` to `defaultMode: 292` (unix 0444). I dont see anything in the file that would be a security risk, but the problem arises if the `speaker` image is run as a non-root user. In my case, I am trying to start the images as non-root (uid/gid of 65532), and because of that speaker cannot start up because the volume mount has permissions so only root can read.

A workaround is to add `--set speaker.securityContext.runAsUser=0` to the helm install but this defeats the purpose of using a hardened image.

If there is another solution besides modifying the manifest configuration I would love to hear it, but either way I think that manifest should probably change since this will be an issue with any image customization that tries to make the speaker image more secure.

### To Reproduce

1. Create customized speaker image that sets the user to a not root user
2. Use `--set speaker.image.repository` to point to a customized image and install via helm chart

### Expected Behavior

Expect the speakers to start properly without a fatal error.

### Additional Context

None

akashchamp:
/kind bug

Fixed #2960

The speaker mounts the non-secret `metallb-excludel2` ConfigMap with mode `0400`. A non-root speaker cannot read that file, so startup fails with a permission error. Set the mode to `0444` in both the bundled manifest source and Helm chart, regenerate the six derived manifests, and add a Helm policy that prevents either rendered path from regressing.

- `memberlist` remains a Secret with its existing `0644` mode; this change only affects the non-secret exclusion ConfigMap.
- Validation included `inv generatemanifests`, `helm lint charts/metallb`, Helm Conftest (42/42), and native plus hardened Helm renders that each produce `defaultMode: 292`.
- A deliberately reverted `256` mode fails the new policy. A rootless Podman check running as UID 65532 was denied at `0400` and read the same config successfully at `0444`.

**Release note**:

```release-note
Fix speaker startup when running as non-root by making the exclude-L2 ConfigMap readable.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated MetalLB speaker configuration permissions so the exclude-interfaces configuration can be read by all required processes.
  * Added validation to ensure the configuration volume uses the expected read-only permission mode.

### DIFF
--- a/charts/metallb/policy/speaker.rego
+++ b/charts/metallb/policy/speaker.rego
@@ -32,3 +32,12 @@ deny[msg] {
   not input.spec.template.spec.tolerations[0] == { "key": "node-role.kubernetes.io/master", "effect": "NoSchedule", "operator": "Exists" }
   msg = "controller tolerations does not include node-role.kubernetes.io/master:NoSchedule"
 }
+# validate exclude-L2 config is readable by a non-root speaker
+deny[msg] {
+  input.kind == "DaemonSet"
+  endswith(input.metadata.name, "-speaker")
+  volume := input.spec.template.spec.volumes[_]
+  volume.name == "metallb-excludel2"
+  not volume.configMap.defaultMode == 292
+  msg = "speaker metallb-excludel2 ConfigMap defaultMode must be 292"
+}
--- a/charts/metallb/templates/speaker.yaml
+++ b/charts/metallb/templates/speaker.yaml
@@ -194,7 +194,7 @@ spec:
       {{- if .Values.speaker.excludeInterfaces.enabled }}
         - name: metallb-excludel2
           configMap:
-            defaultMode: 256
+            defaultMode: 292
             name: metallb-excludel2
       {{- end }}
         - name: memberlist-config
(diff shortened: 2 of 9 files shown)


fedepaol:
LGTM, thanks!

